Issue #51, opened 2017-12-12
Trying to parse any map with duplicate keys (sometimes) results in a non-helpful exception


belyaev-mikhail:
Let's take the following code:
``` kotlin
@Serializable
data class WrappedMap(val mp: Map<String, Int>)
...
println(run<WrappedMap>{ JSON.parse("""{"mp": { "x" : 23, "x" : 42, "y": 4 }}""") })
```

This results in a following exception:
```
Exception in thread "main" kotlinx.serialization.SerializationException: Elements should be in order, unexpected index 3
	at kotlinx.serialization.internal.ListLikeSerializer.load(CollectionSerializers.kt:74)
	at kotlinx.serialization.KInput.readSerializableValue(Serialization.kt:240)
	at kotlinx.serialization.ElementValueInput.readSerializableElementValue(Serialization.kt:402)
	at ru.spbstu.competition.WrappedMap$$serializer.load(Main.kt)
	at ru.spbstu.competition.WrappedMap$$serializer.load(Main.kt:73)
	at kotlinx.serialization.KInput.read(Serialization.kt:206)
	at kotlinx.serialization.json.JSON.parse(JSON.kt:46)
        <...my code...>

```

This is not specific to JSON and is somehow related to the way `Map`s are deserialized from any input type where duplicate keys are possible. 

You should either be exact about whether you support duplicate keys in such formats or at least provide a better exception, 'cos this is not helpful at all. Even more, if you remove `y` from the data, it is parsed successfully with the value of `x` being `42`.

belyaev-mikhail:
Btw, having duplicate keys in JSON is kinda semi-valid according to [this SO answer](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object/23195243#23195243), which has all further links.

belyaev-mikhail:
I think I know how to fix this, do you accept pull requests for runtime?

I have a very weird use case where this bug is critically important)

elizarov:
PRs are welcome.

PR #55, opened 2017-12-14
Trying to fix issue #51

belyaev-mikhail:
#51 

The idea is to write a separate builder for every standard set- or map-like structure to distinguish builder size (the number of entries added) and set size (which may be smaller). 

The (arguably) better way would be to change how `ListLikeSerializer` works, but that would be a more global change with higher chance of breaking other pieces of the library.

### DIFF
--- a/runtime/common/src/main/kotlin/kotlinx/serialization/internal/CollectionSerializers.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/serialization/internal/CollectionSerializers.kt
@@ -127,7 +127,7 @@ class ArrayListSerializer<E>(element: KSerializer<E>) : ListLikeSerializer<E, Li
 class LinkedHashSetSerializer<E>(eSerializer: KSerializer<E>) : ListLikeSerializer<E, Set<E>, LinkedHashSetSerializer.Builder<E>>(eSerializer) {
     override val serialClassDesc = LinkedHashSetClassDesc
 
-    data class Builder<E>(val set: LinkedHashSet<E> = linkedSetOf(), var actualSize: Int = 0) {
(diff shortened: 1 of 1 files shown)


sandwwraith:
In which way do you want to change `ListLikeSerializer`? I'm currently thinking about just removing this check, because it seems redundant for Maps and Sets; and for Arrays `IndexOutOfBoundException` will be raised automatically when index would be out-of-order (except for duplicating ones; however, it's almost impossible to meet a format where arrays serialized as maps `index:value`)

belyaev-mikhail:
Well, I don't quite understand why you have this check in the first place, but I assumed that it is needed for lists somewhere else. I only touched the parts of the library that are related to my usecase. 

Removing the check makes my changes obsolete, but I still don't understand what you need builder size for, besides this check.

sandwwraith:
I think restriction is like a "construction leftover" and may be safely removed. Also, builder size is currently don't needed too, but it turns out to be useful in implementation of so-called "merge" functionality (e.g. for Protobuf arrays) - when you want to write to the end of the array. You can take a look on the prototype here: https://github.com/Kotlin/kotlinx.serialization/blob/1ba0e0034b06f112b4cb9a8a6fb7ebcd72f3e5bf/runtime/common/src/main/kotlin/kotlinx/serialization/internal/CollectionSerializers.kt#L54 to see what I'm trying to achieve

belyaev-mikhail:
So, the questions is, does it make any sense for sets/maps and if it does, what should it be: the number of inserted elements or the real size of the set. If the number of inserted elements matters, my changes make sense. If it doesn't, you should just remove the check and discard my PR.

Btw, I think that my code breaks your `update` function and I need to address this right away.

sandwwraith:
Seems it is a little obsolete and unrelated now, with that size check removed and update functionality added.

PR #84, opened 2018-02-15
Private class serialization support for JVM

xsveda:
Add support for @Serializable classes that are private and live out of kotlinx.serialization package. In such case the Companion field is not visible and must be set accessible before use.
If not an exception is thrown:

``` 
java.lang.IllegalAccessException: Class kotlinx.serialization.SerializationKt can not access a member of class kotlinx.serialization.privateclasstest.DataPrivate with modifiers "public static final"
        at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
        at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
        at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
        at java.lang.reflect.Field.get(Field.java:390)
        at kotlinx.serialization.SerializationKt.invokeSerializerGetter(Serialization.kt:38)
        at kotlinx.serialization.SerializationKt.serializer(Serialization.kt:22)
        at kotlinx.serialization.privateclasstest.PrivateClassOutOfSerializationLibraryPackageTest.testDataPrivate(PrivateDataOutOfKotlinXSerializationPackageTest.kt:40)
```


sandwwraith:
Nice one, thanks!

PR #85, opened 2018-02-15
Fix ProGuard rules docs for serialization 0.4.1 version

xsveda:
Kotlin Serialization 0.4.1 version on JVM assumes that the @Serializable class has a field with name `Companion` that has a method with name `serializer` (see [Serialization.kt](../tree/v0.4.1/runtime/jvm/src/main/kotlin/kotlinx/serialization/Serialization.kt)). To keep these and its names another ProGuard rules must be applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ to add this to your `proguard-rules.pro`:
     *** Companion;
 }
 -keepclasseswithmembers class com.yourcompany.yourpackage.** { # <-- change package name to your app's
-    kotlinx.serialization.KSerializer serializer();
+    kotlinx.serialization.KSerializer serializer(...);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sandwwraith:
This fix is important, but there is on caveat in this solution:

When `@Serializable` class has generic type parameters, its `Companion.serializer` method accepts arguments – to create proper serializers of actual type. E.g. if you have `data class Box<T>(val a: T)`, it's companion object will have method `serializer(serial1: KSerializer<T>): KSerializer<Box<T>>`, and to create a `Box<Int>` serializer, you should call `Box.Companion.serializer(IntSerializer)`. 

Is it possible for proguard to keep all overloaded/with the same name methods?

xsveda:
Good point, I've missed that option. I'll update the rule appropriately...

sandwwraith:
Thank you for that

PR #99, opened 2018-03-15
Support custom container serializer

neworld:
I stuck on such problem because I have custom serializer in the middle of objects tree. I found that compiler plugin is supporting this, but runtime not. I am getting an error:
> kotlinx.serialization.ContextSerializer.<init>(Lkotlin/reflect/KClass;Lkotlinx/serialization/KSerializer;)V

If this solution is OK, I will add some notes to the documentation.


neworld:
I added support for multiple inner serializers. This intended to be work with JS as well, but currently, JS version of the plugin doesn't support it. Generated code (using 1.2.30):
```javascript
    output.writeSerializableElementValue_k4al2t$(this.serialClassDesc, 0, new ContextSerializer(Kotlin.getKClass(CustomContainerSerializersTest$Container)), obj.container);
    output.writeSerializableElementValue_k4al2t$(this.serialClassDesc, 1, new ContextSerializer(Kotlin.getKClass(CustomContainerSerializersTest$KeyValue)), obj.keyValue);
```

Also, I moved from reflection to factory because the reflection of JS support neither `constructors, neither `members`.

PR #102, opened 2018-03-21
Improved performance of JSON.parse/stringify

elizarov:
The most of performance gain is due to moving move of parse state into
local variables instead of fields for the duration of nextToken call and
getting rid of Reader class (now source string is directly parsed).
All the other refactorings (making constants fully static by moving
them to top level and avoiding extra copy of chars for strings that
do not use escapes) have minor effects on parsing of big JSON strings
but could be benefitial in other cases.

### DIFF
--- a/runtime/common/src/main/kotlin/kotlinx/serialization/internal/CommonStrings.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/serialization/internal/CommonStrings.kt
@@ -16,4 +16,10 @@
 
 package kotlinx.serialization.internal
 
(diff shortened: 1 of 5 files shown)


sandwwraith:
@elizarov this request fails one test on JS:
```
not ok 15 kotlinx.serialization > JSONTest > testParseEscapedSymbols
  ---
  message: "Expected <Url(url=https://t.co/M1uhwigsMT)>, actual <Url(url=[h, t, t, p, s, :, /, /, t, ., c, o, /, M, 1, u, h, w, i, g, s, M, T])>."
```

PR #105, opened 2018-03-27
Add null check to Companion when looking up serializer.

drmoose:
This one-byte change eliminates the rather unhelpful error

`TypeError: get_js(...).Companion is undefined`

in favor of the intended

`uncaught exception: SerializationException: Can't locate default serializer for class`

error in the case when one tries to serialize an erased type.

Example triggering problem:

```
interface Foo
@Serializable
data class Bar(var baz: String): Foo

fun tryToSerialize(foo: Foo) = JSON.stringify(foo)

tryToSerialize(Bar("hello"))
```


sandwwraith:
Good point

Issue #95, opened 2018-03-08
Can't locate default serializer for class <...> after upgrading from 0.4 to 0.4.2


Whathecode:
Did anything change to the way serializers for classes are retrieved in 0.4.2 compared to 0.4?

After upgrading to 0.4.2, serialization no longer works:
```
kotlinx.serialization.SerializationException: Can't locate default serializer for class class <...>

	at kotlinx.serialization.SerializationKt.serializer(Serialization.kt:23)
	at <...>.can't deserialize unknown types(StudyProtocolSnapshotTest.kt:130)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:436)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$6(TestMethodTestDescriptor.java:170)
	at org.junit.jupiter.engine.execution.ThrowableCollector.execute(ThrowableCollector.java:40)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:166)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:113)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:58)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor$NodeExecutor.lambda$executeRecursively$3(HierarchicalTestExecutor.java:112)
	at org.junit.platform.engine.support.hierarchical.SingleTestExecutor.executeSafely(SingleTestExecutor.java:66)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor$NodeExecutor.executeRecursively(HierarchicalTestExecutor.java:108)
[...]

PR #130, opened 2018-05-14
Bugfix #95: Can't locate default serializer for classes with named co…

Whathecode:
…mpanion classes

### DIFF
--- a/runtime/jvm/src/main/kotlin/kotlinx/serialization/Serialization.kt
+++ b/runtime/jvm/src/main/kotlin/kotlinx/serialization/Serialization.kt
@@ -35,12 +35,24 @@ actual fun <T: Any, E: T?> ArrayList<E>.toNativeArray(eClass: KClass<T>): Array<
 
 @Suppress("UNCHECKED_CAST")
 internal fun <T> Class<T>.invokeSerializerGetter(vararg args: KSerializer<Any>): KSerializer<T>? {
-    val companion: Any = try {
-        this.getField("Companion").apply { isAccessible = true }.get(null)
-    } catch (e: NoSuchFieldException) {
-        return null
-    }
-    return companion.javaClass.methods
-            .find {it.name == "serializer" && it.parameterTypes.size == args.size && it.parameterTypes.all {it == KSerializer::class.java } }
+    var serializer: KSerializer<T>? = null
+
+    // Search for serializer defined on companion object.
+    val companion = declaredFields.singleOrNull {it.name == "Companion"}?.apply {isAccessible = true}?.get(null)
+    if (companion != null)
+    {
+        serializer = companion.javaClass.methods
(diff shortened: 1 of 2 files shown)


runtime/jvm/src/main/kotlin/kotlinx/serialization/Serialization.kt
```diff
@@ -35,12 +35,24 @@ actual fun <T: Any, E: T?> ArrayList<E>.toNativeArray(eClass: KClass<T>): Array<
...
-        return null
-    }
-    return companion.javaClass.methods
-            .find {it.name == "serializer" && it.parameterTypes.size == args.size && it.parameterTypes.all {it == KSerializer::class.java } }
+    var serializer: KSerializer<T>? = null
+
+    // Search for serializer defined on companion object.
+    val companion = declaredFields.singleOrNull {it.name == "Companion"}?.apply {isAccessible = true}?.get(null)
+    if (companion != null)
+    {
```
sandwwraith: Please format braces around `if` according whole codestyle (and braces around lambdas too)

runtime/jvm/src/main/kotlin/kotlinx/serialization/Serialization.kt
```diff
@@ -35,12 +35,24 @@ actual fun <T: Any, E: T?> ArrayList<E>.toNativeArray(eClass: KClass<T>): Array<
...
-        this.getField("Companion").apply { isAccessible = true }.get(null)
-    } catch (e: NoSuchFieldException) {
-        return null
-    }
-    return companion.javaClass.methods
-            .find {it.name == "serializer" && it.parameterTypes.size == args.size && it.parameterTypes.all {it == KSerializer::class.java } }
+    var serializer: KSerializer<T>? = null
+
+    // Search for serializer defined on companion object.
+    val companion = declaredFields.singleOrNull {it.name == "Companion"}?.apply {isAccessible = true}?.get(null)
```
sandwwraith: I wonder if iterating all fields will not slow down lookup dramatically. @elizarov what do you think?

Whathecode: The alternative would be catching the exception thrown by `getField` to handle this logic, which is very likely to be slower.

Most likely under the covers `getField()` traverses the fields just as well. Doubt this would impact performance very much. I chose `singleOrNull` over `firstOrNull` since it seemed 'unexpected behavior to me in case two companion objects would be defined. But unfortunately I am not familiar with the specification, so this might be a totally wrong presumption.

Whathecode: Apparently you can have only one companion object. I get a warning in the editor when trying to add a second. Oddly enough, [this does not seem documented](https://kotlinlang.org/docs/reference/object-declarations.html), so perhaps my idea there could be several was not too far-fetched.

In case performance _would_ be an issue, the result of this code using reflection (which by itself typically incurs an impact) should likely be cached. Functionality first, optimization second. :)

Issue #49, opened 2017-12-12
Generic Data Structure Serialize / Deserialize


mattinger:
When parsing JSON, it is not always desirable, or required to map the parse result onto a custom data structure.  There are many libraries (gson comes to mind), that allow the user to parse JSON data both to a custom data structure and a generic one.  For instance:

        val input = "{}"
        val parser = JsonParser()
        val root = parser.parse(input)
        val prop = when (root) {
            is JsonObject -> root.get("prop")?.asString
            else -> null
        }

(yes i'm aware this is less than optimal kotlin code, but it illustrates the point, especially if you needed to do something different in the array case)

The current classes in kotlinx.serialization are focused on mapping to custom data structures.

A nice bonus of having this as part of the kotlinx.serialization package would be that it would be cross platform.  Currently, as far as i can tell, there is no cross platform way of parsing JSON data.  As a result, i'm using expect and actual for something that really could be platform agnostic.

elizarov:
+1. We need design a class hierarchy for Json with `JsonObject` and others and support it in our `JSON` and `CBOR` parsers.

czeidler:
Another related use case is to parse json dynamically. For example, consider the following json data structure:
`{ type: "type", data: {...}}`
In this example the data field depends on the type field, i.e. depending on the type I like to deserialize different data classes.

Parsing json data dynamically could work as follows:

```
@Serializable
class Container(val type: String, val data: JsonObject)

val container = JSON.parse(jsonString)

when (container.type) {
    "data1" -> JSON.parse<Data1>(container.data)
    "data2" -> JSON.parse<Data2>(container.data)
}
```

czeidler:
A `JsonObject` would also nice for serialization. Using a `JsonObject` the `KOutput` interface can be simplified to:

```
interface KOutput {
    fun <T> save(object: T): JsonObject
}
```

As far as I can see it is currently not possible to manually serialize a class that cannot be stored into a single variable using the current `KOutput` interface. For example if I want to serialize:
```
class Example(val value1: String = "v1",
              val value2: String = "v2",
              val value3: String = "v3")
```
to
```
{ value1 = "v1", merged = "v2/v3" }
```

This would be trivial to do with the proposed `KOutput` interface.

JakeWharton:
[...]

PR #132, opened 2018-05-18
Json syntax tree, parsers and mappers

sandwwraith:
#49, but not so generic


runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonTreeAndMapperTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        if ("error" in tree) return DummyEither.Left(tree.getAsValue("error").str)
+        return DummyEither.Right(JsonTreeMapper().readTree(tree, Payload.serializer()))
+    }
+
+    override fun save(output: KOutput, obj: DummyEither) {
+        val tree = when (obj) {
+            is DummyEither.Left -> JsonObject(mapOf("error" to JsonString(obj.errorMsg)))
+            is DummyEither.Right -> JsonTreeMapper().writeTree(obj.data, Payload.serializer())
+        }
+        val jsonWriter = output as? JSON.JsonOutput
```
hastebrot: Maybe `jsonWrite` could go to the top of this method to fail fast and mirror that `jsonReader` is also on the top in `fun load()`.

Issue #141, opened 2018-05-23
JSON serializer assumes that the first element it writes is the one at index 0 


pdvrieze:
The JSON serializer hardcodes a check to index 0 whether to write a comma. This assumes that fields are written in order. That is the behaviour of generated code, but may not be the behaviour of custom serializers. It probably also breaks if that first field is an optional field.

PR #142, opened 2018-05-25
Fix issue #141

pdvrieze:
A test and a fix for bug #141, don't let the JSON serializer assume that the first written element always has index 0


pdvrieze:
I can

pdvrieze:
I have added more tests to the file that test various list scenarios with optionals or inverted writing. Lists of optionals, optional lists of optionals or of primitives.
In addition I have ported the CustomSerializersTest class to test on protocol buffers (adding serialid's for all classes) and removing those tests that cannot work (using list/map types at the root-that cannot work due to how protocol buffers work).

Issue #143, opened 2018-05-25
Open up base serializers for extension


Whathecode:
Would it make sense to open base serializers, such as `ReferenceArraySerializer`, for extension? Currently, they are final, so it is not possible to extend from them. However, I find myself in a situation where I would like to write the following:
`object CustomSomeObjectArraySerializer : ReferenceArraySerializer<SomeObject, SomeObject>( SomeObject::class, CustomSomeObjectSerializer )`

The reason is, I want to specify a custom serializer for (polymorph) elements in an array using the `@Serializable` attribute:
```
@Serializable
data class ToSerialize(
    @Serializable( with = CustomSomeObjectArraySerializer::class )
    val someObjectArray: Array<SomeObject> )
```

Of course, I can write a wrapper for `ReferenceArraySerializer` instead:
```
object CustomSomeObjectArraySerializer : KSerializer<Array<SomeObject>>
{
    override val serialClassDesc: KSerialClassDesc
        get() = ArrayClassDesc

    override fun save( output: KOutput, obj: Array<SomeObject> )
    {
        val saver = ReferenceArraySerializer( SomeObject::class, CustomSomeObjectSerializer )
        saver.save( output, obj )
    }

    override fun load( input: KInput ): Array<SomeObject>
    {
        val loader = ReferenceArraySerializer( SomeObject::class, CustomSomeObjectSerializer )
        return loader.load( input )
    }
}
```

This is what I do now, and I can likely also make it generic so that I don't need to implement this for each scenario where I want to specify a custom serializer for arrays. But, I would then have to do something similar for `ArrayListSerializer` and any other existing collection serializers. Instead, would this not be a common enough use case (specifying a custom serializer for items in collections) to warrant opening up these existing serializers for extension?

Maybe I am overlooking a different way of going about what I want to do here. Some further notes for clarification:
- I noticed Kotlin does not allow specifying generics in `SomeObject::class` notations (I believe a JVM limitation), thus it is my understanding extending somewhere using concrete type parameters is needed as I have done above.
- My `CustomSomeObjectSerializer` will become an 'exotic' `PolymorphSerializer` if you will: it will deserialize types known at runtime as it is done in the current `PolymorphSerializer`, but will also support deserializing types unknown at runtime (yet extending from `SomeObject`) by wrapping them in an `UnknownSomeObject` wrapper, exposing some base properties/fields from the base type (`SomeObject`).

Whathecode:
[...]

PR #155, opened 2018-06-19
Open base collection serializers so they can be extended

Whathecode:
I ran into a scenario [where I had to specify a custom serializer for collections which defines a specific serializer to be used for items within the collection]((https://github.com/Kotlin/kotlinx.serialization/issues/143)). For example, in the case of an array:

```
object CustomSomeObjectArraySerializer : ReferenceArraySerializer<SomeObject, SomeObject>( SomeObject::class, CustomSomeObjectSerializer )

@Serializable
data class ToSerialize(
    @Serializable( with = CustomSomeObjectArraySerializer::class )
    val someObjectArray: Array<SomeObject> )
```

Unfortunately, this is not possible since the built-in collection serializers are final. @sandwwraith [commented on the reason for this](https://github.com/Kotlin/kotlinx.serialization/issues/143#issuecomment-393850525):

> Base collection serializers contain a lot of internal machinery, which I believe, should not be open in order not to break invariants.

As a fallback mechanism, I currently wrap the built-in collection serializers using delegation. This works quite well, although it does feel somewhat like introducing unnecessary types (the fact that it is hard to choose an appropriate name for these wrappers is indicative).

In this pull request I investigated what the required changes are to make the built-in collection serializers extendable (open), yet protect the invariants which are currently in place. By making several methods final throughout the type hierarchy this ended up being less dramatic of a change [than I originally anticipated](https://github.com/Kotlin/kotlinx.serialization/issues/143#issuecomment-398379124).

Perhaps making the built-in collection serializers more public/reusable like this is worthwhile? If so, they should likely be moved from the current 'internal' folder which would then be a bit of a misnomer. I only see one reason not to: besides the use case I needed them for (a custom polymorph serializer for items in collections), I can't really imagine other use cases. Although, for a framework, I believe it is not too much to ask to have a mechanism in place to allow intercepting the default behavior. Opening up the built-in serializers (along with some minimal documentation) would make it more transparent how to do so.


runtime/common/src/main/kotlin/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -37,9 +39,7 @@ sealed class ListLikeSerializer<E, C, B>(open val eSerializer: KSerializer<E>) :
     abstract fun B.checkCapacity(size: Int)
     abstract fun B.insert(index: Int, element: E)
 
-    open val typeParams: Array<KSerializer<*>> = arrayOf(eSerializer)
```
Whathecode: There was no need for this to be open. Instead, this is now passed through the constructor.

sandwwraith:
I actually don't see any profit in making classes open with all functions marked as `final override`. 
Why 

`object CustomSomeObjectArraySerializer : ReferenceArraySerializer<SomeObject, SomeObject>( SomeObject::class, CustomSomeObjectSerializer )` 

is more favourable for you than 

`object CustomSomeObjectArraySerializer : KSerializer<Array<SomeObject>> by ReferenceArraySerializer( SomeObject::class, CustomSomeObjectSerializer )`

?

NB: there is no possibility to write `val CustomSomeObjectArraySerializer = ReferenceArraySerializer<SomeObject, SomeObject>( SomeObject::class, CustomSomeObjectSerializer )` for now, but I think it is a good approach too. Probably worth implementing.

Whathecode:
Indeed, the Kotlin delegation syntax does simplify implementing this as composition. But, it is still a bit more unwieldy as you portray it; at least in my use case, I need to specify the types explicitly.
```
object CustomSomeObjectArraySerializer
    : KSerializer<Array<SomeObject>> by ReferenceArraySerializer**<SomeObject, SomeObject>**( SomeObject::class, createSomeObjectSerializer { .... } )
```

But, I presume it all boils down to _semantics_ (and as a result, clarity of the codebase/framework). Surely, composition should be preferred over inheritance in cases where the main argument is mere code reuse and modeling the problem with a 'has a' relationship makes as much sense. However, in this specific case there is a very clear 'is a' relationship and modeling this as a 'has a' relationship sounds contrived.

In fact, the only reason to inherit rather than to instantiate the list serializers directly is a limitation of the language (due to type erasure on the JVM). Ideally, you could write `@Serializable(with=ReferenceArraySerializer<SomeObjectSerializer>::class)` to specify the serializer for a `val someObjects: Array<SomeObject>` property. Or, you could use complicated constructs where you provide all the information necessary to instantiate the list serializer (including the constructor arguments it takes); [something I toyed around with extensively in .NET and called 'attribute metabehavior'](https://whatheco.de/2011/09/22/attribute-metabehavior/).

This would be overkill, of course. :) The point is, by all means this is a 'is a' relationship which is best modeled using inheritance, where the sole use of inheriting is to specify fixed parameters which are passed to the base class. If extending this class poses risks of breaking invariants, this might say more about the underlying class hierarchy than whether or not extending from list serializers makes sense. For example, perhaps `typeParams` was modeled inappropriately and composition _in this case_ did make more sense. I'm not overly familiar with the code base, so you will have to be the final judge. :) Just sharing my thoughts.

sandwwraith:
Serializers now are mostly internal and created via factory functions — to be able to change their internals and methods without affecting public ABI. So this makes little sense now.

PR #174, opened 2018-07-16
Prohibit NaN and infinite values in strict JSON mode

qwwdfsad:


runtime/common/src/main/kotlin/kotlinx/serialization/json/JSON.kt
```diff
@@ -17,15 +17,15 @@
...
-        private val unquoted: Boolean = false,
-        private val indented: Boolean = false,
-        private val indent: String = "    ",
-        internal val nonstrict: Boolean = false,
-        val updateMode: UpdateMode = UpdateMode.OVERWRITE,
-        val context: SerialContext? = null
+    private val unquoted: Boolean = false,
+    private val indented: Boolean = false,
+    private val indent: String = "    ",
+    internal val strictMode: Boolean = true,
```
sandwwraith: Can we afford such breaking change in API?

sandwwraith: There will be at least one supporting release (together with 1.2.60), so seems this should be postponed until core refactoring?

qwwdfsad: Yes, I'm fine with it

sandwwraith:
Corresponding [docs](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/runtime_usage.md#json) should be updated, too

sandwwraith:
Merged manually: https://github.com/Kotlin/kotlinx.serialization/commit/63d223960ff6f80fdb030f484910d45e7f6524e4

PR #176, opened 2018-07-19
Add reproduction test of polymorphic container case

neworld:
JVM (with [runtime patch](https://github.com/neworld/kotlinx.serialization/commit/e7b9ebbbd19c6504fb3fdd4244d08bdb5cbf4258)) and JS tests should succeed in `0.5.0`. Altought I didn't tests in this project, but I have very similar case in my project.

`0.6.0` will fail both JVM and JS. JVM due changed output:
```diff
-{"items":["foo","bar"]}
+{"items":[["kotlinx.serialization.PolymorphicContainer","foo"],["kotlinx.serialization.PolymorphicContainer","bar"]]}
```

I could live with it as long as JS works the same.

JS compiliation is crashing with message:
```
Caused by: java.lang.IllegalArgumentException: Can't locate polymorphic serializer definition
```

### DIFF
--- a/runtime/common/src/test/kotlin/kotlinx/serialization/PolymorphicCaseInsideContainerWithCustomSerializerTest.kt
+++ b/runtime/common/src/test/kotlin/kotlinx/serialization/PolymorphicCaseInsideContainerWithCustomSerializerTest.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


sandwwraith:
Ah, I think I've got it. There was one change in requirements for polymorphic serializer (https://github.com/JetBrains/kotlin/commit/a5301469855214c50073121f917825eda3538d1a): before 0.6.0, it was required only for open classes, and now it is required for open, abstract and sealed (because base sealed class is actually abstract, too – it seems more logical). 

You can workaround this problem by using custom serializer and specifying it on property (`@Serializable(with=...)`. Refer to #143 to find info on how to create list serializer from custom object serializer).

I think in future, implicit usage of polymorphic serializer will be disabled, because it brings a lot of subtle bugs like this one.

neworld:
That makes sense to me. Thanks for the explanation!

Issue #182, opened 2018-07-31
Null nested object causing NoSuchElementException


percula:
When I have a null nested object and mapNullable/unmapNullable it, an exception is thrown. 

Example classes:
```
@Serializable
data class Category(var name: String? = null,
                    var subCategory: SubCategory? = null)

@Serializable
data class SubCategory(var name: String? = null)
```

Code that causes exception:
```
val category = Category(name = "Name")
val map = Mapper.mapNullable(category)
val recreatedCategory = Mapper.unmapNullable(map) <= EXCEPTION
```
Trace:
```
java.util.NoSuchElementException: Key subCategory is missing in the map.

	at kotlin.collections.MapsKt__MapWithDefaultKt.getOrImplicitDefaultNullable(MapWithDefault.kt:24)
	at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:263)
	at kotlinx.serialization.Mapper$InNullableMapper.readTaggedNotNullMark(Tagged.kt:363)
	at kotlinx.serialization.Mapper$InNullableMapper.readTaggedNotNullMark(Tagged.kt:360)
	at kotlinx.serialization.TaggedInput.readNotNullMark(Tagged.kt:226)
	at kotlinx.serialization.KInput.readNullableSerializableValue(Serialization.kt:248)
	at kotlinx.serialization.TaggedInput$readNullableSerializableElementValue$1.invoke(Tagged.kt:265)
	at kotlinx.serialization.TaggedInput.tagBlock(Tagged.kt:278)
	at kotlinx.serialization.TaggedInput.readNullableSerializableElementValue(Tagged.kt:265)
	at com.haystax.constellation.ui.apps.fieldreports.models.Category$$serializer.load(Category.kt)
	at com.haystax.constellation.ui.apps.fieldreports.models.Category$$serializer.load(Category.kt:7)
	at kotlinx.serialization.KInput.read(Serialization.kt:211)
```

I believe this could be fixed by changing Tagged.kt:363 from

```
override fun readTaggedNotNullMark(tag: String): Boolean = map.getValue(tag) != null
```

to 

```
override fun readTaggedNotNullMark(tag: String): Boolean = map.get(tag) != null
```

sandwwraith:
Issue was not reproducible with code you provided (see test in attached commit), but exception occurs when Category looks like `Category(name = "Name", subCategory = SubCategory())`. I've fixed it.

PR #187, opened 2018-08-03
Fix Issue #182

percula:
`getValue()` will throw an exception if the map key does not exist. This causes problems for `@Optional` fields. Changing it to `get()` fixes this issue.


percula:
Upon further testing, Tagged.kt line 361 also needs to be changed to get() but this necessitates a lot changes to accommodate null return types.

PR #200, opened 2018-08-25
Fix namespace spelling mistake

rmaclean:
Fixed a typo in the namespace in the example which could cause confusion.


qwwdfsad:
Thanks!

PR #205, opened 2018-09-02
remove empty primary constructor of some annotations

satoshun:
i think, empty primary constructor is unnecessary. (and IDE raises warning)


sandwwraith:
Indeed, good point. Thank you!

PR #212, opened 2018-09-12
support WildcardType in serializerByTypeToken

NikkyAI:
not sure if the helper function reifiedType belong there in the tests
i assume it could be useful for the library in general

### DIFF
--- a/runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
+++ b/runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
@@ -19,6 +19,7 @@ package kotlinx.serialization
 import kotlinx.serialization.internal.*
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
+import java.lang.reflect.WildcardType
 import kotlin.reflect.KClass
 
 
@@ -67,7 +68,12 @@ fun serializerByTypeToken(type: Type): KSerializer<Any> = when(type) {
     }
     is ParameterizedType -> {
         val rootClass = (type.rawType as Class<*>)
-        val args = (type.actualTypeArguments)
+        val args = (type.actualTypeArguments).map { argument ->
+            when (argument) {
+                is WildcardType -> argument.upperBounds.first() as Class<*>
+                else -> argument
+            }
+        }
         when {
             List::class.java.isAssignableFrom(rootClass) -> ArrayListSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
             Set::class.java.isAssignableFrom(rootClass) -> HashSetSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
--- a/runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
+++ b/runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
@@ -59,7 +59,8 @@ fun <E> serializerBySerialDescClassname(className: String, context: SerialContex
 fun <E> serializerByClass(klass: KClass<*>, context: SerialContext? = null): KSerializer<E> =
     SerialCache.lookupSerializer(mapJavaClassNameToKotlin(klass.java.canonicalName ?: ""), klass, context)
 
-open class TypeBase<T>
+@PublishedApi()
+internal open class TypeBase<T>
 
 inline fun <reified T> typeTokenOf(): Type {
     val base = object : TypeBase<T>() {}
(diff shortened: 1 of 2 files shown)


runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
```diff
@@ -67,7 +68,12 @@ fun serializerByTypeToken(type: Type): KSerializer<Any> = when(type) {
     }
     is ParameterizedType -> {
         val rootClass = (type.rawType as Class<*>)
-        val args = (type.actualTypeArguments)
+        val args = (type.actualTypeArguments).map { argument ->
+            when (argument) {
+                is WildcardType -> argument.upperBounds.first() as Class<*>
```
sandwwraith: Are you sure that this cast will always succeed? can't upper bound be `ParameterizedType` or `WildcardType` too?

NikkyAI: no i am not, added more typechacked and fallback to the uncasted argument, that might fail later, but adding branches should be trivial now

runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
```diff
@@ -67,7 +68,19 @@ fun serializerByTypeToken(type: Type): KSerializer<Any> = when(type) {
     }
     is ParameterizedType -> {
         val rootClass = (type.rawType as Class<*>)
-        val args = (type.actualTypeArguments)
+        val args = (type.actualTypeArguments).map { argument ->
+            when (argument) {
+                is WildcardType -> {
+                    argument.upperBounds.first().let { firstUpperBound ->
+                        when(firstUpperBound) {
+                            is Class<*> -> firstUpperBound
```
sandwwraith: Well, same question. Is it always `Class<*>`? Would this work with `List<List<List<...>>>`, etc? 

I can suggest you an approach to handle WildCard types as a separate clause, not inside ParameterizedType clause. This probably will look like this: `is ParameterizedType -> serializerByTypeToken(type.upperBounds.first())`. Probably with some additional checks around.

NikkyAI: that ended up being the smarter choice indeed
well i reordered it a bit, factored unwrapping of WildcardType into a private extension val
it now supports nested generic containers (list, map), nested arrays and the mix of both
the code to support GenericArray ended up being a bit too ugly for my taste, thanks to java's types

runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
```diff
@@ -47,38 +48,76 @@ private fun mapJavaClassNameToKotlin(s: String): String = when(s) {
...
 
-fun <E> serializerBySerialDescClassname(className: String, context: SerialContext? = null): KSerializer<E> = SerialCache.lookupSerializer(className, context = context)
+fun <E> serializerBySerialDescClassname(className: String, context: SerialContext? = null): KSerializer<E> =
+    SerialCache.lookupSerializer(className, context = context)
+
+fun <E> serializerByClass(klass: KClass<*>, context: SerialContext? = null): KSerializer<E> =
+    SerialCache.lookupSerializer(mapJavaClassNameToKotlin(klass.java.canonicalName ?: ""), klass, context)
 
-fun <E> serializerByClass(klass: KClass<*>, context: SerialContext? = null): KSerializer<E> = SerialCache.lookupSerializer(mapJavaClassNameToKotlin(klass.java.canonicalName ?: ""), klass, context)
+open class TypeBase<T>
```
sandwwraith: I think this declaration could be `internal` instead of `public`, because there is no need for this class outside of this file. However, because it is used in the inline function, you have to mark it as `@PublishedApi` so the compiler will not complain about it. See how it's done in ktor: https://github.com/ktorio/ktor/blob/6c6965c9cafdcc1f6fae2d5635b0ffcf2839afa3/ktor-client/ktor-client-core/ktor-client-core-jvm/src/io/ktor/client/call/TypeInfoJvm.kt#L9

runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
```diff
@@ -47,38 +48,76 @@ private fun mapJavaClassNameToKotlin(s: String): String = when(s) {
...
-    is ParameterizedType -> {
-        val rootClass = (type.rawType as Class<*>)
-        val args = (type.actualTypeArguments)
-        when {
-            List::class.java.isAssignableFrom(rootClass) -> ArrayListSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
-            Set::class.java.isAssignableFrom(rootClass) -> HashSetSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
-            Map::class.java.isAssignableFrom(rootClass) -> HashMapSerializer(serializerByTypeToken(args[0]), serializerByTypeToken(args[1])) as KSerializer<Any>
-            Map.Entry::class.java.isAssignableFrom(rootClass) -> MapEntrySerializer(serializerByTypeToken(args[0]), serializerByTypeToken(args[1])) as KSerializer<Any>
+fun serializerByTypeToken(type: Type): KSerializer<Any> {
+    return when (type) {
```
sandwwraith: Since this function contains a single `return`, it can be transformed to an expression body.

runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
```diff
@@ -47,38 +48,76 @@ private fun mapJavaClassNameToKotlin(s: String): String = when(s) {
...
+                else -> {
+                    val varargs = args.map { serializerByTypeToken(it) }.toTypedArray()
+                    (rootClass.invokeSerializerGetter(*varargs) as? KSerializer<Any>) ?: serializerByClass<Any>(
+                        rootClass.kotlin
+                    )
+                }
             }
         }
+        is WildcardType -> serializerByTypeToken(type.upperBounds.first())
+        else -> throw IllegalArgumentException("typeToken should be instance of Class<?> or ParametrizedType, actual type is $type ${type::class}")
```
sandwwraith: Exception text is a bit obsolete now because you've added more clauses. It can look like `"typeToken should be an instance of Class<?>, GenericArray, ParametrizedType or WildcardType, but actual type is $type ${type::class}"`

runtime/jvm/src/main/kotlin/kotlinx/serialization/Resolving.kt
NikkyAI: had a println there before for debugging purposes, totally forgot about it

sandwwraith:
Thank you for your time!

PR #254, opened 2018-10-30
Expose JsonElement reading/writing from JsonTreeMapper

kris7t:
I wanted to use `JsonElement`-based (de)serialization in a _recursive_ manner, i.e., inspect `JsonElements` when deserializing a field of an object that itself is deserialized using `JsonTreeMapper` (see https://github.com/kris7t/kotlinx.serialization/commit/ef6967b7f295772604c06e80bfbf1b14ce3e9f3c#diff-694566cfe8a0b5cba08e72424def64fdR178 for such a use-case). However, this is currently impossible, because `JsonTreeMapper` does not expose an API similar to `readAsTree` and `writeTree` in `JSON`.

* This patch adds two new interfaces, `JSON.JsonElementReader` and `JSON.JsonElementWriter` to be implemented by classes in both `JSON` and `JsonTreeMapper` to expose common functionality for inspecting `JsonElement`s during (de)serialization.
* To ensure backwards compatibility, a new method `JSON.JsonOutput#writeJsonElement` was added instead of `JSON.JsonOutput#writeTree`. The method `writeJsonElement` now respects indentation and quotation and avoids a sudden "mode switch" when resorting to `JsonElement`-based serialization. To faciliate this, `JsonLiteral#isString` is now `internal` instead of `private`.
* The methods `TaggedEncoder#popTag` and `TaggedDecoder#popTag` were set to `protected` instead of `private` to be able to implement the new interfaces in `JsonTreeMapper`.

### DIFF
--- a/runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonElementReaderWriterTest.kt
+++ b/runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonElementReaderWriterTest.kt
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2018 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kotlinx.serialization.formats.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.context.SimpleModule
+import kotlinx.serialization.internal.SerialClassDescImpl
+import kotlinx.serialization.json.*
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class JsonElementReaderWriterTest {
+    private val inputDataString = """{"id":0,"payload":{"msg":"Hello world","from":42,"to":43},"timestamp":1000}"""
+    private val inputErrorString = """{"id":1,"payload":{"error":"Connection timed out"},"timestamp":1001}"""
+    private val inputDataJson = JsonTreeParser.parse(inputDataString)
+    private val inputErrorJson = JsonTreeParser.parse(inputErrorString)
+    private val inputRecursive = """{"type":"b","children":[{"type":"a","value":1},{"type":"a","value":2},{"type":"b","children":[]}]}"""
+    private val outputRecursive = DummyRecursive.B(
+            listOf(DummyRecursive.A(1), DummyRecursive.A(2), DummyRecursive.B(emptyList()))
+    )
+
+    @Test
+    fun testParseDataString() {
+        val ev = JSON.parse(Event.serializer(), inputDataString)
+        with(ev) {
+            assertEquals(0, id)
+            assertEquals(DummyEither.Right(Payload(42, 43, "Hello world")), payload)
+            assertEquals(1000, timestamp)
+        }
+    }
+
+    @Test
+    fun testParseErrorString() {
+        val ev = JSON.parse(Event.serializer(), inputErrorString)
+        with(ev) {
+            assertEquals(1, id)
+            assertEquals(DummyEither.Left("Connection timed out"), payload)
+            assertEquals(1001, timestamp)
+        }
+    }
+
+    @Test
+    fun testWriteDataString() {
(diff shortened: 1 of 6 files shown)


runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonElementReaderWriterTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        val ev = JsonTreeMapper().writeTree(outputData, Event.serializer())
+        // JsonObject#equals seems to be broken here.
+        assertEquals(inputDataJson.toString(), ev.toString())
+    }
+
+    @Test
+    fun testWriteErrorJson() {
+        val outputError = Event(1, DummyEither.Left("Connection timed out"), 1001)
+        val ev = JsonTreeMapper().writeTree(outputError, Event.serializer())
+        // JsonObject#equals seems to be broken here.
```
sandwwraith: I've fixed this in https://github.com/Kotlin/kotlinx.serialization/commit/6d4016bd8b2f45c3d2167f16055c6cdb59ea28ad, so you probably should reopen this PR against dev.

runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonElementReaderWriterTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        override fun serialize(output: Encoder, obj: DummyRecursive) {
+            if (output !is JSON.JsonElementWriter) throw SerializationException("This class can be saved only by JSON")
+            val mapper = JsonTreeMapper()
+            val (tree, typeName) = when (obj) {
+                is DummyRecursive.A -> mapper.writeTree(obj, DummyRecursive.A.serializer()) to typeNameA
+                is DummyRecursive.B -> mapper.writeTree(obj, DummyRecursive.B.serializer()) to typeNameB
+            }
+            val contents: MutableMap<String, JsonElement> = mutableMapOf(typeAttribute to JsonPrimitive(typeName))
+            contents.putAll(tree.jsonObject.content)
+            val typedTree = tree.jsonObject.copy(content = contents)
```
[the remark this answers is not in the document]
kris7t: I was going by the example code that @tokou kindly provided (btw, I though I thanked for his contribution in the commit message, but seems I forgot, I should absolutely fix that!) after a discussion on the kotlinlang slack: https://gist.github.com/tokou/1660906bd103b0a68f2da283a9c6522c But you're correct, in this case, I could just call the `JsonObject` constructor.

runtime/jvm/src/test/kotlin/kotlinx/serialization/formats/json/JsonElementReaderWriterTest.kt
kris7t: Thanks, will rebase on current `dev` and get rid of this hack.

kris7t:
Thanks for your review! I agree, naming ended up a bit messy.

I propose the following naming:
* Move `JsonElementReader` and `JsonElementWriter` to the `kotlinx.serialization.json` package into a separate file.
* Rename `JsonElementReader#readAsTree` to `decodeJson`. This would introduce a breaking change in `JSON.JsonInput`, so its `readAsTree` method has to stay, but can delegate to `decodeElement`.
* Rename `JsonElementWriter#writeJsonElement` to `encodeJson`. Here, we don't have any method to keep as a delegate, because `JSON.JsonOutput#writeTree` is already separate due to different indentation handling (and messing with indentation can break downstream tests, the very least).

Will get back with a fresh PR when I have some more time to work on this.

sandwwraith:
Good idea. `readAsTree` can be marked as `@Deprecated` for further removal, as in Deprecated.kt.
I think method `writeTree` should be considered erroneous (because non-respecting indentation is not what we want from proper JSON implementation) and can be removed, because a more correct implementation, encodeJson, is present. 

Also, `ElementReader/ElementWriter` in interfaces names can be replaced with `Decoder/Encoder`. It may look more consistent if we cast `Encoder` object to something that ends with 'encoder', e.g. `JsonEncoder`.

sandwwraith:
@kris7t Do you have any plans on this? We are going to merge #271 soon, so it will likely have conflicts. If you don't have free time and don't mind, I can cherry-pick your commits manually and fix them later.

kris7t:
Sorry, I didn't have the time to get back a this -- mostly I had to spend my time supervising students at uni, so I didn't have any for the project that involves working with weird polymorphic JSON. :) If you want, you can cherry-pick my commits, but I'd also be willing to rebase my commits on `dev` once you merge #271 and clean them up myself.

kris7t:
I see #276 added this functionality in a much more elegant way, so I guess this PR can be closed then.

sandwwraith:
@kris7t Despite this functionality indeed was already merged, your tests helped us a lot for validating it. I've included them in #294.

sandwwraith:
#294 merged

PR #268, opened 2018-11-20
Common polymorphic serialization for JVM, JS and Native

sandwwraith:
* Polymorphic serializer moved to common and rewritten
* New functions to register polymorphic subtypes in the serial context
* Several new modules


sandwwraith:
@pdvrieze They're required to build the project on a CI server. It seems that dependency on a snapshot compiler build and regular updates of it in the `dev` branch is generally a good idea which helps to move faster and to contribute easier. `master` branch, of course, will release with stable versions.

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
```diff
@@ -3,45 +3,175 @@
...
+     */
     operator fun <T: Any> get(kclass: KClass<T>): KSerializer<T>?
 
+    /**
+     * Returns a serializer associated with KClass which given [value] has.
+     *
+     * This method is used in context-sensitive operations
+     * on a property marked with [ContextualSerialization], by a [ContextSerializer]
+     */
     fun <T: Any> getByValue(value: T): KSerializer<T>?
```
qwwdfsad: should it really be part of the interface which everybody has to implement?
It can be used in-place via `get(value::class)`, why expose it in the base interface?

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
```diff
@@ -3,45 +3,175 @@
...
-        val klass = t::class
+        val klass = value::class
         return get(klass) as? KSerializer<T>
     }
 
     override fun <T: Any> get(kclass: KClass<T>): KSerializer<T>? = classMap[kclass] as? KSerializer<T> ?: parentContext?.get(kclass)
 }
 
+
+internal object StandardSubtypesOfAny {
```
qwwdfsad: A lot of classes for a single file, please split.
Contexts have a separate package, so it shouldn't be a big deal, but users will see only "public" documented API without undocumented implementation.

qwwdfsad: Currently, this is a common approach for `kotlinx.serialization`, but one public entity <-> one file is much easier to understand and separate implementation from declarations

runtime/common/src/main/kotlin/kotlinx/serialization/context/SerialModules.kt
```diff
@@ -0,0 +1,168 @@
...
+                k as KClass<Base>,
+                s as KSerializer<Base>
+            )
+        }
+    }
+
+    /**
+     * @see PolymorphicModule.unaryPlus
+     */
+    public fun <T : Base> addSubclass(subclass: KClass<T>, serializer: KSerializer<T>) {
```
qwwdfsad: add -> register to keep it consistent

qwwdfsad: Also, it is worth to add reified version of `addSubclass` with `ImplicitReflectionSerializer`

runtime/common/src/main/kotlin/kotlinx/serialization/context/SerialModules.kt
```diff
@@ -0,0 +1,168 @@
...
+     */
+    fun registerIn(context: MutableSerialContext)
+}
+
+/**
+ * A [SerialModule] which registers one class with one serializer for [ContextSerializer].
+ *
+ * @see MutableSerialContext.registerSerializer
+ */
+class SimpleModule<T: Any>(val kClass: KClass<T>, val kSerializer: KSerializer<T>): SerialModule {
```
qwwdfsad: `Simple` is a polluted word. Consider e.g. `SingletonModule`.

Its usage looks a bit boilerplate:
```
Json().apply { install( SimpleModule(Payload::class, PayloadSerializer)) }
```

what if we introduce extensions on serial format such as
```
Json().apply { install(Payload::class, PayloadSerializer) } // 1
Json().register(Payload::class, PayloadSerializer) // 2
```

If we could provide a reified version, even better: `Json().register(PayloadSerializer)`

runtime/common/src/main/kotlin/kotlinx/serialization/context/SerialModules.kt
```diff
@@ -0,0 +1,168 @@
...
+    override fun registerIn(context: MutableSerialContext) {
+        context.registerSerializer(kClass, kSerializer)
+    }
+}
+
+/**
+ * A [SerialModule] which registers multiple classes with its serializers for [ContextSerializer].
+ */
+@Suppress("UNCHECKED_CAST")
+class MapModule(val map: Map<KClass<*>, KSerializer<*>>): SerialModule {
```
qwwdfsad: unused.
Should be tester or removed

runtime/common/src/main/kotlin/kotlinx/serialization/context/SerialModules.kt
sandwwraith: While we register serializers in _contexts_, this is a _module_, which is defined as a 'collection of serializers', and `add` is a natural verb for collection. It also emphasises that actual registering in some context would only be performed after module installation.

runtime/common/src/main/kotlin/kotlinx/serialization/CommonPolymorphic.kt
```diff
@@ -0,0 +1,115 @@
...
+ * Yet PolymorphicSerializer for request should only allow RequestA and RequestB serializers, and none of the response's serializers.
+ * This is achieved via [MutableSerialContext.registerPolymorphicSerializer] function, which accepts two KClass references.
+ *
+ * By default (without special support from [Encoder]), polymorphic values are serialized as list with
+ * two elements: classname (String) and the object itself.
+ *
+ * @see MutableSerialContext.registerPolymorphicSerializer
+ */
+@Suppress("UNCHECKED_CAST")
+class PolymorphicSerializer<T : Any>(private val basePolyType: KClass<T>) : KSerializer<Any> {
```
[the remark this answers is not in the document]
sandwwraith: I think that `polymorphicType` loses a meaning that it is a common supertype. `baseType` is better, but I still think it misses the crucial part connected with 'polymorphic hierarchy'. I think that `basePolymorphicType` is ok, if it's not too long. 'basePolyType` is also used in KEEP with this meaning.

runtime/common/src/main/kotlin/kotlinx/serialization/context/SerialModules.kt
qwwdfsad: >WDYT about passing SerialModule as a constructor parameter in all serial formats?

 I am fine with it

runtime/common/src/test/kotlin/kotlinx/serialization/features/ContextAndPolymorphicTest.kt
```diff
@@ -49,54 +47,59 @@ class CustomSerializersJvmTest {
...
     private val obj = EnhancedData(Data(100500), Payload("string"), Payload("binary"))
     private lateinit var json: Json
 
-    @Before
+    @BeforeTest
     fun initContext() {
-        val scope = SimpleModule(Payload::class, PayloadSerializer)
-        json = Json(unquoted = true).apply { install(scope) }
+        val scope = SingletonModule(Payload::class, PayloadSerializer)
+        val bPolymorphicModule = object : SerialModule {
```
qwwdfsad: This pattern is already frequent in tests.
It would be nice to have something like
`fun PolymorphicModule(baseClass, actualClass, serializer)`, maybe with reified overloads

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
```diff
@@ -41,7 +57,45 @@ fun <T: Any> SerialContext?.getOrDefault(klass: KClass<T>) = this?.let { get(kla
...
+     * Registers a [concreteSerializer] for a [concreteClass] in a scope of [basePolyType].
+     *
+     * Given [concreteSerializer] will be used in polymorphic serialization
+     * if statically known type of property is [basePolyType] and runtime type is a [concreteClass].
+     *
+     * For serialization process, classname is determined using name of a serializer descriptor.
+     *
+     * All standard types and collections are registered in advance as a subtypes in a scope of [kotlin.Any].
+     */
+    fun <Base : Any, Sub : Base> registerPolymorphicSerializer(
```
qwwdfsad: also, it is a good idea to throw an exception if users try to register multiple serializers for the same class

runtime/common/src/main/kotlin/kotlinx/serialization/PlatformUtils.kt
```diff
@@ -40,3 +40,17 @@ expect fun <E: Enum<E>> KClass<E>.enumClassName(): String
...
+ * This check is a replacement for [KClass.isInstance] because
+ * on JVM it requires kotlin-reflect.jar in classpath
+ * (see https://youtrack.jetbrains.com/issue/KT-14720).
+ *
+ * On JS and Native, this function delegates to aforementioned
+ * [KClass.isInstance] since it is supported there out-of-the box;
+ * on JVM, it falls back to java.lang.Class.isInstance which causes
+ * difference when applied to function types with big arity.
+ */
+internal expect fun isInstance(kclass: KClass<*>, obj: Any): Boolean
```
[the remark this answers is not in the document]
sandwwraith: Wouldn't it bloat the completion since it is extension on `Any`?

runtime/common/src/main/kotlin/kotlinx/serialization/context/CompositeModule.kt
```diff
@@ -0,0 +1,49 @@
...
+    constructor(vararg modules: SerialModule) : this(modules.toList())
+
+    private val modules: MutableList<SerialModule> = modules.toMutableList()
+
+    override fun registerIn(context: MutableSerialContext) {
+        modules.forEach { it.registerIn(context) }
+    }
+
+    public operator fun plusAssign(module: SerialModule): Unit { modules += module }
+    public fun addModule(module: SerialModule) = plusAssign(module)
```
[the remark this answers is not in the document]
sandwwraith: Do we have any convention about functions which alias operator functions? Should we add them or, in contrary, avoid?

runtime/native/src/test/kotlin/kotlinx/serialization/features/ContextAndPolymorphic.kt
```diff
@@ -0,0 +1,98 @@
+package kotlinx.serialization.features
```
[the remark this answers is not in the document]
sandwwraith: I don't think that this is a json test since it mainly tests context features. Json is used here just because this is a simple text format for which 'gold' data can be provided

runtime/common/src/test/kotlin/kotlinx/serialization/features/ContextAndPolymorphicTest.kt
sandwwraith: I think the approach in https://github.com/Kotlin/kotlinx.serialization/pull/268/files/cfce6687f9828d0d7425f461d338199198e6b565?utf8=%E2%9C%93&diff=unified#diff-f4c790c741d52624028aea493fb2a532R49 is quite short and I can rewrite this place to use it. Usually, people want to register more than one subclass serializer.

runtime/common/src/main/kotlin/kotlinx/serialization/PlatformUtils.kt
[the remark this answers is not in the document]
qwwdfsad: It's internal, so it's not an issue

runtime/common/src/main/kotlin/kotlinx/serialization/context/CompositeModule.kt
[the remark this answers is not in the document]
qwwdfsad: I am fine with it, `addModule` is a good alternative for ones who don't like overloaded operators.

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
```diff
@@ -5,34 +5,50 @@ package kotlinx.serialization.context
 import kotlinx.serialization.*
 import kotlin.reflect.KClass
 
+/**
+ * Serial context is a runtime mechanism used by [ContextSerializer] and [PolymorphicSerializer]
+ * to obtain serializers which were not found at compile-time by the serialization plugin.
```
qwwdfsad: >to obtain serializers which were not found at compile-time

This statement is misleading.
Not to obtain missing serializers, but to override ones or to provide them in the runtime.
It's not a fallback, it's opt-in mechanism

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
qwwdfsad: A good default will be to throw an exception by default but provide an additional argument `overwriteExisting` (with proper naming)  and note this parameter in the message of the exception.

WDYT?

runtime/common/src/main/kotlin/kotlinx/serialization/context/MutableSerialContextImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+
+package kotlinx.serialization.context
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.isInstance
+import kotlin.reflect.KClass
+
+private typealias SerializersMap = MutableMap<KClass<*>, KSerializer<*>>
+
+class MutableSerialContextImpl(private val parentContext: SerialContext? = null):
```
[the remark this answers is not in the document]
qwwdfsad: It's better to hide implementation details, especially when they are not meaningful. 
E.g. take a look at how channels are created in kx.coroutines:
```
interface Channel
// Factory that looks like constructor
fun Channel(): Channel = ChannelImpl()

private class ChannelImpl() // <- this class now can be changed in any way while it complies Channel contract
```

runtime/common/src/main/kotlin/kotlinx/serialization/context/Context.kt
sandwwraith: Sounds ok, but should also add it to usual `registerSerializer` then

sandwwraith: And since clients of these functions are modules, we should somehow propagate this parameter to all default modules (`SingletonModule`, `MapModule`, `PolymorphicModule`, etc). I wonder how far this can go. Probably we should leave this to design then.

sandwwraith:
I think that somewhere in the observable future we can re-locate serializers in such way so most of the standard ones would live in the `kotlinx.serialization.builtins` package. However, since these two are indeed special, I can move them to separate package, if you insist.

sandwwraith:
I'll now open a separate PR since the concept of modules was changed significantly and this conversation is already bloated with fixed issues

sandwwraith:
#394

PR #270, opened 2018-11-20
Make naming consistent with KEEP

qwwdfsad:


gradle.properties
```diff
@@ -25,5 +25,5 @@ konan.version.snapshot=1.3-SNAPSHOT-native-1.1.0-dev
 
 native.deploy=false
 
-org.gradle.parallel=true
-org.gradle.caching=true
+org.gradle.parallel=false
```
sandwwraith: Roll back accidentally committed local settings

sandwwraith:
Merged manually into `dev`.

PR #275, opened 2018-11-22
Remove ValueTransformer

### LINKED ISSUE #272: Remove Mapper and ValueTransformer
These classes were introduced in order to show capabilities of `kotlinx.serialization`, but carry no value for core runtime library, adding an additional maintenance burden. They are neither documented nor well tested.

We can always add it later to advanced examples (same as hashing).

qwwdfsad:
Fixes #272 

`Mapper` will be here for a while, I will cover it with tests later


sandwwraith:
This class is also used in the samples: https://github.com/Kotlin/kotlinx.serialization/blob/master/example-jvm/src/DataTransformDemo.kt#L11 . We need to either move it there or remove it from there.

qwwdfsad:
Let's remove it from samples as well?

sandwwraith:
Yeah, ok

PR #276, opened 2018-11-23
Unify json

qwwdfsad:


qwwdfsad:
Note: check TODO before merging

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -0,0 +1,183 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.json.*
+
+internal fun <T> Json.readJson(element: JsonElement, deserializer: DeserializationStrategy<T>): T {
+    val descriptor = deserializer.descriptor
+    if (element is JsonNull) {
+        // TODO temporary workaround (?)
+        require(descriptor.isNullable) { "Read JsonNull and expected nullable descriptor, but has $descriptor" }
+        @Suppress("NULL_FOR_NONNULL_TYPE")
```
sandwwraith: Are you sure that this suppress can't trigger e.g. SEGFAULT on Native or something undesirable on other platforms?

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -0,0 +1,183 @@
...
+    : NamedValueDecoder(), JsonInput {
+
+    init {
+        context = json.context
+    }
+
+    override val updateMode: UpdateMode
+        get() = json.updateMode
+
+    override fun readTree(): JsonElement {
```
sandwwraith: It looks to me like this function should always either `return obj` or be equivalent to `currentElement()`

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/TreeJsonOutput.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    override fun encodeTaggedNull(tag: String) = putElement(tag, JsonNull)
+
+    override fun encodeTaggedInt(tag: String, value: Int) = putElement(tag, JsonLiteral(value))
+    override fun encodeTaggedByte(tag: String, value: Byte) = putElement(tag, JsonLiteral(value))
+    override fun encodeTaggedShort(tag: String, value: Short) = putElement(tag, JsonLiteral(value))
+    override fun encodeTaggedLong(tag: String, value: Long) = putElement(tag, JsonLiteral(value))
+
+    override fun encodeTaggedFloat(tag: String, value: Float) {
+        if (json.strictMode && !value.isFinite()) {
```
sandwwraith: This check probably could be leveraged to somewhere around `JsonOutput`, since it is used in regular json, too

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+}
+
+private object JsonObjectSerializer : KSerializer<JsonObject> {
+
+    override val descriptor: SerialDescriptor = JsonObjectDescriptor
+
+    override fun serialize(encoder: Encoder, obj: JsonObject) {
+        val composite = encoder.beginCollection(descriptor, obj.size, StringSerializer, JsonElementSerializer)
```
sandwwraith: I assume you can use delegation to MapSerializer for serialize, deserialize and descriptor, like: 

```kotlin
override fun serialize(encoder: Encoder, obj: JsonObject) {
    LinkedHashMapSerializer(StringSerializer, JsonElementSerializer).serialize(obj.content)
}
```

And use delegation to `LinkedHashMapClassDesc` to override descriptor name.

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -0,0 +1,137 @@
...
+}
+
+private object JsonPrimitiveSerializer : KSerializer<JsonPrimitive> {
+
+    override fun serialize(encoder: Encoder, obj: JsonPrimitive) {
+        if (obj is JsonNull) {
+            return encoder.encodeString(obj.content)
+        }
+
+        val integer = obj.intOrNull
```
sandwwraith: Hm, i think we need also to check `isString` first here, because we want `JsonPrimitive("42", isString=true)` be saved as `"42"`, not `42`.

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        init {
+            addElement("JsonPrimitive")
+        }
+    }
+}
+
+private object JsonArraySerializer : KSerializer<JsonArray> {
+
+    override fun serialize(encoder: Encoder, obj: JsonArray) {
```
sandwwraith: Same here as for `JsonObject`, you could make use of `ArrayListSerializer`.

runtime/common/src/test/kotlin/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -0,0 +1,94 @@
...
+            stringify(JsonElementSerializer, tree)
+        }
+    }
+
+    @ImplicitReflectionSerializer
+    inline fun <reified T : Any> Json.stringify(list: List<T>, useStreaming: Boolean): String {
+        return if (useStreaming) {
+            stringify(list)
+        } else {
+            stringify(context.getOrDefault(T::class).list, list)
```
sandwwraith: Looks strange to me. Why switch streaming/non-streaming should wrap in list additionally? Did you want `JsonArraySerializer` or `JsonElementSerializer.list`?

runtime/common/src/test/kotlin/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            parse((context.getOrDefault(K::class) to context.getOrDefault(V::class)).map, content, useStreaming)
+        }
+    }
+
+    protected fun parametrizedTest(test: (Boolean) -> Unit) {
+        val streamingResult = kotlin.runCatching { test(true) }
+        val treeResult = kotlin.runCatching { test(false) }
+        assertEquals(streamingResult, treeResult)
+        streamingResult.getOrThrow() }
```
sandwwraith: brace formatting. I assume we can return from function value which was returned by `getOrThrow`?

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -28,14 +28,13 @@ object JsonElementSerializer : KSerializer<JsonElement> {
             get() = UnionKind.SEALED
 
         init {
-            addElement("JsonObject")
-            addElement("JsonLiteral")
-            addElement("JsonArray")
+            addElement("JsonElement")
```
sandwwraith: Why so?

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -68,13 +67,68 @@ private object JsonObjectSerializer : KSerializer<JsonObject> {
...
+        return if (obj is JsonNull) {
+            JsonNullSerializer.serialize(encoder, JsonNull)
+        } else {
+            JsonLiteralSerializer.serialize(encoder, obj as JsonLiteral)
+        }
+    }
+
+    override fun deserialize(decoder: Decoder): JsonPrimitive {
+        val nullable = decoder.decodeNullable(StringSerializer)
+        return if (nullable == null) JsonNull else JsonPrimitive(nullable)
```
sandwwraith: Hm, I came up with a better idea: we have `decodeNotNullMark` which says if next element is null. You can check it and call `JsonNullSerializer.deserialize(decoder)` or `JsonLiteral.deserialize(decoder)`.

runtime/common/src/test/kotlin/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -15,6 +15,22 @@ abstract class JsonTestBase {
...
     protected val nonstrict = Json(strictMode = false)
 
+    init {
+        strict.register()
+        unquoted.register()
+        nonstrict.register()
+    }
+
+    private fun Json.register() {
+        val mutableCtx = (context as MutableSerialContext)
```
sandwwraith: This is an implementation detail. Are you using it intentionally? Otherwise, please use `install` (maybe some shorthand builder worth implementing and adding to the library)

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
```diff
@@ -93,23 +147,25 @@ private object JsonPrimitiveSerializer : KSerializer<JsonPrimitive> {
         encoder.encodeString(obj.content)
     }
 
-    override fun deserialize(decoder: Decoder): JsonPrimitive {
-        return JsonPrimitive(decoder.decodeString())
+    override fun deserialize(decoder: Decoder): JsonLiteral {
```
sandwwraith: Todo: how we can deserialize ints, floats and so on with preserving info about it?

runtime/common/src/test/kotlin/kotlinx/serialization/json/JsonTestBase.kt
qwwdfsad: No, I wanted particular public extensions (ones from `SerialImplicits`) to be tested.
I will leave a comment here

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonInput.kt
```diff
@@ -1,119 +1,10 @@
...
-public class JsonInput internal constructor(private val json: Json, private val mode: WriteMode,
-                                            private val parser: JsonParser) : ElementValueDecoder() {
-    private var curIndex = -1
-    private var entryIndex = 0
+public interface JsonInput : Decoder, CompositeDecoder {
 
-    init {
-        context = json.context
-    }
+    public val json: Json
```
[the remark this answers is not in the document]
qwwdfsad: See `EitherSerializer`, it is required to call parse/stringify on it

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonTreeParser.kt
```diff
@@ -1,85 +1,13 @@
...
- * limitations under the License.
+ * Copyright 2017-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization.json
 
-import kotlinx.serialization.json.internal.*
-
-class JsonTreeParser internal constructor(private val parser: JsonParser) {
+@Deprecated(level = DeprecationLevel.ERROR, message = "Use Json.parseJson instead")
```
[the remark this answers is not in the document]
qwwdfsad: To avoid a long deprecation cycle.
The compilation will break, replacement is clear, no need to introduce constructor or `read` method.

Can return it back if you insist.

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
qwwdfsad: Not in this API shape.
We can't distinguish them because we are not doing so on the type system level.

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/TreeJsonInput.kt
qwwdfsad: Yes, I have tests for it.
It will be fixed around #287

runtime/common/src/test/kotlin/kotlinx/serialization/json/JsonTestBase.kt
sandwwraith: Just an idea. We can leave it as is for now.

runtime/common/src/main/kotlin/kotlinx/serialization/json/serializers/JsonElementSerializer.kt
qwwdfsad: Yes

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonElementSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+    override fun serialize(encoder: Encoder, obj: JsonElement) {
+        when (obj) {
+            is JsonPrimitive -> JsonPrimitiveSerializer.serialize(encoder, obj)
+            is JsonObject -> JsonObjectSerializer.serialize(encoder, obj)
+            is JsonArray -> JsonArraySerializer.serialize(encoder, obj)
+        }
+    }
+
+    override fun deserialize(decoder: Decoder): JsonElement {
+        val input = decoder as? JsonInput ?: error("JsonElement is serializable only when used by Json")
```
sandwwraith: maybe better correct this error message to 'deserializable only'? Because it can be serialized

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -1,27 +1,54 @@
...
+ * Is is typically used by constructing an application-specific instance, registering
+ * custom serializers via [Json.install] and then using it either as regular [SerialFormat] or [StringFormat]
+ * or for converting objects to [JsonElement] back and forth.
+ *
+ * This is the only serial format which has first-class [JsonElement] support.
+ * Any serializable class can be serialized to or from [JsonElement] with [Json.fromJson] and [Json.toJson] respectively or
+ * serialize properties of [JsonElement] type.
+ *
+ * Json configuration parameters:
+ * [unquoted] specifies keys and values should be quoted, used mostly for testing.
```
sandwwraith: "specifies whether keys and..." ?

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -1,27 +1,54 @@
...
+ * [indented] specifies whether resulting JSON should be pretty-printed.
+ * [indent] specifies which indent string to use with [indented] mode.
+ * [strictMode] enables strict mode, which prohibits unknown keys and infinite values in floating point numbers.
+ *
+ * Example of usage:
+ * ```
+ * @Serializable
+ * class DataHolder(val id: Int, val data: String, val extensions: JsonElement)
```
sandwwraith: All samples here use `@ImplicitReflectionSerializer` functionality, which is discouraged in general. For this concrete sample it is probably ok (in order not to repeat `DataHolder.serializer()` everywhere), but better add a note about it.

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonBuilder.kt
```diff
@@ -1,102 +0,0 @@
-/*
```
sandwwraith: Bring back accidentally renamed `JsonObjectBuilder.kt` since it also contains `JsonArrayBuilder`

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonTreeMapper.kt
```diff
@@ -1,249 +0,0 @@
...
-import kotlinx.serialization.CompositeDecoder.Companion.READ_DONE
-import kotlinx.serialization.context.*
-import kotlinx.serialization.internal.EnumDescriptor
-
-@Suppress("USELESS_CAST") // contracts not working in Native
-class JsonTreeMapper(val encodeDefaults: Boolean = true): AbstractSerialFormat() {
-    @ImplicitReflectionSerializer
-    inline fun <reified T : Any> readTree(tree: JsonElement): T = readTree(tree, context.getOrDefault(T::class))
-
-    fun <T> readTree(obj: JsonElement, deserializer: DeserializationStrategy<T>): T {
```
sandwwraith: Better leave these methods with deprecation

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -41,14 +92,60 @@ class Json(
...
+    }
+
+    /**
+     * Deserializes [json] element into a corresponding object of type [T] using serializer registered in the context.
+     * @throws [JsonException] subclass in case of serialization error.
+     */
+    @ImplicitReflectionSerializer
+    public inline fun <reified T : Any> fromJson(tree: JsonElement): T = fromJson(tree, context.getOrDefault(T::class))
+
+    @Deprecated(level = DeprecationLevel.WARNING, message = "Global Json is deprecated without replacement as it is a stateful object. " +
```
sandwwraith: We can deprecate it if we won't find good solution with mutable modules

PR #299, opened 2018-12-07
Longs in DynamicParser for JS and LongAsStringSerializer

sandwwraith:


runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -56,6 +60,18 @@ class DynamicObjectParser(): AbstractSerialFormat() {
...
 
+        override fun decodeTaggedLong(tag: String): Long {
+            val obj = getByTag(tag)
+            val number = obj as? Double ?: throw SerializationException("$obj is not a Number")
+            val canBeConverted = number.isFinite() && floor(number) == number
+            if (!canBeConverted)
+                throw SerializationException("$number can't be represented as Long")
+            val inBound = abs(number) <= MAX_SAFE_INTEGER
+            if (!inBound)
+                throw SerializationException("$number does not fit in Long due to a potential precision loss")
```
anton-bannykh: Maybe "does not fit" => "could not be deserialized"? Otherwise the message might be confusing...

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -17,8 +17,12 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.CompositeDecoder.Companion.READ_DONE
-import kotlinx.serialization.context.*
+import kotlinx.serialization.context.getOrDefault
 import kotlinx.serialization.internal.EnumDescriptor
+import kotlin.math.abs
+import kotlin.math.floor
+
+internal const val MAX_SAFE_INTEGER: Double = 9007199254740991.toDouble() // 2^53 - 1
```
anton-bannykh: Suggestion: add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -56,6 +60,18 @@ class DynamicObjectParser(): AbstractSerialFormat() {
             }
         }
 
+        override fun decodeTaggedLong(tag: String): Long {
+            val obj = getByTag(tag)
+            val number = obj as? Double ?: throw SerializationException("$obj is not a Number")
+            val canBeConverted = number.isFinite() && floor(number) == number
```
anton-bannykh: FYI there are built-in [`Number.isInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) and [Number.isSafeInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger)
Just thought you might find them useful =)
Current implementation has the benefit of being supported on IE though... So your choice =)

runtime/common/src/test/kotlin/kotlinx/serialization/features/LongAsStringTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        val str = strict.stringify(HasLong.serializer(), original, useStreaming)
+        assertEquals("""{"l":"9223372036854775806"}""", str)
+        val restored = strict.parse(HasLong.serializer(), str, useStreaming)
+        assertEquals(original, restored)
+    }
+
+    @Test
+    fun canNotDeserializeInvalidString() = parametrizedTest { useStreaming ->
+        val str = """{"l": "this is definitely not a long"}"""
+        assertFailsWith<NumberFormatException> { strict.parse(HasLong.serializer(), str, useStreaming) }
```
anton-bannykh: Minor suggestion: add a test for number literal, which doesn't fit into a Long, e.g. `1000000000000000000000` Just to make sure the exception makes sense (and that it is actually raised).

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -56,6 +60,18 @@ class DynamicObjectParser(): AbstractSerialFormat() {
             }
         }
 
+        override fun decodeTaggedLong(tag: String): Long {
+            val obj = getByTag(tag)
+            val number = obj as? Double ?: throw SerializationException("$obj is not a Number")
```
anton-bannykh: Tests for the `throw` case seem to be missing?

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -56,6 +60,18 @@ class DynamicObjectParser(): AbstractSerialFormat() {
...
+            val canBeConverted = number.isFinite() && floor(number) == number
+            if (!canBeConverted)
+                throw SerializationException("$number can't be represented as Long")
```
anton-bannykh: Maybe "$number is not an integer"?

runtime/js/src/test/kotlin/kotlinx/serialization/DynamicToLongTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        test(js("{l:-1}"), Result.success(-1))
+    }
+
+    @Test
+    fun ignoresIncorrectValues() {
+        shouldFail(js("{l:0.5}"))
+        shouldFail(js("{l: Math.PI}"))
+        shouldFail(js("{l:Infinity}"))
+        shouldFail(js("{l:+Infinity}"))
+        shouldFail(js("{l:-Infinity}"))
```
anton-bannykh: Probably `NaN` is worth checking as well. Pretty sure it is handled correctly, but still...

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
sandwwraith: Yes, I've looked them up in docs before writing this. I wanted to distinguish cases 'not a number' from 'not an integer' from 'not a safe int' without a lot of `js(...)` code blocks – all these methods are not exposed from `kotlin.Number`.

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
sandwwraith: I think it might confuse people with `kotlin.Int`?

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
anton-bannykh: Ok)

runtime/js/src/main/kotlin/kotlinx/serialization/DynamicObjectParser.kt
anton-bannykh: Maybe "$number is not finite, or has a non-zero fractional part"?

PR #302, opened 2018-12-10
Delegate `JsonArray.equals` to the wrapped list

### LINKED ISSUE #301: JsonArray doesn't equals-compare with other lists correctly.
As far as I can tell, the (implicit?) convention for Kotlin collections is that different instances of `List` (or `Set`) should be equal if the (sequence of) elements is equal -- no matter what the concrete type is.

For example, this test (added to `kotlinx/serialization/json/JsonTreeTest.kt`) is green:

```kotlin
@Test
fun equalCollections() {
    val hashSet: Set<JsonElement> = HashSet(listOf(JsonLiteral(3), JsonLiteral(4)))
    val linkedHashSet: Set<JsonElement> = LinkedHashSet(listOf(JsonLiteral(3), JsonLiteral(4)))
    assertEquals(hashSet, linkedHashSet)
}
```

However, comparing a `JsonArray` -- which implements `List<JsonElement>` -- with another `List` we get a red test:

```kotlin
val jsonList: List<JsonElement> = JsonArray(listOf(JsonLiteral(3), JsonLiteral(4)))
val arrayList: List<JsonElement> = ArrayList(listOf(JsonLiteral(3), JsonLiteral(4)))
assertEquals(jsonList, arrayList)
```

reitzig:
Fixes #301

### DIFF
--- a/runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
@@ -341,6 +341,9 @@ data class JsonArray(val content: List<JsonElement>) : JsonElement(), List<JsonE
     inline fun <reified J : JsonElement> getAsOrNull(index: Int): J? = content.getOrNull(index) as? J
 
     override fun toString() = content.joinToString(prefix = "[", postfix = "]")
+    
+    override fun equals(other: Any?): Boolean =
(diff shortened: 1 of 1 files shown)


sandwwraith:
Could you add some tests for this?

reitzig:
Will do. Wanted to check whether this matches the intent first.

reitzig:
--> #305

PR #305, opened 2018-12-11
Delegates `equals`, `hashCode` of `JsonObject` and `JsonArray`.

reitzig:
Fixes issue #301.


reitzig:
Well, there's also `copy()` but I see your point. 

My arguments for keeping the modifier would be a) for declaring intent (to humans and compiler) and b) to profit from future additions to data classes.

sandwwraith:
Hm, preserving `copy` is a good idea, I think. It could be useful. Thanks for your contribution!

PR #309, opened 2018-12-17
Add YAML parser to list of supported formats

charleskorn:
I've started work on a (very, very basic) YAML parser that works with this library - [kaml](https://github.com/charleskorn/kaml).

It's still very limited (only supports deserialisation at the moment), and is very rough (it works, but I've taken some shortcuts that will hurt performance, error messages might not be great etc.).

I'm not sure if this file is the best place to mention it, so I'm open to suggestions / requests to add this somewhere else (or not add it at all).

### DIFF
--- a/formats/README.md
+++ b/formats/README.md
@@ -16,3 +16,12 @@ Allows deserialization of `Config` object from popular [lightbend/config](https:
 into Kotlin objects.
 You can learn about "Human-Optimized Config Object Notation" or HOCON from library's [readme](https://github.com/lightbend/config#using-hocon-the-json-superset).
 
+## Other community-supported formats
+
+### YAML
+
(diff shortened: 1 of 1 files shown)


sandwwraith:
Ok, thanks for answers.

Regarding locations reporting: I personally think this is highly format-dependent. E.g. binary formats can have only byte offset in a stream, string formats can either have both line and column or only char offset (json can be treated as one big line, and usually unformatted json from rest apis doesn't contain indentation or line breaks). Exceptions also are very likely to be format-specific. Although, it may be possible to commonize an error location (as `Location` you have), I'm not sure that it will be sufficient for all formats. If you have ideas on how it could be correctly designed, feel free to open a separate issue.

PR #330, opened 2019-01-12
Add constructor expects for top level classes InputStream and OutputStream

tvierling:
It's not possible to subclass these in other code without the constructors being visible.

### DIFF
--- a/runtime/common/src/main/kotlin/kotlinx/io/Streams.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/io/Streams.kt
@@ -21,7 +21,7 @@ expect open class IOException: Exception {
     constructor(message: String)
 }
 
-expect abstract class InputStream {
+expect abstract class InputStream() {
     open fun available(): Int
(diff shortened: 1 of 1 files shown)


tvierling:
Hm. So I looked at kotlinx-io and it does offer some useful primitives, but it's totally foreign to those coming from a Java world (and, obviously, won't work with code simply converted from Java to Kotlin using auto-conversion, which is part of my use case). That explains why it didn't show up in searches: the Java class names aren't in there at all.

If these get dropped here, I'll probably create a third party emulation layer that exposes the Java-styled streams, maybe based on kotlinx-io under the hood.

sandwwraith:
kotlinx.io API is aimed to address problems of performance and design of Java's Inputs/Readers/etc. AFAIR, it has methods to convert kotlinx.io's Input to Java InputStream and vice versa.

However, a compatibility layer would indeed be helpful for those people which are trying to migrate JVM codebase.

PR #331, opened 2019-01-12
Optimize the Writer.write(String) case.

tvierling:
The old implementation called .toList().toCharArray() which created an
intermediate representation with extra overhead. Instead, simply copy
the representation into a char array directly.


tvierling:
(Fixed commit: end-of-file newline was added originally.)

PR #332, opened 2019-01-12
Reader.read(): Int should return -1 on EOF.

tvierling:
(Currently it will blindly keep returning 0 (from the unused array) indefinitely.)


sandwwraith:
Btw, it would be nice to add tests for added functionality

PR #339, opened 2019-01-20
Update reference to kaml to indicate it supports serialization as well

charleskorn:
This is supported as of kaml version 0.5.


sandwwraith:
Good to hear!

PR #350, opened 2019-01-28
Serializers for common java types

qwwdfsad:


runtime/jvm/src/main/kotlin/kotlinx/serialization/java/JavaTypesModule.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.atomic.*
+import kotlin.reflect.*
+
+/**
+ * Module with external serializers for the following Java types:
+ * Atomics: [AtomicInteger], [AtomicLong], [AtomicBoolean].
+ * Identifiers: [URI], [URL] and [UUID].
+ * Numbers: [BigInteger] and [BigDecimal].
+ * Strings: [StringBuilder].
+ * Dates: [DateSerializer].
```
JakeWharton: Should be `[Date]`

runtime/jvm/src/main/kotlin/kotlinx/serialization/java/JavaTypesModule.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.context.*
+import java.math.*
+import java.net.*
+import java.util.*
+import java.util.concurrent.atomic.*
+import kotlin.reflect.*
+
+/**
+ * Module with external serializers for the following Java types:
```
JakeWharton: How are you going to combat this class from playing whack-a-mole forever in the future from people wanting their type included. For example, why isn't `AtomicIntegerArray` included? Why isn't `java.time.*`. Why isn't `File`?

sandwwraith:
@qwwdfsad IMO we should move all this stuff to another module to publish separately, just like add-on formats or schema generators. People who are using the library in multiplatform context don't need all these `java.*` bits. We can also incrementally grow this module without fear of bloating main library itself.

runtime/jvm/src/main/kotlin/kotlinx/serialization/java/JavaTypesModule.kt
qwwdfsad: I don't mind including other types on a demand-driven basis when it is reasonable, an additional bonus is that this kind of features is a good "first-time contribution".

The only question is what is a "reasonable" request. E.g. `java.sql.Date` is a good candidate, while `java.io.File` is definitely not (at least because its representation is OS/machine/JVM specific). I know no strict criteria for a type to be included except common sense.

runtime/jvm/src/main/kotlin/kotlinx/serialization/java/DateSerializer.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
+import java.text.*
+import java.util.*
+
+/**
+ * Serializer for [java.util.Date] which serializers date as format string in default US locale.
+ */
+public object DateSerializer : KSerializer<Date> {
+    private val format = DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)!!
```
fvasco: Why a non standard format, without a time zone reference?

qwwdfsad:
TODO: discuss FQN in descriptors

runtime/jvm/src/main/kotlin/kotlinx/serialization/java/DateSerializer.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    override fun serialize(encoder: Encoder, obj: Date) {
+        val string = synchronized(format) { format.format(obj) }
```
sandwwraith: Why you've preferred `synchronized` on a format instead of making it ThreadLocal?

runtime/jvm/src/main/kotlin/kotlinx/serialization/java/DateSerializer.kt
pdvrieze: And, there are actually a number of different relevant standards. You probably want to create serializers for each of them. As well as with or without time. Of course maybe dates/calendars/instants (whichever date type you want to use) are not the best starting point for this.

pdvrieze:
> @qwwdfsad IMO we should move all this stuff to another module to publish separately, just like add-on formats or schema generators. People who are using the library in multiplatform context don't need all these `java.*` bits. We can also incrementally grow this module without fear of bloating main library itself.

I think so too. I have my own suggested additions as well (serializers to generate hashcode?). But that doesn't really belong in the main runtime support library.

PR #352, opened 2019-01-30
Schema generators sample

sandwwraith:
* Json schema and protobuf schema represented as samples in the side project, which uses included build
* Visitors.kt left in the main library as an interface and a skeleton implementation

Also fixed a number of bugs near enums.


example-visitors/src/main/kotlin/kotlinx/serialization/schema/JsonSchema.kt
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2018 JetBrains s.r.o.
```
qwwdfsad: The year is expired.
You can use `2018-$today.year` macro in copyrights

qwwdfsad:
We have four example packages, I think it is time to merge it, isn't it? (I don't insist, we can do it in a separate PR)

runtime/common/src/main/kotlin/kotlinx/serialization/Shorthands.kt
```diff
@@ -47,6 +52,16 @@ fun SerialDescriptor.getElementIndexOrThrow(name: String): Int {
...
 }
 
-@Deprecated("Obsolete name from preview version of library.", ReplaceWith("elementsCount"))
+/**
+ * Searches for annotation of type [A] in annotations, obtained via
+ * [SerialDescriptor.getElementAnnotations] at given [elementIndex]
+ *
+ * Returns null if there is 0 or more than 1 annotation with such type.
+ */
+inline fun <reified A: Annotation> SerialDescriptor.findAnnotation(elementIndex: Int): A? {
```
qwwdfsad: Let's move it to samples?
Currently, it does not carry any value in runtime core

runtime/common/src/main/kotlin/kotlinx/serialization/schema/Visitors.kt
```diff
@@ -0,0 +1,72 @@
...
+
+import kotlinx.serialization.*
+
+/**
+ * Base interface for a visitor of a [SerialDescriptor] tree,
+ * which can return the result of processing the tree.
+ *
+ * Tree is formed from descriptors using [SerialDescriptor.getElementDescriptor] method.
+ */
+public interface DescriptorVisitor<R> {
```
qwwdfsad: Again, I'd rather move it to the samples.
We can always make it a core functionality as soon as there is any demand on that, but it is better not to publicly expose any API beyond a demand.

example-visitors/src/main/kotlin/kotlinx/serialization/schema/JsonSchema.kt
sandwwraith: I have it in macro, but have forgotten to update it after copying files

runtime/common/src/main/kotlin/kotlinx/serialization/Shorthands.kt
sandwwraith: I think it can be useful in some applications. I can try to identify where I can use this function in the runtime library itself.

runtime/common/src/main/kotlin/kotlinx/serialization/schema/Visitors.kt
sandwwraith: OK

sandwwraith:
I have a separate branch where the examples are merged

runtime/common/src/main/kotlin/kotlinx/serialization/Shorthands.kt
qwwdfsad: Let's make it internal for a while

Issue #303, opened 2018-12-10
JsonLiteral doesn't expose the type of unquoted JSON literal


reitzig:
The public interface of `JsonLiteral` does not expose the type of the JSON literal. `content` is always available (via `toString()`), and e.g. `int[OrNull]` are delegated to `toInt()` which make integer values available even if the JSON literal was not one.

It is thus impossible to distinguish

```json
{
	"foo": 7
}
```

from

```json
{
	"foo": "7"
}
```

_Use case:_ implementing a JSON Schema validator (cf. [here](https://github.com/reitzig/kovalscj)). In order to verify whether a `:"type": "integer"` assertion is violated, one needs to be able to distinguish values/literals `"7"` and `7`.

reitzig:
This is indeed baked into `JsonTreeParser`: there are no separate token classes for the separate JSON types; those that are not `string` or `null` -- specifically, `number`, `true`, `false` all fall under `TC_OTHER` (and not even _that_ difference is exposed).

Several approaches can be taken, including:

 1. Keep the current parser, but
    a. expose the type of `JsonLiteral.body` directly or
    b. provide `isNumberLiteral`, `isBooleanLiteral`, ... on `JsonLiteral`.
 2. Change the parser to distinguish literal types and have it create instances of corresponding subtypes of `JsonLiteral`.

sandwwraith:
I suppose this is kinda fixed in `dev` since property `isString` of `JsonLiteral` is now public: https://github.com/kotlin/kotlinx.serialization/blob/dev/runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt#L147

However, there are indeed no tools to determine a type of the unquoted literal. You can fallback for such checks for now: https://github.com/kotlin/kotlinx.serialization/blob/dev/runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonElementSerializer.kt#L105

shanshin:
Class `JsonPrimitive` contains `isString` property for this purpose

Tihon-Ustinov:
> Class `JsonPrimitive` contains `isString` property for this purpose

and if this number or boolean?

qwwdfsad:
We do not have out-of-the-box extensions for that, but they can be written according to your needs.
[...]

PR #386, opened 2019-02-21
JsonLiteral retains type of literal

reitzig:
Here's a proposal for a `JsonLiteral` that knows which type of value the original JSON had (cf. #303).

I tried not to break its interface; users of the class need to add the wildcard type parameter `<*>`.

Does making `JsonLiteral` generic have consequences beyond what the tests cover? Can't say I have an overview over the actual object serialization parts of the library.

I can see potential changes to the interface but I'm not sure how far to go in an experimental library, so I stayed conservative.
On concrete thing: `JsonNull` could be a `JsonLiteral` now, couldn't it?

### DIFF
--- a/runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
@@ -146,24 +146,11 @@ public sealed class JsonPrimitive : JsonElement() {
  * and `JsonLiteral("42", true)` will be serialized as `42` and `"42"` respectively.
  */
 @Serializable(JsonLiteralSerializer::class) // TODO val for body is workaround for plugin bug
-public class JsonLiteral internal constructor(val body: Any, public val isString: Boolean) : JsonPrimitive() {
+sealed class JsonLiteral<T> private constructor(val body: T) : JsonPrimitive() {
+    @Deprecated("Check types instead")
+    public abstract val isString: Boolean
     public override val content = body.toString()
     public override val contentOrNull: String = content
-
-    /**
-     * Creates number literal
-     */
-    public constructor(number: Number) : this(number, false)
-
-    /**
-     * Creates boolean literal
-     */
-    public constructor(boolean: Boolean) : this(boolean, false)
-
-    /**
-     * Creates quoted string literal
-     */
-    public constructor(string: String) : this(string, true)
 
     public override fun toString() =
         if (isString) buildString { printQuoted(content) }
@@ -174,7 +161,7 @@ public class JsonLiteral internal constructor(val body: Any, public val isString
         if (this === other) return true
         if (other == null || this::class != other::class) return false
 
-        other as JsonLiteral
+        other as JsonLiteral<*>
 
         if (isString != other.isString) return false
         if (content != other.content) return false
@@ -186,6 +173,61 @@ public class JsonLiteral internal constructor(val body: Any, public val isString
         var result = isString.hashCode()
         result = 31 * result + content.hashCode()
         return result
+    }
+
+    public class JsonNumberLiteral(body: Number) : JsonLiteral<Number>(body) {
+        override val isString: Boolean = false
+    }
+
+    public class JsonBooleanLiteral(body: Boolean) : JsonLiteral<Boolean>(body) {
+        override val isString: Boolean = false
+    }
+
+    public class JsonStringLiteral(body: String) : JsonLiteral<String>(body) {
+        override val isString: Boolean = true
+    }
+
+    @Deprecated("Included to not break the API, but should never occur.")
+    class JsonOtherLiteral(body: Any) : JsonLiteral<Any>(body) {
+        override val isString: Boolean = false
+    }
+
+    companion object {
+        @Deprecated("Use single-parameter variants.", ReplaceWith("JsonLiteral(body)"))
+        internal operator fun invoke(body: Any, isString: Boolean) : JsonLiteral<*> =
+                when (body) {
+                    is Number -> invoke(body)
+                    is Boolean -> invoke(body)
+                    is String -> invoke(body)
+                    else -> JsonOtherLiteral(body)
+                }
+
+// Interestingly (?), this doesn't type-check:
+//        internal operator fun <T> invoke(body: T, isString: Boolean) : JsonLiteral<T> =
+//                when (body) {
+//                    is Number -> invoke(body)
+//                    is Boolean -> invoke(body)
+//                    is String -> invoke(body)
+//                    else -> JsonOtherLiteral(body)
+//                }
+
+        /**
+         * Creates number literal
+         */
+        public operator fun invoke(number: Number) : JsonLiteral<Number> =
+                JsonNumberLiteral(number)
+
+        /**
+         * Creates boolean literal
+         */
+        public operator fun invoke(boolean: Boolean) : JsonLiteral<Boolean> =
+            JsonBooleanLiteral(boolean)
+
+        /**
+         * Creates quoted string literal
+         */
+        public operator fun invoke(string: String) : JsonLiteral<String> =
+                JsonStringLiteral(string)
     }
 }
 
--- a/runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonReader.kt
+++ b/runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonReader.kt
@@ -23,6 +23,8 @@ import kotlin.jvm.*
 
 // special strings
 internal const val NULL = "null"
+internal const val FALSE = "false"
+internal const val TRUE = "true"
 
 // special chars
 internal const val COMMA = ','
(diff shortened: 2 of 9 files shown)


reitzig:
This is a straight-up implementation of the standard now, which breaks quite a few tests.

 1. The tests seem to assume that output of `unquoted` can be parsed again. That _seems_ like relying on undocumented behaviour that contradicts the JSON spec -- or was it a conscious design decision?
 2. `nonstrict` is not supported fully. `NaN`, `[-]Infinity` are not parsed as numbers.
 3. Anything that I'm missing?

Assuming that non-strict JSON is a requirement, I think both parsing unquoted strings and non-strict mode can be done following the current approach; maybe `JsonReader` needs to get option flags.

Both at the same time, however, can not be parsed unambiguously: `NaN`, for example, might be a number (haha) or a string. While the deserialization consumers would know which conversion to use (based on the expected type), a stand-alone JSON parser (such as `Json.plain.parseJson)`) wouldn't know which literal type to choose.

Please advise as to what the expected/required behaviour is here, in particular for item 1 above.

reitzig:
@sandwwraith and @qwwdfsad, are you waiting for a complete pull request before reviewing it, or are you having little time for it? Trying to decide whether to wait for your feedback or just continue to forge ahead.

qwwdfsad:
Hi @reitzig 
I will review it tomorrow properly

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -146,25 +146,12 @@ public sealed class JsonPrimitive : JsonElement() {
  * and `JsonLiteral("42", true)` will be serialized as `42` and `"42"` respectively.
  */
 @Serializable(JsonLiteralSerializer::class) // TODO val for body is workaround for plugin bug
-public class JsonLiteral internal constructor(val body: Any, public val isString: Boolean) : JsonPrimitive() {
+sealed class JsonLiteral<T> private constructor(val body: T) : JsonPrimitive() {
```
qwwdfsad: please do not remove explicit public modifier, it is crucial to have a proper understanding of what is and what is not public API (see [KEEP#45](https://github.com/Kotlin/KEEP/issues/45))

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -146,25 +146,12 @@ public sealed class JsonPrimitive : JsonElement() {
...
+    @Deprecated("Check types instead")
+    public abstract val isString: Boolean
```
qwwdfsad: This property was not released, so you can safely remove it. 
Also, it should not be used in the implementation if it is deprecated

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -76,12 +88,17 @@ internal val C2TC = ByteArray(CTC_MAX).apply {
...
+    initC2TC(TRUE.first(), TC_BOOL)
+    initC2TC(NUMBER_NEG, TC_NUMBER)
+    NUMBER_DIGITS.forEach { initC2TC(it, TC_NUMBER) }
 }
 
 // object instead of @SharedImmutable because there is mutual initialization in [initC2ESC]
 internal object EscapeCharMappings {
     @JvmField
-    public val ESCAPE_2_CHAR = CharArray(ESC2C_MAX)
+    val ESCAPE_2_CHAR = CharArray(ESC2C_MAX)
```
qwwdfsad: ditto`public`

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -37,6 +39,14 @@ internal const val STRING_ESC = '\\'
 internal const val INVALID = 0.toChar()
 internal const val UNICODE_ESC = 'u'
 
+internal const val NUMBER_NEG = '-'
+internal const val NUMBER_POS = '+'
+internal const val NUMBER_SEP = '.'
+internal const val NUMBER_EXP = 'e'
+internal val NUMBER_DIGITS = arrayOf('0', '1', '2', '3', '4', '5', '6', '7', '8', '9') // TODO: ?
```
qwwdfsad: It is better to use `CharArray` instead of `Array<Char>` for the performance sake

qwwdfsad:
>nonstrict is not supported fully. NaN, [-]Infinity are not parsed as numbers.

I will fix it separately, thanks for catching it

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonElement.kt
sandwwraith: it is `public` in master now

sandwwraith:
I also doubt that this feature worth adding a generic parameter to `JsonLiteral` – see, in most of the places where it is used you had to erase this information with `<*>` because you simply do not care. It is also not a 'minimal breakage'. So an additional property with some representation of primitive kind and possibly some converter functions look more favorable to me.

diegomontoya:
I also do not believe we actually need a generic JsonLiteral<T>. To FIX "10" vs 10 issue which is very realistic, we just need to implement a isQuoted property and expose it via public API.

reitzig:
Thanks for the feedback! It's been a while, and I haven't really gotten around to this.

Tl;dr: based on the comments this PR received, it's become clear to me that the library was never intended to provide a cross-platform JSON parser as such. I went at this with wrong expectations; closing.

---

> There is a lot of heavy-lifting in JsonParser/JsonReader just to properly support this small feature. Moreover, due to this feature, regular JSON parser will experience a notable performance hit.

What makes you say that? There's a bit more code, but aside from that one, easily fixed `Array<Char>` usage aside, the only bit that changed the actual performance is the [number literal check](https://github.com/Kotlin/kotlinx.serialization/blob/f5d9b7c950780fcaadcd3a2b8cfdd989432a7791/runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/JsonReader.kt#L325) that could be circumvented if it's a bottleneck.
Note how even that change contributes something that's IMHO worthwhile (required, even): it checks for correct number literals at _parse time_, not when consuming the parse result as before (IIRC).

> The public API surface can (has to) be reduced to a single class change: JsonLiteral.

> I also doubt that this feature worth adding a generic parameter to JsonLiteral – see, in most of the places where it is used you had to erase this information with <*> because you simply do not care.

That was a kind-of uncomfortable compromise. I _actually_ wanted to remove `JsonLiteral` in favor of what then became its subtypes, but clearly _that_ would have been a "very breaking" change. Adding a type parameter technically breaks usages, but those are easily fixed by adding `<*>`, as you noted.

The whole point of this PR is to enable _new_ types of usage; of course, those (internal) usages that already exist didn't care about literal types, for better or worse, so they go with `<*>`. The schema validator I was working on would have relied heavily on the full types!

(FWIW, the "has to" in the first quote feels off to me, talking about an experimental library.)

> Keep API small and simple

Agreed. But those proposals are exactly the opposite, aren't they:

> So an additional property with some representation of primitive kind and possibly some converter functions look more favorable to me.

> To FIX "10" vs 10 issue which is very realistic, we just need to implement a isQuoted property and expose it via public API and let the end-user deal with it.

In summary, I still don't understand why the parser was written the way it was (it doesn't conform to the JSON specification)

qwwdfsad:
@reitzig 

> it's become clear to me that the library was never intended to provide a cross-platform JSON parser as such

It's obviously not true, could you please elaborate on what made you think that way?

>FWIW, the "has to" in the first quote feels off to me, talking about an experimental library

Yes, the library is still experimental, but we are working hard towards its stabilization. It includes stabilization of runtime features such as JSON, Kotlin feature-completeness (objects, sealed classes, inline classes, delegates etc.) and re-evaluation of design-decisions in the code-generating plugin, including its architecture and code-generation strategy (to make it future-proof and to still have opportunities for optimizations and to change code-generation strategy even when serialization is stable). Thus now we are much more mindful about each new public signature or semantic change.

>I still don't understand why the parser was written the way it was (it doesn't conform to the JSON specification)

It was indeed written as an experiment in the first place, thus the code quality. But again, we are working towards changing that (e.g. #497). Now it should **mostly** conform the spec modulo some corners I don't really want to explore. I would even say that we are on par with GSON when talking about spec conformance.

>What makes you say that? There's a bit more code, but aside from that one, easily fixed Array<Char> usage aside, the only bit that changed the actual performance is the number literal check that could be circumvented if it's a bottleneck.

A few points here:
- More token classes to check for/parse. It may be insignificant though, but it's hard to tell without measuring it first.
- Double (and other types as well, double is just the biggest contributor) parsing when reading `JsonElement`, even if one doesn't need a precise type.
- `nextNumber` itself (that indeed can be optimized)

It's arguable whether it is really a blocker, but just something one should be aware about.


Regarding the change in general:

I realize this feature is important to you, but unfortunately, it does not mean that we should include this feature. And it doesn't mean that we are not interested in development of the library, being feature-rich or being contribution-friendly either! It's all about compromises (often mixed with personal preferences regarding API shape) that sometimes do not match your expectations.

For example, you are developing schema-validation. We were thinking about it as well and decided that it should work via `SerialDescriptor` hierarchy (very simplified example: https://github.com/Kotlin/kotlinx.serialization/blob/master/examples/example-visitors/src/test/kotlin/kotlinx/serialization/schema/JsonSchemaTest.kt) instead of relying on raw `JsonElement` types. Maybe this is what suits your needs instead of typed `JsonElement`?

reitzig:
> It's obviously not true, could you please elaborate on what made you think that way?

IMHO, a general-purpose JSON parser would focus being conformant to the standard(s). I would expect the same library to be able to generate identical JSON (up to formatting) from its internal representation, without external information; `marshal(unmarshal(x) == x` and `unmarshal(marshal(x)) == x` seem like reasonable requirements in that sense.

Here, the priority lies on one specific use-case -- object/data serialization -- and efficiency. The assumption that separate type information is available (from a Kotlin class, say) seems inherently baked into the approach.
(This is speaking from what I say back when I created the PR. If things have changed, never mind.)

That's all fair -- as you say, different goals mean different trade-offs. It just meant that I saw that this library was not intended to serve the purpose I expected, so instead of trying to shoehorn it I had to go a different route.

PR #413, opened 2019-03-22
Do not allow to mutate SerialModule in formats

qwwdfsad:


runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/StreamingJsonOutput.kt
```diff
@@ -14,11 +15,12 @@ internal class StreamingJsonOutput(private val composer: Composer, override val
         modeReuseCache: Array<JsonOutput?>
     ) : this(Composer(output, json), json, mode, modeReuseCache)
 
+    public override val context: SerialModule = json.context
```
sandwwraith: Why you've preferred getter over initialization almost everywhere but not here?

qwwdfsad: Getter in abstract and open classes, property in final ones

PR #421, opened 2019-04-02
Json polymorphism

qwwdfsad:


qwwdfsad:
Let's discuss what we are going to do with update mode

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -0,0 +1,37 @@
...
+
+internal inline fun <T> JsonOutput.encodePolymorphically(serializer: SerializationStrategy<T>, value: T, ifPolymorphic: () -> Unit) {
+    if (serializer !is PolymorphicSerializer<*> || json.configuration.useArrayPolymorphism) {
+        serializer.serialize(this, value)
+        return
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    val actualSerializer = serializer.findPolymorphicSerializer(this, value as Any) as KSerializer<Any>
+    if (actualSerializer is EnumSerializer<*>) {
```
sandwwraith: Should there be the same check for primitives? Also, any serializer which would represent classes as value, not as a composite seems to be in trouble here. Probably we can make use of checking a `descriptor.kind`

runtime/common/src/testWithNative/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
```diff
@@ -1,136 +0,0 @@
...
-import kotlinx.serialization.*
-import kotlinx.serialization.json.Json
-import kotlinx.serialization.modules.SerializersModule
-import kotlinx.serialization.modules.plus
-import kotlinx.serialization.test.assertStringFormAndRestored
-import kotlinx.serialization.test.isJs
-import kotlin.test.*
-
-
-class PolymorphismWithAnyTest {
```
sandwwraith: Why you've deleted this file?

runtime/common/src/test/kotlin/kotlinx/serialization/json/polymorphic/JsonListPolymorphismTest.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2017-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.json.polymorphic
```
sandwwraith: Tests in this folder are run only for JVM and JS. Consider moving all tests to `testsWithNative`, which is eventually will be the main test folder (and renamed back to `test`). To do this, you'll need to pass serializer explicitly in some cases,  because `@ImplicitReflectionSerializer` is not available on Native. You can comment out build.gradle#43, reimport project and IDE will highlight to you problematic functions.

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -17,11 +17,16 @@ internal class StreamingJsonInput internal constructor(
...
     private var currentIndex = -1
     private var entryIndex = 0
-
+    private val configuration = json.configuration
+    
     public override fun decodeJson(): JsonElement = JsonParser(reader).read()
 
     override val updateMode: UpdateMode
-        get() = json.updateMode
+        get() = UpdateMode.OVERWRITE
```
sandwwraith: Are there any objections to adding `UpdateMode` to `JsonConfiguration`? I know we are going to redesign it, but for now, let's do not touch unrelated things

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -61,14 +48,34 @@ private val defaultJsonModule = serializersModuleOf(mapOf<KClass<*>, KSerializer
  * Note that `@ImplicitReflectionSerializer` are used in order to omit `DataHolder.serializer`, but this is a temporary limitation.
  */
 public class Json(
-    @JvmField internal val unquoted: Boolean = false,
-    @JvmField internal val indented: Boolean = false,
-    @JvmField internal val indent: String = "    ",
-    @JvmField internal val strictMode: Boolean = true,
-    val updateMode: UpdateMode = UpdateMode.OVERWRITE,
-    val encodeDefaults: Boolean = true,
+    @JvmField internal val configuration: JsonConfiguration = JsonConfiguration(),
```
sandwwraith: AFAIR, our main intention for this change was to provide out of the box a 'proper' default for `useArrayPolymorphism=true`. This parameter has a default value, and default value in JsonConfiguration of this flag is also `false`. It is good that users who use argument-less constructor will not experience semantic change, but it is bad that default semantics is not the one that we want. The proposal was to remove default value for this parameter, so users would always explicitly specify their configuration – and at least they will see that they have option of `useArrayPolymorphism` and it is set to `true`. Regarding default instances (plain, unquoted, etc)..., they should have old semantics but probably some warning about unstable/obsolete semantics could be added with the help of deprecated/experimental annotation.

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -151,3 +158,34 @@ public class Json(
             plain.parse(deserializer, string)
     }
 }
+
+/**
+ * Builder to conveniently build Json instances.
+ * Properties of this builder are directly matched with properties of [JsonConfiguration].
+ */
+public class JsonBuilder {
+    public var encodeDefaults: Boolean = true
```
sandwwraith: Using this builder is the same as using an old json constructor with all arguments, just replace `( )` with `{ }`. I'm not sure if this is advantage or drawback, but it seems as overkill to me (discussable)

runtime/common/src/main/kotlin/kotlinx/serialization/Polymorphic.kt
```diff
@@ -128,6 +128,25 @@ public class PolymorphicSerializer<T : Any>(private val baseClass: KClass<T>) :
         compositeDecoder.endStructure(descriptor)
         return requireNotNull(value) { "Polymorphic value have not been read for class $klassName" }
     }
+
+    /**
+     * Lookup of the polymorphic serializer by given [fqnClassName] in the context of [decoder] by the base class [basePolyType].
+     * Throws [SerializationException] if serializer is not found.
+     */
+    public fun findPolymorphicSerializer(
```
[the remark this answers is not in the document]
qwwdfsad: Merge leftover, thanks.
Because they are used in JSON (and JSON should be implementable without internal API) and required for custom serializers

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
qwwdfsad: `{}` has multiple advantages compared with the constructor:
 - Easier to read: `Json { foo = true }` vs `Json(JsonConfiguration(foo=true))`
 - Easier to provide migrations for builders while preserving source compatibility and deprecation cycles
 - Builder can be more flexible, e.g. in the face of custom values for class discriminators

runtime/common/src/main/kotlin/kotlinx/serialization/Polymorphic.kt
```diff
@@ -2,19 +2,17 @@
  * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-@file:Suppress("RedundantVisibilityModifier")
```
sandwwraith: Why remove suppress?

runtime/common/src/main/kotlin/kotlinx/serialization/json/Json.kt
```diff
@@ -138,9 +162,9 @@ public class Json(
 
     companion object : StringFormat {
         val plain = Json()
-        val unquoted = Json(unquoted = true)
-        val indented = Json(indented = true)
-        val nonstrict = Json(strictMode = false)
+        val unquoted = Json(JsonConfiguration(unquoted = true, useArrayPolymorphism = true))
```
sandwwraith: Here we can either preserve semantics for default instances or break it and annotate them with `@UnstableDefault`. I'm not sure which way is better. Preserving semantics seems right to me since most users are not using any advanced features, but eventually, closer to the release, we still would need to change it.

runtime/common/src/main/kotlin/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -0,0 +1,53 @@
...
+ * Options list:
+ * [encodeDefaults] specifies whether default values are encoded.
+ * [strictMode] enables strict mode, which prohibits unknown keys in the JSON and non-numerical values in floating point numbers.
+ * [unquoted] specifies whether keys and values should be quoted;
+ * [prettyPrint] specifies whether resulting JSON should be pretty-printed.
+ * [indent] specifies indent string to use with [prettyPrint] mode.
+ * [useArrayPolymorphism] switches polymorphic serialization to the default array format.
+ * [classDiscriminator] name of the class descriptor property in polymorphic serialization.
+ */
+public class JsonConfiguration(
```
sandwwraith: Btw, this constructor has also some kind of unstable behavior if we will add new flags which turn off some features by default (e.g. `allowNonTrivialKeysInMaps = false`). Is there any possible way to avoid it or this is inevitable evil?

runtime/common/src/main/kotlin/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -0,0 +1,42 @@
...
+    checkKind(kind)
+
+    ifPolymorphic()
+    actualSerializer.serialize(this, value)
+}
+
+fun checkKind(kind: SerialKind) {
+    if (kind is UnionKind.ENUM_KIND) error("Enums cannot be serialized polymorphically with 'type' parameter. You can use 'JsonConfiguration.useArrayPolymorphism' instead")
+    if (kind is PrimitiveKind) error("Primitives cannot be serialized polymorphically with 'type' parameter. You can use 'JsonConfiguration.useArrayPolymorphism' instead")
+    if (kind is UnionKind.POLYMORPHIC) error("Unexpected kind 'POLYMORPHIC'")
```
sandwwraith: Maybe write a message like `Polymorphic serializer can't be written polymorphically, it would lead to stack overflow. Make sure you have not registered one by mistake.` or smth like that

runtime/common/src/main/kotlin/kotlinx/serialization/SerialImplicits.kt
```diff
@@ -18,31 +18,40 @@ import kotlinx.serialization.internal.HexConverter
...
+public annotation class ImplicitReflectionSerializer
+
+/**
+ * This annotation marks declarations with default parameters that are subject to semantic change without a migration path.
+ *
+ * For example, [JsonConfiguration.Default] marked as unstable, thus indicating that it can change its default values
+ * in the next releases (e.g. disable strict-mode by default), leading to a semantic (not source code level) change.
+ */
+@Experimental
+public annotation class UnstableDefault
```
sandwwraith: We can also lower level from `ERROR` to `WARNING`

Issue #19, opened 2017-10-05
@Optional annotation seems excessive


shafirov:
For `@Optional` annotation default constructor parameter value is required. I guess we could consider all of parameter with default values as optional

elizarov:
We've decided to require it there to be explicit about optionallity (at least initially). It also helps when you start moving your properties between primary constructor and class body. The parameters in the class body must always have an initialization expression, but it may not be conceptually appropriate to tread all of them as implicitly optional. Or maybe it is appropriate. We can reconsider this, but it really needed a larger design discussion.

apatrida:
I prefer we infer `Optional` and you override with `Required` in the cases where you don't like the default behaviour.  This would also match some other implementations of data binding that have been around a bit setting expectations :-)

pdvrieze:
In the current state where optional values are always written anyway this may be an option, but if we have optional mean that writing may be omitted by the format(encoder/decoder) a default optionality can create surprising effects.

rawnsley:
I second the idea of removing @@Optional and replacing it with @@Required.

At the very least @@Optional should be implied for nullable types.

Diolor:
+1 to @rawnsley 's `At the very least Optional should be implied for nullable types`.

If you have strong reasons to keep the existing functionality by default, another way is to make a global strategy and one to chose for himself:

```kotlin
@Serializable
data class Foo(
		val name: String?
)

Json(implyOptionalsDefaults = true).parse(Foo.serializer(), "{ }") // this should print Foo(name=null)
```

avently:
@elizarov  @pdvrieze @sandwwraith
Can you take look at this issue once more? 

I think that every field can be optional by default (or this option can be specified via Json constructor) and doesn't throw exceptions if it has default value. 
In my situation I have client-server architecture. Configuration files can be edited by a user. He can make a mistake and then default value should be used since I don't want to give up and allow to server to die. Default values exist for every field and they are safe for every user. 

What do you think now after year while this issue exist?

pdvrieze:
The way the system works is that missing values are actually verified by the (generated) (de)serializationstrategy. Any format would struggle supporting this as it would need to keep its own account of missing non-read values and implicitly read nulls. This is actually what my XML library does - it treats missing values as null (for now it doesn't yet look at optionality as that was not visible in older versions).

It may be possible to create proxy strategies that make this work with default formats. I think that would probably be the way to go.

sandwwraith:
[...]

PR #435, opened 2019-04-16
Fix optional property example

0xBADDCAFE:
For someone using older kotlin by some reason, I think below comment should be documented.

https://github.com/Kotlin/kotlinx.serialization/issues/19#issuecomment-483331467

> Please note that you have to use Kotlin 1.3.30 or higher.

Also should consider about documenting deprecated `@Optional` annotation.


sandwwraith:
Thanks!

PR #440, opened 2019-04-17
Infrastructure to build kotlinx.serialization against Kotlin snapshots

qwwdfsad:


build.gradle
```diff
@@ -3,7 +3,24 @@
  */
 
 buildscript {
-    def deployNative = property('native.deploy') == 'true'
```
sandwwraith: It was used, please bring it back

build.gradle
```diff
@@ -3,7 +3,24 @@
...
+    /*
+     * These property group is used to build kotlinx.serialization against Kotlin compiler snapshot.
+     * When build_snapshot_train is set to true, kotlin_version property is overridden with kotlin_snapshot_version.
+     * DO NOT change the name of these properties without adapting kotlinx.train build chain.
+     */
+    def prop = rootProject.properties['build_snapshot_train']
+    ext.build_snapshot_train = prop != null && prop != ""
+    if (build_snapshot_train) {
+        ext.kotlin_version = rootProject.properties['kotlin_snapshot_version']
```
sandwwraith: It seems you have to set `ext.compilerVersion`, `ext.serializationPluginVersion` and `  ext.librariesVersion` instead of this property.

There is also `ext.konanVersion`, since serialization uses old mpp. What Native version we have to use in snapshots?

Issue #418, opened 2019-03-26
JsonArrayBuilder's unaryPlus extensions don't work for Number?


jcayzac:
Using `kotlinx-serialization-runtime:0.10.0` the following code produces an empty array:

```kt
jsonArray {
    +2
    +4
    +6
    +8
}
```

Being more verbose works as expected:

```kt
jsonArray {
    +JsonPrimitive(2)
    +JsonPrimitive(4)
    +JsonPrimitive(6)
    +JsonPrimitive(8)
}
```

It seems that https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/common/src/test/kotlin/kotlinx/serialization/json/serializers/JsonArraySerializerTest.kt lacks tests for the the `unaryPlus` overloads defined in `JsonArrayBuilder`, maybe.

jcayzac:
The overload for `String?` works, this returns an array of size `3`:

```kt
jsonArray {
    +"foo"
    +"bar"
    +"baz"
}
```

jcayzac:
I ended up using `JsonArray(listOf(2, 4, 6, 8).map(::JsonPrimitive))` for now…

q-litzler:
```kotlin
jsonArray {
  +(1 as Number)
}
```

This will work. The problem is that when you only write `+4`, you don't overload the correct `unaryPlus` operator.

toddobryan:
Can we get some movement on this. I was just bitten by it. There's a solution in dev, but it hasn't been merged back into master. In the comments on the commit, `add` is the much less verbose choice, especially when you consider that you can do `add(Number)`, `add(String)`, add(Boolean)`, and `add(null)` for all the primitives, and `add(JsonElement)` for non-primitives. It's too bad that `unaryPlus` gets confused with `Number`s, but it seems like `add` is the most concise and consistent way of doing this given that issue.

PR #462, opened 2019-05-21
Deprecate unaryPlus for Json array builder in favor of `add`.

sandwwraith:
Mainly because of #418

There are three ways for it:

1. Do nothing.
2. Deprecate `unaryPlus` only for `Number` (and probably other primitives)
3. Deprecate all overloads, including one with `JsonElement`.

This PR implements 3), but I suggest considering 2) as a resolution, because using `add` everywhere seems verbose and overhead-ish to me (see diff of `JsonArraySerializersTest`). However, this will lead to inconsistency: `JsonElement` would be added via `unaryPlus` and other primitives via `add`. WDYT, @qwwdfsad ?

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 @file:Suppress("RedundantVisibilityModifier")
@@ -34,6 +34,7 @@ public class JsonArrayBuilder internal constructor() {
     /**
      * Adds [this] value to the current [JsonArray] as [JsonPrimitive].
      */
+    @Deprecated("Deprecated due to ambiguity", ReplaceWith("add(this)"), DeprecationLevel.ERROR)
     public operator fun String?.unaryPlus() {
         content.add(JsonPrimitive(this))
     }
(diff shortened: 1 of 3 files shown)


qwwdfsad:
This is a subtle topic. I agree that `add` is too verbose` and I am fine with a option when both `add` and `+` are present for every element type except `Number` to leave some options for ones who prefer consistency over verbosity.

But I think we can do better. Let's take a look at the `JsonArraySerializersTest`.
```
return jsonArray {
    add(1)
    add(JsonNull)
    add(jsonArray {
        add("nested literal")
    })
    add(jsonArray { })
    add(json {
        "key" to "value"
    })
}
```

what do we want to ideally see here? A consistent picture without a boilerplate.
Say, something like this:
```
return jsonArray {
    jsonLiteral(1)
    jsonNull // <- I have doubts about this one
    jsonArray {
        JsonLiteral("nested literal")
    }
    jsonArray { } 
    jsonObject {
        "key" to "value"
    }
}
```

what do you think about that approach? I tend to think that `+` and `add` are really unnecessary here and the only choice we have to make is whether these builders should look like plain json elements `JsonArray { ... }` or like functions `jsonArray { ... }`

sandwwraith:
It looks nice but seems a bit overkill, since this DSL would not be used often (and when it is used, I assume, it would contain one-two calls in a `for` loop). There is also a problem which can cause a mistake you've already made in the example: `jsonArray { jsonLiteral(1) }` works, but `jsonArray { JsonLiteral(1) }` (notice capital letter) does not.

qwwdfsad:
>It looks nice but seems a bit overkill, since this DSL would not be used often

I am not sure this is a valid point. Either we should provide a best-effort DSL or nothing at all.

>There is also a problem which can cause

Then probably it's worth to make builder extensions undistinguishable from constructors?

sandwwraith:
Closing in favor of #794

Issue #287, opened 2018-11-26
Top-level primitive serialization for tagged encoders


qwwdfsad:
Then fix `JsonPrimitiveSerializerTest` as protected `pushTag`

sandwwraith:
Fixed with #465

PR #465, opened 2019-05-21
Support primitives (ints, strings, JsonLiterals, etc) on a top-level 

### LINKED ISSUE #451: Primitive values cannot be serialized to JsonElement
**Describe the bug**
While `Json.plain.stringify` can be used on primitives and complex objects alike, `Json.plain.toJson` fails on primitive objects with an exception:

```
kotlinx.serialization.SerializationException: No tag in stack for requested element
	at kotlinx.serialization.TaggedEncoder.popTag(Tagged.kt:160)
	at kotlinx.serialization.TaggedEncoder.encodeBoolean(Tagged.kt:95)
	at kotlinx.serialization.internal.BooleanSerializer.serialize(Primitives.kt:54)
	at kotlinx.serialization.internal.BooleanSerializer.serialize(Primitives.kt:51)
	at kotlinx.serialization.json.internal.AbstractJsonTreeOutput.encodeSerializableValue(TreeJsonOutput.kt:174)
	at kotlinx.serialization.CoreKt.encode(Core.kt:73)
	at kotlinx.serialization.json.internal.TreeJsonOutputKt.writeJson(TreeJsonOutput.kt:29)
	at kotlinx.serialization.json.Json.toJson(Json.kt:129)
```

**To Reproduce**
Run (e.g. `./gradlew run`) the attached demo project from
[primitive-serialization.zip](https://github.com/Kotlin/kotlinx.serialization/files/3151857/primitive-serialization.zip)

**Expected behavior**
I would expect both Strings and Booleans to be serialized to JsonPrimitive instances.

**Environment**
 - Kotlin version: 1.3.31
 - Library version: 0.11.0
 - Kotlin platforms: JVM
 - Gradle version: 4.10.3

sandwwraith:
…when saving/restoring JSON AST

Fixes #451

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
@@ -1,17 +1,5 @@
 /*
- * Copyright 2018 JetBrains s.r.o.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization.internal
@@ -65,6 +53,14 @@ open class CommonEnumSerializer<T>(val serialName: String, val choices: Array<T>
         check(index in choices.indices)
             { "$index is not among valid $serialName choices, choices size is ${choices.size}" }
         return choices[index]
+    }
+
+    companion object {
+        /**
+         * Factory method for easier instantiation from user/test code.
+         */
+        inline fun <reified E: Enum<E>> create(serialName: String): CommonEnumSerializer<E> =
+            CommonEnumSerializer(serialName, enumValues<E>(), enumValues<E>().map { it.name }.toTypedArray())
     }
 }
 
--- a/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
@@ -54,14 +54,6 @@ open class CommonEnumSerializer<T>(val serialName: String, val choices: Array<T>
             { "$index is not among valid $serialName choices, choices size is ${choices.size}" }
         return choices[index]
     }
-
-    companion object {
-        /**
-         * Factory method for easier instantiation from user/test code.
-         */
-        inline fun <reified E: Enum<E>> create(serialName: String): CommonEnumSerializer<E> =
-            CommonEnumSerializer(serialName, enumValues<E>(), enumValues<E>().map { it.name }.toTypedArray())
-    }
 }
 
 // Binary backwards-compatible with plugin
--- a/runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
@@ -1,27 +1,28 @@
+/*
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 @file:Suppress("LeakingThis")
 
 package kotlinx.serialization.json.internal
 
 import kotlinx.serialization.*
-import kotlinx.serialization.internal.*
+import kotlinx.serialization.internal.EnumDescriptor
 import kotlinx.serialization.json.*
-import kotlinx.serialization.modules.*
-import kotlin.jvm.*
+import kotlinx.serialization.modules.SerialModule
+import kotlin.jvm.JvmField
 
 internal fun <T> Json.readJson(element: JsonElement, deserializer: DeserializationStrategy<T>): T {
-    val descriptor = deserializer.descriptor
-    if (element is JsonNull) {
-        // TODO temporary workaround (?)
-        require(descriptor.isNullable) { "Read JsonNull and expected nullable descriptor, but has $descriptor" }
-        @Suppress("NULL_FOR_NONNULL_TYPE")
-        return null
-    }
-
-    val input = when (descriptor.kind) {
-        StructureKind.LIST -> JsonTreeListInput(this, cast(element))
-        StructureKind.MAP -> JsonTreeMapInput(this, cast(element))
-        is PrimitiveKind -> JsonPrimitiveInput(this, cast(element))
-        else -> JsonTreeInput(this, cast(element))
+    val input = when (element) {
+        is JsonObject -> JsonTreeInput(this, element)
+        is JsonArray -> JsonTreeListInput(this, element)
+        is JsonLiteral -> JsonPrimitiveInput(this, element)
+        is JsonNull -> {
+            deserializer.descriptor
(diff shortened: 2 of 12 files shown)


runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -66,6 +54,14 @@ open class CommonEnumSerializer<T>(val serialName: String, val choices: Array<T>
             { "$index is not among valid $serialName choices, choices size is ${choices.size}" }
         return choices[index]
     }
+
+    companion object {
+        /**
+         * Factory method for easier instantiation from user/test code.
+         */
+        inline fun <reified E: Enum<E>> create(serialName: String): CommonEnumSerializer<E> =
+            CommonEnumSerializer(serialName, enumValues<E>(), enumValues<E>().map { it.name }.toTypedArray())
```
qwwdfsad: 1) Unnecessary type argument in `enumValue`
2) This API is only used in tests, thus probably shouldn't be public
3) More Kotlin-ish API would be constructor-like functions:  `inline fun <reified E: Enum<E>>CommonEnumSerializer(serialName: String)` without companion

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -1,27 +1,28 @@
...
-        else -> JsonTreeInput(this, cast(element))
+    val input = when (element) {
+        is JsonObject -> JsonTreeInput(this, element)
+        is JsonArray -> JsonTreeListInput(this, element)
+        is JsonLiteral -> JsonPrimitiveInput(this, element)
+        is JsonNull -> {
+            deserializer.descriptor
+                .also { desc -> require(desc.isNullable) { "Read JsonNull and expected nullable descriptor, but has $desc" } }
+            @Suppress("NULL_FOR_NONNULL_TYPE")
+            return null
```
qwwdfsad: Semantically, it is an unchecked cast: `@Suppress("UNCHECKED_CAST") return null as T`
I should've written this in the first place

runtime/commonTest/src/kotlinx/serialization/json/serializers/JsonNativePrimitivesTest.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package kotlinx.serialization.json.serializers
+
+import kotlinx.serialization.SampleEnum
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.json.JsonTestBase
+import kotlin.test.Test
+
+class JsonNativePrimitivesTest : JsonTestBase() {
```
qwwdfsad: Please add tests for nullable primitives with `null` as data as well

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonOutput.kt
```diff
@@ -118,6 +111,28 @@ private sealed class AbstractJsonTreeOutput(
...
+    companion object {
+        const val primitive = "primitive"
+    }
+
+    init {
+        pushTag(primitive)
+    }
+
+    override fun putElement(key: String, element: JsonElement) {
+        require(key == primitive)
```
qwwdfsad: 1) [nitpick] you can use referential comparison here :) 
2) Please add a corresponding error message 
3) It is better to use top-level const to avoid blowing the binary with unnecessary empty classes

runtime/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt
```diff
@@ -78,26 +78,26 @@ class JsonPrimitiveSerializerTest : JsonTestBase() {
         assertEquals(JsonPrimitiveWrapper(JsonLiteral("239")), strict.parse(JsonPrimitiveWrapper.serializer(), string, useStreaming))
     }
 
-    @Test // TODO Top-level nulls are not supported in tagged encoders
-    fun testTopLevelPrimitive() {//} = parametrizedTest { useStreaming ->
-        val string = strict.stringify(JsonPrimitiveSerializer, JsonLiteral(42))
+    @Test
```
qwwdfsad: `JsonNullSerializerTest` contains commented code waiting for this change as well

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonOutput.kt
```diff
@@ -68,8 +56,13 @@ private sealed class AbstractJsonTreeOutput(
     }
 
     override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) {
-        encodePolymorphically(serializer, value) {
-            writePolymorphic = true
+        if (serializer.descriptor.kind !is PrimitiveKind && serializer.descriptor.kind !== UnionKind.ENUM_KIND || currentTagOrNull != null) {
+            encodePolymorphically(serializer, value) { writePolymorphic = true }
+        } else {
+            JsonTreePrimitiveOutput(json, nodeConsumer).apply {
```
qwwdfsad: IMO, `apply` is not applicable here semantically, because it creates a false sense that you are somehow changing it in `block`, but this is not true. And every indentation level makes code less readable.
E.g.
```
if (serializer.descriptor.kind !is PrimitiveKind && serializer.descriptor.kind !== UnionKind.ENUM_KIND || currentTagOrNull != null) {
    return encodePolymorphically(serializer, value) { writePolymorphic = true }
}
val output = JsonTreePrimitiveOutput(json, nodeConsumer)
output.encodeSerializableValue(serializer, value)
output.endEncode(serializer.descriptor)
```

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonOutput.kt
```diff
@@ -68,8 +56,13 @@ private sealed class AbstractJsonTreeOutput(
     }
 
     override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) {
-        encodePolymorphically(serializer, value) {
-            writePolymorphic = true
+        if (serializer.descriptor.kind !is PrimitiveKind && serializer.descriptor.kind !== UnionKind.ENUM_KIND || currentTagOrNull != null) {
```
qwwdfsad: Please leave a short comment why this check is here (especially `currentTagOrNull`)

qwwdfsad:
Also, it closes #287

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: It would be convenient if someone would be willing to create this serializer manually. However, I'm not sure that this use-case will be very popular

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonOutput.kt
sandwwraith: Well, then `JsonPrimitiveInput` should be refactored as well. I think I can share the constant between these classes

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonOutput.kt
sandwwraith: Well, actually, calls to `encode` change output state somehow, so I don't see problems with `apply`. Maybe something like this?

```kotlin
if (/**...**/) {
    encodePolymorphically(serializer, value) { writePolymorphic = true }
} else JsonTreePrimitiveOutput(json, nodeConsumer).apply {
    encodeSerializableValue(serializer, value)
    endEncode(serializer.descriptor)
}
```

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
qwwdfsad: Then it's better not to introduce it in public API I suppose

sandwwraith:
Merged manually

Issue #454, opened 2019-05-08
Better Error Messages For Mapper Deserialization


jeffreyfjohnson:
**What is your use-case and why do you need this feature?**
I am using `Mapper` to serialize and deserialize. Right now if I mark a field in my classes as a `Double`, but when the type of that field sent down from the SDK we use to fetch data is actually a `Long`, we get an error message along the lines of "MyClassSerializer: Long cannot be cast to Double". This is just the most simple example, it applies to any deserialization error.

**Describe the solution you'd like**
It would be very helpful to add the `currentTag` to the error message, e.g. "MyClassSerializer: Error when deserializing tag myNumber: Long cannot be cast to Double". I think I could achieve this by subclassing `Mapper` and wrapping all the `decode` calls in a `try catch` but it would be great if that was just a feature of the library.

I think this would be especially helpful because, at least if I am remembering correctly, `currentTag` will accurately reflect the nesting level of the thing it failed to decode, i.e. `myNumbers.7`. Here we would know that the value at index 7 in the array `myNumbers` was the issue.

Having this would speed up debugging significantly

jeffreyfjohnson:
This is what I have so far. I don't have the full context on this problem, so if someone who days can guide me I am happy to expand on this and put in a PR

```
    override fun decodeTaggedUnit(tag: String): Unit = tryDecode(tag)
    override fun decodeTaggedBoolean(tag: String): Boolean = tryDecode(tag)
    override fun decodeTaggedByte(tag: String): Byte = tryDecode(tag)
    override fun decodeTaggedShort(tag: String): Short = tryDecode(tag)
    override fun decodeTaggedInt(tag: String): Int = tryDecode(tag)
    override fun decodeTaggedLong(tag: String): Long = tryDecode(tag)
    override fun decodeTaggedFloat(tag: String): Float = tryDecode(tag)
    override fun decodeTaggedDouble(tag: String): Double = tryDecode(tag)
    override fun decodeTaggedChar(tag: String): Char = tryDecode(tag)
    override fun decodeTaggedString(tag: String): String = tryDecode(tag)
    override fun decodeTaggedEnum(tag: String, enumDescription: EnumDescriptor): Int = tryDecode(tag)

    private inline fun <reified T : Any> tryDecode(tag: String): T {
        return try {
            decodeTaggedValue(tag) as T
        } catch (e: Exception) {
[...]

PR #475, opened 2019-05-30
Improve Unmap Exception Message

jeffreyfjohnson:
Add the `tag` of the value that failed to deserialize in `TaggedDecoder.decodeTagged*` to the thrown exception.

Addresses #454


sandwwraith:
Closing as obsolete, since `TaggedEncoder` is now an internal implementation detail and Mapper was renamed

Issue #480, opened 2019-06-07
Complex map keys cannot be polymorphically deserialized (JSON)


aquamme:
**Describe the bug**
When attempting to use polymorphic deserialization on a type which contains a map with complex keys, the following error is thrown:

```
Expected class kotlinx.serialization.json.JsonObject but found class kotlinx.serialization.json.JsonArray
java.lang.IllegalStateException: Expected class kotlinx.serialization.json.JsonObject but found class kotlinx.serialization.json.JsonArray
	at kotlinx.serialization.json.internal.AbstractJsonTreeInput.beginStructure(TreeJsonInput.kt:189)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.patch(CollectionSerializers.kt:41)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.deserialize(CollectionSerializers.kt:61)
	at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:33)
	at kotlinx.serialization.json.internal.AbstractJsonTreeInput.decodeSerializableValue(TreeJsonInput.kt:48)
	at kotlinx.serialization.TaggedDecoder$decodeSerializableElement$1.invoke(Tagged.kt:257)
	at kotlinx.serialization.TaggedDecoder.tagBlock(Tagged.kt:270)
	at kotlinx.serialization.TaggedDecoder.decodeSerializableElement(Tagged.kt:257)
	at sample.SampleTests$Derived$$serializer.deserialize(SampleTests.kt)
	at sample.SampleTests$Derived$$serializer.deserialize(SampleTests.kt:15)
	at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:33)
	at kotlinx.serialization.json.internal.AbstractJsonTreeInput.decodeSerializableValue(TreeJsonInput.kt:48)
	at kotlinx.serialization.CoreKt.decode(Core.kt:79)
	at kotlinx.serialization.json.internal.TreeJsonInputKt.readJson(TreeJsonInput.kt:27)
	at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:41)
	at kotlinx.serialization.json.internal.StreamingJsonInput.decodeSerializableValue(StreamingJsonInput.kt:29)
	at kotlinx.serialization.CoreKt.decode(Core.kt:79)
[...]

PR #492, opened 2019-06-20
Implement 'allowStructuredMapKeys' flag.

### LINKED ISSUE #424: Deprecate complex keys in map serializer and provide an additional opt-in option instead
This feature currently is fragile, produces unexpected JSON format by default and interferes with other format-specific features in bizarre ways.

Such behaviour should require explicit opt-in

sandwwraith:
* disable map2array transformation by default for Json
* Fix #424
* Improve string representation of kinds
* Rise deprecation warning for Json constructor.
* Rework JsonException hierarchy


runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -214,14 +187,24 @@ public class JsonBuilder {
     public var encodeDefaults: Boolean = true
     public var strictMode: Boolean = true
     public var unquoted: Boolean = false
+    public var allowStructuredMapKeys: Boolean = false
```
qwwdfsad: I would stick with the verb "enable", we enable features, not allow them

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -44,3 +45,9 @@ public class JsonParsingException(position: Int, message: String) : JsonExceptio
  * ```
  */
 public class JsonElementTypeMismatchException(key: String, expected: String) : JsonException("Element $key is not a $expected")
+
+public class JsonMapInvalidKeyKind(keyDescriptor: SerialDescriptor) : JsonException(
```
qwwdfsad: Please do not introduce new public exception type for a rare non-recoverable situation, which is caught and fixed by the developer in unit-tests. A verbose exception is enough

runtime/commonTest/src/kotlinx/serialization/json/JsonMapKeysTest.kt
```diff
@@ -0,0 +1,58 @@
...
+@Serializable
+private data class WithMap(val map: Map<Long, Long>)
+
+@Serializable
+private data class WithEnum(val map: Map<SampleEnum, Long>)
+
+@Serializable
+private data class WithComplexKey(val map: Map<IntData, String>)
+
+class JsonMapKeysTest : JsonTestBase() {
```
qwwdfsad: Now polymorphic serialization with complex keys should work out of the box (#480).
Please uncomment `JsonMapPolymorphismTest.testPolymorphicKeys` and add more tests for that

runtime/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
```diff
@@ -15,15 +21,26 @@ internal enum class WriteMode(@JvmField val begin: Char, @JvmField val end: Char
...
     }
+
+internal inline fun <T, R1 : T, R2 : T> Json.selectMapMode(
+    mapDescriptor: SerialDescriptor,
+    ifMap: () -> R1,
+    ifList: () -> R2
+): T {
+    val keyDescriptor = mapDescriptor.getElementDescriptor(0)
+    val keyKind = keyDescriptor.kind
+    return if (keyKind is PrimitiveKind || keyKind == UnionKind.ENUM_KIND) {
```
qwwdfsad: Should this place be tweaked when we support object serialization?

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: I'm not sure that `enableStructuredMapKeys` is OK, because structured map keys exist (are enabled) in the user's data model regardless of JSON settings. `enableTransformationOfStructuredMapKeys` sounds more correct, but it's too long

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
sandwwraith: Problem is that generic `JsonException` is a sealed (i.e. abstract) class. I think we need to revisit json exception hierarchy

runtime/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
sandwwraith: Probably, I'll take a look at it

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
qwwdfsad: Let's discuss it tomorrow then

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptionHelpers.kt
```diff
@@ -0,0 +1,33 @@
...
+            "You can disable strict mode to serialize such values"
+)
+
+internal fun JsonInvalidValueInStrictModeException(value: Double) = JsonEncodingException(
+    "$value is not a valid double as per JSON spec.\n" +
+            "You can disable strict mode to serialize such values"
+)
+
+
+internal fun jsonUnknownKeyException(position: Int, key: String) = JsonDecodingException(
```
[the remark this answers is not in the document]
sandwwraith: Declaration from capital letter clashes with deprecated (compatibility) `typealias JsonUnknownKeyException = JsonDecodingException` since they both have the signature (Int, String)

runtime/commonMain/src/kotlinx/serialization/Exceptions.kt
```diff
@@ -1,29 +1,18 @@
...
  */
 
-@file:Suppress("OPTIONAL_DECLARATION_USAGE_IN_NON_COMMON_SOURCE")
 package kotlinx.serialization
 
 import kotlin.jvm.JvmOverloads
 
-open class SerializationException @JvmOverloads constructor(message: String, cause: Throwable? = null)
-    : RuntimeException(message, cause)
+open class SerializationException @JvmOverloads constructor(message: String, cause: Throwable? = null) :
```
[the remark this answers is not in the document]
sandwwraith: It is for simpler instantiation on the plugin side

sandwwraith:
Rebased on `dev` & merged manually

PR #497, opened 2019-06-24
Adapted JSON-specification test suite and follow-up fixes

qwwdfsad:
* Trailing, leading and excessive commas are prohibited now in `JsonArray`
 * Prohibit missing commas in arrays, e.g. `[1 2]` is no longer valid `JsonArray`
 * Add more contextual information to exceptions
 * Remove irrelevant exception


runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -175,32 +175,58 @@ internal class JsonReader(private val source: String) {
         length += addLen
     }
 
+    fun nextJsonKey() {
+        val source = source
+        var currentPosition = currentPosition
+        val maxLen = source.length
+        // TODO this one can be optimized when we oll our benchmarks
```
sandwwraith: *roll ?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -175,32 +175,58 @@ internal class JsonReader(private val source: String) {
         length += addLen
     }
 
+    fun nextJsonKey() {
```
sandwwraith: Can't you reuse `nextToken` with additional check `tokenClass == TC_OTHER || tokenClass == TC_STRING`?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -175,32 +175,58 @@ internal class JsonReader(private val source: String) {
...
+        val source = source
+        var currentPosition = currentPosition
+        val maxLen = source.length
+        // TODO this one can be optimized when we oll our benchmarks
+        while (true) {
+            if (currentPosition >= maxLen) {
+               fail("Expected next key name, but had trailing comma instead", currentPosition)
```
sandwwraith: Why not `Unexpected EOF`? Json can't end on trailing comma without `}` or `]` anyway

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -64,38 +64,55 @@ internal class StreamingJsonInput internal constructor(
...
-                    return when (entryIndex++) {
-                        0 -> 0
-                        1 -> 1
-                        else -> {
-                            entryIndex = 0
-                            CompositeDecoder.READ_DONE
-                        }
+            }
+            WriteMode.MAP -> {
+                if (currentIndex % 2 == 0 && reader.tokenClass == TC_COLON) reader.nextToken()
```
sandwwraith: Should we also ban leading/trailing commas in maps in objects?

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonOutput.kt
```diff
@@ -147,15 +147,15 @@ internal class StreamingJsonOutput(private val composer: Composer, override val
 
     override fun encodeFloat(value: Float) {
         if (configuration.strictMode && !value.isFinite()) {
-            throw JsonInvalidValueInStrictModeException(value)
```
sandwwraith: You can make `JsonInvalidValueInStrictModeException` constructor-like function instead of copy-pasting exception text

runtime/commonTest/src/kotlinx/serialization/json/JsonParserTest.kt
```diff
@@ -74,4 +74,18 @@ class JsonParserFuzzerTest {
...
+    @Test
+    fun testTrailingComma() {
+        testTrailingComma("{\"id\":0,}")
+        testTrailingComma("{\"id\":0  ,}")
+        testTrailingComma("{\"id\":0  , ,}")
+    }
+
+
+    private fun testTrailingComma(content: String) {
+        val e = assertFailsWith<JsonParsingException> {  Json.plain.parseJson(content) }
```
sandwwraith: Make this a `parameterizedTest` to test both streaming and tree parsing?

runtime/jvmTest/resources/corner_cases/listing.txt
```diff
@@ -0,0 +1,18 @@
+number_1.0.json
```
sandwwraith: Do we have a 'golden value' for this tests or this is just a parser check, without correctness verification?

qwwdfsad: We don't, unfortunately. I, of course, can hardcode it manually, but I am not sure this is a work that makes any sense

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
qwwdfsad: Good catch!
I've left only polymorphic array as is as the least used code path

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: I am convinced that it is much more maintainable and readable to have two similar (but not exactly, not only in the additional check) methods

runtime/commonTest/src/kotlinx/serialization/json/JsonParserTest.kt
qwwdfsad: JsonParserTest tests only `parseJson` method, it is not applicable here

qwwdfsad:
>rendered on GitHub as a binary file, bug or strange permissions?

It is because it has invalid UTF-8 symbols in a comment :)

>Can we run spec tests for all platforms, not only JVM? I see that we have to work with resources somehow, but our users will have the same problem sooner or later.

We probably can, but it requires a lot of effort without real gains.
1) We should either work with resources (prerequisites: files, IO, resources in the bundle) or rewrite each test manually
2) We should handle different kinds of UTF-8 (not yet designed charsets) **consistently** 
3) This code is common-only, so the most meaningful part of the parser is covered enough

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
```diff
@@ -13,47 +13,50 @@
...
         annotations.add(mutableListOf())
     }
 
-    fun pushAnnotation(a: Annotation) {
+    public fun pushAnnotation(a: Annotation) {
         annotations.last().add(a)
     }
 
-    fun pushDescriptor(desc: SerialDescriptor) {
+    public fun pushClassAnnotation(a: Annotation) {
```
pdvrieze: Any reasons that these changes are in the patch set? I'm not saying the change isn't good, but especially marking things public (when they are by default) is unrelated.

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -4,36 +4,18 @@
...
-    "Strict JSON encountered unknown key: $key\n" +
-            "You can disable strict mode to skip unknown keys")
-
-
 /**
  * Exception thrown when [Json] has failed to parse provided JSON.
  * Such exception usually indicate that [Json] input is not a valid JSON.
  */
-public class JsonParsingException(position: Int, message: String) : JsonException("Invalid JSON at $position: $message")
+public class JsonParsingException(message: String) : JsonException("Invalid JSON: $message")
```
pdvrieze: Why drop the position from the exception?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
pdvrieze: What is an option is to have an inline function that implements the common code and then use parameters to split out the non-shared code. Something like:

```
inline fun nextTokenImpl(onElse: (Char) -> Unit, onEnd: () -> Unit) {
    val source = source
    val currentPosition = currentPosition
    val maxLen = source.length
    while (currentPosition<maxLen) {
        val ch = source[currentPosition]
        when(charToTokenClass(ch) {
            TC_WS -> CurrentPosition++ // skip whitespace
            TC_OTHER -> {
                nextLiteral(source, currentPosition)
                return
            }
            TC_STRING -> {
                nextLiteral(source, currentPosition)
                return
            }
            else -> {
                onElse(ch)
            }
        }
    }
    onEnd()
}

fun nextJsonKey() = nextTokenImpl(onElse = {
    fail("Expected next key name, but had trailing comma instead", currentPosition)
}, onEnd = {
    fail("Unexpected EOF", currentPosition)
})

fun nextToken() = nextTokenImpl(onElse = {
    this.tokenPosition = currentPosition
    this.tokenClass = tc
    this.currentPosition = currentPosition + 1
    return
}, onEnd = {
    tokenPosition = currentPosition
    tokenClass = TC_EOF
})
````

File: runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt [hunk 10]
```diff
@@ -221,35 +241,35 @@ internal class JsonReader(private val source: String) {
         tokenClass = if (rangeEquals(source, offset, length, NULL)) TC_NULL else TC_OTHER
     }
 
-    private fun nextString(source: String, startPos: Int) {
-        tokenPosition = startPos
+    private fun nextString(source: String, statPosition: Int) {
```
pdvrieze: typo. Should be sta**r**tPosition

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -221,35 +241,35 @@ internal class JsonReader(private val source: String) {
...
+        tokenPosition = statPosition
         length = 0 // in buffer
-        var curPos = startPos + 1
-        var lastPos = curPos
-        val maxLen = source.length
+        var currentPosition = statPosition + 1
+        var lastPosition = currentPosition
+        val length = source.length
```
pdvrieze: Inconsistent name with previous code. Probably use maxLength

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufReaderTest.kt
```diff
@@ -14,49 +14,49 @@ class ProtobufReaderTest {
 
     @Test
     fun readSimpleObject() {
-        ProtoBuf.loads<TestInt>(TestInt.serializer(), "08ab02") shouldBe t1
+        ProtoBuf.loads(TestInt.serializer(), "08ab02") shouldBe t1
```
pdvrieze: Removing the type parameters here is valid in principle, but shouldn't really be part of this pull request/patch set.

runtime/jvmTest/resources/corner_cases/listing.txt
pdvrieze: What is quite possible is to test through roundtrips. In particular for valid values you can just serialize the result of deserialization and check that it is the same (and or equivalent - maybe ignore line breaks or something).

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: Yes, it is possible. But I don't think that it makes the code more readable or understandable, rather the opposite. The ultimate goal here is not to write as much generic as possible code, but rather as much straightforward as possible

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
qwwdfsad: It's just a maintenance work that is much easier to do when you pass by.
We are generally ok with such changes within the team

qwwdfsad: >but especially marking things public

To clarify, this is the code style we (libraries team) stick with: if something is `public`, then it has to have explicit `public` modifier (see also [KEEP-45](https://github.com/Kotlin/KEEP/issues/45))

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: Removed this method, resolving

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
```diff
@@ -108,5 +103,10 @@ open class SerialClassDescImpl @JvmOverloads constructor(
         return result
     }
 
-    override fun toString() = "$name$names"
+    override fun toString(): String {
+        return indices.entries.joinToString(", ", "$name(", ")") { it.key + ": " + getElementDescriptor(it.value) }
```
sandwwraith: I don't think that including whole descriptor from `getElementDescriptor` to the String output is reasonable – it will recursively dump whole descriptor tree and bloat output. I think `getElementDescriptor(it.value).name` is sufficient

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -165,94 +164,86 @@ internal class JsonReader(private val source: String) {
...
     private fun nextLiteral(source: String, startPos: Int) {
         tokenPosition = startPos
         offset = startPos
-        var curPos = startPos
-        val maxLen = source.length
-        while (true) {
-            curPos++
-            if (curPos >= maxLen || charToTokenClass(source[curPos]) != TC_OTHER) break
+        var currentPosition = startPos
+        while (currentPosition < source.length && charToTokenClass(source[currentPosition]) == TC_OTHER) {
```
sandwwraith: Will `source.length` be optimized down to the local var to avoid getter call how it was with `val maxLen = source.length`?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -165,94 +164,86 @@ internal class JsonReader(private val source: String) {
...
-                curPos = newPos
-                lastPos = newPos
+        var currentPosition = startPosition + 1
+        var lastPosition = currentPosition
+        val length = source.length
+        while (source[currentPosition] != STRING) {
+            if (currentPosition >= length) fail("Unexpected end in string", currentPosition)
+            if (source[currentPosition] == STRING_ESC) {
+                appendRange(source, lastPosition, currentPosition)
+                val newPos = appendEsc(source, currentPosition + 1)
```
sandwwraith: `newPosition` so it will look consistent?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -165,94 +164,86 @@ internal class JsonReader(private val source: String) {
...
-            } else if (source[curPos] == STRING_ESC) {
-                appendRange(source, lastPos, curPos)
-                val newPos = appendEsc(source, curPos + 1)
-                curPos = newPos
-                lastPos = newPos
+        var currentPosition = startPosition + 1
+        var lastPosition = currentPosition
+        val length = source.length
+        while (source[currentPosition] != STRING) {
+            if (currentPosition >= length) fail("Unexpected end in string", currentPosition)
```
sandwwraith: "Unexpected end **in** string" or "**of** string"? I know this is not your text, but since we are fixing exceptions messages anyway...

runtime/commonTest/src/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -133,4 +134,9 @@ abstract class JsonTestBase {
             assertEquals(data, deserialized)
         }
     }
+
+    inline fun <reified T : Throwable> assertFailsWith(message: String, block: () -> Unit) {
```
sandwwraith: I suggest renaming to `assertFailsWithMessage` because stdlib has very similar overload (https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html#assertfailswith), but `message` there is message used in test assertion.

runtime/commonTest/src/kotlinx/serialization/json/serializers/JsonArraySerializerTest.kt
```diff
@@ -40,6 +36,47 @@ class JsonArraySerializerTest : JsonTestBase() {
         assertEquals(string, prebuiltJson.toString())
     }
 
+    @Test
+    fun testMissingCommas() = parametrizedTest { useStreaming ->
+        val message = "Expected end of the array or comma"
+        assertFailsWith<JsonDecodingException>(message) { nonStrict.parse(JsonArraySerializer, "[a b c]", useStreaming) }
```
sandwwraith: Well, this too can be factored out (probably with usage of local functions) like

```kotlin
fun doTest(input: String) = assertFailsWith<JsonDecodingException>(message) { nonStrict.parse(JsonArraySerializer, input, useStreaming) }
doTest(...)
```

sandwwraith:
> We probably can, but it requires a lot of effort without real gains.
We should either work with resources

Yes, isn't this a time when we should raise a design discussion about **resources** in *Kotlin Multiplatform Programming*?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: Yes, it will.
Note that such low-level optimizations should be profile-guided in the first place

PR #506, opened 2019-07-03
Fix decoding for ProtoBuf when there are missing properties in the model.

kylejbrock:
Noticed that when skipping over tags that weren't defined in the kotlin model, that readTag is being called twice which moves the stream into in invalid position.  Since nextInt, nextLong, and nextObject all call readTag before returning, it's unnecessary and invalid to call again.

### DIFF
--- a/runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
@@ -1,0 +1,138 @@
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.Decoder
+import kotlinx.serialization.Encoder
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerialDescriptor
+import kotlinx.serialization.SerialId
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.internal.SerialClassDescImpl
+import kotlinx.serialization.json.JsonInput
+import kotlinx.serialization.json.JsonOutput
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+
+class ProtobufMissingFieldsTest {
+
+    private val buffer = byteArrayOf(10, 30, 8, 11, 16, 2, 26, 3, 115, 112, 97, 26, 6, 115, 112, 97, 95, 101, 115, 26, 2, 101, 115, 26, 5, 101, 115, 95, 101, 115, 32, 1, 16, 25)
+
+    @Test
+    fun deserialize() {
+        val items = ProtoBuf.load(Items.serializer(), buffer)
+        assertEquals(25, items.pageSize)
+        assertFalse(items.nextPage)
+        assertEquals(1, items.items.size)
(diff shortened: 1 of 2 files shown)


sandwwraith:
Would you mind to add a test for that?

kylejbrock:
It'll probably be on Monday, but yes, I'll add some tests.

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    @SerialId(2) @Serializable(with = ItemPlatformSerializer::class)
+    val platform: ItemPlatform = ItemPlatform.Unknown,
+    @SerialId(3)
+    val language: List<String> = emptyList(),
+    @SerialId(4) @Serializable(with = ItemContextSerializer::class)
+    val context: ItemContext = ItemContext.Unknown
+)
+
+@Serializable
+data class ItemsWithoutFields(
```
sandwwraith: `ItemsWithoutPageSize` sounds more precise to me

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    @SerialId(1)
+    val items: List<ItemWithWithout> = emptyList(),
+    @SerialId(3)
+    val nextPage: Boolean = false
+)
+
+@Serializable
+data class ItemWithWithout(
```
sandwwraith: What's `WithWithout` means? Maybe `WithoutPlatform`?

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
```diff
@@ -0,0 +1,138 @@
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.Decoder
+import kotlinx.serialization.Encoder
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerialDescriptor
+import kotlinx.serialization.SerialId
```
sandwwraith: Please use star-import when a number of imported declarations from a package is more than 3.

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
kylejbrock: Sounds good.

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
kylejbrock: Definite typo.

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufMissingFieldsTest.kt
kylejbrock: Ok.

PR #509, opened 2019-07-05
ByteArray with efficient writing in protobuf. Rest of array serializers

sandwwraith:


runtime/commonMain/src/kotlinx/serialization/internal/CollectionDescriptors.kt
```diff
@@ -66,6 +66,10 @@ internal const val HASH_SET_NAME = "kotlin.collections.HashSet"
 internal const val LINKED_HASH_MAP_NAME = "kotlin.collections.LinkedHashMap"
 internal const val HASH_MAP_NAME = "kotlin.collections.HashMap"
 
+class PrimitiveArrayDescriptor(primitive: PrimitiveDescriptor): ListLikeDescriptor(primitive) {
```
qwwdfsad: should be internal?

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,6 +140,45 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
 
+abstract class PrimitiveArraySerializer<TElement, TCollection, TBuilder : PrimitiveArraySerializer<TElement, TCollection, TBuilder>.Builder>
+internal constructor(
+    primitiveSerializer: KSerializer<TElement>,
+    primitiveDescriptor: PrimitiveDescriptor
+) : ListLikeSerializer<TElement, TCollection, TBuilder>(primitiveSerializer) {
+    final override val descriptor = PrimitiveArrayDescriptor(primitiveDescriptor)
+
+    abstract inner class Builder internal constructor() {
+        abstract val pos: Int
```
qwwdfsad: Please do not use shorthands for overridable properties (same for `objSize` and `objIterator`)

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,6 +140,45 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+        abstract fun ensureCapacity(requiredCapacity: Int = pos + 1)
```
qwwdfsad: All array parsed can be accidentally quadratic. Please use `position * 2` explicitly. I've seen it is used in each particular implementation, this is too error-prone. Additionally, such behaviour (`coerceAtLeast(size * 2)`) prevents precise sizing when size is known in advance (e.g. for "fixed size array" type)

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,6 +140,45 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+    )
+
+    protected abstract fun writeContents(encoder: CompositeEncoder, content: TCollection, size: Int)
+
+    final override fun serialize(encoder: Encoder, obj: TCollection) {
+        val size = obj.objSize()
+        @Suppress("NAME_SHADOWING")
+        val encoder = encoder.beginCollection(descriptor, size, *typeParams)
+        writeContents(encoder, obj, size)
+        encoder.endStructure(descriptor)
```
qwwdfsad: A bit out of the current scope. Could you please remind why do we have separate `beginCollection` and `beginStructure`, but generic `endStructure`?

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -58,15 +58,15 @@ class ProtoBuf(context: SerialModule = EmptyModule) : AbstractSerialFormat(conte
...
             // encode maps as collection of map entries, not merged collection of key-values
-            if (serializer.descriptor is MapLikeDescriptor) {
+            serializer.descriptor is MapLikeDescriptor -> {
                 val serializer = (serializer as MapLikeSerializer<Any?, Any?, T, *>)
                 val mapEntrySerial = MapEntrySerializer(serializer.keySerializer, serializer.valueSerializer)
                 HashSetSerializer(mapEntrySerial).serialize(this, (value as Map<*, *>).entries)
-            } else {
-                serializer.serialize(this, value)
             }
+            serializer.descriptor == ByteArraySerializer.descriptor -> encoder.writeBytes(value as ByteArray, popTag().first)
```
qwwdfsad: non-referential comparison is used on purpose?

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,6 +140,45 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
     }
 }
 
+abstract class PrimitiveArraySerializer<TElement, TCollection, TBuilder : PrimitiveArraySerializer<TElement, TCollection, TBuilder>.Builder>
```
qwwdfsad: internal?

runtime/commonMain/src/kotlinx/serialization/internal/CollectionDescriptors.kt
qwwdfsad: or explicitly public, then documented and with internal ctor

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
qwwdfsad: Also: it isn't `TCollection`, it's `TArray`.
Additionally, I would omit `T` as it does not carry any additional information, but affects readability

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,6 +140,45 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+        abstract fun ensureCapacity(requiredCapacity: Int = pos + 1)
+        abstract fun build(): TCollection
+    }
+
+    final override fun TBuilder.builderSize() = pos
+    final override fun TBuilder.toResult(): TCollection = build()
+    final override fun TBuilder.checkCapacity(size: Int) = ensureCapacity(size)
+
+    // these methods lead to boxing
+    final override fun TCollection.objIterator(): Iterator<TElement> = throw AssertionError("Must not be used")
```
qwwdfsad: Why not? Document this aspect properly.
Please use `error()` to throw ISE, AE is used when internal invariant is violated, not during API misuse

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
sandwwraith: Public classes (such as `IntArraySerializer`) can't have internal supertypes

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
sandwwraith: `beginCollection` has separate parameter `size`

`endStructure` does not need this information

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
sandwwraith: Yes. Since `ByteArraySerializer` is not an object (it has only companion object for convenience, but itself it is a regular class), the descriptor can be instantiated multiple times.

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,27 +128,74 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+) :
+    ListLikeSerializer<Element, Array<Element>, ArrayList<Element>>(eSerializer) {
     override val descriptor = ArrayClassDesc(eSerializer.descriptor)
 
-    override fun Array<E>.objSize(): Int = size
-    override fun Array<E>.objIterator(): Iterator<E> = iterator()
-    override fun builder(): ArrayList<E> = arrayListOf()
-    override fun ArrayList<E>.builderSize(): Int = size
+    override fun Array<Element>.collectionSize(): Int = size
+    override fun Array<Element>.collectionIterator(): Iterator<Element> = iterator()
```
qwwdfsad: `collection` is probably redundant prefix here

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
```diff
@@ -0,0 +1,398 @@
...
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.internal
+
+import kotlinx.serialization.*
+
+private const val INITIAL_SIZE = 0
+
+// it could be an object, but https://youtrack.jetbrains.com/issue/KT-14628
```
qwwdfsad: Making builder outer will allow us to declare serializers as objects

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
```diff
@@ -0,0 +1,398 @@
...
+ * Encode elements one-by-one, as regular list,
+ * unless format's Encoder/Decoder have special handling for this serializer.
+ */
+public class ByteArraySerializer :
+    PrimitiveArraySerializer<Byte, ByteArray, ByteArraySerializer.Builder>(ByteSerializer, ByteDescriptor),
+    KSerializer<ByteArray> {
+    inner class Builder internal constructor() :
+        PrimitiveArraySerializer<Byte, ByteArray, ByteArraySerializer.Builder>.Builder() {
+
+        private var buf: ByteArray = ByteArray(INITIAL_SIZE)
```
qwwdfsad: Performance-wise observation:

1) Most of the arrays definitely have the size bigger than 1, first K resizes will be wasteful. Let's stick to the Java and make it 10 by default.
2) 0 in size requires you to resize as `* 2 + 1`  instead of plain `* 2`
3) Resize is currently quadratic because on each append you ensure that size is at least "position * 2 + 1", effectively doing resize on **each** append

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
```diff
@@ -0,0 +1,398 @@
...
+        override var position: Int = 0
+            private set
+
+        constructor(filledBuf: ByteArray): this() {
+            buf = filledBuf
+            position = filledBuf.size
+        }
+
+        override fun ensureCapacity(requiredCapacity: Int) {
```
qwwdfsad: This method should check that provided capacity is >= -1 if invoked from other classes and shouldn't if invoked from `append`

runtime/commonMain/src/kotlinx/serialization/internal/CollectionDescriptors.kt
```diff
@@ -66,6 +70,16 @@ internal const val HASH_SET_NAME = "kotlin.collections.HashSet"
 internal const val LINKED_HASH_MAP_NAME = "kotlin.collections.LinkedHashMap"
 internal const val HASH_MAP_NAME = "kotlin.collections.HashMap"
 
+/**
+ * Descriptor for primitive arrays, such as [IntArray], [DoubleArray], etc...
+ *
+ * Can be obtained from corresponding serializers (e.g. [ByteArraySerializer.descriptor])
+ */
+public class PrimitiveArrayDescriptor internal constructor(primitive: PrimitiveDescriptor) :
```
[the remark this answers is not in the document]
sandwwraith: Well, it's IDEA auto-formatting on long lines. I don't mind to split it further.

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,27 +128,74 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+
+    abstract inner class Builder internal constructor() {
+        abstract val position: Int
+        abstract fun ensureCapacity(requiredCapacity: Int = position * 2 + 1)
+        abstract fun build(): Array
+    }
+
+    final override fun Builder.builderSize() = position
+    final override fun Builder.toResult(): Array = build()
+    final override fun Builder.checkCapacity(size: Int) = ensureCapacity(size)
```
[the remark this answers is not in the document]
sandwwraith: It would clash with member function on substitution `Builder = ArrayList`

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,27 +128,74 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
+        val encoder = encoder.beginCollection(descriptor, size, *typeParams)
+        writeContents(encoder, obj, size)
+        encoder.endStructure(descriptor)
+    }
+}
+
 // todo: can be more efficient when array size is know in advance, this one always uses temporary ArrayList as builder
-class ReferenceArraySerializer<T: Any, E: T?>(private val kClass: KClass<T>, eSerializer: KSerializer<E>):
-        ListLikeSerializer<E, Array<E>, ArrayList<E>>(eSerializer) {
+class ReferenceArraySerializer<ElementKlass : Any, Element : ElementKlass?>(
```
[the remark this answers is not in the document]
sandwwraith: Since it's effectively compiled into the client's code, it is incorrect to insert internal here.

It is also not unused: https://github.com/kotlin/kotlinx.serialization/blob/45aa7c7ab97f25f205408b46a7069886408bce6b/runtime/jvmMain/src/kotlinx/serialization/JvmResolving.kt#L58

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
sandwwraith: Same motivation as for `builderSize`

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
sandwwraith: Buffer size is not negative, so `buf.size < requiredCapacity` is always false on capacity < 0. Why do we need an additional check?

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -140,27 +128,74 @@ sealed class MapLikeSerializer<TKey, TVal, TCollection, TBuilder: MutableMap<TKe
...
 }
 
+public abstract class PrimitiveArraySerializer<Element, Array, Builder : PrimitiveArraySerializer<Element, Array, Builder>.Builder>
+internal constructor(
+    primitiveSerializer: KSerializer<Element>,
+    primitiveDescriptor: PrimitiveDescriptor
+) : ListLikeSerializer<Element, Array, Builder>(primitiveSerializer) {
+    final override val descriptor = PrimitiveArrayDescriptor(primitiveDescriptor)
+
+    abstract inner class Builder internal constructor() {
```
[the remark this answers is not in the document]
sandwwraith: It would not simplify things much, just one generic parameter will be shorter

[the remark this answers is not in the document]
sandwwraith: It would be also simplier to discover – since it is a type argument of a public class, it would be impossible to make builders (implementation detail) internal

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
sandwwraith: It would not be quadratic due to the `coerceAtLeast` part, would it? If size is known in advance and was set during `ensureCapacity(readSize())`, then `if (buf.size < requiredCapacity)` would never be true and resizing would not be called

sandwwraith: Well, it would be true in one case: when INITIAL_SIZE = 10 and `readSize()` returned something from 10 to 20 (say, 15), we'll still allocate 20 slots. I'm not sure that this is a big concern though.

sandwwraith: > Please use position * 2 explicitly.

That will cause resize on each append, which you correctly noticed later in comments

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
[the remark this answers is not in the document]
qwwdfsad: Now we get rid of one more nesting level (that does not help in readability and maintainability).
Making them internal is not an option, because the type is exposed in generic bounds :(

But It would be nice top mark them as `@InternalSerializationApi` when #541 is merged

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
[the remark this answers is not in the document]
qwwdfsad: Thanks.
`@InternalSerializationApi`?

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
qwwdfsad: Because that's clearly a programmer error (e.g. off-by-one error).
And the best action here is to notify the programmer about such error, not to ignore it.

sandwwraith:
Good. I'll wait until you merge #541, then rebase+squash and add `@InternalSerializationApi` in proper places.

Issue #448, opened 2019-05-01
Polymorphic serializer, ignore unregistered objects when deserializing list of polymorphic objects


luca992:
I have made a serializer class for the json:api specificaiton.

```
@Serializable
data class JsonApiResponseItem(
        override val jsonapi: JsonApiVersion,
        override val included: List<@Polymorphic JsonApiObject>? = null,
        @Polymorphic val data: JsonApiObject): JsonApiResponse() {

}
```

`included` can return many different types of objects which I am successfully using the polymorphic feature to deserialize. But this only works if all possible object types are registered in the scope of class JsonApiObject  for polymorphic serialization.

If any types are not registered an error like is thrown

```
io.ktor.client.call.ReceivePipelineException: Fail to run receive pipeline: 
kotlinx.serialization.SerializationException: example-type is not registered for polymorphic 
serialization in the scope of class api.jsonapi.JsonApiObject
```

Is it possible to ignore deserializing polymorphic type objects which are not registered for polymorphic serialization without throwing an error? This would really be helpful because it would prevent deserialization from throwing errors if any new objects added to the `included` list... and it would be convenient if don't actually need to deserialize some of the included objects

sandwwraith:
I think this feature is more about to __skip deserialization of element in the list if an error occured__. Because polymorphic serializer does not know whether it is in the list now or not, and it has to return something (e.g. if the same error occurs when deserializing the `data` field from your example, there is no meaningful way to recover). Also, this requires from each encoder the support of __skipping malformed element__, so the list deserializer could start deserializing next one.

luca992:
Yeah, that sounds right. 
Maybe another option would be to allow this in nonstrict mode by allowing a backup serializer to be registered for polymorphic classes in SerializersModule.

maybe something along the lines of:

```
@Serializable
data class EmptyJsonApiObject : JsonApiObject()

Json(context = SerializersModule {
    polymorphic(JsonApiObject::class) {
        JsonApiObjectUser::class with JsonApiObjectUser.serializer()
        JsonApiObjectEmployee::class with JsonApiObjectEmployee.serializer()
        addBackUpSerializer(EmptyJsonApiObject.serializer())
    }
},
        configuration = JsonConfiguration(
                strictMode = false,
        )
)

//addBackUpSerializer would require EmptyJsonApiObject to extend JsonApiObject
```


[...]

PR #514, opened 2019-07-10
Default polymorphic serializer for unregistered types

kklorenzotesta:
Allows to configure a default serializer that will be used in serialization/deserialization when a subtype that is not registered is found.

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
@@ -13,6 +13,7 @@ import kotlin.reflect.*
  * A builder which registers all its content for polymorphic serialization in the scope of [baseClass].
  * If [baseSerializer] is present, registers it as a serializer for [baseClass] (which will be used if base class is serializable).
  * Subclasses with its serializers can be added via [addSubclass] or [with].
+ * A Default polymorphic serializer can be set via [setDefaultSerializer], it will be used in case a type is not registered.
  *
  * To obtain an instance of this builder, use [SerializersModuleBuilder.polymorphic] DSL function.
  */
@@ -22,6 +23,7 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
     private val baseSerializer: KSerializer<Base>? = null
 ) {
     private val subclasses: MutableList<Pair<KClass<out Base>, KSerializer<out Base>>> = mutableListOf()
+    private val default: MutableList<KSerializer<Base>> = mutableListOf()
 
     internal fun buildTo(module: SerialModuleImpl) {
         if (baseSerializer != null) module.registerPolymorphicSerializer(baseClass, baseClass, baseSerializer)
@@ -30,6 +32,12 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
                 baseClass,
                 kclass as KClass<Base>,
                 serializer as KSerializer<Base>
+            )
+        }
+        default.forEach { defaultSerializer ->
+            module.registerPolymorphicDefaultSerializer(
+                baseClass,
+                defaultSerializer
             )
         }
     }
@@ -57,6 +65,13 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
      * @see addSubclass
      */
     public infix fun <T : Base> KClass<T>.with(serializer: KSerializer<T>): Unit = addSubclass(this, serializer)
+
+    /**
+     * Adds a default [serializer] to the resulting module under the initial [baseClass]
+     */
+    public fun setDefaultSerializer(serializer: KSerializer<Base>) {
+        default.add(serializer)
+    }
 
 
     /**
--- a/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
@@ -73,6 +73,19 @@ public class SerializersModuleBuilder internal constructor(@JvmField internal va
     }
 
     /**
+     * Adds [defaultSerializer] as the default serializer in the scope of [baseClass] for polymorphic serialization.
+     * It will be used in case a type that is not registered is found.
+     * Throws [SerializationException] if a module already has a default serializer in the scope of [baseClass].
+     * To overwrite an already registered serializer, [SerialModule.overwriteWith] can be used.
+     */
+    override fun <Base : Any> polymorphicDefault(
+        baseClass: KClass<Base>,
+        defaultSerializer: KSerializer<Base>
+    ) {
+       impl.registerPolymorphicDefaultSerializer(baseClass, defaultSerializer)
+    }
+
+    /**
      * Copies contents of [other] module into current builder.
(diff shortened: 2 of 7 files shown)


pdvrieze:
The code itself seems fine. However I'm not sure that this is a good idea. The test suite only implements a default deserializer that provides a default value. In addition, a deserializer is also responsible for reading in the data for the field - if that doesn't happen the deserialization is going to be in an invalid state. It is possible for a format to help with "skipping" elements (depending on the format - it needs to be able to know the end of the unknown data), but in that case it needs to be signalled that it should do so.

If you really want to support unknown child types you also want to pass on the type information to the deserializer. This is not possible for a `KSerializer<*>`. It may be possible to have a different class that can handle unknown types and gets the correct information. It may then also have a pair result that also provides information on the expected behaviour by the format.

Serializing unknown values is even harder. Can that really work? Normally in a serialization framework you want to support round trips. As in, you are able to write out what you read in. If you have a serializer-like class that gets the type it may be possible to do so. You could have it correspond to a type that just records all information in a generic (bytestream) way. Such a generic way would however be format specific, something like `JsonElement`.

kklorenzotesta:
Thank you, I'm not and english speaker, so I apologize if I misunderstood something:

> The code itself seems fine. However I'm not sure that this is a good idea. 

The problem I'm trying to solve is more or less the same as #448 , I need to deserialize a List across different version of an application, so in some older version a subclass can be missing, in this case I want to provide a default value. Something similar to [JsonTypeInfo#defaultImpl](https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonTypeInfo.html#defaultImpl()) in Jackson.

>In addition, a deserializer is also responsible for reading in the data for the field - if that doesn't happen the deserialization is going to be in an invalid state. 

I didn't think about that and I'm not an expert of this library, isn't enough to call `beginStructure` and `endStructure` to consume the remaining fields?

> If you really want to support unknown child types you also want to pass on the type information to the deserializer. This is not possible for a `KSerializer<*>`. It may be possible to have a different class that can handle unknown types and gets the correct information. 

You mean something like a new class with a method from the unknown type name to a KSerializer?

> It may then also have a pair result that also provides information on the expected behaviour by the format.

I don't understand this sorry, can you provide me some example of possible expected behaviours?

> Serializing unknown values is even harder. Can that really work? Normally in a serialization framework you want to support round trips. As in, you are able to write out what you read in. If you have a serializer-like class that gets the type it may be possible to do so. You could have it correspond to a type that just records all information in a generic (bytestream) way. Such a generic way would however be format specific, something like `JsonElement`.

I didn't originally mean to support serializing unknown types, but because of how SerialModule works it can't be avoided. I don't know if it's useful to support round trips, in general I don't think it makes really sense to serialize an unknown type, since if you have the implementation you can simply register it. Furthermore the `KSerializer` can easily get the type name in serialization from the concrete instance.

pdvrieze:
> The problem I'm trying to solve is more or less the same as #448 , I need to deserialize a List across different version of an application, so in some older version a subclass can be missing, in this case I want to provide a default value. Something similar to [JsonTypeInfo#defaultImpl](https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonTypeInfo.html#defaultImpl()) in Jackson.
> 
Thanks for that, this seems like a valid issue. While in XML it is (relatively) easy to **drop** unsupported subtags, the polymorphic markers used in Json make this harder.


> > In addition, a deserializer is also responsible for reading in the data for the field - if that doesn't happen the deserialization is going to be in an invalid state.
> 
> I didn't think about that and I'm not an expert of this library, isn't enough to call `beginStructure` and `endStructure` to consume the remaining fields?

No, it wouldn't be, depending on the format implementation. Dense binary formats do not always have start and end markers for their structures and require the parser to know the valid structure. This means
that all bits in the inputstream must be consumed. Even for XML this means that the format needs to know to skip elements as it assumes that bits are called correctly. Of course if you are parsing an in memory structure rather than a stream it is easier, but many formats will work on streams.

> > If you really want to support unknown child types you also want to pass on the type information to the deserializer. This is not possible for a `KSerializer<*>`. It may be possible to have a different class that can handle unknown types and gets the correct information.
> 
> You mean something like a new class with a method from the unknown type name to a KSerializer?

Actually I had a look. By far the easiest way to solve your problem (even without requiring any library changes) is to implement a custom `SerialModule`. When `PolymorphicSerializer` tries to look up the serializer for a given type this goes to this SerialModule. In the case there is no registered serializer it is able to return a default serializer (perhaps created on demand for the given type) as you want. Of course this still does not solve the consume content problem in the general case (for specific formats it is possible to get the format specific inputs to skip them).

> 
> > It may then also have a pair result that also provides information on the expected behaviour by the format.
> 
> I don't understand this sorry, can you provide me some example of possible expected behaviours?

The fallback serializer can basically do 3 things:

- consume the entire structure from the format specific input
- not consume any part of the structure from the input (reading is still at the beginning of the structure). The format could then possibly consume the structure. Note that this is format dependent and does not work for all formats.
- throw some sort of exception to signal an error

> 
> > Serializing unknown values is even harder. Can that really work? Normally in a serialization framework you want to support round trips. As in, you are able to write out what you read in. If you have a serializer-like class that gets the type it may be possible to do so. You could have it correspond to a type that just records all information in a generic (bytestream) way. Such a generic way would however be format specific, something like `JsonElement`.
> 
> I didn't originally mean to support serializing unknown types, but because of how SerialModule works it can't be avoided. I don't know if it's useful to support round trips, in general I don't think it makes really sense to serialize an unknown type, since if you have the implementation you can simply register it. Furthermore the `KSerializer` can easily get the type name in serialization from the concrete instance.

If you use a custom SerialModule implementation you can put the fallback on reading only.

kklorenzotesta:
>No, it wouldn't be, depending on the format implementation. Dense binary formats do not always have start and end markers for their structures and require the parser to know the valid structure

I understand but I still think that this is a problem of the user who writes the default serializer. Its his responsability to handle correctly the Decoder according to the format. If the format doesn't support to skip a full structure I think nothing can be done about it.

>Actually I had a look. By far the easiest way to solve your problem (even without requiring any library changes) is to implement a custom SerialModule

Actually I can't. Before creating this PR I tried that, but in the `Json` class constructor the given module is dumped into the default `SerialModuleImpl` (because to the super constructor is passed `context + defaultJsonModule`), so it's impossible to use a custom module implementation unless I reimplement the `Json` class.

pdvrieze:
> 
> Actually I can't. Before creating this PR I tried that, but in the `Json` class constructor the given module is dumped into the default `SerialModuleImpl` (because to the super constructor is passed `context + defaultJsonModule`), so it's impossible to use a custom module implementation unless I reimplement the `Json` class.

This is really unfortunate. There seem to be good reasons why SerialModule works that way, but the fallback option would be welcome. In the meantime two other options are either a custom serializer for the list or a custom format.

jnorthrup:
Interchange File Format https://g.co/kgs/6S9r8z  may have a couple of hints how decoders can use what they understand and pass over what they lack in keys.  Box format defines a skippable atom chunk with a uint or byte[4] type key to signal to decoders

sandwwraith:
Closing in favor of #726

PR #518, opened 2019-07-16
wasm32 is added as a build target

ghasemz:
Just to be able to use the serializer inside wasm32 projects,

```gradle
wasm32Main {
            dependencies {
                implementation kotlin("stdlib")
                implementation "org.jetbrains.kotlinx:kotlinx-serialization-runtime-wasm32:$kotlin_serialization"                  
            }
        }
```

### DIFF
--- a/runtime/build.gradle
+++ b/runtime/build.gradle
@@ -29,7 +29,6 @@ def deployMode = property('native.deploy') == 'true'
 
 kotlin {
     infra {
-        target('wasm32')
         target('macosX64')
         target('linuxX64')
         if (deployMode) {
@@ -37,10 +36,9 @@ kotlin {
             target('iosArm64')
             target('iosArm32')
             target('iosX64')
+            target('wasm32')
         }
     }
-
-    wasm32 {}
 
     targets {
         fromPreset(presets.jvm, 'jvm') {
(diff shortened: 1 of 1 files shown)


runtime/build.gradle
```diff
@@ -29,6 +29,7 @@ def deployMode = property('native.deploy') == 'true'
 
 kotlin {
     infra {
+        target('wasm32')
```
sandwwraith: It sufficient to include only this line, without the rest; `infra` block will automatically handle source sets.

Please put it under `if (deployMode)`. See, for regular development (when you write code, then run tests) for a faster feedback loop better to include only one target – host platform (either linux or macos). Other targets (Windows, ios, webasm) are included under `deployMode` so they're tested and built only on CI in order not to waste developer's time.

ghasemz:
> * Wasm doesn't support exceptions ([JetBrains/kotlin-native#3194](https://github.com/JetBrains/kotlin-native/issues/3194)), so library tests are basically unusable. If you get the binary, it would not be tested, and any exception like `JsonDecodingException` could crash your app

* I think the input to this library needs validation somehow to avoid those corner cases, assuming exceptions are not available in wasm32 right now
* Could we say relying on tests running on many different targets could make this library ready for "experimental" integration in wasm32?

sandwwraith:
If you validate your input manually before using the library, it is probably safe to assume that it would work correctly, given that it was tested on Linux and Mac targets.

PR #523, opened 2019-07-25
Basic version for MPP serializer resolving from KType (typeOf() support)

sandwwraith:
– Standard collections
– No user-defined generic classes

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.internal.*
+import kotlin.reflect.*
+
+/**
+ * Reified version of [serializerByKType], provided for convenience.
+ * [KType] is obtained via [typeOf] function, consult its documentation
+ * to understand limitations.
+ *
+ * [typeOf] and [KType] API currently don't work on Kotlin/JS.
+ */
+@Suppress("UNCHECKED_CAST", "NO_REFLECTION_IN_CLASS_PATH")
+@ImplicitReflectionSerializer
+public inline fun <reified T> serializer(): KSerializer<T> {
+    return serializerByKType(typeOf<T>()) as KSerializer<T>
+}
+
+/**
+ * This method is able to construct serializer for a given [type].
+ * Since [KType], in contrary to [KClass], contains information about generic type parameters,
+ * it is possible to construct serializers for collections using this method.
+ * However, user-defined generic classes are not supported for now.
+ *
+ * Keep in mind that this is a 'heavy' call, so result probably should be cached somewhere else.
+ *
+ * [typeOf] and [KType] API currently don't work on Kotlin/JS.
+ *
+ * @see typeOf
+ */
+@Suppress("UNCHECKED_CAST", "NO_REFLECTION_IN_CLASS_PATH", "UNSUPPORTED")
+@ImplicitReflectionSerializer
+public fun serializerByKType(type: KType): KSerializer<Any?> {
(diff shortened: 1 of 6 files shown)


runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.internal.*
+import kotlin.reflect.*
+
+/**
+ * Reified version of [serializerByKType], provided for convenience.
+ * [KType] is obtained via [typeOf] function, consult its documentation
```
qwwdfsad: `consult` -> `refer to`.
Also, as this method will be much more popular, it is better to make its documentation self-contained.

runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Keep in mind that this is a 'heavy' call, so result probably should be cached somewhere else.
+ *
+ * [typeOf] and [KType] API currently don't work on Kotlin/JS.
+ *
+ * @see typeOf
+ */
+@Suppress("UNCHECKED_CAST", "NO_REFLECTION_IN_CLASS_PATH", "UNSUPPORTED")
+@ImplicitReflectionSerializer
+public fun serializerByKType(type: KType): KSerializer<Any?> {
```
qwwdfsad: `fun serializer(type: KType)`?

`serializer(typeOf<Clz>())` is already self-desribing

runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
```diff
@@ -0,0 +1,67 @@
...
+        } as KClass<Any>
+        return when {
+            type.arguments.isEmpty() -> rootClass.serializer()
+            else -> {
+                val args = type.arguments
+                    .map { requireNotNull(it.type) { "Star projections are not allowed" } }
+                    .map(::serializerByKType)
+                // Array is not supported, see KT-32839
+                when (rootClass) {
+                    List::class, ArrayList::class -> ArrayListSerializer(args[0])
```
qwwdfsad: Do we want to support all potential list supertypes by inspecting superclasses of the given type?

runtime/commonTest/src/kotlinx/serialization/TypeTokenResolvingTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        val myList = arrayListOf(1, 2, 3)
+        assertSerializedWithType<ArrayList<Int>>("[1,2,3]", myList)
+        assertSerializedWithType<List<Int>>("[1,2,3]", myList)
+    }
+
+    @Test
+    fun equalityOfProjectedLists() = run {
+        val myList = arrayListOf(1, 2, 3)
+        assertSerializedWithType<List<Int>>("[1,2,3]", myList)
+        assertSerializedWithType<MutableList<out Int>>("[1,2,3]", myList)
```
qwwdfsad: `in` list is missing

runtime/commonMain/src/kotlinx/serialization/TypeTokenResolving.kt
sandwwraith: What do you mean by supertypes? Any or something?

If you mean subtypes, the design decision was especially no – we provide default list serializers only for default list types. Moreover, KType does not contain `isSubtype/isSupertype` API

qwwdfsad: Yeah, I've meant subtypes.
Ok, works for me.

>Moreover, KType does not contain isSubtype/isSupertype API

`KClass.superclass` + casting to `KClass` should work

sandwwraith:
Merged manually.

Next step would finalize design decision about how we are gonna retrieve serializers for user's generic classes on JS and Native. Then we can start integration with formats and finally make explicit passing of serializer optional.

PR #541, opened 2019-08-22
Make makeNullable extension, make NullableSerializer internal

qwwdfsad:
* Extension is more expressive, readable and is aligned with Kotlin conventions
    * NullableSerializer is an implementation detail and should not be exposed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,12 +14,14 @@ buildscript {
     }
     ext.experimentalsEnabled = ["-progressive", "-Xuse-experimental=kotlin.Experimental",
                                 "-Xuse-experimental=kotlin.ExperimentalMultiplatform",
+                                "-Xuse-experimental=kotlinx.serialization.InternalSerializationApi"
     ]
 
     ext.experimentalsInTestEnabled = ["-progressive", "-Xuse-experimental=kotlin.Experimental",
                                       "-Xuse-experimental=kotlin.ExperimentalMultiplatform",
                                       "-Xuse-experimental=kotlinx.serialization.ImplicitReflectionSerializer",
-                                      "-Xuse-experimental=kotlinx.serialization.UnstableDefault"
+                                      "-Xuse-experimental=kotlinx.serialization.UnstableDefault",
+                                      "-Xuse-experimental=kotlinx.serialization.InternalSerializationApi"
     ]
 
     /*
--- a/runtime/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -4,7 +4,7 @@
 
 package kotlinx.serialization
 
-import kotlin.reflect.KClass
+import kotlin.reflect.*
 
 /**
  * Instructs to use specific serializer for class, property or type argument.
@@ -99,3 +99,15 @@ public annotation class UseSerializers(vararg val serializerClasses: KClass<out 
  */
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.TYPE, AnnotationTarget.CLASS)
 public annotation class Polymorphic
+
+
+/**
+ * Public API marked with this annotation is effectively **internal**, which means
+ * it should not be used outside of `kotlinx.serialization`.
+ * Signature, semantics, source and binary compatibilities are not guaranteed for this API
+ * and will be changed without any warnings or migration aids.
+ * If you cannot avoid using internal API to solve your problem, please report your use-case to serialization's issue tracker.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY, AnnotationTarget.FUNCTION)
+@Experimental(level = Experimental.Level.ERROR)
+public annotation class InternalSerializationApi
--- a/runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
@@ -1,39 +1,42 @@
 /*
  * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-
 package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
-import kotlin.reflect.KClass
 
-fun <T : Any> makeNullable(actualSerializer: KSerializer<T>): KSerializer<T?> = NullableSerializer(actualSerializer)
 
-class NullableSerializer<T : Any>(public val actualSerializer: KSerializer<T>) : KSerializer<T?> {
-    private class SerialDescriptorForNullable(val original: SerialDescriptor): SerialDescriptor by original {
-        override val isNullable: Boolean
-            get() = true
-
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other !is SerialDescriptorForNullable) return false
-
-            if (original != other.original) return false
-
-            return true
-        }
-
-        override fun hashCode(): Int {
-            return original.hashCode() * 31
-        }
+/**
+ * Returns a nullable serializer for the given serializer of non-null type.
+ */
+public val <T : Any> KSerializer<T>.nullable: KSerializer<T?>
+    get() {
+        @Suppress("UNCHECKED_CAST")
+        return if (descriptor.isNullable) (this as KSerializer<T?>) else NullableSerializer(this)
     }
 
-    override val descriptor: SerialDescriptor = SerialDescriptorForNullable(actualSerializer.descriptor)
+@Deprecated(
+    message = "Deprecated in the favor of extension",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("actualSerializer.nullable)")
+)
+fun <T : Any> makeNullable(actualSerializer: KSerializer<T>): KSerializer<T?> {
+    return NullableSerializer(actualSerializer)
+}
+
+/**
+ * Use [KSerializer.nullable][nullable] instead.
+ * @suppress internal API
+ */
+@InternalSerializationApi
+public class NullableSerializer<T : Any>(private val serializer: KSerializer<T>) : KSerializer<T?> {
+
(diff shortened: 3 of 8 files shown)


runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -5,50 +5,64 @@
...
-
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other !is SerialDescriptorForNullable) return false
-
-            if (original != other.original) return false
+/**
+ * Returns a nullable serializer for the given serializer of non-null type.
+ */
+public fun <T : Any> KSerializer<T>.nullable(): KSerializer<T?> {
```
sandwwraith: `KSerializer.list` and `KSerializer.map` are extension properties, so I propose to make `nullable` a property too for the sake of consistency

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -5,50 +5,64 @@
...
+}
 
-            return true
-        }
+@Deprecated(
+    message = "Deprecated in the favor of extension",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("actualSerializer.nullable()")
+)
+fun <T : Any> makeNullable(actualSerializer: KSerializer<T>): KSerializer<T?> {
```
sandwwraith: Note that this declaration is used directly by JVM compiler plugin, so it is impossible to remove it for now without plugin change.

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -5,50 +5,64 @@
...
+    return NullableSerializer(actualSerializer)
+}
 
-        override fun hashCode(): Int {
-            return original.hashCode() * 31
-        }
-    }
+internal class NullableSerializer<T : Any>(private val serializer: KSerializer<T>) : KSerializer<T?> {
```
sandwwraith: It can't have `internal` modifier – plugins for Native and JS use invocations of this constructor directly

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
qwwdfsad: I have questions for this design (and `Pair<Ser, Ser>.map` extension in particular).
Let's postpone it (or you could raise them at our DM), renamed for the sake of consistency

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
qwwdfsad: Sure. Could you please take care of it?

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
qwwdfsad: It's not the first time I bump into this.
Introduced `InternalSerializationApi`, if it looks good, let's gradually mark all plugin-used API with that (that is supposed to be `internal`, at least)

qwwdfsad: By the way, why does plugin use both `makeNullable` and `NullableSerializer`?
Couldn't it use only `.nullable` extension?

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
+@file:UseExperimental(InternalSerializationApi::class)
```
qwwdfsad: It's a bit weird that we have to opt-in for experimental annotations on the file level.
Why won't we use "-Xuse-experimental=kotlinx.serialization.*` in all main source sets?

sandwwraith: For this particular annotation, it is, of course, more reasonable to enable it in build file for whole source set. `ImplicitReflectionSerializer` is not enabled because it imposes some restrictions on runtime (e.g. does not work on Native) and we, as library developers, want to make sure that it would not 'infect' big parts of the library, for users' convenience.

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
sandwwraith: It's for historical reasons. I'll remove `makeNullable` invocation and replace it with `NullableSerialzier` constructor call

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -1,54 +1,73 @@
...
-            if (this === other) return true
-            if (other !is SerialDescriptorForNullable) return false
-
-            if (original != other.original) return false
+/**
+ * Returns a nullable serializer for the given serializer of non-null type.
+ */
+public val <T : Any> KSerializer<T>.nullable: KSerializer<T?>
+    get() {
+        return NullableSerializer(this)
```
pdvrieze: One question. What happens if the receiver is already the NullableSerializer (or perhaps a custom serializer that handles nulls directly?). I don't think that nested NullableSerializers are correct for all formats (where notnullmark is actually used)

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
sandwwraith: OK, just leave it deprecated, I'll remove it later

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
sandwwraith: Good point. Checking `this.descriptor.isNullable` will also help to save some allocations.

build.gradle
```diff
@@ -19,7 +19,8 @@ buildscript {
     ext.experimentalsInTestEnabled = ["-progressive", "-Xuse-experimental=kotlin.Experimental",
                                       "-Xuse-experimental=kotlin.ExperimentalMultiplatform",
                                       "-Xuse-experimental=kotlinx.serialization.ImplicitReflectionSerializer",
-                                      "-Xuse-experimental=kotlinx.serialization.UnstableDefault"
+                                      "-Xuse-experimental=kotlinx.serialization.UnstableDefault",
+                                      "-Xuse-experimental=kotlinx.serialization.InternalSerializationApi"
```
sandwwraith: I think this particular annotation can be added to main source sets (`experimentalsEnabled`) too. We can use our internal API without limitation)

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -99,3 +99,15 @@ public annotation class UseSerializers(vararg val serializerClasses: KClass<out
  */
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.TYPE, AnnotationTarget.CLASS)
 public annotation class Polymorphic
+
+
+/**
+ * Public API marked with this annotation is effectively **internal**, which means
```
sandwwraith: Should we mention the reason it still has public modifier? Docs without it may leave a user with a question 'why they just didn't use `internal` modifier'

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -99,3 +99,15 @@ public annotation class UseSerializers(vararg val serializerClasses: KClass<out
...
+ * it should not be used outside of `kotlinx.serialization`.
+ * Signature, semantics, source and binary compatibilities are not guaranteed for this API
+ * and will be changed without any warnings or migration aids.
+ * If you cannot avoid using internal API to solve your problem, please report your use-case to serialization's issue tracker.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY, AnnotationTarget.FUNCTION)
+@Experimental(level = Experimental.Level.ERROR)
```
sandwwraith: Someone may be already using `NullableSerializer`, so probably stick to warning is a good place to start.

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
```diff
@@ -1,54 +1,73 @@
...
+}
 
-        override fun hashCode(): Int {
-            return original.hashCode() * 31
-        }
-    }
+/**
+ * @suppress internal API
+ */
+@InternalSerializationApi
```
sandwwraith: I wonder if we could somehow tell users 'use `.nullable` instead'. Do experimental annotations offer replacements? Perhaps we should mention that in particular documentation for this class?

runtime/commonMain/src/kotlinx/serialization/internal/NullableSerializer.kt
sandwwraith: @qwwdfsad Could you take a look at this?

qwwdfsad: Fixed.

> I don't think that nested NullableSerializers are correct for all formats (where notnullmark is actually used)

It should be, if you know the case where this property doesn't hold, please report

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
qwwdfsad: Ktor and coroutines experience shows that it's not really necessary.

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
qwwdfsad: We will have to provide a replacement (=> make it deprecated), then undeprecate it etc. without a significant benefit
Just left a comment

PR #545, opened 2019-08-27
Fix the library to not create a stack overflow exception when creating a MissingDescriptorException.

pdvrieze:
Unfortunately MissingDescriptorException uses the toString of the descriptor. This by default lists all component descriptors. Obviously that creates a stack overflow. This patch adds a simple test that triggers the error, and a fix that uses the name of the violating descriptor rather than the toString. It may be possible to use toString (or to simulate it), but that would introduce significantly more complexity.

### DIFF
--- a/runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
@@ -17,11 +17,13 @@
 package kotlinx.serialization.features
 
 import kotlinx.serialization.*
+import kotlinx.serialization.internal.SerialClassDescImpl
 import kotlinx.serialization.modules.*
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.protobuf.ProtoBuf
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.fail
 
 class PolymorphismTest {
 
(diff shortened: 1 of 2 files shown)


runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
```diff
@@ -17,11 +17,13 @@
 package kotlinx.serialization.features
 
 import kotlinx.serialization.*
+import kotlinx.serialization.internal.SerialClassDescImpl
```
sandwwraith: Is this import really needed?

runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
```diff
@@ -63,4 +65,17 @@ class PolymorphismTest {
...
     }
+
+    @Test
+    fun testElementDescriptors() {
+        val serializer = PolymorphicSerializer(PolyDerived::class)
+        try {
+            serializer.descriptor.getElementDescriptor(1)
+            fail ("Exception expected")
+        } catch (e: Exception) {
+            if (e !is SerializationException) {
```
sandwwraith: Looks like you can use `assertFailsWith`

pdvrieze: Fixed this. Somehow I missed the function when looking for it.

pdvrieze:
Changes applied. I found it as it broke my own library (and its tests)

PR #559, opened 2019-09-15
Update README.md

jershell:
Add the Bson format


sandwwraith:
The project looks great, thank you very much!

PR #561, opened 2019-09-17
Add XML library to list of available formats.

pdvrieze:
Add my xml serialization library to the list of community formats.


sandwwraith:
Amazing work, thank you very much!

PR #569, opened 2019-10-03
Update README.MD

Gabgab2003:
No need for a String Array anymore in any of the new kotlin versions (since 1.3)


sandwwraith:
Thanks!

Issue #400, opened 2019-03-07
@Serializable objects cause compilation errors


pauliusrum:
Hi when i run this code i get `java.lang.NoSuchFieldError: Companion exception`.
```kotlin
fun main() {
    val a = A // Throws java.lang.NoSuchFieldError: Companion
}

@Serializable
object A

@Serializer(forClass = A::class)
object ASerializer : KSerializer<A> {

    override val descriptor = SerialClassDescImpl("ASerializer")

    override fun serialize(encoder: Encoder, obj: A) = encoder.encodeString(obj::class.qualifiedName!!)

    override fun deserialize(decoder: Decoder): A = decoder.decodeString().run { A }
}
```

q-litzler:
You `A` object doesn't have a companion.

I think that what you are trying to achieve could look like this:

```kotlin
@Serializable(A.Companion::class)
object A {

@Serializer(forClass = A::class)
companion object : KSerializer<A> {

    override val descriptor = SerialClassDescImpl("ASerializer")

    override fun serialize(encoder: Encoder, obj: A) = encoder.encodeString(obj::class.qualifiedName!!)

    override fun deserialize(decoder: Decoder): A = decoder.decodeString().run { A }
}
```

pauliusrum:
It works if I install a `SimpleModule`. I could not manage to get it working with annotations it would always throw an exception that serializer could not be found. Is this the correct way to solve this issue?
```Kotlin
@Test
fun main() {
    Json.install(SimpleModule(A::class, ASerializer)
    val a = Json.stringify(A)
}

object A

object ASerializer : KSerializer<A> {

    override val descriptor = SerialClassDescImpl("ASerializer")

    override fun serialize(encoder: Encoder, obj: A) = encoder.encodeString(obj::class.qualifiedName!!)

    override fun deserialize(decoder: Decoder): A = decoder.decodeString().run { A }
}
```

q-litzler:
Did you simply try this ?

```kotlin
@Test
fun main() {
    Json.stringify(A, "A")
    val a = Json.stringify(A)
}

@Serializable(A.Companion::class)
object A {

   @Serializer(forClass = A::class)
   companion object : KSerializer<A> {

       override val descriptor = SerialClassDescImpl("ASerializer")

       override fun serialize(encoder: Encoder, obj: A) = encoder.encodeString(obj::class.qualifiedName!!)

[...]

PR #572, opened 2019-10-07
Generated serializers for objects, sealed classes and enums

### LINKED ISSUE #31: SerialName for Enums
It's quite common case when deserialization of JSON you want to parse some property to Enum constant, but now Enum constant should have the same name and same case as a property value.
Would be nice to allow annotate enums with SerialName to provide custom naming for a serilized format. 
An alternative solution is to use Serializer but looks like overkill for enum classes and SerialName could replace it.

sandwwraith:
* Objects serialization

* Sealed classes serialization & generated serializers for enum classes (@SerialInfo support) (require `@Serializable` annotations on enums for now)

* Move Polymorphic and sealed kinds from UnionKind to special PolymorphicKind

* Add tests for polymorphism in protobuf

* Support @SerialId on enum members in protobuf encoding

* Improve toString() for EnumDescriptor.

Closes #31 , #400 , etc


pdvrieze:
Overall this looks quite good. I have found one issue though that would limit applicability for some use cases (and my XML parser). That is, I need a way to enumerate all sub serializers for a given type. There exists a way to do so, but that is to use a `SerialModule`. This approach bypasses that and thereby limits the ability to introspect the format (so that a tag can be mapped to a type even if the tag name and type don't match - and `SerialName` would not be great there either).

A way to resolve this would be to introduce the concept of a "local" `SerialModule` that allows custom serializers to provide local information (or possibly a user of a polymorphic class - ie. specify children by local annotation). When resolving both local and decoder modules would be used (I'm not 100% on priority, perhaps using a flag for a decoder module).

There are some challenges:

- should serializers "inherit" modules from the serializers that invoke them? 
- in case of inheritance, how does "enhancement" work (by default for generated serializers - or custom serializer) while not making all serializers require multiple instances for different usage contexts.
- perhaps the "local" module should be managed by the encoder (which already handles other context)? Does that mean that a separate serializer "module" class for a single serializer (and base class) should be created. (such a class could be created in the plugin).

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,117 @@
...
+ * ```
+ * val abstractContext = SerializersModule {
+ *     polymorphic(ProtocolWithAbstractClass::class, ProtocolWithAbstractClass.Message::class) {
+ *         ProtocolWithAbstractClass.Message.IntMessage::class with ProtocolWithAbstractClass.Message.IntMessage.serializer()
+ *         ProtocolWithAbstractClass.Message.StringMessage::class with ProtocolWithAbstractClass.Message.StringMessage.serializer()
+ *         // no need to register ProtocolWithAbstractClass.ErrorMessage
+ *     }
+ * }
+ * ```
+ *
```
qwwdfsad: redundant newline

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,117 @@
...
+ */
+public class SealedClassSerializer<T : Any>(
+    serialName: String,
+    override val baseClass: KClass<T>,
+    subclasses: Array<KClass<out T>>,
+    subSerializers: Array<KSerializer<out T>>
```
qwwdfsad: it's rather "serializers" than "subSerializers"

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,117 @@
...
+                ?: super.findPolymorphicSerializer(decoder, klassName)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    override fun findPolymorphicSerializer(encoder: Encoder, value: T): KSerializer<out T> {
+        return backingMap[value::class]
+                ?: super.findPolymorphicSerializer(encoder, value)
+    }
+
+    override val descriptor: SerialDescriptor = SealedClassDescriptor(serialName, subSerializers.map { it.descriptor })
```
qwwdfsad: code style, all properties should be declared in a single place (before functions)

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,117 @@
...
+    subclasses: Array<KClass<out T>>,
+    subSerializers: Array<KSerializer<out T>>
+) : AbstractPolymorphicSerializer<T>() {
+
+    private val backingMap: Map<KClass<out T>, KSerializer<out T>>
+
+    private val inverseMap: Map<String, KSerializer<out T>>
+
+
+    init {
```
qwwdfsad: According to the bytecode, an instance of `SealedClassSerializer ` is created **on each** invocation of `serializer()`.
Taking into account its non-trivial constructor, it should be cached in its companion

runtime/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
```diff
@@ -0,0 +1,61 @@
...
+        return theInstance
+    }
+
+    override val descriptor: SerialDescriptor = ObjectDescriptor(serialName)
+}
+
+/**
+ * Object descriptor has UnionKind.OBJECT and consists of itself:
+ * It has one element with name equals to object name
+ * and [getElementDescriptor] returns the descriptor itself.
```
qwwdfsad: This is not a documentation, but rather the description of the implementation

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -5,9 +5,13 @@
 package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
+import kotlin.jvm.JvmOverloads
 import kotlin.reflect.KClass
 
-public class EnumDescriptor(override val name: String, private val choices: Array<String>) : SerialClassDescImpl(name) {
+public class EnumDescriptor @JvmOverloads constructor(
+    override val name: String,
+    private val choices: Array<String> = emptyArray()
```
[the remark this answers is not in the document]
sandwwraith: WDYT of just "values", as in `Enum.values()`?

runtime/commonMain/src/kotlinx/serialization/Descriptors.kt
```diff
@@ -21,20 +21,37 @@ sealed class PrimitiveKind: SerialKind() {
...
 
-sealed class UnionKind: SerialKind() {
-    object OBJECT: UnionKind()
-    object ENUM_KIND: UnionKind() // https://github.com/JetBrains/kotlin-native/issues/1447
-    object SEALED: UnionKind()
-    object POLYMORPHIC: UnionKind()
+@Suppress("unused", "PropertyName")
+sealed class UnionKind : SerialKind() {
+    object OBJECT : UnionKind()
+    object ENUM_KIND : UnionKind() // https://github.com/JetBrains/kotlin-native/issues/1447
```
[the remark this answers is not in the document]
sandwwraith: Why do you think so? The issue is still open. There's no `@ObjCName` (still) and https://github.com/JetBrains/kotlin-native/pull/2747 seems to fix properties only. I'll investigate it later if you insist.

runtime/commonMain/src/kotlinx/serialization/Descriptors.kt
```diff
@@ -21,20 +21,37 @@ sealed class PrimitiveKind: SerialKind() {
...
+        ReplaceWith("PolymorphicKind.OPEN"),
+        DeprecationLevel.ERROR
+    )
+    val POLYMORPHIC = PolymorphicKind.OPEN
+    @Deprecated(
+        "Moved out from UnionKind to simplify instance check for both POLYMORPHIC and SEALED. You can use 'is PolymorphicKind' now.",
+        ReplaceWith("PolymorphicKind.SEALED"),
+        DeprecationLevel.ERROR
+    )
+    val SEALED = PolymorphicKind.SEALED
```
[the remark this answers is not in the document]
sandwwraith: Why 'nor source code migration'?

runtime/commonTest/src/kotlinx/serialization/features/ObjectSerialization.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.ObjectSerializer
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.test.assertStringFormAndRestored
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+sealed class ApiResponse {
+    @Suppress("PLUGIN_ERROR")
```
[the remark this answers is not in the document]
sandwwraith: Because without `1.3.60-eap-xxx` installed in the IDE, this code would be red. I'll remove them soon.

runtime/jvmMain/src/kotlinx/serialization/Serialization.kt
```diff
@@ -63,6 +63,6 @@ internal fun <T> Class<T>.invokeSerializerGetter(vararg args: KSerializer<Any>):
  * on JVM, it falls back to java.lang.Class.isInstance which causes
  * difference when applied to function types with big arity.
  */
-internal actual fun Any.isInstanceOf(kclass: KClass<*>): Boolean = kclass.java.isInstance(this)
+internal actual fun Any.isInstanceOf(kclass: KClass<*>): Boolean = kclass.javaObjectType.isInstance(this)
```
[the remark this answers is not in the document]
sandwwraith: `Int::class.java.isInstance(java.lang.Integer.valueOf(1)) == false`

When you serialize ints inside collections, they're boxed and check https://github.com/kotlin/kotlinx.serialization/blob/0d830ae4db78ad7a43ce5f707ca86c895895029f/runtime/commonMain/src/kotlinx/serialization/modules/StandardSubtypesOfAny.kt#L58 fails.

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
sandwwraith: 'subclassSerializers'?

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
sandwwraith: It's an optimization works in the plugin which will be performed later.

runtime/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
```diff
@@ -0,0 +1,61 @@
...
+        if (name != other.name) return false
+        return true
+    }
+
+    override fun hashCode(): Int {
+        return name.hashCode()
+    }
+
+    override fun toString(): String {
+        return "$name()"
```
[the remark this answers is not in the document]
sandwwraith: Regular classes (which use `SerialDescImpl`) have parentheses for their members. An object does not have members, therefore empty parenthesis.

runtime/commonTest/src/kotlinx/serialization/features/SealedClassesSerializationTest.kt
```diff
@@ -0,0 +1,231 @@
...
+    arrayOf(SimpleSealed.SubSealedA.serializer(), SimpleSealed.SubSealedB.serializer())
+)
+
+@Serializable
+data class SealedHolder(val s: SimpleSealed)
+
+@Serializable
+data class SealedBoxHolder(val b: Box<SimpleSealed>)
+
+class SealedClassesSerializationTest {
```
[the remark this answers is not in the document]
sandwwraith: I have a plans on a big refactoring for test code, but such refactoring is out of the scope of this PR.

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -5,9 +5,13 @@
 package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
+import kotlin.jvm.JvmOverloads
 import kotlin.reflect.KClass
 
-public class EnumDescriptor(override val name: String, private val choices: Array<String>) : SerialClassDescImpl(name) {
+public class EnumDescriptor @JvmOverloads constructor(
```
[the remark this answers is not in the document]
sandwwraith: I wonder if users would want to create their own `EnumSerialzer` implementations for which they would need this descriptor. Or perform a `checkcast`. However, I can change the implementation of `equals` so custom descriptors with enum kind would also qualify.

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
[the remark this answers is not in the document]
sandwwraith: Yes, it is used directly in the codegen. Unfortunately, it is not possible [for now] to make it non-primary because it would require to change plugin code, too. We anyway have a plan to refactor all such constructors, don't we?

runtime/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
sandwwraith: I'd suggest applying the same logic as with `EnumDescriptor`: don't document, make internal. Some descriptor traits (such as "It has one element with name equals to object name") will be described in `SerialKind` documentation, because such traits must be true for all descriptors with particular kind, regardless of implementation.

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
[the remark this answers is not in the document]
qwwdfsad: I am fine with it

runtime/commonMain/src/kotlinx/serialization/Descriptors.kt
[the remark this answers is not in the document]
qwwdfsad: Because it's properties in the sealed class, not in its companion.
You can access them only with a specific instance (e.g. `UnionKind.OBJECT.SEALED`)

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
qwwdfsad: works for me

runtime/commonTest/src/kotlinx/serialization/features/SealedClassesSerializationTest.kt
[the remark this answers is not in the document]
qwwdfsad: Please make it before the next big feature, it's really hard to keep track on tests right now

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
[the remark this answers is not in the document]
qwwdfsad: Okay, let's then make this constructor internal (via annotation) and provide a secondary constructor from a single map

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,117 @@
...
+        require(subclasses.size == subSerializers.size) { "Arrays of classes and serializers must have the same length" }
```
qwwdfsad: Exception message always should contain data that lead to this error otherwise it's just not helpful.

runtime/commonMain/src/kotlinx/serialization/internal/AbstractPolymorphicSerializer.kt
```diff
@@ -0,0 +1,97 @@
...
+                    "Invalid index in polymorphic deserialization of " +
+                            (klassName ?: "unknown class") +
+                            "\n Expected 0, 1, READ_ALL(-2) or READ_DONE(-1), but found $index"
+                )
+            }
+        }
+
+        compositeDecoder.endStructure(descriptor)
+        @Suppress("UNCHECKED_CAST")
+        return requireNotNull(value) { "Polymorphic value have not been read for class $klassName" } as T
```
qwwdfsad: has not

runtime/commonMain/src/kotlinx/serialization/internal/AbstractPolymorphicSerializer.kt
```diff
@@ -0,0 +1,97 @@
...
+ * This class cannot be implemented by library users. To learn how to use it for your case,
+ * see [PolymorphicSerializer] for interfaces/abstract classes and [SealedClassSerializer] for sealed classes.
+ *
+ * By default, (without special support from [Encoder]), polymorphic values are serialized as list with
+ * two elements: fully-qualified class name (String) and the object itself.
+ */
+@InternalSerializationApi
+public abstract class AbstractPolymorphicSerializer<T : Any> internal constructor() : KSerializer<T> {
+
+    public abstract val baseClass: KClass<T>
```
qwwdfsad: Missing KDoc

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,113 @@
...
+        return backingMap[value::class]
+                ?: super.findPolymorphicSerializer(encoder, value)
+    }
+}
+
+/**
+ * Descriptor for sealed class contains descriptors for all its serializable inheritors
+ * which can be obtained via [getElementDescriptor].
+ */
+public class SealedClassDescriptor(
```
qwwdfsad: internal constructor/internal class/internal annotation on class?

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -5,13 +5,17 @@
 package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
+import kotlin.jvm.JvmOverloads
 import kotlin.reflect.KClass
 
-public class EnumDescriptor(override val name: String, private val choices: Array<String>) : SerialClassDescImpl(name) {
+internal class EnumDescriptor @JvmOverloads constructor(
```
qwwdfsad: Do you still need JvmOverloads here?

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -20,45 +24,53 @@ public class EnumDescriptor(override val name: String, private val choices: Arra
...
         var result = super.hashCode()
         result = 31 * result + name.hashCode()
-        result = 31 * result + choices.contentHashCode()
+        result = 31 * result + values.contentHashCode()
         return result
     }
 }
 
-open class CommonEnumSerializer<T>(val serialName: String, val choices: Array<T>, choicesNames: Array<String>) :
+open class CommonEnumSerializer<T>(serialName: String, val values: Array<T>, valuesNames: Array<String>) :
```
qwwdfsad: Deprecation with error on both classes?

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -5,13 +5,17 @@
...
 
 import kotlinx.serialization.*
+import kotlin.jvm.JvmOverloads
 import kotlin.reflect.KClass
 
-public class EnumDescriptor(override val name: String, private val choices: Array<String>) : SerialClassDescImpl(name) {
+internal class EnumDescriptor @JvmOverloads constructor(
+    override val name: String,
+    private val values: Array<String> = emptyArray()
+) : SerialClassDescImpl(name) {
```
qwwdfsad: Now enums support `SerialInfo`, it's worth to add some tests

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: I just double-checked and found that `EnumDescriptor` is referenced from the generated code. So we should go with `@InternalSerializationApi public EnumDescriptor`. Seems it was working by accident that `internal` classes are available from test sources of the same module.

qwwdfsad: >internal classes are available from test sources of the same module.

Nope, that's actually what friend modules are for right now

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: Basically, `CommonEnumSerializer` is still used when we don't need a full-blown generated serializer (when no `@SerialInfo` annotations are present). `EnumSerializer` can be deprecated, but it is more convenient to use it when we are on JVM (and therefore have all the information about enum we need). I'll suggest going with `@InternalSerializationApi` on both.

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: `SerialName/SerialId` saving is tested here: https://github.com/kotlin/kotlinx.serialization/blob/c95ab9846a5df82ffb98c26d782891d4ddce2fc0/runtime/commonTest/src/kotlinx/serialization/kinds/SerializableEnumTest.kt#L45. Do you think I've missed something?

runtime/commonMain/src/kotlinx/serialization/Descriptors.kt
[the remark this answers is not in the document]
qwwdfsad: Still not addressed

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -78,7 +79,7 @@ public class SealedClassSerializer<T : Any>(
     private val inverseMap: Map<String, KSerializer<out T>>
 
     init {
-        require(subclasses.size == subclassSerializers.size) { "Arrays of classes and serializers must have the same length" }
+        require(subclasses.size == subclassSerializers.size) { "Arrays of classes and serializers must have the same length, got arrays: $subclasses, $subclassSerializers" }
```
qwwdfsad: 1) The line is too long
2) `toString` on arrays does not help here

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -44,6 +45,7 @@ internal class EnumDescriptor @JvmOverloads constructor(
     }
 }
 
+@InternalSerializationApi
 open class CommonEnumSerializer<T>(serialName: String, val values: Array<T>, valuesNames: Array<String>) :
     KSerializer<T> {
```
qwwdfsad: Codestyle, should be 
```
@InternalSerializationApi
open class CommonEnumSerializer<T>(
    serialName: String,
    val values: Array<T>,
    valuesNames: Array<String>
) : KSerializer<T> {
    
    override val descriptor: SerialDescriptor = EnumDescriptor(serialName, valuesNames)
```

runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonPolymorphicClassDescriptor.kt
```diff
@@ -4,8 +4,9 @@
 
 package kotlinx.serialization.json.polymorphic
 
-import kotlinx.serialization.json.*
-import kotlin.test.*
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonTestBase
+import kotlin.test.Test
 
 class JsonPolymorphicClassDescriptor : JsonTestBase() {
```
qwwdfsad: `Test` suffix is missing

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -20,45 +25,53 @@ public class EnumDescriptor(override val name: String, private val choices: Arra
...
         encoder.encodeEnum(descriptor, index)
     }
 
     final override fun deserialize(decoder: Decoder): T {
         val index = decoder.decodeEnum(descriptor)
-        check(index in choices.indices)
-            { "$index is not among valid $serialName choices, choices size is ${choices.size}" }
-        return choices[index]
+        check(index in values.indices)
+        { "$index is not among valid $${descriptor.name} enum values, values size is ${values.size}" }
```
qwwdfsad: code style:
```
check(index in values.indices) {
    "$index is not among valid $${descriptor.name} enum values, values size is ${values.size}"
}
```

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -20,45 +25,53 @@ public class EnumDescriptor(override val name: String, private val choices: Arra
...
+open class CommonEnumSerializer<T>(serialName: String, val values: Array<T>, valuesNames: Array<String>) :
     KSerializer<T> {
-    override val descriptor: EnumDescriptor = EnumDescriptor(serialName, choicesNames)
+    override val descriptor: SerialDescriptor = EnumDescriptor(serialName, valuesNames)
 
     final override fun serialize(encoder: Encoder, obj: T) {
-        val index = choices.indexOf(obj)
-            .also { check(it != -1) { "$obj is not a valid enum $serialName, choices are $choices" } }
+        val index = values.indexOf(obj)
+            .also { check(it != -1) { "$obj is not a valid enum ${descriptor.name}, must be one of ${values.contentToString()}" } }
```
qwwdfsad: Codestyle: `.also` chain helps neither with readability, not with compactness
```
val index = values.indexOf(obj)
check(index != -1) { "$obj is not a valid enum ${descriptor.name}, must be one of ${values.contentToString()}" }
```

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -5,13 +5,18 @@
...
 
 import kotlinx.serialization.*
+import kotlin.jvm.JvmOverloads
 import kotlin.reflect.KClass
 
-public class EnumDescriptor(override val name: String, private val choices: Array<String>) : SerialClassDescImpl(name) {
+@InternalSerializationApi
+public class EnumDescriptor @JvmOverloads constructor(
+    override val name: String,
+    private val values: Array<String> = emptyArray()
```
qwwdfsad: `values` should not be `val`.
All required state for eq/hc is already present in the descriptor elements

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -0,0 +1,114 @@
...
+                ?: super.findPolymorphicSerializer(encoder, value)
+    }
+}
+
+/**
+ * Descriptor for sealed class contains descriptors for all its serializable inheritors
+ * which can be obtained via [getElementDescriptor].
+ */
+internal class SealedClassDescriptor(
+    override val name: String,
```
qwwdfsad: pointless override

runtime/commonMain/src/kotlinx/serialization/Descriptors.kt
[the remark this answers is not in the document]
sandwwraith: Ah, I finally get it. I'd suggest to add `companion object` to `UnionKind` so migration will work.

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: Any tips to set up the IDEA formatter properly? Standard Kotlin code style does not address this.

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
sandwwraith: https://youtrack.jetbrains.com/issue/KT-34776

runtime/commonMain/src/kotlinx/serialization/Polymorphic.kt
```diff
@@ -29,27 +29,31 @@ public object PolymorphicClassDescriptor : SerialClassDescImpl("kotlin.Any") {
 /**
  * This class provides support for multiplatform polymorphic serialization for interfaces and abstract classes.
  *
- * Due to security and reflection usage concerns, all serializable implementations of some abstract class must be registered in advance.
- * However, it allows registering subclasses in runtime, not compile-time. For example, it allows adding additional subclasses to the registry
- * that were defined in a separate module, dependent on the base module with the base class.
+ * To avoid the most common security pitfalls and reflective lookup (and potential load) of an arbitrary class,
+ * all serializable implementations of any polymorphic type must be [registered][SerializersModuleBuilder.polymorphic]
+ * in advance in the scope of base polymorphic type, efficiently preventing unbounded polymorphic serialization
+ * and (de)serialization of an arbitrary type.
```
sandwwraith: You already have `serialization` word before 'and', why '(de)' is in brackets?

qwwdfsad: It was much better in my head :)

PR #586, opened 2019-10-30
Cleanup unused symbols

qwwdfsad:
* SerialTag that is used neither by us nor by community formats
  * Unused private symbols
  * Obsolete deprecations


qwwdfsad:
We would do it if these declarations were actually used/useful or `kx.serialization` was stable.
But none of the kx.serialization libraries use these symbols, so deprecation cycle looks a bit of overkill

Issue #71, opened 2018-01-24
NULLs are not supported when writing to protobuf


sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/issues/58#issuecomment-359719900

```kotlin
@Serializable class SampleMessage {
    @Optional @SerialId(1) var name: String? = null
}

fun main(args: Array<String>) {
    val message = SampleMessage()
    println(ProtoBuf.dumps(message))
}
```

crashes with 

```
Exception in thread "main" kotlinx.serialization.SerializationException: null is not supported
	at kotlinx.serialization.TaggedOutput.writeTaggedNull(Tagged.kt:38)
	at kotlinx.serialization.TaggedOutput.writeNullValue(Tagged.kt:80)
	at kotlinx.serialization.KOutput.writeNullableSerializableValue(Serialization.kt:151)
	at kotlinx.serialization.KOutput.writeNullableSerializableElementValue(Serialization.kt:198)
	at SampleMessage.write$Self(TestOneOf.kt)
	at SampleMessage$$serializer.save(TestOneOf.kt)
	at SampleMessage$$serializer.save(TestOneOf.kt:5)
	at kotlinx.serialization.KOutput.write(Serialization.kt:99)
	at kotlinx.serialization.protobuf.ProtoBuf.dump(ProtoBuf.kt:415)
	at TestOneOfKt.main(TestOneOf.kt:17)
```

wem:
Any news on this issue? Unfortunately, this bug makes Protobuf serialization unusable at the moment.

sandwwraith:
@wem It took longer than I though. I think it would be resolved alongside with #58 , but with one limitation: protobuf doesn't have a concept of `null` anyway, so the single way to record them is to just skip them in output stream. But when you have an optional value which default is not a null value, you can't really say was it missing from the stream because it had default value or because it was null, so such declaration will likely be prohibited by protobuf.

raderio:
I think this https://github.com/Kotlin/kotlinx.serialization/issues/407 may also be a solution

PR #588, opened 2019-10-30
Skip null fields in ProtoBuf serialization

bezmax:
Defaults to skipping null fields during Protobuf serialization. Resolves issue https://github.com/Kotlin/kotlinx.serialization/issues/71 .

The issue blocks any real usage of kotlinx.serialization, as there is no way to serialize an optional field as of right now. I suggest this as an easy and mostly-compatible fix until a long-term solution is devised.

The only real problem I can see with this approach is this:
```
data class Message(val a: String? = "foo bar")
...
val msgIn = Message(null)
val bytes = ProtoBuf.dump(Message.serializer(), msgIn )
val msgOut = ProtoBuf.load(Message.serializer(), bytes)
msgIn shouldBe msgOut //Fails, as it will become a specified default string.
```

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/Tagged.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/Tagged.kt
@@ -64,9 +64,18 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
 
     // ---- Implementation of low-level API ----
 
-    fun encodeElement(desc: SerialDescriptor, index: Int): Boolean {
+    fun <T: Any?> encodeElement(desc: SerialDescriptor, index: Int, value: T): Boolean {
         val tag = desc.getTag(index)
-        val shouldWriteElement = shouldWriteElement(desc, tag, index)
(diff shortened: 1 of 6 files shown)


bezmax:
@sandwwraith Yes, I know about that issue. Did not mention it as it was already there even before this suggested change. The suggestion actually makes it more consistent now in (incorrect) behaviour.

For example, right now you already can do this:
```
data class Message(val a: Int = 42) //Not even nullable
print(ProtoBuf.loads(Message.serializer(), "")) //Will be 42, while should have been 0
```

I see that there's a `shouldWriteElement` on `TaggedEncoder`, instead of adding a property I could just implement it correctly just for classes by overriding this method. In that case for scalar types you could still define your fields as `val a: Int = 0` and be mostly proto compatible, and for classes it will allow nulls by default - which is also compatible. What do you think about this approach? I'll try to modify my PR for a concrete example of how this would look like.

bezmax:
Created a new PR against `dev` with a slightly different (and fully Proto compatible) approach: https://github.com/Kotlin/kotlinx.serialization/pull/591

PR #591, opened 2019-10-31
Skip null fields in ProtoBuf serialization for CLASS type fields

bezmax:
Resubmission of https://github.com/Kotlin/kotlinx.serialization/pull/588

The issue blocks any real usage of kotlinx.serialization, as there is no way to serialize an optional field as of right now. I suggest this as an easy and mostly-compatible fix until a long-term solution is devised.

The solution here introduces separate `TaggedEncoder#shouldWriteNullElement` method, which can be overriden by different Encoders to provide custom null-handling behaviour. For this PR only ProtoBufEncoder overrides it to skip null CLASS type fields (as should be completely compatible with Proto spec).

Other encoders could also benefit from this, for example an extra setting could be added to JSON encoder to skip null fields altogether.

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/Tagged.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/Tagged.kt
@@ -64,9 +64,18 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
 
     // ---- Implementation of low-level API ----
 
-    fun encodeElement(desc: SerialDescriptor, index: Int): Boolean {
+    fun <T: Any?> encodeElement(desc: SerialDescriptor, index: Int, value: T): Boolean {
         val tag = desc.getTag(index)
-        val shouldWriteElement = shouldWriteElement(desc, tag, index)
+        val shouldWriteElement = shouldWriteElement(desc, tag, index, value)
+        if (shouldWriteElement) {
+            pushTag(tag)
+        }
+        return shouldWriteElement
+    }
+
+    fun encodeNullElement(desc: SerialDescriptor, index: Int): Boolean {
+        val tag = desc.getTag(index)
+        val shouldWriteElement = shouldWriteNullElement(desc, tag, index)
         if (shouldWriteElement) {
             pushTag(tag)
         }
@@ -74,7 +83,8 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
     }
 
     // For format-specific behaviour, invoked only on
-    open fun shouldWriteElement(desc: SerialDescriptor, tag: Tag, index: Int) = true
+    open fun <T: Any?> shouldWriteElement(desc: SerialDescriptor, tag: Tag, index: Int, value: T) = true
+    open fun shouldWriteNullElement(desc: SerialDescriptor, tag: Tag, index: Int) = true
 
     final override fun encodeNotNullMark() = encodeTaggedNotNullMark(currentTag)
     final override fun encodeNull() = encodeTaggedNull(popTag())
@@ -122,12 +132,13 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
     final override fun encodeStringElement(desc: SerialDescriptor, index: Int, value: String) = encodeTaggedString(desc.getTag(index), value)
 
     final override fun <T : Any?> encodeSerializableElement(desc: SerialDescriptor, index: Int, serializer: SerializationStrategy<T>, value: T) {
-        if (encodeElement(desc, index))
+        if (encodeElement(desc, index, value))
             encodeSerializableValue(serializer, value)
     }
 
     final override fun <T : Any> encodeNullableSerializableElement(desc: SerialDescriptor, index: Int, serializer: SerializationStrategy<T>, value: T?) {
-        if (encodeElement(desc, index))
+        if (value == null && encodeNullElement(desc, index)
+            || value != null && encodeElement(desc, index, value))
             encodeNullableSerializableValue(serializer, value)
     }
 
--- a/runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
@@ -40,6 +40,9 @@ class ProtoBuf(context: SerialModule = EmptyModule) : AbstractSerialFormat(conte
             else -> throw SerializationException("Primitives are not supported at top-level")
         }
 
+        override fun shouldWriteNullElement(desc: SerialDescriptor, tag: ProtoDesc, index: Int): Boolean =
+            desc.kind !is StructureKind.CLASS
+
         override fun encodeTaggedInt(tag: ProtoDesc, value: Int) = encoder.writeInt(value, tag.first, tag.second)
         override fun encodeTaggedByte(tag: ProtoDesc, value: Byte) = encoder.writeInt(value.toInt(), tag.first, tag.second)
         override fun encodeTaggedShort(tag: ProtoDesc, value: Short) = encoder.writeInt(value.toInt(), tag.first, tag.second)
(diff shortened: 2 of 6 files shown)


runtime/commonMain/src/kotlinx/serialization/Tagged.kt
```diff
@@ -64,17 +64,27 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
...
         if (shouldWriteElement) {
             pushTag(tag)
         }
         return shouldWriteElement
     }
 
     // For format-specific behaviour, invoked only on
-    open fun shouldWriteElement(desc: SerialDescriptor, tag: Tag, index: Int) = true
+    open fun <T: Any?> shouldWriteElement(desc: SerialDescriptor, tag: Tag, index: Int, value: T) = true
+    open fun shouldWriteNullElement(desc: SerialDescriptor, tag: Tag, index: Int) = true
```
pdvrieze: It's not clear that you need those (or need to change `Tagged` at all. In any case you are changing the semantics of `shouldWriteElement` which is data independent. I would say that you want the actual format to just make the appropriate decision when attempting to write the value. If the case is that a null value is written through omitting it (and this is valid for the format) that still means that it is "written".

runtime/commonMain/src/kotlinx/serialization/Tagged.kt
```diff
@@ -122,12 +132,13 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
...
     final override fun <T : Any?> encodeSerializableElement(desc: SerialDescriptor, index: Int, serializer: SerializationStrategy<T>, value: T) {
-        if (encodeElement(desc, index))
+        if (encodeElement(desc, index, value))
             encodeSerializableValue(serializer, value)
     }
 
     final override fun <T : Any> encodeNullableSerializableElement(desc: SerialDescriptor, index: Int, serializer: SerializationStrategy<T>, value: T?) {
-        if (encodeElement(desc, index))
+        if (value == null && encodeNullElement(desc, index)
+            || value != null && encodeElement(desc, index, value))
```
pdvrieze: This is invalid. It violates the design approach of this class and doesn't separate out the different aspects. As such it makes the class design inconsistent.

runtime/commonMain/src/kotlinx/serialization/Tagged.kt
```diff
@@ -122,12 +132,13 @@ abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
     final override fun encodeStringElement(desc: SerialDescriptor, index: Int, value: String) = encodeTaggedString(desc.getTag(index), value)
 
     final override fun <T : Any?> encodeSerializableElement(desc: SerialDescriptor, index: Int, serializer: SerializationStrategy<T>, value: T) {
-        if (encodeElement(desc, index))
+        if (encodeElement(desc, index, value))
```
pdvrieze: This breaks the design of `Tagged`.

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -40,6 +40,9 @@ class ProtoBuf(context: SerialModule = EmptyModule) : AbstractSerialFormat(conte
             else -> throw SerializationException("Primitives are not supported at top-level")
         }
 
+        override fun shouldWriteNullElement(desc: SerialDescriptor, tag: ProtoDesc, index: Int): Boolean =
+            desc.kind !is StructureKind.CLASS
+
```
pdvrieze: This is a very minimal change in `ProtoBuf` but perhaps the format should have more changes (and `Tagged` none). The correct interception point is in `encodeNullableSerializableValue` that has full flexibility on how to encode the nullable value.
Basically I see 4 main approaches for handling null values in a format like ProtoBuf that doesn't support them (the choice should be configurable):

- Disallow them (current solution, safe but not always satisfactory)
- Use a marker flag (most correct solution, but there is no standard - other options should be available when the format is not under full control) -- write an "isNull" flag and only write the value if the flag has a false value, the NullableSerializer class is designed in this philosophy
- Omit null values - this has as disadvantages that no distinction can be made between missing values (which need to be set to the defaults) and null values (which need to be set to `null`)
- Use "null" objects (I don't think this works for protobuf) - where there is an actual object that gets the meaning of null (this means that object itself cannot be serialized)

pdvrieze:
Please also note that the conceptual problem is really only there for nullable values that are `Optional`. In the case they are not `Optional` it is fairly straightforward (when reading) to conclude that all missing nullable properties have a null value and should therefore be set to null. This is the approach I use in my XML format.

bezmax:
@pdvrieze Thanks for the feedback and clarification on design of TaggedEncoder. I'll try to refactor the code to move the null-related stuff into Protobuf tag.

> Please also note that the conceptual problem is really only there for nullable values that are Optional. 
Every field is optional in proto3 encoding. And right now the ProtoBuf implementation does not support nulls during serialization anywhere. Therefore there is no way to encode a `CLASS` field in such a way that it's, well, not encoded. That's basically why I'm trying to fix this, as proto3 is unusable for any nested messages when you want not to pass them.

pdvrieze:
@bezmax I had a look at the protobuf3 specification. In the format specification it does not distinguish between a default value (of null) and the absence of a value (old message). It also specifies how absence is handled (with a set default value). This means that effectively protobuf cannot support properties with a default value that is not equal to this default. It also means that protobuf optional has different semantics from kotlinx.serialization optional.

The way I would attempt to address the issue is by making the format support protobuf semantics: nullable fields are allowed (but nullable numbers must be implemented as length delimited fields with a single content element - be careful of format backwards compatibility). The algorithm works as follows (in the decoder - note that names are not quite correct):

1. on creation of the CompositeDecoder create a list for all seen fields (or unseen fields)
2. when handling a field, mark that field as handled
3. on closing the structure, look at all fields in the structure (using `SerialDesc`), for each of those, determine the type and synthesize the reading of that field with the appropriate default value. 

If you do not wish to strictly follow the semantics of protobuf you can support `@Optional` fields by using the default value in Kotlin. For (nulllable) integer types the story is a bit trickier, there is no default way to handle it in protobufs. It may be best to encode them as packed fields with only a single element instead.

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -40,6 +40,9 @@ class ProtoBuf(context: SerialModule = EmptyModule) : AbstractSerialFormat(conte
             else -> throw SerializationException("Primitives are not supported at top-level")
         }
 
+        override fun shouldWriteNullElement(desc: SerialDescriptor, tag: ProtoDesc, index: Int): Boolean =
+            desc.kind !is StructureKind.CLASS
```
sandwwraith: This check is still not under the protobuf-specific configuration flag, so it would change already existing semantic. Also, I can imagine that it can be more flexible, like `configuration.nullSupportMode = DISABLED` will have old semantic, `configuration.nullSupportMode = DROP_NULLS` will drop `null` values to be compatible with another clients, and `configuration.nullSupportMode = DROP_DEFAULTS` will drop default values as in Json.

runtime/commonTest/src/kotlinx/serialization/protobuf/ProtobufReaderTest.kt
```diff
@@ -59,4 +59,9 @@ class ProtobufReaderTest {
         if (isNative()) return // todo: support update on Native
         ProtoBuf.loads(TestIntWithList.serializer(), "500308960150045005") shouldBe TestIntWithList(150, listOf(3, 4, 5))
     }
+
+    @Test
+    fun readObjectWithNullableMissingFields() {
```
sandwwraith: There are no new tests that will verify that nullable fields with `kind != Kind.CLASS` are indeed prohibited.

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
sandwwraith: Also, it makes sense to move `throw SerializationException("null is not supported")` somewhere here to insert more meaningful error message (with descriptor kind, current tag, etc) and how the user can change configuration (again, see `JsonConfiguration.allowStructuredMapKeys` implementation and how this flag is used)

bezmax:
Will close this for now as I do not have time right now to finish this correctly in reasonable time. Will reopen when I finish the fix.

PR #597, opened 2019-11-06
Support for multiple alternative names during JSON parsing

### LINKED ISSUE #203: Can SerialName annotation support alternate?
Can SerialName annotation support alternate?
Like Gson's SerializedName annotation.

sandwwraith:
* Configurable via `@JsonAlternativeNames` annotation
* Uses new `DescriptorSchemaCache` – class for associating custom format info with descriptors in effective and type-safe way
* Stabilize .toString() for SerialClassDescImpl; because HashMaps on different platforms have different iteration order

Fixes #203

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+private typealias DescriptorData<T> = MutableMap<DescriptorSchemaCache.Key<T>, T>
+
+/**
+ * A type-safe map for storing custom information (such as format schema), associated with [SerialDescriptor].
+ *
+ * This map is not thread safe. For thread safety and performance reasons, it is advised to pass it via stack (e.g. in function parameters)
+ */
+public class DescriptorSchemaCache {
+    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = hashMapOf()
+
+    public operator fun <T : Any> set(descriptor: SerialDescriptor, key: Key<T>, value: T): Unit {
+        map.getOrPut(descriptor, ::hashMapOf)[key as Key<Any>] = value as Any
+    }
+
+    public fun <T : Any> getOrPut(descriptor: SerialDescriptor, key: Key<T>, defaultValue: () -> T): T {
+        get(descriptor, key)?.let { return it }
+        val value = defaultValue()
+        set(descriptor, key, value)
+        return value
+    }
+
+    public operator fun <T : Any> get(descriptor: SerialDescriptor, key: Key<T>): T? {
+        return map[descriptor]?.get(key as Key<Any>) as? T
+    }
+
+    /**
+     * A key for associating user data of type [T] with a given [SerialDescriptor].
+     */
+    public class Key<T : Any> {}
+}
--- a/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
@@ -4,15 +4,18 @@
 
 package kotlinx.serialization
 
+import kotlin.native.concurrent.*
+
 private typealias DescriptorData<T> = MutableMap<DescriptorSchemaCache.Key<T>, T>
 
 /**
  * A type-safe map for storing custom information (such as format schema), associated with [SerialDescriptor].
  *
- * This map is not thread safe. For thread safety and performance reasons, it is advised to pass it via stack (e.g. in function parameters)
+ * This cache uses ConcurrentHashMap on JVM and regular maps on other platforms.
+ * To be able to work with it from multiple threads in Kotlin/Native, use @[ThreadLocal] in appropriate places.
  */
 internal class DescriptorSchemaCache {
-    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = createMapForCache()
+    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = createMapForCache(1)
 
     @Suppress("UNCHECKED_CAST")
     public operator fun <T : Any> set(descriptor: SerialDescriptor, key: Key<T>, value: T) {
--- a/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization
@@ -11,10 +11,11 @@ private typealias DescriptorData<T> = MutableMap<DescriptorSchemaCache.Key<T>, T
  *
  * This map is not thread safe. For thread safety and performance reasons, it is advised to pass it via stack (e.g. in function parameters)
  */
-public class DescriptorSchemaCache {
-    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = hashMapOf()
+internal class DescriptorSchemaCache {
+    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = createMapForCache()
 
-    public operator fun <T : Any> set(descriptor: SerialDescriptor, key: Key<T>, value: T): Unit {
+    @Suppress("UNCHECKED_CAST")
+    public operator fun <T : Any> set(descriptor: SerialDescriptor, key: Key<T>, value: T) {
         map.getOrPut(descriptor, ::hashMapOf)[key as Key<Any>] = value as Any
     }
 
@@ -25,6 +26,7 @@ public class DescriptorSchemaCache {
         return value
     }
 
+    @Suppress("UNCHECKED_CAST")
     public operator fun <T : Any> get(descriptor: SerialDescriptor, key: Key<T>): T? {
         return map[descriptor]?.get(key as Key<Any>) as? T
     }
--- a/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -203,6 +203,7 @@ public class JsonBuilder {
     public var indent: String = "    "
     public var useArrayPolymorphism: Boolean = false
     public var classDiscriminator: String = "type"
+    public var supportAlternativeNames: Boolean = false
     public var serialModule: SerialModule = EmptyModule
 
     public fun buildConfiguration(): JsonConfiguration =
@@ -214,7 +215,8 @@ public class JsonBuilder {
             prettyPrint,
             indent,
             useArrayPolymorphism,
-            classDiscriminator
+            classDiscriminator,
+            supportAlternativeNames
         )
 
     public fun buildModule(): SerialModule = serialModule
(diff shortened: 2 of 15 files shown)


runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * A type-safe map for storing custom information (such as format schema), associated with [SerialDescriptor].
+ *
+ * This map is not thread safe. For thread safety and performance reasons, it is advised to pass it via stack (e.g. in function parameters)
+ */
+public class DescriptorSchemaCache {
+    private val map: MutableMap<SerialDescriptor, DescriptorData<Any>> = hashMapOf()
+
+    public operator fun <T : Any> set(descriptor: SerialDescriptor, key: Key<T>, value: T): Unit {
```
qwwdfsad: unit return type is redundant

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -105,6 +106,12 @@ internal class StreamingJsonInput internal constructor(
         }
     }
 
+    private fun SerialDescriptor.getJsonElementIndex(key: String): Int {
+        if (!json.configuration.supportAlternativeNames) return this.getElementIndex(key)
```
qwwdfsad: Please take into account the following invariants:
    * Fields rarely have alternative names even when this feature is enabled
    * Fields rarely have more than one alternative name

Then you can significantly speedup this method, just replace the nested map with two arrays (`String[]` and `int[]`), lookup with linear search and ensure that main name is always in the first position of the array

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -161,7 +176,8 @@ private class JsonTreeMapInput(json: Json, override val obj: JsonObject) : JsonT
     }
 }
 
-private class JsonTreeListInput(json: Json, override val obj: JsonArray) : AbstractJsonTreeInput(json, obj) {
+private class JsonTreeListInput(json: Json, override val obj: JsonArray, schemaCache: DescriptorSchemaCache) :
+    AbstractJsonTreeInput(json, obj, schemaCache) {
```
qwwdfsad: formatting

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -134,7 +148,8 @@ private open class JsonTreeInput(json: Json, override val obj: JsonObject) : Abs
     }
 }
 
-private class JsonTreeMapInput(json: Json, override val obj: JsonObject) : JsonTreeInput(json, obj) {
+private class JsonTreeMapInput(json: Json, override val obj: JsonObject, schemaCache: DescriptorSchemaCache) :
```
qwwdfsad: formatting

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -104,7 +109,8 @@ private class JsonPrimitiveInput(json: Json, override val obj: JsonPrimitive) :
     }
 }
 
-private open class JsonTreeInput(json: Json, override val obj: JsonObject) : AbstractJsonTreeInput(json, obj) {
+private open class JsonTreeInput(json: Json, override val obj: JsonObject, schemaCache: DescriptorSchemaCache) :
```
qwwdfsad: formatting

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -92,7 +96,8 @@ private sealed class AbstractJsonTreeInput(override val json: Json, open val obj
     override fun decodeTaggedString(tag: String) = getValue(tag).content
 }
 
-private class JsonPrimitiveInput(json: Json, override val obj: JsonPrimitive) : AbstractJsonTreeInput(json, obj) {
+private class JsonPrimitiveInput(json: Json, override val obj: JsonPrimitive, schemaCache: DescriptorSchemaCache) :
```
qwwdfsad: formatting

runtime/commonMain/src/kotlinx/serialization/json/JsonAlternativeNames.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.json
+
+import kotlinx.serialization.*
+
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY)
+public annotation class JsonAlternativeNames(val names: Array<String>)
```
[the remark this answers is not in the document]
qwwdfsad: What do you think about renaming it to `JsonNames`? So our users will have a much clearer code without annotations that are 20 symbols long

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -193,6 +193,7 @@ public class JsonBuilder {
     public var indent: String = "    "
     public var useArrayPolymorphism: Boolean = false
     public var classDiscriminator: String = "type"
+    public var supportAlternativeNames: Boolean = false
```
qwwdfsad: Please consider naming it "useAlternativeNames" to have a shorter option name.
It's supported, users just have to enable/use it :)

runtime/commonMain/src/kotlinx/serialization/json/JsonAlternativeNames.kt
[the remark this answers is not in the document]
sandwwraith: Sounds OK

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
sandwwraith: Even if the property has one alternative name from `@JsonNames`, it means it has two names: one primary and one alternative, i.e., the map will contain two keys which point to the property's index. 

I don't quite get how it is possible to replace the map with two arrays. Put all names into one array, linearly search it, and then use index we found in the second array so it would give us the index of the field? I don't see how it is faster than a map, because even with disabled alternative names, `this.getElementName` still uses a map inside a `SerialClassDescImpl`. However, it is possible to do two-phase lookup: first for a primary name in the descriptor and second in the aforementioned array. But this case can still use a map – we just don't need to allocate it if a property does not have an alternative name.

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -122,7 +122,7 @@ public constructor(
      */
     public override fun <T> parse(deserializer: DeserializationStrategy<T>, string: String): T {
         val reader = JsonReader(string)
-        val input = StreamingJsonInput(this, WriteMode.OBJ, reader)
+        val input = StreamingJsonInput(this, WriteMode.OBJ, reader, DescriptorSchemaCache())
```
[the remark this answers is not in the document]
sandwwraith: Even if some class is encountered only once in a user's code, caching can still be useful, e.g. if this class is used in a collection.

Schema cache bound to the `Json` instance indeed seems to be a more efficient way, but it will make `Json` a stateful class. We should carefully think about whether we can allow such change or not.

runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
```diff
@@ -0,0 +1,36 @@
...
+package kotlinx.serialization
+
+private typealias DescriptorData<T> = MutableMap<DescriptorSchemaCache.Key<T>, T>
+
+/**
+ * A type-safe map for storing custom information (such as format schema), associated with [SerialDescriptor].
+ *
+ * This map is not thread safe. For thread safety and performance reasons, it is advised to pass it via stack (e.g. in function parameters)
+ */
+public class DescriptorSchemaCache {
```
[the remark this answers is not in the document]
sandwwraith: AFAIR, we've discussed some ideas about public cache which associates some specific information with descriptors that every serialization format could benefit from. It looks like a better solution instead of 'let 3rd party format authors write their own cache implementations again and again'.

I can agree that this class is generic and vague now, but we still have time to collect feedback about it. Since the library is still experimental, we can continue to change this class and adapt it for our purposes (next candidate, as you've correctly noticed, is protobuf). If experiments show that such format-specific caches can't be fit in a single abstraction, we can just deprecate and remove it.

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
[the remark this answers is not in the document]
qwwdfsad: > but it will make Json a stateful class

It won't. A cache is an optimization that does not add an observable "stateful" behaviour.

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
qwwdfsad: Good points, two-phase lookup seems to be slightly better.

runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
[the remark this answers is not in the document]
qwwdfsad: >AFAIR, we've discussed some ideas about public cache

Nope, about **just** a cache :)

We are again at the design decision "let's add something, maybe it's useful, maybe it's not, we can potentially remove it someday" which is not really a healthy way to do things, even though the library is still experimental.

We neither know if it has any use at all (at current point, at least), nor we will be able to know at some point. If someone uses this API, we will never know about it. Especially taking into account that there is at most dozen people who are going to write its own format. 

It's much better to adapt in in JSON/ProtoBuf, experiment (and benchmark!) different strategies and only then make the decision, not the vice versa.

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
sandwwraith: Just found out that this strategy would prevent us from detecting alternative name collisions with primary ones (because we do not build cache map until primary name is not found). So we either should choose speed or error detection

runtime/commonMain/src/kotlinx/serialization/SchemaCache.kt
```diff
@@ -0,0 +1,41 @@
...
+
+private typealias DescriptorData<T> = MutableMap<DescriptorSchemaCache.Key<T>, T>
+
+/**
+ * A type-safe map for storing custom information (such as format schema), associated with [SerialDescriptor].
+ *
+ * This cache uses ConcurrentHashMap on JVM and regular maps on other platforms.
+ * To be able to work with it from multiple threads in Kotlin/Native, use @[ThreadLocal] in appropriate places.
+ */
+internal class DescriptorSchemaCache {
```
pdvrieze: This type seems like it would be good to be public. It is something that many formats could reuse.

mochadwi:
really love to wait for this features merged

lic2050:
hi， When can this be used？

sandwwraith:
Postponed for after-1.0.0 (1.1.0 likely)

kevinvandenbreemen:
> Postponed for after-1.0.0 (1.1.0 likely)

Hey there,
I am also on a team that could really use this feature.  We are migrating code from GSon, and some of our serialized fields have alternate names.

mustafaozhan:
Hi there 👋  Same here I am really looking forward to this PR to be merged and released 😞

krottv:
I am also looking for this feature. Also it would be nice if aliases could be processed differently. Like I have "durationFrames": 30, "durationMillis": 1000. So json can contain one of the following items.

mustafaozhan:
Unfortunately 1.1.0 Released without this feature, When It can be used ?

sandwwraith:
@mustafaozhan We plan it for the next release

sandwwraith:
Superseded by #1375

mustafaozhan:
@sandwwraith glad to hear that thanks a lot!

PR #600, opened 2019-11-08
READ_ALL rework

qwwdfsad:
* Deprecate READ_ALL marker, replace it with a new method 'decodeSequentially'
    * Document working with indices in CompositeDecoder
    * Rewrite hand-rolled serializers on 'decodeSequentially'

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/Coders.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/Coders.kt
@@ -4,7 +4,7 @@
 
 package kotlinx.serialization
 
-import kotlinx.serialization.modules.SerialModule
+import kotlinx.serialization.modules.*
 
 interface Encoder {
     val context: SerialModule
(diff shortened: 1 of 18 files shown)


runtime/commonMain/src/kotlinx/serialization/Coders.kt
```diff
@@ -125,31 +125,130 @@ interface Decoder {
...
+     * fun decodeElementIndex(descriptor: SerialDescriptor): Int {
+     *     // Ignore arrays
+     *     val nextKey: String? = myStringJsonParser.nextKey()
+     *     if (nextObjectKey == null) return READ_DONE
+     *     return descriptor.getElementIndex(nextKey) // getElementIndex can return UNKNOWN_FIELD
+     * }
+     * ```
      */
-    fun decodeCollectionSize(desc: SerialDescriptor): Int = -1
+    public fun decodeElementIndex(descriptor: SerialDescriptor): Int
```
qwwdfsad: I would also like to extract Composite* into its own file and rename `desc` parameter to `descriptor` as part of this PR, if you don't mind

runtime/jvmJsTest/src/kotlinx/serialization/MapperTest.kt
```diff
@@ -30,8 +30,7 @@ class MapperTest {
     data class NullableData(val nullable: String?, val nullable2: String?, val property: String)
```
qwwdfsad: `MapperTest` does not pass, but it's not the issue right now.
Let's focus on the general concept and plugin changes, eventually, I will fix it

runtime/commonMain/src/kotlinx/serialization/Coders.kt
```diff
@@ -125,31 +125,130 @@ interface Decoder {
...
+    val context: SerialModule
+    val updateMode: UpdateMode
+
+    fun endStructure(desc: SerialDescriptor) {}
+
     /**
-     *  Returns either index or one of READ_XXX constants
+     * Checks whether the current decoder supports strictly ordered decoding of the data
+     * without calling to [decodeElementIndex].
+     * If the method returns `true`, the caller might skip [decodeElementIndex] calls
```
sandwwraith: `might` or `must`? I.e. can I still call `decodeElementIndex` even if `decodeSequentially` returned `true`?

runtime/commonMain/src/kotlinx/serialization/Coders.kt
```diff
@@ -125,31 +125,130 @@ interface Decoder {
...
+     * ```
+     * This example is a rough equivalent of what serialization plugin generates for serializable pair class.
+     *
+     * Sequential decoding is a performance optimization for formats with strictly ordered schema,
+     * usually binary ones. Regular formats such as JSON or ProtoBuf cannot use this optimization,
+     * because e.g. in the latter example, the same data can be represented both as
+     * `{"i": 1, "d": 1.0}`"` and `{"d": 1.0, "i": 1}` (thus, unordered).
      */
-    fun decodeElementIndex(desc: SerialDescriptor): Int
+    public fun decodeSequentially(): Boolean = false
```
sandwwraith: Probably there should be a parameter `descriptor: SerialDescriptor`, too. We can think about use-cases like 'sequential decoding supported for collections and not supported for regular objects' or smth like that

sandwwraith:
Do you plan to commit benchmarks or its results somewhere?

runtime/commonMain/src/kotlinx/serialization/Coders.kt
qwwdfsad: `might`, that's exactly the point for the potential DCE.
You can either avoid calling `decodeSequentially` at all or ignore its result

qwwdfsad:
>Do you plan to commit benchmark or its results somewhere?

Branch `read-all` though I didn't expect these benchmarks to be read by anyone except me (and I do not expect `read-all` to be eventually merged)

runtime/commonMain/src/kotlinx/serialization/Coders.kt
qwwdfsad: Good point!
Let me think about it for a while, it shouldn't really block prototyping in the plugin

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
```diff
@@ -6,179 +6,160 @@ package kotlinx.serialization.internal
...
-                    bSet = true
-                    c = input.decodeSerializableElement(descriptor, 2, cSerializer)
-                    cSet = true
-                    break@mainLoop
-                }
+            when (val index = composite.decodeElementIndex(descriptor)) {
                 CompositeDecoder.READ_DONE -> {
                     break@mainLoop
                 }
                 0 -> {
```
sandwwraith: Redundant curly braces

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -98,6 +100,8 @@ private class JsonPrimitiveInput(json: Json, override val obj: JsonPrimitive) :
         pushTag(PRIMITIVE_TAG)
     }
 
+    override fun decodeElementIndex(desc: SerialDescriptor): Int = 0
```
sandwwraith: Why do we need this here and why it is equal to `0`?

runtime/jvmJsTest/src/kotlinx/serialization/MapperTest.kt
sandwwraith: I believe we need this test working so we can test that compiler still generates valid code in case `Decoder` supports `decodeSequentially`.

However, `Mapper` is not very good case, because, as you may remember, it should be rewritten to properties decoder which supports optional fields, hence we can't rely on `decodeSequentially`.

sandwwraith: I'll fix `TaggedTest`, leaving `Mapper` and `MapperTest` out of scope.

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
qwwdfsad: I personally find consistent braces in a single `when` much more readable

runtime/commonMain/src/kotlinx/serialization/Coders.kt
Dominaezzz: This should've probably been given an name like `canDecodeSequenctially` and not that imperative name.
I'm implementing a BER sequence decoder and thought `decodeSequentially` would change some internal state but it looks like the result can be ignored.
Whilst it's still experimental, please rename this function to reflect this.

sandwwraith: Good point! I've created https://github.com/Kotlin/kotlinx.serialization/issues/1469

Issue #116, opened 2018-05-02
Serialize Kotlin objects to dynamic JS objects


nlochschmidt:
I saw that it's possible to parse `dynamic` objects in Javascript to data classes using koltinx.serialization https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/runtime_usage.md#dynamic-object-parser-js-only

I think it would be a useful addition to be able to serialize directly into a `dynamic` object as well. 
At the moment I can only do this (and go through an intermediate `String` instance):

```kotlin
val jsonString = kotlinx.serialization.json.JSON.stringify(kotlinData)
val dataAsPureJs = kotlin.js.JSON.parse(jsonString)
```

However it would be nice to be able to do this without the intermediate representation.

fkrauthan:
Any plans for thought when this feature might be added to this library? This would be a huge win for frameworks where you have to work a lot with json objects to persist and restore data.

ankushg:
Is this still not possible at this time?

sandwwraith:
Nope :(

ankushg:
Would it be possible to create a JS-platform-only `AbstractSerialFormat` that spits out `dynamic`?

sandwwraith:
Yep, that's how it should be done. Similar to `DynamicObjectParser`.

exaV:
Would a PR be welcome?

sandwwraith:
Fixed in #628

PR #628, opened 2019-12-16
add Dynamic object serializer that allows to convert kotlin data to dynamic

exaV:
Supports maps with primitive keys, lists, and classes

### DIFF
--- a/docs/runtime_usage.md
+++ b/docs/runtime_usage.md
@@ -217,5 +217,25 @@ val dyn = js("""{s:"foo", d:{a:42}}""")
 val parsed = DynamicObjectParser().parse<DataWrapper>(dyn)
 parsed == DataWrapper("foo", Data(42)) // true
 ```
-
 > Parser does not support kotlin maps with keys other than `String`.
+
+ ### Dynamic object serializer 
+ 
+ Allows you to convert kotlin data structures into their dynamic JS representation.
+
+```kotlin
+
+@Serializable
+data class Data(val a: Int)
+
+@Serializable
+open class DataWrapper(open val s: String, val d: Data?)
+
+val wrapper = DataWrapper("foo", Data(42))
+JSON.stringify(wrapper) // {"s_dsrefg$_0":"foo","d":{"a":42}}
+val plainJS: dynamic = DynamicObjectSerializer().serialize(DataWrapper.serializer(), wrapper)
+plainJS.s == wrapper.s // true
+JSON.stringify(plainJS) // {"s":"foo","d":{"a":42}}
+```
+
+
--- a/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
+++ b/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
@@ -1,0 +1,252 @@
+package kotlinx.serialization
+
+import kotlinx.serialization.builtins.AbstractEncoder
+import kotlinx.serialization.builtins.list
+import kotlinx.serialization.builtins.nullable
+import kotlinx.serialization.json.JsonConfiguration
+import kotlinx.serialization.json.internal.BEGIN_LIST
+import kotlinx.serialization.json.internal.BEGIN_OBJ
+import kotlinx.serialization.json.internal.END_LIST
+import kotlinx.serialization.json.internal.END_OBJ
+import kotlinx.serialization.modules.EmptyModule
+import kotlinx.serialization.modules.SerialModule
+import kotlinx.serialization.modules.getContextualOrDefault
+import kotlin.math.abs
+import kotlin.math.floor
+
+
+/**
+ * Converts Kotlin data structures to plain Javascript objects
+ *
+ *
+ * Limitations:
+ * * Map keys must be of primitive or enum type
+ * * Enums are serialized as the value of `@SerialName` if present or their name, in that order.
+ * * Currently does not support polymorphism
+ *
+ * Example of usage:
+ * ```
+ *  @Serializable
+ *  open class DataWrapper(open val s: String, val d: String?)
+ *
+ *  val wrapper = DataWrapper("foo", "bar")
+ *  val plainJS: dynamic = DynamicObjectSerializer().serialize(DataWrapper.serializer(), wrapper)
+ * ```
+ *
+ * @param encodeNullAsUndefined if true null properties will be omitted from the output
+ */
+public class DynamicObjectSerializer @OptIn(UnstableDefault::class) constructor(
+    public val context: SerialModule = EmptyModule,
+    private val configuration: JsonConfiguration = JsonConfiguration.Default,
+    private val encodeNullAsUndefined: Boolean = false
+) {
+
+    public fun <T> serialize(strategy: SerializationStrategy<T>, obj: T): dynamic {
+        if (strategy.descriptor.kind is PrimitiveKind || strategy.descriptor.kind is UnionKind.ENUM_KIND) {
+            val serializer = DynamicPrimitiveEncoder(configuration)
+            serializer.encode(strategy, obj)
+            return serializer.result
+        }
+        val serializer = DynamicObjectEncoder(configuration, encodeNullAsUndefined)
+        serializer.encode(strategy, obj)
+        return serializer.result
+    }
+
+    @ImplicitReflectionSerializer
+    public inline fun <reified T : Any> serialize(obj: T): dynamic =
+        serialize(context.getContextualOrDefault(T::class), obj)
+
+    @ImplicitReflectionSerializer
+    public inline fun <reified T : Any> serialize(obj: List<T?>): dynamic =
+        serialize(context.getContextualOrDefault(T::class).nullable.list, obj)
+}
+
+private class DynamicObjectEncoder(val configuration: JsonConfiguration, val encodeNullAsUndefined: Boolean) :
+    AbstractEncoder() {
+    private object NoOutputMark
+
+    var result: dynamic = NoOutputMark
+    lateinit var current: Node
+    var currentName: String? = null
+    lateinit var currentDescriptor: SerialDescriptor
+    var currentElementIsMapKey = false
+
+    class Node(val writeMode: WriteMode, val jsObject: dynamic) {
+        var index: Int = 0
+        lateinit var parent: Node
+    }
+
+    enum class WriteMode {
+        OBJ, MAP, LIST
+    }
+
+    override fun encodeElement(descriptor: SerialDescriptor, index: Int): Boolean {
+        current.index = index
+        currentDescriptor = descriptor
+
+        if (current.writeMode == WriteMode.MAP) {
+            currentElementIsMapKey = current.index % 2 == 0
+        } else {
+            currentName = descriptor.getElementName(index)
+        }
+        return true
+    }
+
+    override fun encodeValue(value: Any) {
+        if (currentElementIsMapKey) {
+            currentName = value.toString()
+        } else {
+            current.jsObject[currentName] = value
+        }
+    }
+
+    override fun encodeChar(value: Char) {
+        encodeValue(value.toString())
+    }
+
+    override fun encodeNull() {
+        if (currentElementIsMapKey) {
+            currentName = null
+        } else {
+            if (encodeNullAsUndefined) return // omit element
+
+            current.jsObject[currentName] = null
+        }
+    }
+
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+        encodeValue(enumDescriptor.getElementName(index))
+    }
+
+    override fun encodeLong(value: Long) {
+        val asDouble = value.toDouble()
+        val conversionHasLossOfPrecision = abs(asDouble) > MAX_SAFE_INTEGER
+
+        if (!configuration.isLenient && conversionHasLossOfPrecision) {
+            throw IllegalArgumentException(
+                "$value can't be deserialized to number due to a potential precision loss. " +
+                        "Use the JsonConfiguration option isLenient to serialise anyway"
+            )
+        }
+
+        if (currentElementIsMapKey && conversionHasLossOfPrecision) {
+            throw IllegalArgumentException(
+                "Long with value $value can't be used in json as map key, because its value is larger than Number.MAX_SAFE_INTEGER"
+            )
+        }
+
+        encodeValue(asDouble)
+
+    }
+
+    override fun encodeFloat(value: Float) {
+        encodeDouble(value.toDouble())
+    }
+
+    override fun encodeDouble(value: Double) {
+        if (currentElementIsMapKey) {
+            val hasNonZeroFractionalPart = floor(value) != value
+            if (!value.isFinite() || hasNonZeroFractionalPart) {
+                throw IllegalArgumentException(
+                    "Double with value $value can't be used in json as map key, because its value is not an integer."
+                )
+            }
+        }
+        encodeValue(value)
+    }
+
+    override fun shouldEncodeElementDefault(descriptor: SerialDescriptor, index: Int) = configuration.encodeDefaults
+
+    fun enterNode(jsObject: dynamic, writeMode: WriteMode) {
+        val child = Node(writeMode, jsObject)
+        child.parent = current
+        current = child
+    }
+
+    fun exitNode() {
+        current = current.parent
+        currentElementIsMapKey = false
+    }
+
+    override fun beginStructure(
+        descriptor: SerialDescriptor,
+        vararg typeSerializers: KSerializer<*>
+    ): CompositeEncoder {
+        // we currently do not structures as map key
+        if (currentElementIsMapKey) {
+            throw IllegalArgumentException(
+                "Value of type ${descriptor.serialName} can't be used in json as map key. " +
+                        "It should have either primitive or enum kind, but its kind is ${descriptor.kind}."
+            )
+        }
+
+        val newMode = selectMode(descriptor)
+        if (result === NoOutputMark) {
+
+            result = newChild(newMode)
+            current = Node(newMode, result)
+            current.parent = current
+        } else {
+            val child = newChild(newMode)
+            current.jsObject[currentName] = child
+            enterNode(child, newMode)
+        }
+
+        current.index = 0
+        return this
+    }
+
+    fun newChild(writeMode: WriteMode) = when (writeMode) {
+        WriteMode.OBJ, WriteMode.MAP -> js(BEGIN_OBJ.toString() + END_OBJ)
+        WriteMode.LIST -> js(BEGIN_LIST.toString() + END_LIST)
+    }
+
+    override fun endStructure(descriptor: SerialDescriptor) {
+        exitNode()
+    }
+
+    fun selectMode(desc: SerialDescriptor) = when (desc.kind) {
+        StructureKind.CLASS, StructureKind.OBJECT, UnionKind.CONTEXTUAL -> WriteMode.OBJ
+        StructureKind.LIST, is PolymorphicKind -> WriteMode.LIST
+        StructureKind.MAP -> WriteMode.MAP
+        is PrimitiveKind, UnionKind.ENUM_KIND -> {
+            // the two cases are handled in DynamicObjectSerializer. But compiler does not know
+            error("DynamicObjectSerializer does not support serialization of singular primitive values or enum types.")
+        }
+    }
+}
+
+private class DynamicPrimitiveEncoder(private val configuration: JsonConfiguration) : AbstractEncoder() {
+    var result: dynamic = null
+
+    override fun encodeNull() {
+        result = null
+    }
+
+    override fun encodeLong(value: Long) {
+        val asDouble = value.toDouble()
+
+        if (!configuration.isLenient && abs(value) > MAX_SAFE_INTEGER) {
+            throw IllegalArgumentException(
+                "$value can't be deserialized to number due to a potential precision loss. " +
+                        "Use the JsonConfiguration option isLenient to serialise anyway"
+            )
+        }
+        encodeValue(asDouble)
+    }
+
+    override fun encodeChar(value: Char) {
+        encodeValue(value.toString())
+    }
+
+    override fun encodeValue(value: Any) {
+        result = value
+    }
+
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+        encodeValue(enumDescriptor.getElementName(index))
+    }
+
+    override fun endStructure(descriptor: SerialDescriptor) {
+    }
+}
--- a/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
+++ b/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
@@ -125,7 +125,7 @@ private class DynamicObjectEncoder(val configuration: JsonConfiguration, val enc
         if (!configuration.isLenient && conversionHasLossOfPrecision) {
             throw IllegalArgumentException(
                 "$value can't be serialized to number due to a potential precision loss. " +
-                        "Use the JsonConfiguration option isLenient to serialise anyway"
+                        "Use the JsonConfiguration option isLenient to serialize anyway"
             )
         }
 
--- a/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
+++ b/runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
@@ -124,7 +124,7 @@ private class DynamicObjectEncoder(val configuration: JsonConfiguration, val enc
 
         if (!configuration.isLenient && conversionHasLossOfPrecision) {
             throw IllegalArgumentException(
-                "$value can't be deserialized to number due to a potential precision loss. " +
+                "$value can't be serialized to number due to a potential precision loss. " +
                         "Use the JsonConfiguration option isLenient to serialise anyway"
             )
         }
--- a/runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
+++ b/runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
@@ -1,0 +1,369 @@
+package kotlinx.serialization
+
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonConfiguration
+import kotlinx.serialization.modules.EmptyModule
+import kotlinx.serialization.modules.getContextualOrDefault
+import kotlin.test.*
+
+
+@Suppress("UnsafeCastFromDynamic")
+class DynamicSerializerTest {
+    @Serializable
+    data class Data(val a: Int)
+
+    @Serializable
+    open class DataWrapper(open val s: String, val d: Data? = Data(1))
+
+    @Serializable
+    data class NestedList(val a: String, val list: List<Int>)
+
+    @Serializable
+    data class ListOfLists(val l: List<List<Data>>)
+
+    @Serializable
+    data class MapWrapper(val m: Map<String?, Int>)
+
+    @Serializable
+    data class ComplexMapWrapper(val m: Map<String, Data>)
+
+    @Serializable
+    data class WithChar(val a: Char)
+
+    @Serializable
+    data class WithLong(val l: Long)
+
+    @Serializable
+    data class AllTypes(
+        val b: Byte,
+        val s: Short,
+        val i: Int,
+        val f: Float,
+        val d: Double,
+        val c: Char,
+        val B: Boolean,
+        val S: String
+    )
+
+    @Serializable
+    data class EnumWrapper(val e: Color)
+
+    @Serializable
+    sealed class Sealed {
+        @Serializable
+        data class One(val string: String) : Sealed()
+    }
+
+    @Serializable
+    class WithJsName(@JsName("b") val a: String)
+
+    @Serializable
+    class WithSerialName(@SerialName("b") val a: String)
+
+    @Serializable
+    enum class Color {
+        RED,
+        GREEN,
+
+        @SerialName("red")
+        WITH_SERIALNAME_red
+    }
+
+    @Serializable
+    class MyFancyClass(val value: String) {
+
+        @Serializer(forClass = MyFancyClass::class)
+        companion object : KSerializer<MyFancyClass> {
+
+            override val descriptor: SerialDescriptor = PrimitiveDescriptor("MyFancyClass", PrimitiveKind.STRING)
+            override fun serialize(encoder: Encoder, value: MyFancyClass) {
+                encoder.encodeString("fancy ${value.value}")
+            }
+
+            override fun deserialize(decoder: Decoder): MyFancyClass {
+                return MyFancyClass(decoder.decodeString().removePrefix("fancy "))
+            }
+        }
+    }
+
+    private inline fun <reified T : Any> assertDynamicForm(
+        data: T,
+        serializer: SerializationStrategy<T>? = null,
+        noinline assertions: ((T, dynamic) -> Unit)? = null
+    ) {
+        val effectiveSerializer = serializer ?: EmptyModule.getContextualOrDefault(T::class)
+        val serialized = DynamicObjectSerializer().serialize(effectiveSerializer, data)
+        assertions?.invoke(data, serialized)
+        assertEquals(
+            Json.stringify(effectiveSerializer, data),
+            JSON.stringify(serialized),
+            "JSON.stringify representation must be the same"
+        )
+    }
+
+    @Test
+    fun dynamicSimpleTest() {
+        assertDynamicForm(Data(42)) { data, serialized ->
+            assertEquals(data.a, serialized.a)
+        }
+
+        assertDynamicForm(WithChar('c')) { data, serialized ->
+            assertEquals(data.a.toString(), serialized.a)
+        }
+
+        assertDynamicForm(AllTypes(1, 2, 3, 4.0f, 5.0, 'c', true, "string"))
+
+
+        assertDynamicForm(WithLong(5L))
+        assertDynamicForm(WithLong(MAX_SAFE_INTEGER.toLong()))
+        assertDynamicForm(WithLong(MAX_SAFE_INTEGER.unaryMinus().toLong()))
+    }
+
+    @Test
+    fun wrappedObjectsTest() {
+        assertDynamicForm(DataWrapper("a string", Data(42))) { data, serialized ->
+            assertEquals(data.s, serialized.s)
+            assertNotNull(serialized.d)
+            assertEquals(data.d?.a, serialized.d.a)
+        }
+    }
+
+    @Test
+    fun nullsTest() {
+        val data = DataWrapper("a string", null)
+
+        val serialized = DynamicObjectSerializer(encodeNullAsUndefined = true).serialize(data)
+        assertNull(serialized.d)
+        assertFalse(js("""Object.keys(serialized).includes("d")"""), "should omit null properties")
+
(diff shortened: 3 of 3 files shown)


exaV:
Implements https://github.com/Kotlin/kotlinx.serialization/issues/116

exaV:
I am mainly looking for feedback. There are at least 3 TODOs in the code for which I am looking for feedback.
* Should we serialize nulls by omitting the key and value in the output. Or should we add the key with `null` as the value
* Serialdescriptor.getElementName() sometimes returns an int if the datastructure is a collection. Can I use that or is it just an implementation detail?
* how does WriteMode.POLY_OBJ differ from WriteMode.OBJ and how should we serialize and object in that mode?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import kotlinx.serialization.json.Json
+import kotlin.test.*
+
+
+class DynamicSerializerTest {
+    @Serializable
+    data class Data(val a: Int)
+
+    @Serializable
+    open class DataWrapper(open val s: String, val d: Data?)
```
sandwwraith: Same question about `open`

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    fun dynamicSimpleTest() {
+        val data = Data(42)
+        val serialized = DynamicObjectSerializer().serialize(data)
+        assertEquals(Json.stringify(data), JSON.stringify(serialized))
+        assertEquals(data.a, serialized.a)
+
+        val data2 = WithChar('c')
+        val serialized2 = DynamicObjectSerializer().serialize(data2)
+        assertEquals(Json.stringify(data2), JSON.stringify(serialized2), "should serialize characters as strings")
+        assertEquals(data2.a.toString(), serialized2.a) // FIXME why is toString needed here, both should be char?
```
sandwwraith: Probably, because natively chars are represented as ints in Kotlin/JS. You can leave this `.toString()`

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Test
+    fun nestedListTest() {
+        val data = NestedList("a string", listOf(1, 2, 3, 44))
+        val serialized = DynamicObjectSerializer().serialize(data)
+        assertEquals(Json.stringify(data), JSON.stringify(serialized))
+        assertEquals(data.a, serialized.a)
+        assertNotNull(serialized.list)
+        assertEquals(data.list[0], serialized.list[0])
```
sandwwraith: Why not make a loop?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        assertEquals(data.list[1], serialized.list[1])
+        assertEquals(data.list[2], serialized.list[2])
+        assertEquals(data.list[3], serialized.list[3])
+    }
+
+    @Test
+    fun complexMapWrapperTest() {
+        val data = ComplexMapWrapper(mapOf("key1" to Data(1), "key2" to Data(2)))
+        val serialized = DynamicObjectSerializer().serialize(data)
+        assertEquals(Json.stringify(data), JSON.stringify(serialized))
```
sandwwraith: You can reduce copy-paste by extracting these 3 lines into a function similar to `assertStringFormAndRestored` (e.g. `assertDynamicForm`)

docs/runtime_usage.md
```diff
@@ -220,5 +220,25 @@ val dyn = js("""{s:"foo", d:{a:42}}""")
...
+
+@Serializable
+data class Data(val a: Int)
+
+@Serializable
+open class DataWrapper(open val s: String, val d: Data?)
+
+val wrapper = DataWrapper("foo", Data(42))
+JSON.stringify(wrapper) // {"s_dsrefg$_0":"foo","d":{"a":42}}
+val plainJS: dynamic = DynamicObjectSerializer().serialize(wrapper)
```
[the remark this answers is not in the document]
pdvrieze: In this case are you mainly trying to use serialization to get a representation that removes the name mangling for properties (and thereby is usable with plain javascript)? Any reason not to use `@JsName` instead?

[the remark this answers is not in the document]
sandwwraith: It also affects data structures like lists as well

docs/runtime_usage.md
```diff
@@ -220,5 +220,25 @@ val dyn = js("""{s:"foo", d:{a:42}}""")
...
+ 
+ Allows you to convert kotlin data structures into their dynamic JS representation.
+
+```kotlin
+
+@Serializable
+data class Data(val a: Int)
+
+@Serializable
+open class DataWrapper(open val s: String, val d: Data?)
```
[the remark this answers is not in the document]
exaV: I thought it was important to show how we are unaffected by name mangling here since it is one of the use cases for DynamicObjectSerializer. 
We could also just mention it in the text above and leave out the `open val` from the example if you prefer.

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
exaV: The test shows that the names are not mangled after serialization.

exaV:
Thanks for having a look :)

I have one TODO left in the code on line 115 in DynamicObjectSerializer.kt. I am not sure what WriteMode.POLY_OBJ is about.

exaV:
The implementation encodes maps as JS objects `"{}"` and thus throws an exception when a key is not String or Int. 
We could convert to [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead, which has been supported by major browsers for some time, with an opt-in parameter for DynmicObjectSerializer. However I have not seen any support for Map in all of Kotlin/js so I am a bit hesitant to just add it.

sandwwraith:
@exaV 
POLY_OBJ is basically a mode which is used by useArrayPolymorphism flag (see more here: https://github.com/kotlin/kotlinx.serialization/blob/c4acfb69b8/docs/polymorphism.md#a-bit-of-customizing). So leaving js(BEGIN_LIST.toString() + END_LIST) here will likely get things done.

> The implementation encodes maps as JS objects "{}" and thus throws an exception when a key is not String or Int.

I think this is acceptable behaviour; use case with javascript `Map` is quite rare.

I see you've fixed most of my remarks, but there are some of them left. Do you have further questions?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,191 @@
...
+        val S: String
+    )
+
+    @Serializable
+    data class EnumWrapper(val e: Color)
+
+    private inline fun <reified T : Any> assertDynamicForm(
+        data: T,
+        serializer: SerializationStrategy<T>? = null,
+        noinline assertions: ((T, dynamic) -> Unit)? = null
```
ankushg: Should `assertions` be allowed to be null? It seems like there are currently quite a few test cases that don't assert anything regarding the input/output, but that is likely important?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,191 @@
...
+            , serializer = NestedCollections.serializer()
+        )
+    }
+
+    @Test
+    fun enums() {
+//        assertDynamicForm(Color.GREEN, serializer = CommonEnumSerializer("Color"))
+        assertDynamicForm(EnumWrapper(Color.RED)) { data, serialized ->
+
+        }
```
ankushg:
```suggestion
        assertDynamicForm(EnumWrapper(Color.RED))
```

Doesn't seem like you have any custom assertions, so we might as well omit it?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,191 @@
...
+                    "two" to mapOf("twotwo" to listOf(22, 23))
+                )
+            )
+            , serializer = NestedCollections.serializer()
+        )
+    }
+
+    @Test
+    fun enums() {
+//        assertDynamicForm(Color.GREEN, serializer = CommonEnumSerializer("Color"))
```
ankushg: Should this be uncommented? Are custom serializers for enums supported?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,191 @@
...
+        assertDynamicForm(EnumWrapper(Color.RED)) { data, serialized ->
+
+        }
+    }
+
+    enum class Color {
+        RED, GREEN
+    }
```
ankushg: Should this class be moved to the top, with the rest of the classes that you're testing?

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        current.index = 0
+        return this
+    }
+
+    internal fun newStructure(mode: WriteMode) = when (mode) {
+        WriteMode.OBJ -> js(BEGIN_OBJ.toString() + END_OBJ)
+        WriteMode.MAP -> js(BEGIN_OBJ.toString() + END_OBJ)
+        WriteMode.LIST -> js(BEGIN_LIST.toString() + END_LIST)
+        WriteMode.POLY_OBJ -> TODO()
```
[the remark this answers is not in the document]
ankushg: The reason we don't just use `begin`/`end` from the `WriteMode` itself (and essentially redeclare them here) is because the compiler doesn't treat `begin`/`end` as string constants, correct?

exaV:
Hi all, thanks for the reviews. Special thanks to @sandwwraith, you actually left a very good review the first time around and answered all my questions. I didn't see the answers at first because they were collapsed and marked as outdated by Github, I will definitely check "outdated" comments more carefully in the future.

I (think) I accounted for all the suggestions made. Please let me know if there is anything we should change.

Compared to the first round of reviews, we now:
* support enums
* support singular values e.g. a primitive Int (for consistency's sake)
* reflect the tree like structure of the serialization output in our naming

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
exaV: it was not supported before, I added support for it in the last commit.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+                val value: Char = obj // without the explicit type annotation int value would be used
+                return value.toString()
+            }
+            if (obj is Long) {
+                return obj.toDouble()
+            }
+            return obj
+        }
+        if (strategy.descriptor.kind is UnionKind.ENUM_KIND) {
+            return obj.unsafeCast<Enum<*>>().name
```
ankushg: Does this still work if we have a custom serializer for the enum that _doesn't_ use `name`?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,205 @@
...
+        val c: Char,
+        val B: Boolean,
+        val S: String
+    )
+
+    @Serializable
+    data class EnumWrapper(val e: Color)
+
+    enum class Color {
+        RED, GREEN
```
ankushg: Can we add some `@SerialId` and `@SerialName` test cases to make sure we cover those situations, especially with enums where it's easy to accidentally use `name`?

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,205 @@
...
+                )
+            )
+            , serializer = NestedCollections.serializer()
+        )
+    }
+
+    @Test
+    fun enums() {
+        assertDynamicForm(EnumWrapper(Color.RED))
+        assertDynamicForm(Color.GREEN, serializer = CommonEnumSerializer("Color"))
```
ankushg: The default `CommonEnumSerializer` seems to just serialize using `name`, so I don't think this test is a good indicator for using custom enum serializers.

Can we replace this test with a serializer that behaves differently from the default behavior?

Maybe using something like 

```kt
val serializer = CommonEnumSerializer(
  serialName, 
  enumValues(), 
  enumValues<Color>()
    .map { "${it.name}_${it.ordinal}" }
    .toTypedArray()
)
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    override fun encodeValue(value: Any) {
+        val elementIsMapKey = current.writeMode == WriteMode.MAP && current.index % 2 == 0
+        if (elementIsMapKey) {
+            val elementDescriptor = currentDescriptor.getElementDescriptor(current.index)
+            currentName = if (elementDescriptor.kind is PrimitiveKind) {
+                value.toString()
+            } else if (elementDescriptor.kind is UnionKind.ENUM_KIND && value is Enum<*>) {
+                value.name
```
ankushg: Should we honor custom serializers or `@SerialName`/`@SerialId` annotations for enums when being used as map keys?

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+    fun exitNode() {
+        current.parent
+        current = current.parent
+    }
+
+    override fun beginStructure(desc: SerialDescriptor, vararg typeParams: KSerializer<*>): CompositeEncoder {
+        val newMode = selectMode(desc)
+
+        if (result === NoOutputMark) {
+
```
ankushg:
```suggestion
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: The if block will match any singular enum e.g. `DynamicObjectSerializer.serialize(Color.RED)`.

I am not sure what you are getting at. Are you worried about about a use-case where someone would want their enums to be serialised to a different name?

```
enum class Color {
@SerialName("red")
RED
}
```

The code above would not consider the SerialName annotation, or a custom serialiser if there were one. I think we would have to introduce a `SingularValueEncoder` to handle the case correctly.

exaV: The if block will match any singular enum e.g. `DynamicObjectSerializer.serialize(Color.RED)`.

I am not sure what you are getting at. Are you worried about about a use-case where someone would want their enums to be serialised to a different name?

```
enum class Color {
@SerialName("red")
RED
}
```

The code above would not consider the SerialName annotation, or a custom serialiser if there were one. I think we would have to introduce a separate `DynamicValueEncoder` class to handle the case correctly, because `DynamicObjectEncoder` only deals with classes and collections.

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
exaV: The test case may have been confusing, I only wanted to test the serialization of a singular enum. I will add `@Serializable` to the Color enum and remove the custom one here.

We can also add an additional test with custom serialisers.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: We now have two cases where we (naively) serialize Enum as `Enum.name`. We will have to again resort to something like `DynamicValueEncoder` to do it correctly.

Also if we were supporting custom serialisers here, we would have to support them for all types  iff the serialiser returns a legal javascript object key (that is String or Int).

exaV:
Is there any way to support the annotation `@JsName` in a similar way to `@SerialName`? 

In existing code where I'd like to use `DynamicObjectSerializer` is already using `@SerialName` for interop. Unfortunately that annotation currently has no effect on this serializer. 

IMO the implementation should use `JsName` if `SerialName` is not used and some config property is enabled e.g. `useJsName`

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
ankushg: > Also if we were supporting custom serialisers here, we would have to support them for all types iff the serialiser returns a legal javascript object key (that is String or Int).

Yeah that's fair. I think the most important thing is to be clear about what the limitations are.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
ankushg: > Are you worried about about a use-case where someone would want their enums to be serialised to a different name?

Yeah that's exactly it. 

I know several folks (including myself) who use `@SerialName`, `@SerialId`, or custom `Serializer` implementations for enum classes (and of course other classes) since `ordinal`/`name` might not be consistent across codebases that still need to communicate with each other.

ankushg:
I agree that it would be nice to factor `@JsName` into `dynamic` serialization, but maybe adding new serialization behavior/flags is out of scope for this PR?

I think it's probably best to focus this PR on serializing with the already-standard `@SerialId`/`@SerialName`/`Serializer` interactions, and make a separate PR to propose a new `@JsName` behavior

sandwwraith:
I'm not sure if `@JsName` annotation can be obtained at runtime in any way; afair, it is a compile-time annotation.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
sandwwraith: It would be nice to support enums with `@SerialName` here despite singular customized enums are rare case —because it is already supported in `encodeEnum` for wrapped enums. I've not tested it myself (yet), but i believe that replacing this line with `strategy.descriptor.getElementName(obj.unsafeCast<Enum<*>>().ordinal)` would be sufficient — since custom serial names are stored in descriptor.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
sandwwraith: You can apply here the same technique that I described in previous comment

sandwwraith: However, there's a bigger problem with that: `encodeValue` is a fallback method, i.e. if map key is `Char` or `Enum`, `encodeChar` and `encodeEnum` are called, respectively — and this method won't be called. Seems like you have two options: 1. Move 'elementIsMapKey' check to a separate function and call it in the beginning of `encodeChar/Enum` 2. Move check to a `encodeElement` function (check itself does not require actual value, only descriptor) and then delegate to a `encodeValue` manually (e.g. `override fun encodeEnum(enumDescription: SerialDescriptor, ordinal: Int) = encodeValue(enumDescription.getElementName(ordinal))`

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+            current.jsObject[currentName] = value
+        }
+    }
+
+    override fun encodeChar(value: Char) {
+        current.jsObject[currentName] = value.toString()
+    }
+
+    override fun encodeNull() {
+        if (encodeDefaults &&
```
sandwwraith: You don't need this check since `shouldEncodeElementDefault` is called by serializer _before_ any `encode` calls. So you should just override it: `override fun shouldEncodeElementDefault(desc: SerialDescriptor, index: Int): Boolean = encodeDefaults`

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    fun enterNode(jsObject: dynamic, writeMode: WriteMode) {
+        val child = Node(writeMode, jsObject)
+        child.parent = current
+        current = child
+    }
+
+    fun exitNode() {
+        current.parent
```
sandwwraith: unused statement?

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: Thanks for the suggestions, they helped to arrive a somewhat elegant solution.
We delegate writing operations to `encodeValue`. We also keep track of `currentElementIsMapKey` and use it in `beginStructure` to throw an exception and in  `encodeNull` to write null as the map key if necessary.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: Are you sure? Maybe I am doing something wrong but the shouldEncodeElementDefault method is not called at all if I omit the call here.
Do I need to rebase first?

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: We now support `@Serialname` on enums, both for map keys and for singular enums (and of course for regular properties).

exaV: I also documented what is currently not supported which is 
```
 * Limitations:
 * * Map keys must be of primitive or enum type
 * * Custom serializers are not supported
 * * Enums are serialized as the value of `@SerialName` if present or their name, in that order. Ordinals are not supported.
 * * There is no analogue to DynamicObjectParsers' useArrayPolymorphism flag
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
sandwwraith: No, it's quite old functionality. Maybe you confused `class X(val s: String? = null)` vs `class Y(val s: String?)` – in case of Y, `shouldEncodeElementDefault` indeed would not be called. If you still want to drop nulls regardless whether they are optional or not, (which may confuse `DynamicObjectParser`, which tries to make the difference between property being `null` and missing (`undefined`)), then you indeed should apply custom logic here (which basically would be 'do not write null if encodeDefaults=true)

runtime/jsTest/src/kotlinx/serialization/DynamicSerializerTest.kt
```diff
@@ -0,0 +1,338 @@
...
+        }
+    }
+
+    @Test
+    fun nullsTest() {
+        val data = DataWrapper("a string", null)
+
+        val serialized = DynamicObjectSerializer(encodeDefaults = true).serialize(data)
+        assertNull(serialized.d)
+        assertFalse(js("""Object.keys(serialized).includes("d")"""), "should omit null properties")
```
sandwwraith: Idea of 'omitting defaults' is that we should omit _default property_ value, not _null_. Since the `DataWrapper` has declaration `val d: Data? = Data(1)`, we should omit d only if it is equal to `Data(1)`. That's automatically detected by generated serializer and then `shouldEncodeElementDefault` is called.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,180 @@
...
+
+    override fun encodeChar(value: Char) {
+        encodeValue(value.toString())
+    }
+
+    override fun encodeNull() {
+        if (currentElementIsMapKey) {
+            currentName = null
+        } else {
+            if (shouldEncodeElementDefault(currentDescriptor, current.index) &&
```
sandwwraith: If you want the converter to omit exactly `null`s, not all the default values (although I'm not sure if this is reliable in terms of parsing data back with `DynamicObjectParser`, then you can rename the flag and just check it there, without calling other methods.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: Thanks for clearing things up. I now understand that the intended behaviour of `shouldEncodeElementDefault` is to omit properties that *exactly* have the default value. Which totally makes sense given its name.

IMO the intended behaviour of `shouldEncodeElementDefault` is almost never what I would want from `DynamicObjectSerializer`.

The original problem I wanted to solve was indeed to differentiate between undefined and null. I will add an optional property `omitNullProperties = false` and document it accordingly. We could still leave the optional `shouldEncodeElementDefault = true` and change the implementation to what you suggested above.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: IMO `Dynamic` is not a serialisation format in the traditional sense, therefore I think it is fine to have options which will not play nice with the `DynamicObjectParser `. I will still set the default value of the option to encode null as `null`, so parsing the output of `DynamicObjectSerializer` should work by default.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+ * Converts Kotlin data structures to plain Javascript objects
+ *
+ *
+ * Limitations:
+ * * Map keys must be of primitive or enum type
+ * * Custom serializers are not supported
+ * * Enums are serialized as the value of `@SerialName` if present or their name, in that order. Ordinals are not supported.
+ * * There is no analogue to DynamicObjectParsers' useArrayPolymorphism flag
+ *
+ * @param encodeDefaults if true any property which equals exactly its default value will be omitted
```
EdwarDDay: if `encodeDefaults` is true, the default values should be encoded and not omitted.
```suggestion
 * @param encodeDefaults if false any property which equals exactly its default value will be omitted
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+ * @param encodeNullAsUndefined if true null properties will be omitted from the output
+ */
+class DynamicObjectSerializer(
+    private val encodeDefaults: Boolean = true,
+    private val encodeNullAsUndefined: Boolean = false,
```
EdwarDDay: Maybe create a configuration class, like `JsonConfiguration`.

exaV: It seems like an overkill to me. I'd rather discuss whether we need the options in the first place. It's hard for me to imagine a use case for `encodeDefaults` (in the context of DynamicObjectSerializer) and while I have a use case for `encodeNullAsUndefined` it is also a bit out there.

My use-case for `encodeNullAsUndefined`: I want to pass some data to a table component (e.g. https://material-table.com/#/). I know (that for some business reason) some property will always be null, therefore I use `encodeNullAsUndefined ` to omit nulls which means there will be no column for that property in the table (instead of an empty one).

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Converts Kotlin data structures to plain Javascript objects
+ *
+ *
+ * Limitations:
+ * * Map keys must be of primitive or enum type
+ * * Custom serializers are not supported
+ * * Enums are serialized as the value of `@SerialName` if present or their name, in that order. Ordinals are not supported.
+ * * There is no analogue to DynamicObjectParsers' useArrayPolymorphism flag
```
sandwwraith: I'd rephrase this to 'currently does not support polymorphism'. It seems it doesn't work even with array polymorphism now. I have some thoughts how to fix that, so you can leave that to me.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
sandwwraith: I think it is fine to have two separate flags. Maybe it's even ok to put full `JsonConfiguration` itself here, see https://github.com/Kotlin/kotlinx.serialization/pull/694/files#r376207786 , but I do not insist.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fun selectMode(desc: SerialDescriptor) =
+        when (desc.kind) {
+            StructureKind.CLASS, UnionKind.OBJECT -> WriteMode.OBJ
+            StructureKind.LIST, is PolymorphicKind -> WriteMode.LIST
+            StructureKind.MAP -> WriteMode.MAP
+            is PrimitiveKind, UnionKind.ENUM_KIND -> {
+                // the two cases are handled in DynamicObjectSerializer. But compiler does not know
+                throw IllegalArgumentException("DynamicObjectSerializer does not support serializing singular values of primitive or enum type.")
```
sandwwraith:
```suggestion
                error("DynamicObjectSerializer does not support serialization of singular primitive values or enum types.")
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Converts Kotlin data structures to plain Javascript objects
+ *
+ *
+ * Limitations:
+ * * Map keys must be of primitive or enum type
+ * * Custom serializers are not supported
+ * * Enums are serialized as the value of `@SerialName` if present or their name, in that order. Ordinals are not supported.
+ * * There is no analogue to DynamicObjectParsers' useArrayPolymorphism flag
+ *
```
sandwwraith: It also may be useful to put code sample here from runtime_usage.md, how it is done for JSON: https://github.com/kotlin/kotlinx.serialization/blob/85c0b12df4d7656a81bd4997b1b0a21ab91da35b/runtime/commonMain/src/kotlinx/serialization/json/Json.kt#L25

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+            StructureKind.CLASS, UnionKind.OBJECT -> WriteMode.OBJ
+            StructureKind.LIST, is PolymorphicKind -> WriteMode.LIST
+            StructureKind.MAP -> WriteMode.MAP
+            is PrimitiveKind, UnionKind.ENUM_KIND -> {
+                // the two cases are handled in DynamicObjectSerializer. But compiler does not know
+                throw IllegalArgumentException("DynamicObjectSerializer does not support serializing singular values of primitive or enum type.")
+            }
+        }
+
+    override fun toString(): String {
```
sandwwraith: Since it is useful only as a debug information while developing, I suggest removing this before merging.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,182 @@
...
+import kotlinx.serialization.modules.getContextualOrDefault
+
+
+/**
+ * Converts Kotlin data structures to plain Javascript objects
+ *
+ *
+ * Limitations:
+ * * Map keys must be of primitive or enum type
+ * * Custom serializers are not supported
```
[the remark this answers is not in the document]
exaV: Thanks, I did not realise. I added a `DynamicPrimitiveEncoder : ElementValueEncoder` which takes care of serialising the primitive and enum cases. Now we support custom serialisers in all cases.

exaV:
I also discovered that Longs were not serialised correctly. I now convert them to double before assigning to dynamic. This works for longs where abs(value) <= MAX_SAFE_INTEGER. Larger longs will be converted with a loss of precision if `JsonConfiguration.isLenient` is set, otherwise an exception is thrown.

exaV:
I rebased onto kotlinx/serialisation:dev, updated my api usage (replace deprecated stuff i used and so on) and squashed my commits.
I also discovered and fixed an edge case where Double is used as a map key with non-integer values.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: we now support custom serialisers everywhere :)

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,252 @@
...
+ * Example of usage:
+ * ```
+ *  @Serializable
+ *  open class DataWrapper(open val s: String, val d: String?)
+ *
+ *  val wrapper = DataWrapper("foo", "bar")
+ *  val plainJS: dynamic = DynamicObjectSerializer().serialize(DataWrapper.serializer(), wrapper)
+ * ```
+ *
+ * @param encodeNullAsUndefined if true null properties will be omitted from the output
```
sandwwraith: This param doc is obsolete now

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,252 @@
...
+        encodeValue(enumDescriptor.getElementName(index))
+    }
+
+    override fun encodeLong(value: Long) {
+        val asDouble = value.toDouble()
+        val conversionHasLossOfPrecision = abs(asDouble) > MAX_SAFE_INTEGER
+
+        if (!configuration.isLenient && conversionHasLossOfPrecision) {
+            throw IllegalArgumentException(
+                "$value can't be deserialized to number due to a potential precision loss. " +
```
sandwwraith:
```suggestion
                "$value can't be serialized to number due to a potential precision loss. " +
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
```diff
@@ -0,0 +1,252 @@
...
+                        "Use the JsonConfiguration option isLenient to serialise anyway"
```
sandwwraith:
```suggestion
                        "Use the JsonConfiguration option isLenient to serialize anyway"
```

runtime/jsMain/src/kotlinx/serialization/DynamicObjectSerializer.kt
exaV: The parameter is still there as well. Is there a well suited parameter that's already part of JsonConfiguration we could use?

sandwwraith: Indeed, my bad. There is no suitable parameter [yet], so you may leave it as is.

sandwwraith:
Thank you once again for your help and work!

Issue #273, opened 2018-11-21
build.gradle.kts Gradle Kotlin DSL example


jrhy:
Could you please provide a Gradle example using the Kotlin DSL?  I am trying this:

```
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
import org.jetbrains.kotlin.gradle.dsl.Coroutines

buildscript {
	val kotlin_version = "1.3.10"
	repositories {
		mavenCentral()
	}
	dependencies {
		classpath(kotlin("gradle-plugin", kotlin_version))
		classpath("org.jetbrains.kotlin:kotlin-serialization:$kotlin_version")
	}
}

plugins {
    kotlin("jvm") version "1.3.10"
    "kotlinx-serialization"
}

dependencies {
    compile(kotlin("stdlib-jdk8"))
    compile(kotlin("reflect"))
    testCompile(kotlin("test"))
    testCompile(kotlin("test-junit"))

    compile("com.fasterxml.jackson.module:jackson-module-kotlin:2.9.5")
    testCompile("io.kotlintest:kotlintest-runner-junit5:3.1.7")
    compile("redis.clients:jedis:2.9.0")
    compile("org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.9.0")
}

tasks.withType<KotlinCompile> {
    kotlinOptions {
        jvmTarget = "1.8"
        freeCompilerArgs = listOf("-Xjsr305=strict")
    }
}

repositories {
    mavenCentral()
    maven (url="https://kotlin.bintray.com/kotlinx" )
}
```

But when I try to compile the example like `JSON.stringify(Data.serializer(), Data(42))`, I get the error:

```
e: src/main/kotlin/HelloWorld.kt: (260, 45): Unresolved reference: serializer
> Task :compileKotlin FAILED

FAILURE: Build failed with an exception.
```

sandwwraith:
I suppose that plugin was not applied because you need something like `id("kotlinx-serialization") version ("1.3.0")` instead of just `"kotlinx-serialization"` line which does nothing. Consult Kotlin Gradle DSL documentation for a precise syntax of applying plugin.

reitzig:
I have a working multi-platform example over at [reitzig/kovalscj](https://github.com/reitzig/kovalscj). I imagine that a subset of that Gradle file will work for single-platform projects?

sandwwraith:
@reitzig I think this is a correct assumption

[...]

PR #638, opened 2019-12-23
Update README with Gradle Kotlin DSL

lukaszkalnik:
This fills the gap with missing documentation of configuration for Gradle Kotlin DSL (as described in https://github.com/Kotlin/kotlinx.serialization/issues/273).
Also divides the "Gradle" chapter into 3 subchapters for easier navigation (Adding the plugin, Applying the plugin, Dependency on the runtime library).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ You can open example projects for [JS](examples/example-js), [JVM](examples/exam
 * [Current status](#current-project-status)
 * [Library installing](#setup)
     + [Gradle](#gradle)
-    + [Gradle (with `plugins` block)](#gradle-with-plugins-block)
     + [Android/JVM](#androidjvm)
     + [Multiplatform (common, JS, Native)](#multiplatform-common-js-native)
     + [Maven/JVM](#mavenjvm)
@@ -91,11 +90,35 @@ Example projects on JVM are available for [Gradle](examples/example-jvm/build.gr
 
 ### Gradle
 
-#### Adding the plugin to classpath
+#### Using the `plugins` block
 
-You have to add the serialization plugin to your classpath as the other [compiler plugins](https://kotlinlang.org/docs/reference/compiler-plugins.html):
+You can setup the serialization plugin with the Kotlin plugin using [Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block):
 
 Kotlin DSL:
+
+```kotlin
+plugins {
+    kotlin("multiplatform") // or kotlin("jvm") or any other kotlin plugin
+    id("kotlinx-serialization")
+}
+```
+Groovy DSL:
+
+```gradle
+plugins {
+    id 'org.jetbrains.kotlin.multiplatform' version '1.3.61' // or any other kotlin plugin
+    id 'org.jetbrains.kotlin.plugin.serialization' version '1.3.61'
+}
+```
+
+Note: plugin marker for serialization has been published in Kotlin 1.3.50. If you need to use the earlier Kotlin version, see [KT-27612](https://youtrack.jetbrains.com/issue/KT-27612) for workaround with [plugin resolution rules](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_resolution_rules).
+
+#### Using `apply plugin` (the old way)
+
+First, you have to add the serialization plugin to your classpath as the other [compiler plugins](https://kotlinlang.org/docs/reference/compiler-plugins.html):
+
+Kotlin DSL:
+
 ```kotlin
 buildscript {
     repositories { jcenter() }
@@ -107,7 +130,9 @@ buildscript {
     }
 }
 ```
+
 Groovy DSL:
+
 ```gradle
 buildscript {
     ext.kotlin_version = '1.3.61'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,41 +120,60 @@ buildscript {
 }
 ```
 
-Don't forget to apply the plugin:
+#### Applying the plugin
+
+Next, apply the plugin to your project.
 
+You can setup serialization plugin with the kotlin plugin using [Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) **instead** of traditional `apply plugin`:
```
sandwwraith: If one using `plugins {}` block, there's no need to add plugins as build script dependencies – plugins are resolved directly from Gradle Plugin Portal via plugin markers. Therefore I'd recommend moving this section before `Adding the plugin to classpath` and probably make it on a higher level (`###`) or join sections 'adding the plugin to the classpath' and 'applying the plugin in an old way' as it was before.

README.md
```diff
@@ -105,41 +120,60 @@ buildscript {
...
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    compile "org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.14.0" // JVM dependency
+    implementation(kotlin("stdlib-jdk8", KotlinCompilerVersion.VERSION))
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.14.0") // JVM dependency
 }
 ```
-
-### Gradle (with `plugins` block)
```
sandwwraith: Please update the table of contents accordingly

README.md
```diff
@@ -105,41 +120,60 @@ buildscript {
...
+```kotlin
 repositories {
     // artifacts are published to JCenter
     jcenter()
 }
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    compile "org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.14.0" // JVM dependency
+    implementation(kotlin("stdlib-jdk8", KotlinCompilerVersion.VERSION))
```
sandwwraith: You're using different Kotlin stdlibs in KotlinDSL and GroovyDSL examples

```suggestion
    implementation(kotlin("stdlib", KotlinCompilerVersion.VERSION)) // or "stdlib-jdk8"
```

README.md
```diff
@@ -105,41 +120,60 @@ buildscript {
...
+repositories {
+    jcenter()
 }
-```
 
-Note: plugin marker for serialization has been published in Kotlin 1.3.50. If you need to use the earlier Kotlin version, see [KT-27612](https://youtrack.jetbrains.com/issue/KT-27612) for workaround with [plugin resolution rules](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_resolution_rules).
-
-Runtime library should be added to dependencies the same way as before.
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
sandwwraith:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version" // or "kotlin-stdlib-jdk8"
```

README.md
lukaszkalnik: I have just tried it out and if I remove the
`classpath(kotlin("serialization", version = kotlinVersion))` from my buildscript dependencies then I get the following error:
```
Plugin [id: 'kotlinx-serialization'] was not found in any of the following sources:

- Gradle Core Plugins (not a core plugin, please see https://docs.gradle.org/6.1-milestone-2/userguide/standard_plugins.html for available core plugins)
- Plugin Repositories (plugin dependency must include a version number for this source)
```

lukaszkalnik:
Thank you for your remarks. Please let me know if it's ok now or I need to adjust something.

lukaszkalnik:
Also, I just realized that the plugin IDs were outdated. I have updated them with the correct ones from the official Gradle plugins search examples:
https://plugins.gradle.org/plugin/org.jetbrains.kotlin.plugin.serialization

README.md
```diff
@@ -91,11 +90,52 @@ Example projects on JVM are available for [Gradle](examples/example-jvm/build.gr
...
+#### Using the `plugins` block
+
+You can setup the serialization plugin with the Kotlin plugin using [Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block):
+
+Kotlin DSL:
+
+```kotlin
+plugins {
+    kotlin("multiplatform") // or kotlin("jvm") or any other kotlin plugin
+    id("org.jetbrains.kotlin.plugin.serialization") version "1.3.61"
```
sandwwraith: I haven't tried it myself, but does it work just as `kotlin("serialization")`? (judging by the following 'old way' example in Kotlin DSL)

README.md
```diff
@@ -105,41 +145,43 @@ buildscript {
 }
 ```
 
-Don't forget to apply the plugin:
+Then you can `apply plugin` (example in Groovy):
 
 ```gradle
 apply plugin: 'kotlin' // or 'kotlin-multiplatform' for multiplatform projects
-apply plugin: 'kotlinx-serialization'
+apply plugin: 'org.jetbrains.kotlin.plugin.serialization'
```
sandwwraith: Again, not tested assumption: `apply plugin: 'kotlinx-serialization'` should still work. Probably, it doesn't work for Kotlin DSL but works for Gradle DSL. Or works when dependency declared as  `classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"` but not as `classpath(kotlin("serialization", version = kotlinVersion))`... Gradle is somewhat complicated mechanism.

README.md
lukaszkalnik: Actually, looking at the source code of the `kotlin()` extension function I could reduce it:
```
fun PluginDependenciesSpec.kotlin(module: String): PluginDependencySpec =
    id("org.jetbrains.kotlin.$module")
```
Good hint, thanks 👍

README.md
lukaszkalnik: There is no difference between the two `classpath` syntaxes you mentioned - the `kotlin()` extension function is here again a convenience function which just appends the proper group, module prefix and version number:
```
fun DependencyHandler.kotlin(module: String, version: String? = null): Any =
    "org.jetbrains.kotlin:kotlin-$module${version?.let { ":$version" } ?: ""}"
```

lukaszkalnik: Regarding the first part of your question, the old plugin id
`apply plugin: 'kotlinx-serialization'`
indeed still does work. I will restore the example.

sandwwraith:
Thank you for your help!

PR #642, opened 2019-12-26
Extract ProtoBuf And Cbor formats to separate modules

qwwdfsad:
* Including SerialId annotation that is clearly ProtoBuf-specific
 * Copy-paste some of the test (but not main!) sources to make it work. These classes will be extracted to mediator module to workaround KT-35073 when CBOR is extracted as well
 * Replace @SerialId in core tests with @Id from test sources
 * Postpone readme and publication for a while
 * Make some types public for the sake of simplicity, will be revisited them during cleanup phase


qwwdfsad:
It is mostly ready, only module-level documentation is missing

whyoleg:
@qwwdfsad Will JSON Will be separated in format in this or anoother PR?

qwwdfsad:
@whyoleg nope, JSON is likely to stay in the core module (no commitment though)

sandwwraith:
Preliminary comments:

If we consider `@SerialId` protobuf-specific, let's reflect that in its name, e.g. `@ProtoId`.
Have you checked that other 3rd party formats (yaml, avro) do not use this annotation? Do we have migration aids?
_NB_: `@SerialName` is considered format-agnostic, but `@SerialId` is format-specific. We need to develop clear criteria for that.

It's better not to create any additional modules nor copy-paste code. Can we create a separate folder and add it to both `commonTest` source sets?

There's a lot of copy-paste in build scripts. Let's somehow unify them.

Why one declarations are deprecated and others marked as `@InternalSerializationApi`? Do you have the criteria for that?

qwwdfsad:
>If we consider @SerialId protobuf-specific, let's reflect that in its name, e.g. @ProtoId.

Good idea

>Have you checked that other 3rd party formats (yaml, avro) do not use this annotation?

Yes, I did. They don't.

>It's better not to create any additional modules nor copy-paste code. Can we create a separate folder and add it to both commonTest source sets?

Taking into account different level of quality and plans to properly maintain/support/document these modules, I'd rather avoid adding a premature abstraction

>Why one declarations are deprecated and others marked as @InternalSerializationApi? Do you have the criteria for that?

Yes. I deprecate everything I want to remove and that shouldn't be in the library in the first place and mark effectively internal declarations with `InternalSerializationApi`

README.md
```diff
@@ -1,32 +1,21 @@
-# Kotlin cross-platform / multi-format reflectionless serialization
+# Kotlin multiplatform / multi-format serialization reflectionless serialization
```
sandwwraith: Word 'serialization' is doubled

README.md
```diff
@@ -45,10 +34,9 @@ You can open example projects for [JS](examples/example-js), [JVM](examples/exam
 * [Instructions for old versions under Kotlin 1.2 and migration guide](docs/old12.md)
 
 
-## Quick example
+## Quick start
```
sandwwraith: Are you sure that 'start' is the correct word? This section does not mention any ways to include/download the library, and without it, it's rather hard to start exploring library...

README.md
```diff
@@ -132,7 +131,7 @@ You can setup serialization plugin with the kotlin plugin using [Gradle plugins
 
 ```gradle
 plugins {
-    id 'org.jetbrains.kotlin.multiplatform' version '1.3.60' // or any other kotlin plugin
+    id 'org.jetbrains.kotlin.multiplatform' version '1.3.60' // or kotlin plugin
```
sandwwraith: we have `kotlin.js` plugin, don't we? so 'any other' is still ok

README.md
```diff
@@ -238,7 +237,7 @@ Include kotlinx bintray repository for library:
 </repositories>
 ```
 
-You also can use JCenter.
+You can also use JCenter.
```
sandwwraith: Probably it's time to move JCenter to the first place, since even eap releases automatically synced.
Also, we support Maven Central now.

build.gradle
```diff
@@ -71,8 +71,6 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"
         classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:$bintray_version"
-
-        // Protobuf is udes in JVM tests
```
sandwwraith: Why remove the comment?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -294,36 +303,54 @@ class Cbor(val updateMode: UpdateMode = UpdateMode.BANNED, val encodeDefaults: B
...
+            val res = Double.fromBits(readLong())
             readByte()
             return res
         }
 
+        private fun readLong(): Long {
+            var result = 0L
+            for (i in 0..7) {
+                val byte = input.read()
+                result = (result shl 8) or byte.toLong()
```
sandwwraith: Won't `byte.toLong` make the sign-extending conversion? It should be `byte.toLong() and 0xFFL`, probably.

formats/cbor/commonTest/src/kotlinx/serialization/cbor/SampleClasses.kt
```diff
@@ -0,0 +1,31 @@
...
+
+package kotlinx.serialization.cbor
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class Simple(val a: String)
+
+@Serializable
+data class SmallZoo(
```
sandwwraith: Since you don't like word `Zoo`, you can rename these classes

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoId.kt
```diff
@@ -0,0 +1,18 @@
...
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.*
+
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY)
+@Deprecated(message = "SerialId is renamed to ProtoId to better reflect its semantics",
+    level = DeprecationLevel.WARNING, replaceWith = ReplaceWith("ProtoId(id)")
+)
+annotation class SerialId(val id: Int)
```
sandwwraith: `SerialId` was in another package

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoId.kt
```diff
@@ -0,0 +1,18 @@
...
+
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY)
+annotation class ProtoId(val id: Int)
```
sandwwraith: You also can add documentation here

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufPrimitivesTest.kt
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2017 JetBrains s.r.o.
```
sandwwraith: Old copyright

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufPrimitivesTest.kt
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.*
+import kotlin.test.*
+
+// TODO these tests are far from being comoplete
```
sandwwraith: I don't understand what 'complete' means here. Better put a clear action point after `TODO`, e.g. "add more tests for floats, doubles and other primitives..."

gradle/configureSourceSets.gradle
```diff
@@ -0,0 +1,92 @@
...
+            useExperimentalAnnotation("kotlin.ExperimentalStdlibApi")
+            useExperimentalAnnotation("kotlinx.serialization.InternalSerializationApi")
+        }
+    }
+
+    sourceSets {
+        commonMain {
+            dependencies {
+                api 'org.jetbrains.kotlin:kotlin-stdlib-common'
+//                api project(":kotlinx-serialization-runtime")
```
sandwwraith: Better remove this line at all or comment about it (smth like 'runtime can't depend on itself' or whatever)

runtime/commonMain/src/kotlinx/io/Streams.kt
```diff
@@ -16,11 +16,15 @@
...
 
+import kotlinx.serialization.*
+
+@Deprecated(message = message, level = DeprecationLevel.ERROR)
 expect open class IOException: Exception {
     constructor()
     constructor(message: String)
 }
 
+@InternalSerializationApi
```
sandwwraith: It should also be deprecated then, isn't it? We plan to remove all this stuff in favor of real `kotlinx-io`

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
```diff
@@ -29,6 +29,7 @@ interface BinaryFormat: SerialFormat {
     fun <T> load(deserializer: DeserializationStrategy<T>, bytes: ByteArray): T
 }
 
+// TODO migrate
```
sandwwraith: migrate to where?

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
```diff
@@ -29,6 +29,7 @@ interface BinaryFormat: SerialFormat {
...
 fun <T> BinaryFormat.dumps(serializer: SerializationStrategy<T>, obj: T): String =
     HexConverter.printHexBinary(dump(serializer, obj), lowerCase = true)
```
sandwwraith: btw, we should eventually make a decision on this `HexConverter` class, because it is public now. And we don't plan to provide any analogs in `kotlinx-io`, do we?

runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
```diff
@@ -41,16 +40,16 @@ class PolymorphismTest {
...
-        val obj = Wrapper(
-            PolyBase(2),
-            PolyDerived("b")
-        )
-        val bytes = protobuf.dumps(Wrapper.serializer(), obj)
-        val restored = protobuf.loads(Wrapper.serializer(), bytes)
-        assertEquals(obj, restored)
-    }
+//    @Test
+//    fun testInheritanceProtobuf() {
```
sandwwraith: Why comment this?

runtime/commonTest/src/kotlinx/io/ByteBufferTest.kt
```diff
@@ -1,78 +0,0 @@
...
-package kotlinx.io
-
-import kotlinx.serialization.internal.HexConverter
-import kotlinx.serialization.test.shouldBe
-import kotlin.test.Test
-import kotlin.test.assertEquals
-
-class ByteBufferTest {
-    @Test
-    fun longsAndDoubles() {
```
sandwwraith: I'd suggest deleting tests with deleting a class, not only because its deprecated

README.md
```diff
@@ -1,32 +1,21 @@
...
 
 [![JetBrains incubator project](https://jb.gg/badges/incubator.svg)](https://confluence.jetbrains.com/display/ALL/JetBrains+on+GitHub)
 [![GitHub license](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](http://www.apache.org/licenses/LICENSE-2.0)
 [![TeamCity build](https://img.shields.io/teamcity/http/teamcity.jetbrains.com/s/KotlinTools_KotlinxSerialization_Ko.svg)](https://teamcity.jetbrains.com/viewType.html?buildTypeId=KotlinTools_KotlinxSerialization_Ko&guest=1)
 [![Download](https://api.bintray.com/packages/kotlin/kotlinx/kotlinx.serialization.runtime/images/download.svg) ](https://bintray.com/kotlin/kotlinx/kotlinx.serialization.runtime/_latestVersion)
 
-Kotlin serialization consists of a compiler plugin, which automatically produces visitor code for classes, and runtime library, which uses generated code to serialize objects without reflection.
+Kotlin serialization consists of a compiler plugin, that generates visitor code for serializable classes,
+ runtime library with core serialization API and JSON format, and support libraries with ProtoBuf, CBOR and properties formats.
```
sandwwraith: WDYT of adding the corresponding links to `formats/README.md#protobuf` etc?

sandwwraith:
Have you tested publication manually? e.g. upload library to your personal bintray and see if integration test passes with commented `mavenLocal()`

build.gradle
qwwdfsad: Because it's obvious from the presence of `protobuf` module

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoId.kt
qwwdfsad: Good catch!

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoId.kt
qwwdfsad: No, it's not my goal. Especially taking into account the fact that I am not really familiar with out protobuf implementation

README.md
qwwdfsad: Not now, because it's not really that useful

runtime/commonTest/src/kotlinx/io/ByteBufferTest.kt
qwwdfsad: Usually, it's a good policy, but here it's mostly about aliases that are going to be removed in the next release. And we are not afraid of breaking anything here (as it is a standalone code)

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
qwwdfsad: to a proper name. Removed this TODO for a while

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
qwwdfsad: We do. `HexConverter` should be deprecated and eventually hidden from `kotlinx.serialization`

runtime/commonMain/src/kotlinx/io/Streams.kt
qwwdfsad: I will do it in a follow-up PR. It requires a lot of rewriting in CBOR and ProtoBuf

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
qwwdfsad: No, because `InputStream.read` returns `Int` type with a values within -1..255 range

qwwdfsad:
Tested publication with `intergation-tests` using https://bintray.com/qwwdfsad/kotlinx/kotlinx.serialization/0.14.1-q

runtime/commonMain/src/kotlinx/io/Streams.kt
sandwwraith: We can still use deprecated classes, so no need to rewriting

README.md
sandwwraith: Why not useful?

formats/README.md
```diff
@@ -16,6 +16,20 @@ Allows deserialization of `Config` object from popular [lightbend/config](https:
 into Kotlin objects.
 You can learn about "Human-Optimized Config Object Notation" or HOCON from library's [readme](https://github.com/lightbend/config#using-hocon-the-json-superset).
 
+## ProtoBuf
+
+* Artifact id: `kotlinx-serialization-protobuf`
```
sandwwraith: btw, it's not clear that these artifacts have the same naming conventions (e.g. `kotlinx-serialization-protobuf-js` for js, `-native` for native etc). Better to highlight this somewhere (maybe in `Platform:` section?)

formats/protobuf/commonMain/src/kotlinx/serialization/SerialId.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package kotlinx.serialization
+
+
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY)
+@Deprecated(message = "SerialId is renamed to ProtoId to better reflect its semantics",
```
sandwwraith: Can we communicate somehow that protobuf was extracted to separate module?

runtime/commonMain/src/kotlinx/io/Streams.kt
qwwdfsad: That's why it's marked as `@InternalSerializationApi` in the first place.

README.md
qwwdfsad: Because it contains no really useful information.

formats/README.md
qwwdfsad: I am not sure it is worth documenting MPP usages in every library.

README.md
sandwwraith: It contains artifact-id for protobuf and cbor modules (and basically, says that they **are** separate artifacts, instead of vague 'support libraries')

formats/README.md
sandwwraith: It's about discoverability

formats/protobuf/commonMain/src/kotlinx/serialization/SerialId.kt
qwwdfsad: We can add `SerialId` back to the `runtime` module with a message that mentions that, but then its deprecation level should be increased to `ERROR`.
Adding anything else in `runtime` module is a no-go, because it will seriously complicate UX when protobuf module is already included

formats/README.md
qwwdfsad: Let's do it properly in a separate PR then: add `README` to every platform-specific source set with a way to add the dependency, and mention that in the root readme with a proper references

formats/protobuf/commonMain/src/kotlinx/serialization/SerialId.kt
sandwwraith: Yeah, I think it's better with message like "SerialId is deprecated in favor of ProtoId from kotlinx-serialization-protobuf module"

formats/README.md
sandwwraith: Yeah, okay. Maybe we'll come up with a better documentation structure (or eventually publish metadata for common artifact)

PR #651, opened 2020-01-09
Allows to set `ProtoBuf.shouldEncodeElementDefault` to false

### LINKED ISSUE #71: NULLs are not supported when writing to protobuf

### LINKED ISSUE #397: Make skipping null values default behavior for protobuf
Based on #58.

My understanding is that since `0.10.0` there is support for skipping optional values, which can be enabled for JSON with `Json(encodeDefaults = false)`, but it's not clear how to enable this behavior for `ProtoBuf`. In my opinion this should also be the default behavior for protobuf, since the only way to transfer null values is to omit them in the binary representation.

brudaswen:
Porting a project using the *Java ProtoBuf library* to *kotlinx.serialization* can become quite hard since optional properties are not supported in the same way.

On the one hand it is not possible to omit optional properties during serialization (this is possible by not setting an optional property while building the message in Java).
On the other hand it is not possible to check if a property has been omitted by the sender during de-serialization (this is possible via `Message.hasA()` in Java).

Allowing to set `ProtoBuf.shouldEncodeElementDefault()` to `false` allows to create `@Serializable data classes` that support optional properties during serialization and de-serialization. Additionally, it makes `null` values possible for optional properties as well.

Closes #397 Make skipping null values default behavior for protobuf
Closes #71 NULLs are not supported when writing to protobuf

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
@@ -15,11 +15,95 @@ import kotlinx.serialization.protobuf.ProtoBuf.Varint.decodeSignedVarintLong
 import kotlinx.serialization.protobuf.ProtoBuf.Varint.decodeVarint
 import kotlinx.serialization.protobuf.ProtoBuf.Varint.encodeVarint
 
-class ProtoBuf(context: SerialModule = EmptyModule) : AbstractSerialFormat(context), BinaryFormat {
+/**
+ * The main entry point to work with ProtoBuf serialization.
+ * It is typically used by constructing an application-specific instance, with configured ProtoBuf-specific behaviour
+ * ([encodeDefaults] constructor parameter) and, if necessary, registered
+ * custom serializers (in [SerialModule] provided by [context] constructor parameter).
+ *
+ * ## Usage Example
+ * Given a ProtoBuf definition with one required field, one optional field and one optional field with a custom default
+ * value:
+ * ```
+ * message MyMessage {
+ *     required int32 first = 1;
+ *     optional int32 second = 2;
+ *     optional int32 third = 3 [default = 42];
+ * }
+ * ```
+ *
+ * The corresponding [Serializable] class should match the ProtoBuf definition and should use the same default values:
+ * ```
+ * @Serializable
+ * data class MyMessage(val first: Int, val second: Int = 0, val third: Int = 42)
+ *
+ * // Serialize to ProtoBuf hex string
+ * val encoded = ProtoBuf.dumps(MyMessage.serializer(), MyMessage(15)) // "080f1000182a"
+ *
+ * // Deserialize from ProtoBuf hex string
+ * val decoded = ProtoBuf.loads<MyMessage>(MyMessage.serializer(), encoded) // MyMessage(first=15, second=0, third=42)
+ *
+ * // Serialize to ProtoBuf bytes (omitting default values)
+ * val encoded2 = ProtoBuf(encodeDefaults = false).dump(MyMessage.serializer(), MyMessage(15)) // [0x08, 0x0f]
+ *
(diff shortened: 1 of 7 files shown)


brudaswen:
> Can I ask you to add the documentation for the new flag?

Sure, I will add some documentation to describe the requirements when generating a `@Serializable` class.

## Some more questions came up in the meanwhile: 

### Order of configuration flags
For all other AbstractSerialFormats, the `context: SerialModule` is the last argument and all configuration parameters come first. I added the new parameter for ProtoBuf to the end to avoid a breaking API change.

Would you prefer a consistent API that comes with a breaking API change?
```
class ProtoBuf(
        private val encodeDefaults: Boolean = true,
        context: SerialModule = EmptyModule
) : AbstractSerialFormat(context), BinaryFormat {
```
Or a non-breaking API change that introduces some inconsistencies?
```
class ProtoBuf(
        context: SerialModule = EmptyModule,
        private val encodeDefaults: Boolean = true
) : AbstractSerialFormat(context), BinaryFormat {
```

Since the project is still in incubation state, I would prefer a consistent API.

### Visibility of configuration flags
For Cbor the `encodeDefaults` configuration flag is public. Should it be public for ProtoBuf as well? 
For me public seems to be fine here.

sandwwraith:
Yes, since the project is marked as experimental, it is possible to introduce some breaking changes to the API, so I also would prefer consistency.

`public` visibility of the flag seems fine for me

brudaswen:
I made the changes and added documentation to the ProtoBuf class. Should I squash those commits?

runtime/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -15,12 +15,96 @@ import kotlinx.serialization.protobuf.ProtoBuf.Varint.decodeSignedVarintLong
...
+ * }
+ * ```
+ *
+ * The corresponding [Serializable] class should match the ProtoBuf definition and should use the same default values:
+ * ```
+ * @Serializable
+ * data class MyMessage(val first: Int, val second: Int = 0, val third: Int = 42)
+ *
+ * // Serialize to ProtoBuf hex string
+ * val encoded = ProtoBuf.dumps(MyMessage(15)) // "080f1000182a"
```
sandwwraith: Please don't use `@ImplicitReflectionSerializer` methods in the examples, i.e. use `ProtoBuf.dumps(MyMessage.serializer(), MyMessage(15))` / `ProtoBuf.load(MyMessage.serializer(), encoded)` etc

brudaswen:
I updated the documentation and squashed all commits.

sandwwraith:
Thank you, this was very helpful!

Issue #298, opened 2018-12-07
Move default serializers for numbers, strings, lists, etc from 'internal'


sandwwraith:
Name `internal` can be confusing if being understood as "Do not touch".

Some really internal classes like `SerialClassDescImpl` should remain in `internal`.

sandwwraith:
Recent decision: do not introduce 'builtins' package, leave public-facing serializers as-is in `kotlinx.serialization.*`

sandwwraith:
Things that have to be moved, reviewed and documented:

- [x] Primitive serializers
- [ ] Primitive array serializers
- [ ] Collection serializers
- [ ] Misc serializers: Pair, Triple, LongAsString

PR #665, opened 2020-01-20
Abi cleanup

sandwwraith:
* Refactor EnumSerializer 
* Get rid of obsolete constructor arguments (removed @JvmOverloads)
* Optimize space allocation for SerialClassDescImpl by passing actual serializable properties count in the generated code
* Supply correct child descriptors for Pair and Triple
* Fix annoying cmdline compiler warnings (mainly 'The corresponding parameter in the supertype 'XXX' is named 'descriptor'. This may cause problems when calling this function with named arguments.'). Format code somewhere.
* Fix some test function bodies which got commented or erased during merges.

Works on my machine. Need to merge changes in plugin & publish dev version


runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -1,22 +1,16 @@
...
 
 package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
-import kotlin.jvm.*
-import kotlin.reflect.*
 
 @InternalSerializationApi
-public class EnumDescriptor @JvmOverloads constructor(
+public class EnumDescriptor constructor(
```
qwwdfsad: `constructor` is redundant

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -1,22 +1,16 @@
...
     name: String,
-    values: Array<String> = emptyArray()
-) : SerialClassDescImpl(name) {
-
-    init {
-        values.forEach { addElement(it) }
-    }
+    elementsCount: Int = 1
```
qwwdfsad: do we really need default value here?

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
```diff
@@ -56,38 +57,44 @@ public sealed class KeyValueSerializer<K, V, R>(
     }
 }
 
+// todo: move from internal package and add documentation
```
qwwdfsad: Could you please create a corresponding issue for each TODO?
It would be much easier to find a quick/small pieces of work, chances that we forget something are lower and reviews are quicker

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: It's not very convenient when it's used in tests like this: https://github.com/kotlin/kotlinx.serialization/blob/d0fb63d5002c85632cfb524aabbb0a8883242c32/runtime/commonTest/src/kotlinx/serialization/features/BinaryPayloadExampleTest.kt#L15 (and user code)

Once this constructor won't be user-facing, there will be no need in default value

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
sandwwraith: I've created a task list here: https://github.com/Kotlin/kotlinx.serialization/issues/298#issuecomment-577272487

sandwwraith:
I'll put corresponding changes into plugin and merge this PR when build would be ready

PR #668, opened 2020-01-22
Polymorphism name collision

qwwdfsad:


runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -75,26 +74,45 @@ public class SealedClassSerializer<T : Any>(
...
                     " got arrays: ${subclasses.contentToString()}, ${subclassSerializers.contentToString()}"
         }
-        backingMap = subclasses.zip(subclassSerializers).toMap()
-        inverseMap = backingMap.values.associateBy { serializer -> serializer.descriptor.serialName }
+        class2Serializer = subclasses.zip(subclassSerializers).toMap()
+        serialName2Serializer = class2Serializer.entries.groupingBy { it.value.descriptor.serialName }
+            .aggregate<Map.Entry<KClass<out T>, KSerializer<out T>>, String, Map.Entry<KClass<*>, KSerializer<out T>>>
+            { key, accumulator, element, _ ->
+                if (accumulator != null) {
+                    error("Multiple sealed subclasses of $baseClass have the same serial name '$key: ${accumulator.key}, ${element.key}")
```
pdvrieze: Just nitpicky, but there is a missing end single quote for the name.

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -75,26 +74,45 @@ public class SealedClassSerializer<T : Any>(
...
+        return class2Serializer[value::class] ?: super.findPolymorphicSerializer(encoder, value)
+    }
+
+    internal fun validate(actualSerializer: KSerializer<Any>, classDiscriminator: String) {
+        val properties = serializer2AllProperties[actualSerializer] ?: return
+        if (classDiscriminator !in properties) return
+        val clazz = class2Serializer.entries.first { it.value === actualSerializer }.key
+        error(
+            "Sealed $clazz cannot be serialized as base class $baseClass because it has property name that conflicts with JSON class discriminator. " +
+                    "You can either change class discriminator in JsonConfiguration, " +
```
pdvrieze: Don't refer to JSON as it is/should be format agnostic. If this is a JSON specific issue then it should be somehow handled in the format, not in the agnostic code.

qwwdfsad: This exact method is JSON-specific (and internal anyway).
In order to make it format-agnostic, we should expose backing maps (with properties and KClasses) to public and check them from within in JSON.
I don't mind doing it, but only when we have a compelling use-case to make it public.

qwwdfsad:
Note to self: merge only after changing base of the #670

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -141,11 +140,45 @@ public class SerializersModuleBuilder internal constructor(@JvmField internal va
...
+            baseClass,
+            concreteClass
+        )
+        baseClassSerializers[concreteClass] = concreteSerializer
+
+        val names = polyBase2NamedSerializers.getOrPut(baseClass, ::hashMapOf)
+        val previous = names.put(name, concreteSerializer)
+        if (previous != null) {
+            val conflictingClass = polyBase2Serializers[baseClass]!!.filter { it.value === previous }.keys.firstOrNull()
+            throw IllegalArgumentException("Multiple polymorphic serializers for base class '$baseClass' " +
```
sandwwraith: I'd suggest moving this check and exception throw a bit lines up – before actual writing to a map. By that, the module builder will be in consistent state even if someone catches this exception

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleImpl.kt
```diff
@@ -1,79 +1,49 @@
...
     override fun <T : Any> getPolymorphic(baseClass: KClass<T>, serializedClassName: String): KSerializer<out T>? {
-        (if (baseClass == Any::class) StandardSubtypesOfAny.getDefaultDeserializer(
-            serializedClassName
-        ) else null)?.let { return it as KSerializer<out T> }
-        return inverseClassNameMap[baseClass]?.get(serializedClassName) as? KSerializer<out T>
+        val standardPolymorphic =
+            if (baseClass == Any::class) StandardSubtypesOfAny.getDefaultDeserializer(serializedClassName)
+            else null
+
+        if (standardPolymorphic != null) return standardPolymorphic as KSerializer<out T>
```
sandwwraith: `standardPolymorphic` still will be selected instead of custom registered one

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
```diff
@@ -13,6 +14,7 @@ import kotlin.reflect.KClass
  * Registering in the same module twice is prohibited. In case you are combining modules
  * with [SerialModule.plus], consider using [SerialModule.overwriteWith] if you want overwriting behaviour.
  */
+@InternalSerializationApi // Will be hidden in the next release
```
sandwwraith: Why so?

runtime/commonTest/src/kotlinx/serialization/modules/SerialNameCollisionInSealedClassesTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        Json("type", true).stringify(Base.serializer(), Base.Child("a"))
+    }
+
+    @Serializable
+    sealed class BaseCollision {
+        @Serializable
+        class Child() : BaseCollision()
+
+        @Serializable
+        @SerialName("$prefix.BaseCollision.Child")
```
sandwwraith: You can specify `@SerialName` on both to avoid handling prefix

runtime/commonTest/src/kotlinx/serialization/modules/SerialNameCollisionTest.kt
```diff
@@ -0,0 +1,127 @@
...
+    @Test
+    fun testNoCollisionWithArrayPolymorphism() {
+        val module = SerializersModule {
+            polymorphic(Base::class) {
+                Derived::class with Derived.serializer()
+            }
+        }
+
+        Json("type", module, true)
+        Json("type", module, true)
```
sandwwraith: same line 3 times. typo?

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleImpl.kt
sandwwraith: Sry, messed up with another PR

runtime/commonMain/src/kotlinx/serialization/json/internal/ContextValidator.kt
```diff
@@ -0,0 +1,34 @@
...
+        actualClass: KClass<Sub>,
+        actualSerializer: KSerializer<Sub>
+    ) {
+        val descriptor = actualSerializer.descriptor
+        for (i in 0 until descriptor.elementsCount) {
+            val name = descriptor.getElementName(i)
+            if (name == discriminator) {
+                throw IllegalArgumentException(
+                    "Polymorphic serializer for $actualClass has property '$name' that conflicts " +
+                            "with JSON class discriminator. You can either change class discriminator in JsonConfiguration, " +
```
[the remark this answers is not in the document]
qwwdfsad: It's literally the same as `$name`

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -41,23 +39,24 @@ public fun serializersModuleOf(map: Map<KClass<*>, KSerializer<*>>): SerialModul
...
 /**
  * A builder class for [SerializersModule] DSL.
  */
-public class SerializersModuleBuilder internal constructor(@JvmField internal val impl: SerialModuleImpl) :
-    SerialModuleCollector {
-
+public class SerializersModuleBuilder internal constructor() : SerialModuleCollector {
+    private val class2Serializer: MutableMap<KClass<*>, KSerializer<*>> = hashMapOf()
+    private val polyBase2Serializers: MutableMap<KClass<*>, MutableMap<KClass<*>, KSerializer<*>>> = hashMapOf()
+    private val polyBase2NamedSerializers: MutableMap<KClass<*>, MutableMap<String, KSerializer<*>>> = hashMapOf()
```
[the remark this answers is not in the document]
qwwdfsad: Well, it is a map of maps, and calling `name -> serializer` as `named serializers` doesn't look like a problem to me

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
qwwdfsad: Because it's one more internal exception without proper use. `IllegalArgumentException` is enough to throw

runtime/commonTest/src/kotlinx/serialization/modules/SerialNameCollisionInSealedClassesTest.kt
qwwdfsad: I wanted one of the classes to have its default name intentionally

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
sandwwraith: Then you can deprecate it and replace all usages. I doubt anyone actually tried to catch that

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
```diff
@@ -31,6 +31,7 @@ public open class SerialClassDescImpl(
     // this array is only used when serializer is written by hand
     private var descriptors: MutableList<SerialDescriptor>? = null
     private var flags = BooleanArray(elementsCount)
+    internal val namesSet: Set<String> by lazy { names.toHashSet() }
```
qwwdfsad: I start having a feeling that we'll need something like `SerialLocal` at some point :)

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
qwwdfsad: Let it be for a release, according to the issues, at least one user tried to catch it :)

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
sandwwraith: We could leverage `DescriptorSchemaCache` from https://github.com/Kotlin/kotlinx.serialization/pull/597

qwwdfsad: I'd rather avoid it for such a small task

runtime/commonMain/src/kotlinx/serialization/Shorthands.kt
```diff
@@ -26,12 +26,15 @@ public fun SerialDescriptor.elementDescriptors(): List<SerialDescriptor> {
...
 
 /**
- * Creates a [List] out of a serial names of [this] descriptor's elements [SerialDescriptor.getElementDescriptor].
- *
- * Size of a list is equal to [SerialDescriptor.elementsCount].
+ * Returns a [Set] out of all serial names of serial descriptor [elements][SerialDescriptor.getElementDescriptor].
  */
-public fun SerialDescriptor.elementNames(): List<String> {
-    return List(elementsCount) { getElementName(it) }
+public fun SerialDescriptor.elementNames(): Set<String> {
```
sandwwraith: I may be wrong, but for some usages, element order might be crucial – that's why there was a `List`

runtime/commonMain/src/kotlinx/serialization/json/internal/ContextValidator.kt
```diff
@@ -0,0 +1,34 @@
...
+        // Nothing here
+    }
+
+    override fun <Base : Any, Sub : Base> polymorphic(
+        baseClass: KClass<Base>,
+        actualClass: KClass<Sub>,
+        actualSerializer: KSerializer<Sub>
+    ) {
+        val descriptor = actualSerializer.descriptor
+        for (i in 0 until descriptor.elementsCount) {
```
sandwwraith: Can be simplified to `discriminator in descriptor.elementNames()` then?

runtime/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -16,12 +16,31 @@ internal inline fun <T> JsonOutput.encodePolymorphically(serializer: Serializati
...
+    serializer: KSerializer<*>,
+    actualSerializer: KSerializer<Any>,
+    classDiscriminator: String
+) {
+    if (serializer !is SealedClassSerializer<*>) return
+    if (classDiscriminator in actualSerializer.descriptor.elementNames()) {
+        val baseName = serializer.descriptor.serialName
+        val actualName = actualSerializer.descriptor.serialName
+        error(
+            "Sealed class '$actualName' cannot be serialized as base class $baseName because" +
```
sandwwraith: `'$actualName'` has quotes, `$baseName` does not. pick one.

runtime/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -16,12 +16,31 @@ internal inline fun <T> JsonOutput.encodePolymorphically(serializer: Serializati
...
+                    " it has property name that conflicts with JSON class discriminator. " +
```
sandwwraith: now you can mention the actual name/discriminator value

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -141,11 +140,48 @@ public class SerializersModuleBuilder internal constructor(@JvmField internal va
...
+        val name = concreteSerializer.descriptor.serialName
+        val baseClassSerializers = polyBase2Serializers.getOrPut(baseClass, ::hashMapOf)
+        if (!allowOverwrite && concreteClass in baseClassSerializers) throw SerializerAlreadyRegisteredException(
+            baseClass,
+            concreteClass
+        )
+        // Check for serial name conflicts
+        val names = polyBase2NamedSerializers.getOrPut(baseClass, ::hashMapOf)
+        val previous = names[name]
+        if (previous != null) {
```
sandwwraith: Hmmm, do we have a test that this check works correctly with `allowOverwrite=true` when we try to overwrite the previous serializer by a new one with the same name? Or should we prohibit such operation? 

(btw, aren't we supposed to delete old serial name from the map on overwrite action?)

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
```diff
@@ -101,12 +102,14 @@ public open class SerialClassDescImpl(
         if (this === other) return true
         if (other !is SerialClassDescImpl) return false
         if (serialName != other.serialName) return false
+        // TODO get rid of it during SerialClassDescImpl rework
```
sandwwraith: are you talking about `elementDescriptors` function?

qwwdfsad: I am talking about comparing each individual element because we already decided that `serialName` is **unique** identifier of SD

runtime/commonMain/src/kotlinx/serialization/json/internal/ContextValidator.kt
qwwdfsad: It will create an undesired footprint. 
I will create `cachedElementNames` separately and left `elementNames` as is, so users won't be able to increase footprint for no real reason

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
qwwdfsad: Reworked this, good point, one more old bug caught :)

runtime/commonMain/src/kotlinx/serialization/internal/SerialClassDescImpl.kt
sandwwraith: you can mention that then

runtime/commonMain/src/kotlinx/serialization/Shorthands.kt
```diff
@@ -26,9 +26,8 @@ public fun SerialDescriptor.elementDescriptors(): List<SerialDescriptor> {
 }
 
 /**
- * Creates a [List] out of a serial names of [this] descriptor's elements [SerialDescriptor.getElementDescriptor].
- *
- * Size of a list is equal to [SerialDescriptor.elementsCount].
+ * Returns a [List] out of all serial names of serial descriptor [elements][SerialDescriptor.getElementDescriptor].
+ * TODO always allocates, also revisit
```
sandwwraith: I'm not sure `TODO` is a good line for documentation, maybe put it as a separate regular `//` comment?)

sandwwraith: btw, it can be improved also with `if (this is SerialClassDescImpl) return names` (assuming `names` visibility is raised to internal and we do not care if someone tries to cast the result to an `ArrayList`)

PR #670, opened 2020-01-23
Allow duplicating serializers during SerialModule concatenation if th…

### LINKED ISSUE #616: Throwing of exception on creating of Json from decoder context
**Description**

If you are using custom serializer and trying to create new `Json` format (via `Json` with `context` constructor) and parent decoder is from `Json` formatter - it will throw exception.

**To reproduce**

```kotlin
object : KSerializer<T> {
    ...
    override fun deserialize(decoder: Decoder): T {
        val jsonFormat = Json(context = decoder.context)
    }
    ...
}
```

**Expected behavior**

It is expected, that will be created new format with context of decoder.

**Environment**
 - Kotlin version: 1.3.61
 - Library version: 0.14.0
 - Kotlin platforms: JVM
 - Gradle version: 6.0.1

qwwdfsad:
…ey are equal.

Additionally, fix a bug when builtin serializer was selected for known subtypes of Any even when custom one was registered

Fixes #616

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
@@ -7,6 +7,7 @@
 package kotlinx.serialization.modules
 
 import kotlinx.serialization.*
+import kotlin.jvm.*
 import kotlin.reflect.*
 
 /**
@@ -152,15 +153,28 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
         }
     }
 
+    @JvmName("registerSerializer") // Don't mangle method name for prettier stack traces
     internal fun <T : Any> registerSerializer(
         forClass: KClass<T>,
         serializer: KSerializer<T>,
         allowOverwrite: Boolean = false
     ) {
-        if (!allowOverwrite && forClass in class2Serializer) throw SerializerAlreadyRegisteredException(forClass)
+        if (!allowOverwrite) {
+            val previous = class2Serializer[forClass]
+            if (previous != null && previous != serializer) {
+                // TODO when working on SD rework, provide a way to properly stringify serializer as its FQN
+                val currentName = serializer.descriptor.serialName
+                val previousName = previous.descriptor.serialName
+                throw SerializerAlreadyRegisteredException(
+                    "Serializer for $forClass already registered in this module: $previous ($previousName), " +
+                            "attempted to register $serializer ($currentName)"
+                )
+            }
+        }
         class2Serializer[forClass] = serializer
     }
 
+    @JvmName("registerPolymorphicSerializer") // Don't mangle method name for prettier stack traces
     internal fun <Base : Any, Sub : Base> registerPolymorphicSerializer(
         baseClass: KClass<Base>,
         concreteClass: KClass<Sub>,
@@ -184,7 +198,12 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
         }
         // Overwrite prohibited
         if (previousSerializer != null) {
-            throw SerializerAlreadyRegisteredException(baseClass, concreteClass)
+            if (previousSerializer != concreteSerializer) {
+                throw SerializerAlreadyRegisteredException(baseClass, concreteClass)
+            } else {
+                // Cleanup name mapping
+                names.remove(previousSerializer.descriptor.serialName)
+            }
         }
         val previousByName = names[name]
         if (previousByName != null) {
--- a/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleImpl.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleImpl.kt
@@ -20,8 +20,13 @@ internal class SerialModuleImpl(
 
     override fun <T : Any> getPolymorphic(baseClass: KClass<T>, value: T): KSerializer<out T>? {
         if (!value.isInstanceOf(baseClass)) return null
-        (if (baseClass == Any::class) StandardSubtypesOfAny.getSubclassSerializer(value) else null)?.let { return it as KSerializer<out T> }
-        return polyBase2Serializers[baseClass]?.get(value::class) as? KSerializer<out T>
+        val custom = polyBase2Serializers[baseClass]?.get(value::class) as? KSerializer<out T>
+        if (custom != null) return custom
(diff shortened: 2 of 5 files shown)


runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleImpl.kt
```diff
@@ -20,8 +20,13 @@ internal class SerialModuleImpl(
 
     override fun <T : Any> getPolymorphic(baseClass: KClass<T>, value: T): KSerializer<out T>? {
         if (!value.isInstanceOf(baseClass)) return null
-        (if (baseClass == Any::class) StandardSubtypesOfAny.getSubclassSerializer(value) else null)?.let { return it as KSerializer<out T> }
```
qwwdfsad: Let's please avoid such one-liners in the future. IMO, it is really hard to read and requires much more effort than it should

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -148,15 +149,24 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
...
+    @JvmName("registerSerializer") // Don't mangle method name for prettier stack traces
     internal fun <T : Any> registerSerializer(
         forClass: KClass<T>,
         serializer: KSerializer<T>,
         allowOverwrite: Boolean = false
     ) {
-        if (!allowOverwrite && forClass in class2Serializer) throw SerializerAlreadyRegisteredException(forClass)
+        if (!allowOverwrite) {
+            val previous = class2Serializer[forClass]
+            if (previous != null && previous != serializer)
```
sandwwraith: Still, these serializers would be tested for referential equality (which will likely fail for generic serializers, since they're instantiated each time). I remember we had an idea for comparing serializers by their descriptors

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -148,15 +149,24 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
...
+                throw SerializerAlreadyRegisteredException(
+                    "Serializer for $forClass already registered in this module: $previous, " +
```
sandwwraith: I'm not sure that `previous` would have meaningful `.toString()` in all of the cases. We could probably add `descrptor.serialName` here?

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
```diff
@@ -15,7 +15,7 @@ import kotlin.reflect.KClass
  * with [SerialModule.plus], consider using [SerialModule.overwriteWith] if you want overwriting behaviour.
  */
 @InternalSerializationApi // Will be hidden in the next release
-public class SerializerAlreadyRegisteredException private constructor(msg: String) : IllegalArgumentException(msg) {
+public class SerializerAlreadyRegisteredException internal constructor(msg: String) : IllegalArgumentException(msg) {
```
sandwwraith: You could move exception message to secondary constructors instead

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
qwwdfsad: Filed #677 to resolve it separately. Because it's likely that we need changes in the plugin as well. Let it be the way it is right now

runtime/commonMain/src/kotlinx/serialization/modules/SerializerAlreadyRegisteredException.kt
qwwdfsad: Nah, I'd prefer to keep the logic where it belongs

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
sandwwraith: I propose do not call `.equals()` on KSerializer directly, only on `SerialDescriptor`

qwwdfsad: Well, it will complicate the life of format authors, that's why I've created a separate issue for that.
Maybe suggestion to use `this.descriptor == (other as? KSerializer<*>).descriptor` would be better

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -148,15 +149,28 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
...
+                throw SerializerAlreadyRegisteredException(
+                    "Serializer for $forClass already registered in this module: $previous ($previousName), " +
+                            "attempted to register $serializer ($currentName)"
+                )
+            }
+        }
         class2Serializer[forClass] = serializer
     }
 
+    @JvmName("registerPolymorphicSerializer") // Don't mangle
```
sandwwraith: [nit] you not always read file from top to bottom continuously, so I'd suggest copy-paste comment `// Don't mangle method name for prettier stack traces` here completely

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -166,10 +180,16 @@ public class SerializersModuleBuilder internal constructor() : SerialModuleColle
...
-        if (!allowOverwrite && concreteClass in baseClassSerializers) throw SerializerAlreadyRegisteredException(
-            baseClass,
-            concreteClass
-        )
+        if (!allowOverwrite) {
+            val previous = baseClassSerializers[concreteClass]
+            if (previous != null) {
+                if (previous == concreteSerializer) return
+                throw SerializerAlreadyRegisteredException(
+                    "Polymorphic serializer for $concreteClass already registered in this module: $previous, " +
```
sandwwraith: You've decided not to extract `serialName`s from descriptors in this message. Why?

qwwdfsad: Added it back to be consistent with the rest of the changes

PR #671, opened 2020-01-24
Reified extensions for registering serializers in SerialModule

### LINKED ISSUE #669: Provide fluent typeOf-based API for SerialModule builders

qwwdfsad:
Fixes #669

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
@@ -44,8 +44,27 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
     /**
      * @see addSubclass
      */
+    @Deprecated(
+        message = "Use 'subclass(serializer)' instead",
+        level = DeprecationLevel.WARNING,
+        replaceWith = ReplaceWith("subclass(serializer)")
+    )
     public inline fun <reified T : Base> addSubclass(serializer: KSerializer<T>): Unit =
         addSubclass(T::class, serializer)
+
+    public inline fun <reified T : Base> subclass(serializer: KSerializer<T>): Unit =
+        addSubclass(T::class, serializer)
+
+    /**
+     * @see addSubclass
+     */
+    @Deprecated(
+        message = "Use 'subclass' instead",
+        level = DeprecationLevel.WARNING,
+        replaceWith = ReplaceWith("subclass<T>()")
+    )
+    @ImplicitReflectionSerializer
+    public inline fun <reified T : Base> addSubclass(): Unit = addSubclass(T::class, T::class.serializer())
 
     /**
      * @see addSubclass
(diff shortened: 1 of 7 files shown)


runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -51,7 +51,7 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
      * @see addSubclass
      */
     @ImplicitReflectionSerializer
-    public inline fun <reified T : Base> addSubclass(): Unit = addSubclass(T::class, T::class.serializer())
+    public inline fun <reified T : Base> addSubclass(): Unit = addSubclass(T::class, serializer())
```
sandwwraith: maybe better rename this function to `subclass` alone:

```kotlin
polymorphic(Any::class, PolyBase::class) {
  subclass<PolyDerived>() // Looks more DSL-ish
}
```

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
```diff
@@ -23,6 +23,13 @@ public fun <T : Any> serializersModuleOf(kClass: KClass<T>, serializer: KSeriali
 public inline fun <reified T : Any> serializersModule(serializer: KSerializer<T>): SerialModule =
     serializersModuleOf(T::class, serializer)
 
+/**
+ * Shortcut for [serializersModuleOf] function with type parameter.
+ */
+@ImplicitReflectionSerializer
+public inline fun <reified T : Any> serializersModule(): SerialModule =
```
sandwwraith: I'm not sure this is useful. Why would one want to use a module of a single default serializer? They wouldn't choose `@ContextualSerialization` in the first place.

runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
qwwdfsad: Great catch!

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt
qwwdfsad: For the same reason we have `serializersModule(serializer: KSerializer<T>)` and `serializersModuleOf(kClass: KClass<T>, serializer: KSerializer<T>)`

sandwwraith: Functions you've mentioned have explicit `serializer` parameter, i.e. user-defined one

qwwdfsad: Good point, I'll remove this one.
Do you have any other comments regarding this PR or I can merge it?

runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -51,7 +51,7 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
      * @see addSubclass
      */
     @ImplicitReflectionSerializer
-    public inline fun <reified T : Base> addSubclass(): Unit = addSubclass(T::class, T::class.serializer())
+    public inline fun <reified T : Base> subclass(): Unit = addSubclass(T::class, serializer())
```
sandwwraith: It's likely someone already used this function, so we'd better provide deprecation alias.

In other parts, PR is ok (modulo function removing we've already discussed)

PR #674, opened 2020-01-27
Object serializer support in KType and Type based serializer lookups …

### LINKED ISSUE #656: Getting serializer by object's KClass or KType instance
For object classes, it should be a way to find a serializer. Currently, both `KClass<*>.serializer()` and `serializer<T>()` do not work.

The reason why we need this is that we should somehow serialize objects (that is accepted to be allowed). However, it is not possible if an object is a serialization root.

qwwdfsad:
…on JVM

Fixes #656

### DIFF
--- a/runtime/jvmMain/src/kotlinx/serialization/Serialization.kt
+++ b/runtime/jvmMain/src/kotlinx/serialization/Serialization.kt
@@ -4,7 +4,8 @@
 
 package kotlinx.serialization
 
-import kotlin.reflect.KClass
+import java.lang.reflect.*
+import kotlin.reflect.*
 
 @Suppress("UNCHECKED_CAST")
 @ImplicitReflectionSerializer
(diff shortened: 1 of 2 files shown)


sandwwraith:
I'm not sure limiting this to JVM is a good idea. Are we going to document that? Probably users will be only frustrated by this limitation. Is this for our internal usage? Would it do any good for us since we can use reflection on JVM anyway?

qwwdfsad:
>Is this for our internal usage?

It's both for Ktor (only JVM part _for now_) and external usages.

>Are we going to document that?

We can, but I'd rather keep current documentation to `serializer(KType)` compact and simple. Instead, we should extend support of associated objects for objects on JVM and JS.

> Probably users will be only frustrated by this limitation. Would it do any good for us since we can use reflection on JVM anyway?

Let's not forget that apart from MPP users, we have plenty of JVM-only users as well and it would be really nice if they had a proper experience without reflection (even though it is used under the hood).

sandwwraith:
Then create a separate ticket to support this functionality on Native and JS

sandwwraith:
> Then create a separate ticket to support this functionality on Native and JS

because this PR does not _completely_ fix #656

sandwwraith:
> Then create a separate ticket to support this functionality on Native and JS

because this PR does not _completely_ fix #656

Issue #661, opened 2020-01-20
RFC: serialization delegates


pdvrieze:
**What is your use-case and why do you need this feature?**
There are certain design limitations to serialization that make it hard/impossible/much less convenient to have serialization work directly on certain complex types (for example if they have bidirectional links to other objects). These objects can often be handled by custom serializers but this looses convenience, and more importantly annotations. Another option would be to use a plain datastructure (data class-ish) that intermediates the serialization.

Overall there are many cases where an intermediate helper object, purely for serialization, would be beneficial for handling more complex cases where the in-memory representation of an object does not align well with a serialization representation. While a custom serializer can theoretically handle this, it may be more beneficial to add some support for this in the plugin (and library)

**Describe the solution you'd like**
Below is an idea of how to it may be possible to address the issue. I'm all open on implementation details, especially names.

I would like the plugin/library to support serializing the following code:

```
interface SerialDelegate<T> {
  fun create(): T
}

@Serializable
class ComplexClass {
  val a: String
  val b: Int
  val c: SomeChild

  @SerializeThrough
  private fun serialDelegate(): ComplexDelegate = TODO("impl omitted")

  private class ComplexDelegate: SerialDelegate<ComplexClass> {
    override fun create(): ComplexClass = TODO("Create ComplexClass from delegate")
  }
}
```
The way I imagine this could work is: the plugin finds the `@SerializeThrough` annotation and
based upon that will reject all other serialization information. The generated serializer has the following properties:

- The delegate and delegation code can be private
- The serializer delegates all invocations to the serializer of the delegate (with appropriate creation of the delegate object) and creation of the non-delegate at the end of deserialization.
- (maybe) type annotations on the original may be able to override type annotations on the delegate -- I'm not sure this is feasible or really that much of a benefit.

sandwwraith:
Hi, thanks for your suggestion. I'm not sure I understood the idea. Code sample is understandable, however, it's not clear for me how it solves the stated problem: 'objects can often be handled by custom serializers but this looses convenience, and more importantly annotations'. Is `SerialDelegate` used instead of `KSerializer`? If so, where's actual serialization code (calls to Encoder/Decoder)? If not, then how would you like to use `KSerializer` inside of `SerialDelegate`?

pdvrieze:
Basically (in pseudocode) the generated serializer works somewhat like the following serializer (as companion object):
```
    @Serializer(ComplexClass::class)
[...]

Issue #524, opened 2019-07-29
Invalid codegeneration for serialDesc for custom serializer


pdvrieze:
When using an Companion object to stand in as custom serializer it is possible to make it inherit GeneratedSerializer (it does this anyway). Unfortunately the initialisation for the companion object is incorrect. Look at the following snippet:

```
  public kotlinx.serialization.SerialDescriptor getDescriptor();
    Code:
       0: getstatic     #69                 // Field $$serialDesc:Lkotlinx/serialization/SerialDescriptor;
       3: areturn

  static {};
    Code:
       0: new           #125                // class kotlinx/serialization/internal/SerialClassDescImpl
       3: dup
       4: ldc           #127                // String childModel
       6: getstatic     #106                // Field nl/adaptivity/process/processModel/engine/XmlChildModel.Companion:Lnl/adaptivity/process/processModel/engine/XmlChildModel$Companion;
       9: checkcast     #7                  // class kotlinx/serialization/internal/GeneratedSerializer
      12: invokespecial #130                // Method kotlinx/serialization/internal/SerialClassDescImpl."<init>":(Ljava/lang/String;Lkotlinx/serialization/internal/GeneratedSerializer;)V

```
Unfortunately, there is a circular reference here. In the class initialiser the Companion field in the owner object is not (can not) yet initialised, resulting in a null value.

This could be solved by generating the code as instance field/method on the companion object, where instead of a field reference a this reference is used.

sandwwraith:
It is not recommended to inherit `GeneratedSerializer` manually

sandwwraith:
In case of custom serializer, better to define `descriptor` manually

pdvrieze:
This is a custom serializer. Unfortunately there are annotations that are needed and almost the only way to provide annotations is to get them from actual annotations that have been pulled into a descriptor that is automatically serialized. As to the use of `GeneratedSerializer` explicitly. This was actually not in the code in question, but is needed for a workaround (to register the child serializers in the descriptor)

pdvrieze:
I will have a look at fixing the actual code generation bug. In the meantime it seems that it may be an option to allow somehow generating a list of members with annotations that can be used for a custom serializer without having to resort to a synthetic class whose only purpose is to extract these properties for serialization purposes..

pdvrieze:
[...]

PR #682, opened 2020-01-30
Introduce DSL for creating user-defined serial descriptors

qwwdfsad:
* Migrate all descriptors to this DSL from SerialClassDescImpl
    * Fix a few bugs in descriptors that violated SD invariants


runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,240 @@
...
+ * SerialDescriptor("my.package.Data", 3) {
+ *     // intField is deliberately ignored by serializer -- not present in the descriptor as well
+ *     element<Long>("_longField") // longField is named as _longField
+ *     element("stringField", listDescriptor<String>())
+ * }
+ * ```
+ */
+public fun SerialDescriptor(
+    serialName: String,
+    elementsCount: Int,
```
sandwwraith: I do not see why we should force users to pre-calculate this size if it is clearly can be deduced from `element` calls count

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,240 @@
...
+    /**
+     * [Serial][SerialInfo] annotations on a target type.
+     */
+    public var annotations: List<Annotation> = emptyList()
+
+    internal val elementNames: MutableList<String> = ArrayList(elementsCount)
+    private val uniqueNames: MutableSet<String> = HashSet(elementsCount)
+    internal val elementDescriptors: MutableList<SerialDescriptor> = ArrayList(elementsCount)
+    internal val elementAnnotations: MutableList<List<Annotation>> = ArrayList(elementsCount)
+    internal val elementOptionality: MutableList<Boolean> = ArrayList(elementsCount)
```
sandwwraith: `BitSet` crying in a distance

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * Retrieves descriptor of type [T] using reified [serializer] function.
+ */
+@ImplicitReflectionSerializer
+public inline fun <reified T> SerialDescriptorBuilder.descriptor(): SerialDescriptor = serializer<T>().descriptor
+
+/**
+ * Creates descriptor for the type `List<T>` where `T` is the type associated with [typeDescriptor].
+ */
+public fun SerialDescriptorBuilder.listDescriptor(typeDescriptor: SerialDescriptor): SerialDescriptor {
```
sandwwraith: We have convention functions like `.list` to do `IntSerializer.list`, maybe make this API similar?

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    override fun getElementName(index: Int): String = elementNames.getChecked(index)
+    override fun getElementIndex(name: String): Int = name2Index[name] ?: CompositeDecoder.UNKNOWN_NAME
+    override fun getElementAnnotations(index: Int): List<Annotation> = elementAnnotations.getChecked(index)
+    override fun getElementDescriptor(index: Int): SerialDescriptor = elementDescriptors.getChecked(index)
+    override fun isElementOptional(index: Int): Boolean = elementOptionality.getChecked(index)
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
```
sandwwraith: I'm not sure we really have to prohibit comparison to other `SerialDescriptor` implementations

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,240 @@
...
+        other as SerialDescriptorImpl
```
sandwwraith: cast is unnecessary to compare by `serialName`

runtime/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
```diff
@@ -13,8 +13,10 @@ import kotlinx.serialization.*
  * By default, a singleton is serialized as an empty structure, e.g. `{}` in JSON
  */
 @InternalSerializationApi
+@Suppress("DEPRECATED") // Used by compiler plugin
```
sandwwraith: There are probably some user-facing usages (e.g. serialize particular class instance as a singleton), so I suggest to leave this in internal api without deprecating

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
```diff
@@ -63,18 +63,12 @@ public class MapEntrySerializer<K, V>(
...
-    private inner class MapEntryClassDesc : SerialClassDescImpl("kotlin.collections.Map.Entry") {
-        override val kind = StructureKind.MAP
-
-        init {
-            addElement("key")
-            pushDescriptor(keySerializer.descriptor)
-            addElement("value")
-            pushDescriptor(valueSerializer.descriptor)
-        }
+    // TODO map kind is most likely to be incorrect
```
sandwwraith: It allows to write it in JSON literally as map entry `"key":"value"`. It is likely used in protobuf, too

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Serializable
+    @SerialName("Box")
+    class Box<T>(val value: T, val list: List<T>)
+
+    class CustomBoxSerializer<T>(val typeSerializer: KSerializer<T>) {
+        val descriptor: SerialDescriptor = SerialDescriptor("Box", 2, StructureKind.CLASS) {
+            element("value", typeSerializer.descriptor)
+            element("list", ArrayListSerializer(typeSerializer).descriptor)
```
sandwwraith: use `listSerializer` function?

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorSpecificationTest.kt
```diff
@@ -24,23 +25,20 @@ class SerialDescriptorSpecificationTest {
         testHolderDescriptor(HolderDescriptor)
     }
 
-    private object HolderDescriptor : SerialClassDescImpl(
-        "kotlinx.serialization.SerialDescriptorSpecificationTest.Holder"
-    ) {
-        init {
-            addElement("a")
-            pushDescriptor(IntSerializer.nullable.descriptor)
+    private object StaticHolder {
```
sandwwraith: why do we need `StaticHolder`?

runtime/commonTest/src/kotlinx/serialization/features/ObjectSerialization.kt
```diff
@@ -48,16 +46,4 @@ class ObjectSerializationTest : JsonTestBase() {
             json
         )
     }
-
-    @Test
-    fun correctDescriptor() {
```
sandwwraith: why you've deleted this test?

runtime/commonTest/src/kotlinx/serialization/json/JsonInputOutputRecursiveTest.kt
```diff
@@ -155,49 +164,52 @@ class JsonInputOutputRecursiveTest : JsonTestBase() {
...
 
-    private object RecursiveSerializer: KSerializer<DummyRecursive> {
+    private object RecursiveSerializer : KSerializer<SealedRecursive> {
         private const val typeAttribute = "type"
         private const val typeNameA = "a"
         private const val typeNameB = "b"
 
-        override val descriptor: SerialDescriptor
-            get() = SerialClassDescImpl("DummyRecursive")
+        // TODO in builder is not suitable for recursive descriptors
```
sandwwraith: well, this might be a problem. do you have ideas how can we overcome that?

qwwdfsad: Already done in the next commit

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorSpecificationTest.kt
qwwdfsad: to have a `val` to be referenced from another object via delegation

runtime/commonTest/src/kotlinx/serialization/features/ObjectSerialization.kt
qwwdfsad: I've added a separate test to SerialDescriptorSpecificationTest

altavir:
Is it possible to transform existing generated descriptor to a builder? The typical use case of custom descriptors and serializers is to use existing descriptor and meddle with it a little. For example, serialize two fields into one element or vise versa.

runtime/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
qwwdfsad: Well, that's the point of **internal** API: we do not expect users to use it and thus may change it in the way we want unless it breaks binary compatibility

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
qwwdfsad: Then we'll lose the consistency and discoverability of reified version (`listDescriptor<Int>`).
And using `IntSerializer.descriptor.list` is not really readable

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorSpecificationTest.kt
qwwdfsad: This test was written before this function :)

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
qwwdfsad: This test was written before this function :)

qwwdfsad:
@altavir definitely not in the scope of this PR.
It would be nice if you could elaborate on potential API shape and your use-cases in the separate issue

altavir:
@qwwdfsad OK, I will setup additional issue later. But the basic idea is that I want to reuse compiler-generated descriptor and serializer, changing only several fields. Typical case, I've stated above. For example, you need tor transform two fields in JSON into one field.

It probably won't make sense without automatic generation of serializers from descriptors anyway.

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,234 @@
...
+ * and a corresponding [Decoder] counterpart.
+ *
+ * ### Implementation note
+ *
+ * Serial descriptors for primitive kinds are not expected to have any nested elements, thus its element count should be zero.
+ * If a class is represented as a primitive value, its corresponding serial name *should not* be equal to the corresponding primitive type name.
+ * For the `Color` example, represented as single [Int], its descriptor should have [INT] kind, zero elements and serial name **not equals**
+ * to `kotlin.Int`: `PrimitiveDescriptor("my.package.ColorAsInt", PrimitiveKind.INT)`
+ */
+public sealed class PrimitiveKind : SerialKind() {
```
qwwdfsad: While writing the documentation, I've found a few problems related to our mental model: some kinds do not match each other.

So I propose the following breaking change:
* Make `UnionKind.OBJECT` structure kind instead, because it is not a union in any interpretation of "tagged union"
* Make `PolymorphicKind` a subclass of `UnionKind` because it is, in fact, a union. `SEALED` is just a more specific `ENUM` and `OPEN` is just a more specific `CONTEXTUAL`

sandwwraith:
I'd also suggest you to update `docs/custom_serializers.md` so we keep documentation in somewhat actual state

pdvrieze:
> @qwwdfsad OK, I will setup additional issue later. But the basic idea is that I want to reuse compiler-generated descriptor and serializer, changing only several fields. Typical case, I've stated above. For example, you need tor transform two fields in JSON into one field.
> 
> It probably won't make sense without automatic generation of serializers from descriptors anyway.

There is a challenge to update the descriptor in a custom serializer. I see this mainly as a case where you may want to use a [delegate object](https://github.com/Kotlin/kotlinx.serialization/issues/661) #661  for serialization (it would be good if it could use setters/getters)

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -123,13 +123,13 @@ private open class JsonTreeInput(json: Json, override val obj: JsonObject) : Abs
...
-    override fun endStructure(desc: SerialDescriptor) {
-        if (!configuration.strictMode || desc is PolymorphicClassDescriptor) return
+    override fun endStructure(descriptor: SerialDescriptor) {
+        if (!configuration.strictMode || descriptor.kind is PolymorphicKind.OPEN) return
 
         // Validate keys
-        val names = HashSet<String>(desc.elementsCount)
-        for (i in 0 until desc.elementsCount) {
-            names += desc.getElementName(i)
+        val names = HashSet<String>(descriptor.elementsCount)
```
sandwwraith: Replace with `.cachedNamesSet`?

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    // No documentation, let's decide whether it is a union at all
+    public object OBJECT : UnionKind()
+
+    /**
+     * Represents a Kotlin [Enum] with statically known values.
+     * All enum values should be enumerated in descriptor elements.
+     * Corresponding encoder and decoder methods are [Encoder.encodeEnum] and [Decoder.decodeEnum].
+     */
+    public object ENUM_KIND : UnionKind() // https://github.com/JetBrains/kotlin-native/issues/1447
```
sandwwraith: Also, the fact that `EnumDescriptor` has elements of ObjectKind should be mentioned somewhere

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
```diff
@@ -0,0 +1,82 @@
...
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+@ImplicitReflectionSerializer
```
sandwwraith: we have it enabled in `gradle/configureSourceSets`, don't we?

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
```diff
@@ -0,0 +1,82 @@
...
+            override fun serialize(encoder: Encoder, value: Wrapper) = TODO()
+            override fun deserialize(decoder: Decoder): Wrapper = TODO()
+        }).descriptor
+        original.assertDescriptorEqualsTo(userDefined)
+    }
+
+    @Test
+    fun testMisconfiguration() {
+        assertFailsWith<IllegalStateException> {
+            SerialDescriptor("", StructureKind.CLASS) {
```
sandwwraith: btw, should we prohibit empty serial name?

runtime/commonTest/src/kotlinx/serialization/SerialDescriptorSpecificationTest.kt
sandwwraith: Maybe `companion object`?

runtime/commonTest/src/kotlinx/serialization/features/GenericCustomSerializerTest.kt
```diff
@@ -33,11 +31,10 @@ class CheckedData<T : Any>(val data: T, val checkSum: ByteArray) {
 
 @Serializer(forClass = CheckedData::class)
 class CheckedDataSerializer<T : Any>(private val dataSerializer: KSerializer<T>) : KSerializer<CheckedData<T>> {
-    override val descriptor: SerialDescriptor = object : SerialClassDescImpl("CheckedDataSerializer") {
-        init {
-            addElement("data")
-            addElement("checkSum")
-        }
+    override val descriptor: SerialDescriptor = SerialDescriptor("CheckedDataSerializer") {
+        val typeDescriptor = dataSerializer.descriptor
```
sandwwraith:
```suggestion
        val dataDescriptor = dataSerializer.descriptor
```

runtime/commonTest/src/kotlinx/serialization/json/JsonInputOutputRecursiveTest.kt
sandwwraith: not clear how `defer {}` helps here

runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
```diff
@@ -1,11 +1,31 @@
 /*
  * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-
+@file:Suppress("DEPRECATION_ERROR")
```
sandwwraith: why is it here?

runtime/commonMain/src/kotlinx/serialization/Decoding.kt
```diff
@@ -0,0 +1,509 @@
...
+     *                }
+     *                1 -> {
+     *                    d = composite.decodeDoubleElement(descriptor, 1)
+     *                }
+     *                else -> error("Unexpected index: $index)
+     *            }
+     *
+     *        }
+     *        composite.endStructure(descriptor)
+     *        if (i == null || d == null) throwMissingFieldException()
```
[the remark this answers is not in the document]
qwwdfsad: Yes, this example is not intended to compilable.
MFE expects a name of the missing element and adding this check won't make anything clearer.
I expect to put links to the **complete** examples here as soon as they are written

runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
qwwdfsad: For the sake of EnumSerializer

runtime/commonMain/src/kotlinx/serialization/Polymorphic.kt
```diff
@@ -4,26 +4,23 @@
...
-        addElement("class")
-        addElement("value")
-    }
+@Deprecated(
+    message = "Top-level polymorphic descriptor is deprecated, use descriptor from the instance of PolymorphicSerializer or" +
+            "check for descriptor kind instead", level = DeprecationLevel.WARNING
+)
+public val PolymorphicClassDescriptor = SerialDescriptor("kotlinx.serialization.Polymorphic", PolymorphicKind.OPEN) {
+    element("class", StringSerializer.descriptor)
+    // TODO make a decision on this one, see contextual
```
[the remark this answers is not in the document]
qwwdfsad: Follow up, it's a more complex than just a removal

runtime/commonMain/src/kotlinx/serialization/Encoding.kt
```diff
@@ -0,0 +1,384 @@
...
+    public fun encodeByte(value: Byte)
+
+    /**
+     * Encodes a 16-bit short value.
+     * Corresponding kind is [PrimitiveKind.SHORT].
+     */
+    public fun encodeShort(value: Short)
+
+    /**
+     * Encodes a 16-bit unicode character value.
```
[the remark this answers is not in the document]
qwwdfsad: I wish our characters were code points :)

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -68,10 +68,10 @@ class Cbor(val updateMode: UpdateMode = UpdateMode.BANNED, val encodeDefaults: B
         override fun encodeNull() = encoder.encodeNull()
 
         override fun encodeEnum(
-            enumDescription: SerialDescriptor,
-            ordinal: Int
+            enumDescriptor: SerialDescriptor,
+            index: Int
```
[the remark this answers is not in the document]
qwwdfsad: In this way, it is consistent with the rest of the methods. Moreover, we do not require an index to be related with `Enum.ordinal` in any way

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -0,0 +1,245 @@
...
+     * ```
+     * class Data(
+     *     val intField: Int? = null, // Optional, has default value
+     *     @ProtoId(1) val longField: Long
+     * )
+     *
+     * // Corresponding descriptor
+     * SerialDescriptor("package.Data", 2) {
+     *     element<Int?>("intField", isOptional = true)
+     *     element<Long>("longField", annotations = listOf(protoIdAnnotationInstance))
```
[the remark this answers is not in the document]
qwwdfsad: Still better than nothing. At least we'll see the demand on that :)

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,236 @@
...
+package kotlinx.serialization
+
+/**
+ * Serial kind is an intrinsic property of [SerialDescriptor] that indicates how
+ * the corresponding type is structurally represented by its serializer.
+ *
+ * Kind is used by serialization formats to determine how exactly the given type
+ * should be serialized. For example, JSON format detects the kind of the value and,
+ * depending on that, may write it as a plain value for primitive kinds, open a
+ * curly brace '{' for class-like structures and square bracket '[' for list- and array- like structures.
```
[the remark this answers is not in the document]
qwwdfsad: The same, both it's worth to mention both, We don't really mention arrays often.

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,236 @@
...
+     * application-specific restrictions on specific [LIST] types.
+     *
+     * Example of such application-specific serialization may be class `class ListOfThreeElements() : List<Any>`,
+     * for which an author of the serializer knows that while it is `List<Any>`, in fact, is always has three elements
+     * of a known type (e.g. the first is always a string, the second one is always an int etc.)
+     */
+    public object LIST : StructureKind()
+
+    /**
+     * Structure kind for maps of an arbitrary length.
```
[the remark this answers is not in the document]
qwwdfsad: We don't have any yet.
I'll visit this one more time during list/map serializers documentation

runtime/commonMain/src/kotlinx/serialization/internal/Util.kt
```diff
@@ -72,10 +72,33 @@ object HexConverter {
...
         result += getElementName(i)
     }
     return result
 }
+
+/**
+ * Returns serial descriptor that delegates all the calls to descriptor returned by [deferred] block.
+ * Used to resolve cyclic dependencies between recursive serializable structures.
+ */
+internal fun defer(deferred: () -> SerialDescriptor): SerialDescriptor = object : SerialDescriptor {
```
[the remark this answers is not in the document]
qwwdfsad: I expect them to complain in our issue tracker first.
Then we can come up with a more decent solution

runtime/commonMain/src/kotlinx/serialization/Encoding.kt
```diff
@@ -0,0 +1,384 @@
...
+     *     encoder.encodeStringValue(value)
+     * } else {
+     *     encoder.encodeNull()
+     * }
+     * ```
+     *
+     * This method has a use in highly-performant binary formats and can
+     * be safely ignore by most of the regular formats.
+     */
+    public fun encodeNotNullMark()
```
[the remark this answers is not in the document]
qwwdfsad: I'll address it a bit later (along with the rest of default impls)

runtime/commonMain/src/kotlinx/serialization/Decoding.kt
```diff
@@ -0,0 +1,542 @@
...
+ * representations.
+ *
+ * Decoder provides high-level API that operates with basic primitive types, collections
+ * and nested structures. Internally, decoder represents input storage and operates with its state
+ * and lower level format-specific details.
+ *
+ * To be more specific, serialization asks a decoder for a sequence of "give me an int, give me
+ * a double, give me a list of strings and give me another object that is a nested int", while decoding
+ * transforms this sequence into a format-specific commands such as "parse the part of the string until the next quotation mark
+ * as an int to retrieve an int, parse everything withing the next curly braces to retrieve elements of a nested object etc."
```
sandwwraith:
```suggestion
 * as an int to retrieve an int, parse everything within the next curly braces to retrieve elements of a nested object etc."
```

runtime/commonMain/src/kotlinx/serialization/KSerializer.kt
```diff
@@ -0,0 +1,179 @@
...
+public interface KSerializer<T> : SerializationStrategy<T>, DeserializationStrategy<T> {
+    /**
+     * Describes the structure of the serializable representation of [T], produced
+     * by this serializer. Knowing the structure of the descriptor is required to determine
+     * the shape of the serialized form (e.g. what elements are encoded as lists and what as primitives)
+     * along with its metadata such as alternative names.
+     *
+     * The descriptor is used dynamically, during serialization by encoders and decoders
+     * to introspect the type and metadata of [T]'s elements being encoded or decoded and
+     * statically, to introspect the type, infer the schema or to compare against the predefined schema.
```
sandwwraith: I'm not sure 'dynamic/static' is a good wording. What's dynamic in encoders and what's static in schema generators?

runtime/commonMain/src/kotlinx/serialization/KSerializer.kt
```diff
@@ -0,0 +1,179 @@
...
+ * Serialization is decoupled from the encoding process to make it completely format-agnostic.
+ * Serialization represents a type as its serial form and is abstracted from the actual
+ * format (whether its JSON, ProtoBuf or a hashing) and is not aware of the underlying storage
+ * (whether it is a string builder, byte array or a network socket), while
+ * encoding/decoding is abstracted from a particular type and its serial form and is responsible
+ * for transforming primitives ("here in an int property 'foo'" call from a serializer) into a particular
+ * format-specific representation ("for a given int, append a property name in quotation marks,
+ * then append a semicolon, then append an actual value" for JSON) and how to retrieve a primitive
+ * ("give me an int that is 'foo' property") from the underlying representation ("expect the next string to be 'foo',
+ * parse it, then parse semicolon, then parse a string until the next comma as an int an return it).
```
sandwwraith:
```suggestion
 * parse it, then parse semicolon, then parse a string until the next comma as an int and return it).
```

runtime/commonMain/src/kotlinx/serialization/KSerializer.kt
```diff
@@ -0,0 +1,179 @@
...
+package kotlinx.serialization
+
+/**
+ * KSerializer is responsible for the representation of a serial form of a type [T]
+ * in terms of [encoders][Encoder] and [decoders][Decoder] and for constructing and deconstructing [T]
+ * from/to a sequence of encoding primitives.
+ *
+ * Serialization is decoupled from the encoding process to make it completely format-agnostic.
+ * Serialization represents a type as its serial form and is abstracted from the actual
+ * format (whether its JSON, ProtoBuf or a hashing) and is not aware of the underlying storage
```
sandwwraith: (Grazie suggestion)
```suggestion
 * format (whether its JSON, ProtoBuf or a hashing) and is unaware of the underlying storage
```

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,254 @@
...
+     * Serializers typically encode classes with calls to [Encoder.beginCollection] and [CompositeEncoder.endStructure],
+     * writing the elements of the map between these calls.
+     *
+     * Built-in map serializers treat elements as homogeneous, though application-specific serializers may impose
+     * application-specific restrictions on specific [MAP] types.
+     */
+    public object MAP : StructureKind()
+
+    /**
+     * Structure kind for singleton objects defined with `object` keyword with.
```
sandwwraith:
```suggestion
     * Structure kind for singleton objects defined with `object` keyword.
```

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -0,0 +1,254 @@
...
+     * By default, objects are serialized as empty structures without any states and their identity is preserved
```
sandwwraith:
```suggestion
     * By default, objects are serialized as empty structures without any state and their identity is preserved
```

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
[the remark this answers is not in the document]
sandwwraith: https://github.com/Kotlin/kotlinx.serialization/issues/524#issuecomment-532222969

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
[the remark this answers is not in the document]
sandwwraith: Then you can mention arrays and sets in documentation for `StructureKind.LIST`

runtime/commonMain/src/kotlinx/serialization/internal/Util.kt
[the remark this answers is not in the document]
sandwwraith: well, use-case and reproducer is pretty obvious, why should we wait for issue ticket

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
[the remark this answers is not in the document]
qwwdfsad: ALready done

runtime/commonMain/src/kotlinx/serialization/internal/Primitives.kt
```diff
@@ -70,7 +70,7 @@ private fun checkName(serialName: String) {
 
 @Suppress("UNCHECKED_CAST")
 internal fun <T : Any> KClass<T>.builtinSerializerOrNull(): KSerializer<T>? =
-    BUILTIN_SERIALIZERS[this] as KSerializer<T>
+    BUILTIN_SERIALIZERS[this] as? KSerializer<T>
```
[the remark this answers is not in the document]
qwwdfsad: It will fail with `ClassCastException: Illegal cast` on JS without any meaningful message, for example.
`builtinSerializerOrNull` is expected to return `null` if there is not serializer and caller will throw more specific exception

runtime/commonMain/src/kotlinx/serialization/Decoding.kt
```diff
@@ -0,0 +1,537 @@
...
+ *        val value = composite.decodeStringElement(descriptor, index = 0)
+ *        // Denotes end of the structure
+ *        composite.endStructure(descriptor)
+ *    }
+ * }
+ * ```
+ *
+ * This deserializer does not know anything about the underlying data and will work with any properly-implemented decoder.
+ * JSON, for example, parses an opening bracket `{` during the `beginStructure` call, checks that the next key
+ * after this bracket is `stringValue` (using the descriptor), returns the value after the semicolon as string value
```
EdwarDDay: Normally a colon should follow the property name
```suggestion
 * after this bracket is `stringValue` (using the descriptor), returns the value after the colon as string value
```

runtime/commonMain/src/kotlinx/serialization/Decoding.kt
```diff
@@ -0,0 +1,537 @@
...
+     * The need in such loop comes from unstructured nature of most serialization formats.
+     * For example, JSON for the following input `{"d": 2.0, "i": 1}`, will first read `d` key with index `1`
+     * and only after `i` with index `0`.
+     *
+     * A potential implementation of this method for JSON format can be the following:
+     * ```
+     * fun decodeElementIndex(descriptor: SerialDescriptor): Int {
+     *     // Ignore arrays
+     *     val nextKey: String? = myStringJsonParser.nextKey()
+     *     if (nextObjectKey == null) return READ_DONE
```
EdwarDDay:
```suggestion
     *     if (nextKey == null) return READ_DONE
```

runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
```diff
@@ -57,11 +57,31 @@ package kotlinx.serialization
...
+ * ```
+ * // Class with custom serializer and custom serial descriptor
+ * class Data(
+ *     val intField: Int, // This field is ignored by custom serializer
+ *     val longField: Long, // This field is written as long, but in serialized form is named as "_longField"
+ *     val stringList: List<String> // This field is written as regular list of strings
+ * )
+ *
+ * // Descriptor for such class:
+ * SerialDescriptor("my.package.Data", 3) {
```
EdwarDDay: `SerialDescriptor` has no `Int` argument
```suggestion
 * SerialDescriptor("my.package.Data") {
```

qwwdfsad:
@EdwarDDay thanks for the help!

Issue #489, opened 2019-06-17
Provide way to specify which values to elide


MariusVolkhart:
**What is your use-case and why do you need this feature?**
Consider the following Kotlin class
```kotlin
data class Ball(val action: String = "Bounce")
```

The `action` property has a default value of `Bounce` because in my program, most balls bounce. There may be instances where balls do not `Bounce` however.

When I serialize this class to JSON, my `action` is elided if the value of the action is `Bounce`. This is undesirable. I want the action to be in the JSON, as other programs in my system may not have the same default. Put differently, the default is a trait of my program, not of the data itself. For example, a program for a bowling alley would probably not consider all balls to `Bounce`.

One option I have for making sure that my `action` is always in the JSON is to write a custom `KSerializer<Ball>`. This may be feasible in the simple case presented here, but doesn't scale as the number of properties on `Ball` increases. Maintaining a custom `KSerializer` for the purposes of a single property seems sub-optimal.

**Describe the solution you'd like**
A solution that appears simple on the surface is an annotation that lets me specify what the default value is.

For example,
```kotlin
@Serializable
data class Ball(
    @Default(null) val action: String = "Bounce",
    @Default(5) val diameter: Int = 5,
    @Default(0) val age: Int
)

fun main() {
    val json = Json(JsonConfiguration.Stable.copy(encodeDefaults = false))
    val encoded = json.stringify(Ball.serializer(), Ball(age = 0))
    if (encoded.contains("\"action\"")) {
        throw AssertionError("Action was not elided but should have been.")
    }
    if (encoded.contains("\"diameter\"")) {
        throw AssertionError("Diameter was not elided but should have been.")
    }
    if (encoded.contains("\"age\":")) {
        throw AssertionError("Age was not elided but should have been.")
    }
}
```

A solution that I think addresses more use cases is a way to plugin to the serialization pipeline via `SerialModule`. Some sort of way for the serialization mechanism to tell me "this is Type being serialized, and this is it's value" and for me to respond "do|don't serialize it, use <this> value, and make <this> the name".

For example, when serializing the `Ball` class, I would be told, "a `String` called `action` with value of `Bounce` is being serialized. I could then reply with, "Serialize a `String` called `action` with value `Bounce`. If I were told, "a `String` called `action` with value of `null` is being serialized" I could reply with, "skip this property".

***Contextual Serialization***
[...]

Issue #471, opened 2019-05-28
Polymorphic type determined by side property value


jimirocks:
Some API returns json like this:
```json
{ "control": "type1/X", "value": { "a": "a"}}
{ "control": "type1/Y", "value": { "a": "b"}}
{ "control": "type2/X", "value": { "c": "d"}}
```
Ideally I would write `sealed class Value`, but is there a way to let kotlinx.serialization pick up the correct serializer for value based on the property `control`. Moreover the control property is not an object type discriminator, but different prefixes maps to different values. In Jackson once can use `TypeIdResolver` for this.

I tried to write some custom serializer - while serialization is just fine, the deserialization is problematic - since there is no guarantee the `control` property comes sooner than `value` so the code below doesn't work:

```kotlin
loop@ while (true) {
                when (val i = compositeDecoder.decodeElementIndex(descriptor)) {
                    CompositeDecoder.READ_DONE -> break@loop
                    0 -> control = compositeDecoder.decodeStringElement(descriptor, i)
                    1 -> value = compositeDecoder.decodeSerializableElement(descriptor, i, chooseSerializer(control))
                    else -> throw SerializationException("Unknown index $i")
                }
            }
```

Additionaly I didn't find a way how to "remember raw value" for later decoding based on `control`

Thanks for help

Issue #468, opened 2019-05-27
Json "wrapper object" support


jimirocks:
Is there an easy to use syntax to configure `@Serializable` wrapper object similar to jacksons`
```
@JsonTypeName("someName")
@JsonTypeInfo(include = JsonTypeInfo.As.WRAPPER_OBJECT, use = JsonTypeInfo.Id.NAME)
```

in order to produce read jsons like 
```
{ "someName": {...}}
```
There are many APIs in the wild using this pattern. I know I could achieve this using another "wrapper" class, but that's too verbose.

May be the existing `@SerialName` annotation with some some wrapping indication could be used for that?

jimirocks:
Well here is the code for wrapping object serializer, the ugly thing on that is that I need to explicitly write the inner Serializer instead of let it have generated.

```kotlin

@Serializer(forClass = WrappedObject::class)
objectWrappedObjectSerializer: KSerializer<WrappedObject> {

    private val innerSerializer = object : KSerializer<WrappedObject> {
        override val descriptor: SerialDescriptor = object : SerialClassDescImpl("WrappedObjectInner") {
            init {
                addElement("test")
            }
        }

        override fun deserialize(decoder: Decoder): WrappedObject {
            TODO("not implemented")
        }

        override fun serialize(encoder: Encoder, obj: WrappedObject) {
            encoder.beginStructure(descriptor).apply {
                encodeStringElement(descriptor, 0, obj.test)
                endStructure(descriptor)
            }
        }

    }

    override val descriptor: SerialDescriptor = object : SerialClassDescImpl("WrappedObject") {
        init {
            addElement("someName")
        }
    }

    override fun serialize(encoder: Encoder, obj: WrappedObject) {
        val compositeEncoder = encoder.beginStructure(descriptor)
        compositeEncoder.encodeSerializableElement(descriptor, 0, innerSerializer, obj)
        compositeEncoder.endStructure(descriptor)
    }
}

```

micheljung:
@sandwwraith I looked through the documentation but failed to find a strategy that solves this problem - could you please point me to it?

sandwwraith:
@micheljung There's no built-in strategy for that, but I believe it can be easy implemented using [skeleton classes for Json transformations](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#json-transformations)

jimirocks:
[...]

Issue #419, opened 2019-03-29
Unwrapping single value JSON arrays to objects


slomkowski:
It is a common pattern for APIs to return single object wrapped in an array as in following JSON:

```json
[
  {
    "foo": "bar",
    "lorem": "ipsum",
    "num": 42
  }
]
```
which should be deserialized to:

```kotlin
data class Example(
        val foo: String,
        val lorem: String,
        val num: Int)
```
Jackson already implements this feature, it can be enabled with property `UNWRAP_SINGLE_VALUE_ARRAYS`. More info: https://github.com/FasterXML/jackson-databind/issues/381.

It would be nice to have similar option here.

JakeWharton:
I would advocate against baking any support for this in. You should be able to accomplish this yourself by writing a custom serializer and using `@Serializable(with=UnwrapSingleValueArray::class).

leomillon:
@slomkowski I faced the same issue and I quickly developed a "workaround" but that's not really the best way to implement it... Anyway, hope it will help to unblock some devs:

```kotlin
import kotlinx.serialization.Decoder
import kotlinx.serialization.Encoder
import kotlinx.serialization.KSerializer
import kotlinx.serialization.SerialDescriptor
import kotlinx.serialization.Serializer
import kotlinx.serialization.encode
import kotlinx.serialization.internal.ArrayListSerializer
import kotlinx.serialization.internal.SerialClassDescImpl
import kotlinx.serialization.json.Json
import kotlinx.serialization.json.JsonArray
import kotlinx.serialization.json.JsonElementSerializer

@Serializer(forClass = List::class)
class UnwrapSingleValueArray<T : Any>(private val dataSerializer: KSerializer<T>) : KSerializer<List<T>> {
    override val descriptor: SerialDescriptor = object : SerialClassDescImpl("UnwrapSingleValueArray") {}

    override fun deserialize(decoder: Decoder): List<T> {
        return when (val jsonElt = JsonElementSerializer.deserialize(decoder)) {
            is JsonArray -> jsonElt.content.map { jsonClient.parse(dataSerializer, it.toString()) }
            else -> listOf(jsonClient.parse(dataSerializer, jsonElt.toString()))
        }
    }

    override fun serialize(encoder: Encoder, obj: List<T>) {
        encoder.encode(ArrayListSerializer(dataSerializer), obj)
    }
}
```

Note: the `jsonClient` is just the `kotlinx.serialization.json.Json` object.

[...]

PR #686, opened 2020-02-03
Abstract classes to ease custom json transformations

### LINKED ISSUE #654: Parse single object or list of objects
As described [here](https://stackoverflow.com/questions/53344033/moshi-parse-single-object-or-list-of-objects-kotlin): sometimes api's will return a json value which could contain a single object or a list of objects.

Also see the "data" json values in the json:api [example](https://jsonapi.org/)

**json:api example**
```
{
  "links": {
    "self": "http://example.com/articles",
    "next": "http://example.com/articles?page[offset]=2",
    "last": "http://example.com/articles?page[offset]=10"
  },
  "data": [{
    "type": "articles",
    "id": "1",
    "attributes": {
      "title": "JSON:API paints my bikeshed!"
    },
    "relationships": {
      "author": {
        "links": {
          "self": "http://example.com/articles/1/relationships/author",
          "related": "http://example.com/articles/1/author"
        },
        "data": { "type": "people", "id": "9" }
      },
      "comments": {
        "links": {
          "self": "http://example.com/articles/1/relationships/comments",
          "related": "http://example.com/articles/1/comments"
        },
        "data": [
          { "type": "comments", "id": "5" },
          { "type": "comments", "id": "12" }
        ]
      }
    },
    "links": {
      "self": "http://example.com/articles/1"
    }
  }],
  "included": [{
    "type": "people",
    "id": "9",
    "attributes": {
      "firstName": "Dan",
      "lastName": "Gebhardt",
      "twitter": "dgeb"
    },
    "links": {
      "self": "http://example.com/people/9"
    }
  }, {
    "type": "comments",
    "id": "5",
    "attributes": {
      "body": "First!"
    },
    "relationships": {
      "author": {
        "data": { "type": "people", "id": "2" }
      }
    },
    "links": {
      "self": "http://example.com/comments/5"
    }
  }, {
    "type": "comments",
    "id": "12",
    "attributes": {
      "body": "I like XML better"
    },
    "relationships": {
      "author": {
        "data": { "type": "people", "id": "9" }
      }
    },
    "links": {
      "self": "http://example.com/comments/12"
    }
  }]
}

```

Maybe use of `vararg`, or an `SingleToArray` annotation like in that moshi example to read single objects into a list

I am attempting to write a custom serializer to handle this case... but is there an example anywhere of how to do this in a clean way?

sandwwraith:
Users tend to struggle when writing custom serializers due to complex calling conventions. The ability to manipulate JSON AST directly will ease the task and also cover as much functionality as possible from some community requests without adding baked-in mechanisms. 

fixes #654, #489, #468, #419, #471 and probably some other feature requests.


runtime/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -56,6 +56,14 @@ public sealed class JsonElement {
     public val isNull: Boolean
         get() = this === JsonNull
 
+    /**
+     * Convenience method to check whether element represents a [JsonObject] and contains given [key].
+     * Returns false if element is not a [JsonObject].
+     */
+    public operator fun contains(key: String): Boolean {
```
qwwdfsad: KDoc style, it's better not to add "Convenience method" as it neither informative nor helpful.
For example:
```
Checks whether an element is a [JsonObject] and contains the given [key].
```

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * // Now both statements will yield different subclasses of Payment:
+ *
+ * Json.parse(PaymentSerializer, """{"amount":"1.0","date":"03.02.2020"}""")
+ * Json.parse(PaymentSerializer, """{"amount":"2.0","date":"03.02.2020","reason":"complaint"}""")
+ * ```
+ *
+ * @param T A root class for all classes that could be possibly encountered during serialization and deserialization.
+ * @param baseClass A class token for [T].
+ *
```
qwwdfsad: redundant empty line

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
```diff
@@ -0,0 +1,96 @@
...
+public abstract class JsonParametricSerializer<T : Any>(private val baseClass: KClass<T>) : KSerializer<T> {
+    /**
+     * A descriptor for this set of parametric serializers.
+     * By default, it uses the name composed of [baseClass] simple name,
+     * kind is set to [PolymorphicKind.SEALED] and contains 0 elements.
+     *
+     * However, this descriptor can be overridden to achieve better representation of custom transformed JSON shape
+     * for schema generating/introspection purposes.
+     */
+    override val descriptor: SerialDescriptor =
```
qwwdfsad: Great idea with a pre-built descriptor.
For the sake of consistency, let's change its type to `CONTEXTUAL` because it's not really about sealed classes

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Json.parse(PaymentSerializer, """{"amount":"1.0","date":"03.02.2020"}""")
+ * Json.parse(PaymentSerializer, """{"amount":"2.0","date":"03.02.2020","reason":"complaint"}""")
+ * ```
+ *
+ * @param T A root class for all classes that could be possibly encountered during serialization and deserialization.
+ * @param baseClass A class token for [T].
+ *
+ */
+public abstract class JsonParametricSerializer<T : Any>(private val baseClass: KClass<T>) : KSerializer<T> {
```
qwwdfsad: Maybe let's rename it to `JsonContextualSerializer`?
Not sure how approachable is `parametric`

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    final override fun deserialize(decoder: Decoder): T {
+        val input = decoder.asJsonInput()
+        val tree = input.decodeJson()
+        val actualSerializer = selectSerializer(tree) as KSerializer<T>
+        return input.json.fromJson(actualSerializer, tree)
+    }
+
+    /**
+     * A function that determines a particular serializer looking on a parsed [element].
```
qwwdfsad: KDoc style: `Determines a concrete serializer looking on a parsed JSON [element].`

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
```diff
@@ -0,0 +1,103 @@
...
+
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.json.internal.*
+
+/**
+ * Base class for custom serializers that allows manipulating an abstract JSON
+ * representation of the class.
+ *
+ * Instead of direct interaction with [Encoder] or [Decoder], this class
```
qwwdfsad: less 'you', more 'we': `this class allows manipulations with [JsonElement] directly`.

runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
```diff
@@ -51,3 +51,6 @@ public interface JsonOutput: Encoder, CompositeEncoder {
      */
     public fun encodeJson(element: JsonElement)
 }
+
+internal fun Encoder.asJsonOutput() = this as? JsonOutput
+        ?: throw SerializationException("This transformation can be used only with Json")
```
qwwdfsad: Message assumes that it is used only from the transformer. Either the method should belong to or not to mention transformation.
Please use `JsonException` instead and be a bit more specific in the message: "Expected to have JsonOutput, but has this::class instead"

qwwdfsad: Same for `asJsonInput`

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
```diff
@@ -0,0 +1,103 @@
...
+ *     }
+ * }
+ * // Now these functions both yield correct result:
+ * Json.parse(Example.serializer(), """{"data":["str1"]}""")
+ * Json.parse(Example.serializer(), """{"data":"str1"}""")
+ * ```
+ *
+ * @param T A type for Kotlin property for which this serializer could be applied.
+ *        **Not** the type that you may encounter in JSON. (e.g. if you unwrap a list
+ *        to a single value `T`,use `T`, not `List<T>`)
```
qwwdfsad: indent after comma

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
```diff
@@ -0,0 +1,103 @@
...
+ * @param tSerializer A serializer for type [T]. Determines [JsonElement] which is passed to [writeTransform].
+ *        Should be able to parse [JsonElement] from [readTransform] function.
+ *        Usually, default serializer is sufficient.
+ * @param transformationName A name for the particular implementation to fulfill [SerialDescriptor.serialName] uniqueness guarantee.
+ */
+abstract class JsonTransformingSerializer<T : Any>(
```
qwwdfsad: missing public

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
```diff
@@ -0,0 +1,103 @@
...
+    private val tSerializer: KSerializer<T>,
+    transformationName: String
+) : KSerializer<T> {
```
qwwdfsad: The class itself is great and I'd expected it to be used frequently.
The only concern is a not-so-short name. WDYT about "JsonTransformer" instead?

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
sandwwraith: After #682, because `CONTEXTUAL` is not yet in dev )

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
sandwwraith: Well, I though 'Parametric Polymorphsim' is somewhat used term in Jackson and Gson but I've looked it up and seems it is not used anywhere outside FP. Anyway, 'parametric' kinda may fit because we look at unknown in advance parameter – json property

Contextual doesn't fit either because we have regular `@Contextual` and `ContextSerializer` that work in a completely different way to `JsonContextualSerializer`

All references say something about 'property value' or 'property presence' or 'property definitions'... Maybe we should coin something around words 'property' , 'content' or 'value'

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
sandwwraith: We have 'Serializer' word appended to all `KSerializer`s, AFAIR. I do not have any objections to `JsonTransformer` besides that. 
Also, this name occurs only in supertype position, so maybe we shouldn't pursue shortness at all

runtime/commonMain/src/kotlinx/serialization/json/JsonParametricSerializer.kt
qwwdfsad: Let's stick with parametric then

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
qwwdfsad: I am fine with both options

sandwwraith:
Fixed everything except `CONTEXTUAL`, will add after accepted review & rebase

docs/json_transformations.md
```diff
@@ -1,10 +1,10 @@
 # Json-specific serializers
 
 To affect the shape and contents of JSON output after serialization, or adapt input to deserialization,
-one may write a [custom serializer](custom_serializers.md). However, it may not be convenient to
+it is possible write a [custom serializer](custom_serializers.md). However, it may not be convenient to
```
sandwwraith:
```suggestion
it is possible to write a [custom serializer](custom_serializers.md). However, it may not be convenient to
```

docs/json_transformations.md
```diff
@@ -16,56 +16,53 @@ can boil down the burden of implementing a custom serializer to a problem of man
 
 It is still strongly recommended to be acquainted with [custom serializers guide](custom_serializers.md)
 in its [using serializers](custom_serializers.md#using-custom-serializers) part,
-since the instantiation process and annotations to enable custom serializers remain the same.
-It is also important to remember that since the discussed in this article serializers are manipulating Json AST,
-they have the following limitations:
+since the instantiation process and annotations to enable custom serializers remains the same.
```
sandwwraith: Isn't `remain` correct because it refers to plural ' instantiation process and annotations'?

docs/json_transformations.md
```diff
@@ -16,56 +16,53 @@ can boil down the burden of implementing a custom serializer to a problem of man
...
-See `JsonElement`, `JsonObject`, `JsonArray`, and `JsonPrimitive` documentation for details.
 
-### Samples
+### Transformation examples. List mainupulation
 
-Here are the samples of several popular (judged by open issues) transformations. You can copy-paste them to your project.
+The first example is our own implementation of list wrapping. Consider the API that returns list 
+of objects, or, if there is only one element in the result, then it is a single object, not wrapped in the list.
+ 
+To simplify object model, it is possible to implement transforming serializer that always returns list of object,
```
sandwwraith:
```suggestion
To simplify object model, it is possible to implement transforming serializer that always returns list of objects,
```

docs/json_transformations.md
```diff
@@ -16,56 +16,53 @@ can boil down the burden of implementing a custom serializer to a problem of man
...
+automatically wrapping a single object in a list as well: 
 
 ```kotlin
-// We're going to use this simple class as a target for our transformations
-@Serializable
+@Serializable // Samle class with data
```
sandwwraith:
```suggestion
@Serializable // Sample class with data
```

docs/json_transformations.md
```diff
@@ -74,46 +71,50 @@ data class Example(
     @Serializable(WrappingJsonListSerializer::class) val moreData: List<StringData>
 )
 
-// And be able to parse different shapes of the same data, such as:
-```
+// And for input 
+{"name":"test","data":{"data":"str1"},"moreData": {"data":"str2"} }
```
sandwwraith: Why have you erased other examples?

docs/json_transformations.md
```diff
@@ -74,46 +71,50 @@ data class Example(
...
-    // =>  """{"data":{"data":"str1"}}"""
+// =>  """{"data":{"data":"str1"}}"""
 ```
 
 ## Json parametric polymorphic deserialization
 
-Usually, [polymorphic serialization](polymorphism.md) requires a dedicated `"type"` property
-(a so-called 'class discriminator') in JSON stream to determine actual serializer
+Typically, [polymorphic serialization](polymorphism.md) requires a dedicated `"type"` property
+(also known as class discriminator) in the resulting JSON to determine actual serializer
```
sandwwraith: probably not the 'resulting', but 'incoming'?

runtime/commonMain/src/kotlinx/serialization/json/JsonTransformingSerializer.kt
```diff
@@ -89,15 +88,14 @@ public abstract class JsonTransformingSerializer<T : Any>(
...
     protected open fun readTransform(element: JsonElement): JsonElement = element
 
     /**
      * Transformation which happens during [deserialize] call.
-     *
      * Does nothing by default.
      */
     protected open fun writeTransform(element: JsonElement): JsonElement = element
 }
+
```
sandwwraith: redundant empty line?

docs/json_transformations.md
qwwdfsad: To reduce the amount of cognitive load, typically one example is enough

Issue #687, opened 2020-02-04
DynamicObjectParser: can't parse children of a sealed class


SerVB:
**Describe the bug**

It's not possible to use `DynamicObjectParser.parse` for parsing children of sealed classes.

**To Reproduce**
```kotlin
@Serializable
sealed class AAA

@Serializable
data class BBB(val i: Int) : AAA()

fun main() {
    val dynamicObjectParser = DynamicObjectParser()

    val b = BBB(4)

    val serializedB = Json(JsonConfiguration.Stable).stringify(AAA.serializer(), b)
    println(serializedB)

    val nativelyParsedB = JSON.parse<Any>(serializedB)
    println(nativelyParsedB)

    val parsedB = dynamicObjectParser.parse(nativelyParsedB, AAA.serializer())
    println(parsedB)
}
```

It crashes on the line `val parsedB = ...` with "IllegalArgumentException(Polymorphic value has not been read for class null) at AbstractPolymorphicSerializer.deserialize(Decoder)".

If I replace `AAA.serializer()` with `BBB.serializer()` in both places, it works, but it doesn't suit me because I need polymorphic serialization: in real code I have multiple children of a sealed class.

**Expected behavior**

I want it to make `parsedB == b`.

**Environment**
 - Kotlin version: 1.3.61
 - Library version: 0.14.0
 - Kotlin platforms: JS
 - Gradle version: 5.2.1
 - Other relevant context: Tested in Chromium 78

JakeWharton:
I'm pretty sure this means that the DOP needs to accept a `JsonConfiguration` since things like the class discriminator could also have been changed.

And, ideally, in the future the library will also support serializing _to_ a dynamic object where things settings like default encoding will matter.

sandwwraith:
Obsolete

PR #694, opened 2020-02-07
Allow DynamicObjectParser to handle polymorphic types

JakeWharton:
This is only a partial fix, as only array-based polymorphism works.

Refs #687.

### DIFF
--- a/runtime/jsMain/src/kotlinx/serialization/DynamicObjectParser.kt
+++ b/runtime/jsMain/src/kotlinx/serialization/DynamicObjectParser.kt
@@ -5,6 +5,7 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.CompositeDecoder.Companion.READ_DONE
+import kotlinx.serialization.json.JsonConfiguration
 import kotlinx.serialization.modules.*
 import kotlin.math.abs
 import kotlin.math.floor
@@ -14,7 +15,10 @@ import kotlin.math.floor
  */
 internal const val MAX_SAFE_INTEGER: Double = 9007199254740991.toDouble() // 2^53 - 1
 
-class DynamicObjectParser(context: SerialModule = EmptyModule): AbstractSerialFormat(context) {
+class DynamicObjectParser(
+    context: SerialModule = EmptyModule,
+    internal val configuration: JsonConfiguration = JsonConfiguration.Default
+): AbstractSerialFormat(context) {
     @ImplicitReflectionSerializer
     inline fun <reified T : Any> parse(obj: dynamic): T = parse(obj, context.getContextualOrDefault(T::class))
 
@@ -76,7 +80,14 @@ class DynamicObjectParser(context: SerialModule = EmptyModule): AbstractSerialFo
 
         override fun beginStructure(desc: SerialDescriptor, vararg typeParams: KSerializer<*>): CompositeDecoder {
             val curObj = currentTagOrNull?.let { obj[it] } ?: obj
-            return when (desc.kind) {
+            val kind = when (desc.kind) {
+                is PolymorphicKind -> {
+                    if (configuration.useArrayPolymorphism) StructureKind.LIST
+                    else StructureKind.MAP
+                }
+                else -> desc.kind
+            }
+            return when (kind) {
                 StructureKind.LIST -> DynamicListInput(curObj)
                 StructureKind.MAP -> DynamicMapInput(curObj)
                 else -> DynamicInput(curObj)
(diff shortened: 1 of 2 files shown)


runtime/jsTest/src/kotlinx/serialization/DynamicParserTest.kt
```diff
@@ -168,4 +176,28 @@ class DynamicParserTest {
         val dyn1 = js("({data: 42})")
         assertEquals(NDWrapper(NotDefault(42)), deserializer.parse(dyn1))
     }
+
+    @Test
+    @Ignore // Currently broken!
```
JakeWharton: There's some problem with `pos` being incremented too late / the tag being read too early with maps. "type" is read from index 0 which then calls `decodeStringElement` which ends up calling `getByTag("type")` but because `pos` is at 0, not 1, the tag is returned directly rather than `obj[tag]`.

I couldn't figure out the problem, but figured this was good enough to land as-is because it at least allows array-based polymorphism to work.

runtime/jsMain/src/kotlinx/serialization/DynamicObjectParser.kt
```diff
@@ -14,7 +15,10 @@ import kotlin.math.floor
  */
 internal const val MAX_SAFE_INTEGER: Double = 9007199254740991.toDouble() // 2^53 - 1
 
-class DynamicObjectParser(context: SerialModule = EmptyModule): AbstractSerialFormat(context) {
+class DynamicObjectParser(
+    context: SerialModule = EmptyModule,
+    internal val configuration: JsonConfiguration = JsonConfiguration.Default
```
JakeWharton: This may be a little weird, but JSON is the closest approximation to JS objects so if you squint it doesn't hurt. At least not as bad as duplicating all of the controls? Although I can pursue that as well if you'd like.

sandwwraith: Well, it makes sense in a light of #628 , where 'encodeDefaults' and 'allowStructuredMapKeys' could also be used (although structured map keys have to be supported in Parser, too). It's also possible to implement strict mode in parser, however, no one complained about lack of it yet.

runtime/jsTest/src/kotlinx/serialization/DynamicParserTest.kt
sandwwraith: I have a feeling that it's more than an off-by-one error, because regular polymorphic serializer expects first a class discriminator in _primitive_ string, and then a start of object. Here, class discriminator is located directly inside object, so it likely won't work without additional special handling, similar to one that performed in JSON: https://github.com/kotlin/kotlinx.serialization/blob/95dc4312d8148c5b945bcb3961e2926feaeef5e2/runtime/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L32 (i.e. basically intercepting `deserialize` call to AbstractPolymorphicSerializer and replacing it with custom logic of manipulating json tree).

Since it is some additional amount of work, you can leave this test as is with some more meaningful comment.

pdvrieze: The challenge with this form of polymorphism is that the type discriminator (if it is not merely the presence/absence of certain attributes is that it normally requires reading the entire json subtree into memory.

Thinking about how to "solve" this, it is certainly something that should be handled at format level, not at serializer level (it is json specific). You could use an extension of the json format that would have an additional configuration for polymorphism (possibly dependent on base type exactly like SerialModule). This configuration would be a function that would accept a `JsonElement` or `JsonObject` and would then return an actual value.

In the format it would hook into (`Encoder`|`Decoder`)`.beginStructure` where it can inspect the passed in serial descriptor for the kind (Polymorphic) and override the default behaviour (even if the serializer wants to write the type in a certain way, the format is free to ignore that - my XML format does not require type discriminators - the tag names are already).

JakeWharton: In Moshi we have a special peeking reader which you can fork off from the main reader. This allows you to peek and consume JSON for things like locating the "type" key for polymorphic deserialization. But instead of moving forward in the actual underlying data where going backwards is challening, it merely buffers the data in the original reader as you progress. Once you're done with the peeked reader, you throw it away and resume consuming the original reader (which is still in the same original position, but with N bytes buffered where N was the number of bytes read by the peeked reader). It means JSON in the form `{"type":"a","value":"b"}` and `{"value":"b","type":a"}` can both be correctly supported without having to convert the whole object to a tree-like model.

I'm not familiar enough with the codebase here to know whether such a thing could be applied, but it's an extremely useful concept for cases like this where you need to arbitrarily look ahead.

> Since it is some additional amount of work, you can leave this test as is with some more meaningful comment.

Thanks, will do.

JakeWharton:
Updated. Let me know what you think.

sandwwraith:
Great, thanks!

Issue #595, opened 2019-11-02
Make PolymorphicClassDescriptor retain type of the base class


altavir:
**What is your use-case and why do you need this feature?**
When using custom serializers ant tools one seldom needs to work with descriptors directly without a `Decoder` in those cases, `PolymorphicClassDescriptor` does not provide sufficient information since it is not possible to know, which field does it represent. 

**Describe the solution you'd like**

The simplest solution is to use base class name for descriptor instead of `Any`. It will increase memory consumption because at the moment Polymorphic descriptor is a singleton, but I do not think it is critical. The state could be moved from `PolymorphicSerializer` into descriptor thus making API more uniform.

sandwwraith:
Is serial name of the base class sufficient for your use-case or you need to retain a full `KClass`?

sandwwraith:
Also see https://github.com/Kotlin/kotlinx.serialization/issues/515

altavir:
Full KClass would be better, but string would suffice. Currently the actual case is when you need to manually work with descriptors. We need to be able to find base class of the property from the descriptor without constructing the serializer. A general solution is to move class description from `PolimorphicSerializer` into `PolymorphicClassDescriptor`.

pdvrieze:
I think we need to look carefully at the architecture of the descriptor part of the framework (in conjunction with the serializers). Perhaps it is worthwhile to create a format descriptor/schema system. That should expose all parts where information is not (easily) accessible (or requires looking at internal/implementation details).

altavir:
@pdvrieze I agree. The descriptor should in fact be capable of defining everything needed to work with schema without creating actual serializer. I also think that descriptor should be extensible and in theory allow for advanced features like validation. It would allow to automatically create consistent  advanced schema (like json/XML schema) as well as doing non-serialization tasks like validation with serialization framework.

qwwdfsad:
Fixed in 1.0.0-RC

PR #695, opened 2020-02-07
Provide context aware descriptor

### LINKED ISSUE #515: The baseclass property of PolymorphicSerializer is private
For my XML parsing I'm trying to enumerate all possible polymorphic children of a parent (as declared in the serialmodule). To do so I need to know the actual base class. This information is encapsulated in the baseclass parameter of Polymorhpic serializer.

Currently I'm using a hack where trigger a lookup on a dummy context which will tell me the base class. It's much easier to just get the property value directly.

sandwwraith:
Which should capture KClass in case of contextual/polymorphic serialization, thus allowing to build full schema, provided correct SerialModule is present

Fixes #515 and #595

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.modules.*
+import kotlin.jvm.*
+import kotlin.reflect.*
+
+
+/**
+ * Retrieves [KClass] associated with serializer and its descriptor, if it was captured.
+ *
+ * For schema introspection purposes, [capturedKClass] can be used in [SerialModule] as a key
+ * to retrieve registered descriptor at runtime.
+ * This property is intended to be used on [UnionKind.CONTEXTUAL] and [PolymorphicKind.OPEN] kinds of descriptors,
+ * where actual serializer used for a property can be determined only at runtime.
+ * Serializers which represent contextual serialization and open polymorphism (namely, [ContextSerializer] and
+ * [PolymorphicSerializer]) capture statically known KClass in a descriptor and can expose it via this property.
+ *
+ * This property is `null` for descriptors that are not of [UnionKind.CONTEXTUAL] or [PolymorphicKind.OPEN] kinds.
+ * It _may_ be `null` for descriptors of these kinds, if captured class information is unavailable for various reasons.
+ * It means that schema introspection should be performed in an application-specific manner.
+ *
+ * ### Example
+ * Imagine we need to find all distinct properties names, which may occur in output after serializing a given class
+ * with respect to [`@ContextualSerialization`][ContextualSerialization] annotation and all possible inheritors when the class is
+ * serialized polymorphically.
+ * Then we can write following function:
+ * ```
+ * fun allDistinctNames(descriptor: SerialDescriptor, module: SerialModule) = when (descriptor.kind) {
+ *   is PolymorphicKind.OPEN -> module.getPolymorphicDescriptors(descriptor)
+ *     .map { it.elementNames() }.flatten().toSet()
+ *   is UnionKind.CONTEXTUAL -> module.getContextualDescriptor(descriptor)
+ *     ?.elementNames().orEmpty().toSet()
+ *   else -> descriptor.elementNames().toSet()
+ * }
+ * ```
+ * @see SerialModule.getContextualDescriptor
+ * @see SerialModule.getPolymorphicDescriptors
+ */
+public val SerialDescriptor.capturedKClass: KClass<*>?
+    get() = when (this) {
+        is ContextDescriptor -> kClass
+        is SerialDescriptorForNullable -> original.capturedKClass
+        else -> null
+    }
+
+/**
+ * Wraps [this] in [ContextDescriptor].
+ */
+internal fun SerialDescriptor.withContext(context: KClass<*>): SerialDescriptor =
+    ContextDescriptor(this, context)
+
+/**
+ * Descriptor that captures [kClass] and allows retrieving additional runtime information,
+ * if proper [SerialModule] is provided.
+ */
+private class ContextDescriptor(
+    private val original: SerialDescriptor,
+    @JvmField val kClass: KClass<*>
+) : SerialDescriptor by original {
+    override val serialName = "${original.serialName}<${kClass.simpleName()}>"
+
+    override fun equals(other: Any?): Boolean {
+        val another = other as? ContextDescriptor ?: return false
+        return original == another.original && another.kClass == this.kClass
+    }
+
+    override fun hashCode(): Int {
+        var result = kClass.hashCode()
+        result = 31 * result + serialName.hashCode()
+        return result
+    }
+}
--- a/runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
@@ -4,6 +4,10 @@
 
 package kotlinx.serialization
 
+import kotlinx.serialization.PrimitiveKind.*
+import kotlinx.serialization.StructureKind.*
+import kotlinx.serialization.modules.*
+
 /**
  * Serial kind is an intrinsic property of [SerialDescriptor] that indicates how
  * the corresponding type is structurally represented by its serializer.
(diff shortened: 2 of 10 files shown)


runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+     * The same class that was known statically while generating a serializer.
+     */
+    public val kClass: KClass<*>
+}
+
+internal fun SerialDescriptor.withContext(context: KClass<*>): ContextAwareDescriptor =
+    ContextAwareDescriptorImpl(this, context)
+
+private class ContextAwareDescriptorImpl(val original: SerialDescriptor, override val kClass: KClass<*>) :
+    SerialDescriptor by original, ContextAwareDescriptor {
```
qwwdfsad: Code style, invalid indents

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -223,6 +227,9 @@ public sealed class UnionKind : SerialKind() {
      * Represents an "unknown" type that will be known only at the moment of the serialization.
      * Effectively it defers the choice of the serializer to a moment of the serialization, and can
      * be used for [contextual][ContextualSerialization] serialization.
+     *
+     * To introspect such a descriptor correctly, one may need a [SerialModule].
```
qwwdfsad: `To introspect descriptor of this kind`?

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.modules.*
+import kotlin.reflect.*
+
+/**
+ * Descriptor which allows retrieving additional runtime information, if proper [SerialModule] is provided.
```
qwwdfsad: "Descriptor that"

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ *   is UnionKind.CONTEXTUAL -> module.getContextualDescriptor(descriptor as ContextAwareDescriptor)
+ *     ?.elementNames().orEmpty().toSet()
+ *   else -> descriptor.elementNames().toSet()
+ * }
+ * ```
+ *
+ * @see SerialModule.getContextualDescriptor
+ * @see SerialModule.getPolymorphicDescriptors
+ */
+public interface ContextAwareDescriptor : SerialDescriptor {
```
qwwdfsad: Maybe `ContextualDescriptor` is a better name?
Because this descriptor is, actually, not aware of any contexts and it may be a bit misleading

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ * Usually, descriptors are generated by compiler plugin and know all information statically.
+ * However, that is not the case for [UnionKind.CONTEXTUAL] and [PolymorphicKind.OPEN] kinds of descriptors,
+ * where actual serializer used for property can be determined only at runtime.
+ * This descriptor is used by serializers which represent contextual serialization and open polymorphism (namely,
+ * [ContextSerializer] and [PolymorphicSerializer]), so they can capture statically known [kClass] in this descriptor.
+ * For schema introspection purposes, this [kClass] can be used in [SerialModule] to retrieve actual descriptor in runtime.
+ *
+ * This descriptor must always has either [UnionKind.CONTEXTUAL] or [PolymorphicKind.OPEN] kind,
+ * and vice versa: if some [SerialDescriptor.kind] is either CONTEXTUAL or OPEN, that descriptor
+ * could be safely casted to [ContextAwareDescriptor].
```
qwwdfsad: Is downcast really allowed?
E.g. what about `descriptor.nullable`?

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * ### Example
+ * Imagine we need to find all distinct properties names, which may occur in output after serializing a given class
+ * with respect to @[ContextualSerialization] annotation and all possible inheritors when the class is
```
qwwdfsad: "@[ContextualSerialization]" is rendered weirdly, it's better to use plaing [ContextualSerialization] or `['@ContextualSerialization'][ContextualSerialization]` (`'` is a backtick, can't escape it in GitHub .md)

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ * serialized polymorphically.
+ * Then we can write following function:
+ * ```
+ * fun allDistinctNames(descriptor: SerialDescriptor, module: SerialModule) = when (descriptor.kind) {
```
qwwdfsad: Let's not advocate for using downcasting. 
`fun allDistinctNames(descriptor: ContextAwareDescriptor, module: SerialModule)` and let the authors decide what to do with non-`ContextAwareDescriptor`

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
```diff
@@ -0,0 +1,68 @@
...
+ *   else -> descriptor.elementNames().toSet()
+ * }
+ * ```
+ *
+ * @see SerialModule.getContextualDescriptor
+ * @see SerialModule.getPolymorphicDescriptors
+ */
+public interface ContextAwareDescriptor : SerialDescriptor {
+    /**
+     * A class which should be used in [SerialModule] as a key.
```
qwwdfsad: "which" -> that
https://www.grammarly.com/blog/which-vs-that/

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -234,14 +234,16 @@ class ProtoBuf(
             get() = this@ProtoBuf.context
 
         private val indexByTag: MutableMap<Int, Int> = mutableMapOf()
-        private fun findIndexByTag(desc: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean = false): Int =
-            (0 until desc.elementsCount).firstOrNull {
+        private fun findIndexByTag(desc: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean = false): Int {
+            if (desc.kind is PolymorphicKind) return serialId - 1
```
qwwdfsad: I am really concerned about this one.
For our own polymorphic descriptor, we break our own contract in the following manner:
1) First, we declare it has 0 elements
2) Then, in `AbstractPolymorphicSerializer` we invoke `encodeX(polymorphicDescriptor, index = 0)`, `encodeX(polymorphicDescriptor, index = 1)` and same for decoding. Any sane format will break at this point and always hacking it with `if (kind == 0) treatIndexAsNonExisitng` is far from the good API design. Let's discuss it

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
sandwwraith: You can use `ContextualDescriptor` as a parameter to `getContextualDescriptor` function, which seems fine. But you could also use it as a parameter to `getPolymorphicDescriptors`, which seems unrelated. Maybe `ContextDescriptor`, at least, then?

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
sandwwraith: It is unclear what to do when kind is CONTEXTUAL, but descriptor is not `ContextDescriptor`. I'd say it is possible to put some efforts in this and make `.nullable` aware of this particular subclass. I'm not 100% sure we should pursue this; we can discuss this in DM if you'd like

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
sandwwraith: Then it would be impossible to generalize this function on a descriptors that are not of context kinds? e.g. `allDistinctNames(MyClass.serializer(), EmptyModule)`

runtime/commonMain/src/kotlinx/serialization/ContextAware.kt
qwwdfsad: `ContextDescriptor` is okay

runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleExtensions.kt
```diff
@@ -61,3 +61,42 @@ public infix fun SerialModule.overwriteWith(other: SerialModule): SerialModule =
...
+        override fun <Base : Any, Sub : Base> polymorphic(
+            baseClass: KClass<Base>,
+            actualClass: KClass<Sub>,
+            actualSerializer: KSerializer<Sub>
+        ) {
+            if (baseClass == kClass) builder.add(actualSerializer.descriptor)
+        }
+    })
+    return builder
+}
```
pdvrieze: When reading this code I believe this is almost the missing piece I've been waiting for. For my XML format I must be able to enumerate all possible children of a parent (so I can determine the tag names to recognize and which property they belong to). I need the same information for descriptors of lists (IOW. I need to get the list element descriptor -- btw. I haven't checked that this information isn't available).

PR #696, opened 2020-02-07
Cleanup

sandwwraith:
* finally get rid of `public HexConverter` in protobuf and cbor production files
* cleanup platform utils mess
* get rid of AbstractSerialFormat superclass

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
@@ -11,9 +11,9 @@ import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 import kotlin.experimental.*
 
-class Cbor(
-    val updateMode: UpdateMode = UpdateMode.BANNED,
-    val encodeDefaults: Boolean = true,
+public class Cbor(
+    public val updateMode: UpdateMode = UpdateMode.BANNED,
+    public val encodeDefaults: Boolean = true,
     override val context: SerialModule = EmptyModule
 ) : BinaryFormat {
     // Differs from List only in start byte
@@ -79,7 +79,7 @@ class Cbor(
     }
 
     // For details of representation, see https://tools.ietf.org/html/rfc7049#section-2.1
-    class CborEncoder(val output: OutputStream) {
+    internal class CborEncoder(val output: OutputStream) {
 
         fun startArray() = output.write(BEGIN_ARRAY)
         fun startMap() = output.write(BEGIN_MAP)
@@ -223,7 +223,7 @@ class Cbor(
 
     }
 
-    class CborDecoder(val input: InputStream) {
+    internal class CborDecoder(val input: InputStream) {
         private var curByte: Int = -1
 
         init {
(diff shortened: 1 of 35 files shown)


formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -11,7 +11,11 @@ import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 import kotlin.experimental.*
 
-class Cbor(val updateMode: UpdateMode = UpdateMode.BANNED, val encodeDefaults: Boolean = true, context: SerialModule = EmptyModule): AbstractSerialFormat(context), BinaryFormat {
+public class Cbor(
+    val updateMode: UpdateMode = UpdateMode.BANNED,
```
qwwdfsad: `public` missing from vals

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
```diff
@@ -12,7 +12,7 @@ interface SerialFormat {
 }
 
 @Deprecated(
-    "AbstractSerialFormat is deprecated for removal since it is indistinguishable from SerialFormat interface." +
+    "Deprecated for removal since it is indistinguishable from SerialFormat interface. " +
             "Use SerialFormat instead.", ReplaceWith("SerialFormat"), DeprecationLevel.WARNING
```
qwwdfsad: Let's make it `DeprecationLevel.ERROR` already?

runtime/jvmMain/src/kotlinx/serialization/JvmResolving.kt
```diff
@@ -6,30 +6,43 @@ package kotlinx.serialization
...
+ * also provide information about nullability.
+ *
  * Keep in mind that this is a 'heavy' call, so result probably should be cached somewhere else.
  *
  * This method intended for static, format-agnostic resolving (e.g. in adapter factories) so context is not used here.
  */
 @Suppress("UNCHECKED_CAST")
 @UseExperimental(ImplicitReflectionSerializer::class)
-fun serializerByTypeToken(type: Type): KSerializer<Any> = when (type) {
+// not deprecated because many converters (e.g. retrofit2-kotlinx-serialization-converter) still have to use java's Type
```
qwwdfsad: This comment is a bit redundant, especially before the method body. The same thing can be rephrased in the documentation itself. 

Notes on KDoc:
* "However, since it accepts type token, it is available only on JVM by design" is also redundant, especially for functions from JVM module that have `java.lang.reflect.Type` in its parameters.
* "This method uses X to do Y" -> Does Y" or "Does Y using X"
* "This method intended for static, format-agnostic resolving (e.g. in adapter factories)"  lacking the context what are adapter factories

Fixed with my commit

runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt
sandwwraith: There was no release with WARNING yet. Moreover, this class is not really harmful and hardly can be accidentaly misused (as opposed to internal functions), and used in 3rd party format implementations.

PR #698, opened 2020-02-10
Add option to use HOCON naming convention

EdwarDDay:
[HOCON naming convention](https://github.com/lightbend/config/blob/master/HOCON.md#hyphen-separated-vs-camelcase)

### DIFF
--- a/formats/config/src/main/kotlin/org/jetbrains/kotlinx/serialization/config/ConfigReader.kt
+++ b/formats/config/src/main/kotlin/org/jetbrains/kotlinx/serialization/config/ConfigReader.kt
@@ -12,7 +12,10 @@ import kotlinx.serialization.modules.*
 private val SerialKind.listLike get() = this == StructureKind.LIST || this is PolymorphicKind
 private val SerialKind.objLike get() = this == StructureKind.CLASS || this == StructureKind.OBJECT
 
-class ConfigParser(context: SerialModule = EmptyModule): AbstractSerialFormat(context) {
+class ConfigParser(
+    private val configuration: ConfigParserConfiguration = ConfigParserConfiguration(),
+    override val context: SerialModule = EmptyModule
+) : SerialFormat {
     @ImplicitReflectionSerializer
     inline fun <reified T : Any> parse(conf: Config): T = parse(conf, context.getContextualOrDefault(T::class))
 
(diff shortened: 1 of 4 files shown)


formats/config/src/main/kotlin/org/jetbrains/kotlinx/serialization/config/ConfigReader.kt
```diff
@@ -76,10 +79,14 @@ class ConfigParser(context: SerialModule = EmptyModule): AbstractSerialFormat(co
...
-        override fun SerialDescriptor.getTag(index: Int): String = composeName(
-            currentTagOrNull
-                    ?: "", getElementName(index)
-        )
+        override fun SerialDescriptor.getTag(index: Int): String =
+            composeName(currentTagOrNull ?: "", getConventionElementName(index))
+
+        private fun SerialDescriptor.getConventionElementName(index: Int): String {
+            val originalName = getElementName(index)
+            return if (!configuration.useConfigNamingConvention || getElementAnnotations(index).any { it is SerialName }) originalName
```
sandwwraith: unfortunately, `SerialName` is not stored in element annotations, since it is not a `@SerialInfo` annotation. Instead, it affects descriptor directly — `getElementName(index)` returns value from `@SerialName` annotation, if present. I'm afraid there's currently no way to check whether serial name was altered, since it was made indistinguishable from property name on purpose. Probably, test regex for match would be sufficient for your use case?

EdwarDDay: Yeah, it would be enough. I thought, that it would be nice to use overall the naming convention and still could overwrite the `SerialName`. Maybe this can be achieved in the future with a `@ConfigSerialName` annotation.

pdvrieze: In my XML format I also have the function that I need to know whether SerialName was overridden or not. If there is a default way this is done/visible that would be good.

ps. as context, if the value to be serialized is an xml tag, it will normally want to use the name derived from the type that stores the value (analogous with how XML Schema/DTDs work), not from the attribute that contains it. However if the attribute was renamed, then by default the assumption would be that the user intended for that name to be used.

sandwwraith:
Thank you!

PR #707, opened 2020-02-17
Change MapEntrySerializer.descriptor to be truly map-like. Otherwise,…

qwwdfsad:
… it cannot be properly serialized by TaggedDecoder (-> to JsonObject)

### DIFF
--- a/runtime/commonTest/src/kotlinx/serialization/json/JsonTestBase.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/json/JsonTestBase.kt
@@ -7,7 +7,7 @@ package kotlinx.serialization.json
 import kotlinx.serialization.*
 import kotlinx.serialization.json.internal.*
 import kotlinx.serialization.modules.*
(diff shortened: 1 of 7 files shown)


qwwdfsad:
Contract of `MAP`/`LIST` descriptor is out of the scope of this PR for now

runtime/commonTest/src/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -135,7 +134,7 @@ abstract class JsonTestBase {
         serializer: KSerializer<T>,
         data: T,
         expected: String,
-        json: Json = unquoted
+        json: Json = Json.plain
```
sandwwraith: Maybe just `Json(.Companion)`? Since you're going to deprecate plain anyway

qwwdfsad: I'll merge with it separately

PR #711, opened 2020-02-18
Public api cleanup

qwwdfsad:


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -135,10 +136,11 @@ class ProtoBuf(
         @Suppress("UNCHECKED_CAST", "NAME_SHADOWING")
         override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) = when {
             // encode maps as collection of map entries, not merged collection of key-values
-            serializer.descriptor is MapLikeDescriptor -> {
+            serializer is MapLikeSerializer<*, *, *, *> -> {
+                // TODO it probably requires some common interface exposure
```
qwwdfsad: I will address it in a follow-up PR

runtime/commonMain/src/kotlinx/serialization/builtins/CollectionSerializers.kt
```diff
@@ -0,0 +1,60 @@
...
+
+package kotlinx.serialization.builtins
+
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
+
+/**
+ * Creates a serializer for [`List<T>`][List] for the given serializer of type [T].
+ */
+public val <T> KSerializer<T>.list: KSerializer<List<T>>
```
qwwdfsad: I'd say that (apart from `.nullable`) `.list` and `.set` extensions are neither readable nor conventional.
Now, when we have a replacement, WDYT about deprecating it?

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -158,27 +159,73 @@ public constructor(
...
     public inline fun <reified T : Any> fromJson(tree: JsonElement): T = fromJson(context.getContextualOrDefault(T::class), tree)
 
-    companion object : StringFormat {
-        @Suppress("DEPRECATION")
+    /**
+     * The default instance of [Json] in the form of companion object.
+     */
+    @UnstableDefault
+    public companion object Default: StringFormat {
+        private const val message =
```
qwwdfsad: The rationale behind `: StringFormat` is pretty simple -- it may be convenient to use it with some common methods.
But default instance is not exposed to the public deliberately: if one has to pass `Json` instance somewhere, they should create it manually

formats/properties/commonMain/src/kotlinx/serialization/Properties.kt
```diff
@@ -217,8 +217,7 @@ public class Properties(context: SerialModule = EmptyModule) : AbstractSerialFor
 
     public companion object {
         /**
-         * A [Properties] instance which
-         * does not have any [SerialModule]s installed.
+         * A [Properties] instance which does not have any [SerialModule] installed.
          */
         public val DEFAULT: Properties = Properties()
```
qwwdfsad: The default instance is left deliberately: Properties have (in fact) no configuration and not state, neither it has common denominator like `StringFormat`

runtime/commonMain/src/kotlinx/serialization/builtins/CollectionSerializers.kt
```diff
@@ -0,0 +1,60 @@
...
+    get() = ArrayListSerializer(this)
+
+/**
+ * Creates a serializer for [`List<T>`][List] for the given serializer of type [T].
+ */
+public fun <T> ListSerializer(elementSerializer: KSerializer<T>): KSerializer<List<T>> =
```
whyoleg: Is such shorthand needed, when anyone can just use `serializer<List<T>>()` and it will produce the the same result, no?
What is the purpose of it?

qwwdfsad: It can be used with custom serializer, for example, or with polymorphic/contextual one.
Another reason can be `typeOf` avoidance -- one may want to avoid experimental stdlib API or `typeOf` specifically for whatever reason

whyoleg: oh, I added comment to wrong line, they are too similar...
I want to add it to next function
`public inline fun <reified T> ListSerializer(): KSerializer<List<T>>`
The same question :)

qwwdfsad: Just a handy and a bit more lightweight alias

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonOutput.kt
```diff
@@ -5,13 +5,14 @@
...
 import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
 import kotlinx.serialization.json.*
 import kotlinx.serialization.modules.SerialModule
 import kotlin.jvm.JvmField
 
 
 internal class StreamingJsonOutput(private val composer: Composer, override val json: Json, private val mode: WriteMode,
-                                   private val modeReuseCache: Array<JsonOutput?>) : JsonOutput, ElementValueEncoder() {
+                                   private val modeReuseCache: Array<JsonOutput?>) : JsonOutput, AbstractEncoder() {
```
sandwwraith: formatting

runtime/commonTest/src/kotlinx/serialization/builtins/KeyValueSerializersTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        run {
+            val serializer = TripleSerializer<A, B, C>()
+            val json = default.stringify(serializer, tripleInstance, useStreaming)
+            assertEquals(expectedJson, json)
+            val triple = default.parse(serializer, json, useStreaming)
+            assertEquals(tripleInstance, triple)
+        }
+    }
+
+    class Entry<K, V>(override val key: K, override val value: V) : Map.Entry<K, V> {
```
sandwwraith: `MapEntrySerializer` has its own `Map.Entry` impl. Probably you may reuse it. (also, that impl should be comparable with Map.Entry, so fix this please if you don't mind)

runtime/commonTest/src/kotlinx/serialization/TuplesTest.kt
```diff
@@ -22,16 +23,19 @@ class TuplesTest {
...
+        Json { unquoted = true }
     )
 
     @Test
     fun testStandardPair() = assertStringFormAndRestored(
         "{p:{first:42,second:foo}}",
         PairWrapper(42 to "foo"),
         PairWrapper.serializer(),
-        Json.unquoted
+        Json { unquoted = true }
```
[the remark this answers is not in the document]
sandwwraith: Since unquoted is widely used in our tests, you can do the following:

```kotlin
private val testUnquoted = Json { unquoted = true }

val Json.Default.unquoted: Json get() = testUnquoted
```

Btw, `Default` as a name for companion looks really ugly in this context. Maybe we don't have to rename them?

runtime/commonTest/src/kotlinx/serialization/json/JsonTransformingSerializerTest.kt
```diff
@@ -18,7 +18,7 @@ class JsonTransformingSerializerTest : JsonTestBase() {
     )
 
     object WrappingJsonListSerializer :
-        JsonTransformingSerializer<List<StringData>>(StringData.serializer().list, "WrappingList") {
+        JsonTransformingSerializer<List<StringData>>(serializer(), "WrappingList") {
```
sandwwraith: Not sure that it is a good idea until `typeOf` fully supported in JS

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
qwwdfsad: We can't, because we have to retrieve key and value serializers.
My plan is to add a simple interface and recommend using it, but not as part of this PR.

formats/properties/commonMain/src/kotlinx/serialization/Properties.kt
qwwdfsad: Ok, works for me

benchmark/src/jmh/kotlin/kotlinx/benchmarks/ProtoBaseline.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.openjdk.jmh.annotations.*
+import java.util.concurrent.*
+
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+@Fork(1)
+open class ProtoBaseline {
```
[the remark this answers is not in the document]
qwwdfsad: Part of the `findAnnotation` replacement, I've found myself writing this one second or third time

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
```diff
@@ -17,15 +17,18 @@ public annotation class ProtoType(val type: ProtoNumberType)
...
 
-
 public class ProtobufDecodingException(message: String) : SerializationException(message)
 
-@Suppress("DEPRECATION_ERROR")
-internal fun getProtoId(desc: SerialDescriptor, index: Int): Int?
-        = desc.findAnnotation<ProtoId>(index)?.id ?: desc.findAnnotation<SerialId>(index)?.id
+internal inline fun <reified A: Annotation> SerialDescriptor.findAnnotation(elementIndex: Int): A? {
+    for (a in getElementAnnotations(elementIndex)) {
+        if (a is A) return a
```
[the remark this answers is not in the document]
qwwdfsad: Well, it is broken in binary format and `SerialId` is completely deprecated with replacement, not sure it's worth to keep it in the code.
Returned it to ease the pain of the migration

runtime/commonTest/src/kotlinx/serialization/builtins/KeyValueSerializersTest.kt
qwwdfsad: It's private and I prefer to keep it that way.

>also, that impl should be comparable with Map.Entry

This is impossible because `equals` should be symmetric. Fun fact: `assertEquals(a, b)` works differently on different platforms (one invokes `a == b` and another one `b == a``)

runtime/commonTest/src/kotlinx/serialization/json/JsonTransformingSerializerTest.kt
qwwdfsad: It is a test that works, not sure what did you mean

runtime/commonMain/src/kotlinx/serialization/Shorthands.kt
```diff
@@ -1,70 +0,0 @@
...
-}
-
-/**
- * Searches for annotation of type [A] in annotations, obtained via
- * [SerialDescriptor.getElementAnnotations] at given [elementIndex]
- *
- * Returns null if there are no annotations with such type.
- * Throws [IllegalStateException] if there are duplicated annotations for a given type.
- */
-public inline fun <reified A: Annotation> SerialDescriptor.findAnnotation(elementIndex: Int): A? {
```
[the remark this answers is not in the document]
qwwdfsad: Yes I do

runtime/commonMain/src/kotlinx/serialization/builtins/CollectionSerializers.kt
```diff
@@ -0,0 +1,60 @@
...
+ * Reified version of [SetSerializer]
+ */
+@ImplicitReflectionSerializer
+public inline fun <reified T> SetSerializer(): KSerializer<Set<T>> = SetSerializer(serializer())
+
+/**
+ * Creates a serializer for [`Map<K, V>`][Map] for the given serializers for
+ * its ket type [K] and value type [V].
+ */
+public fun <K, V> MapSerializer(
```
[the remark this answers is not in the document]
qwwdfsad: File an issue so we could provide `HashMapSerializer`

runtime/commonMain/src/kotlinx/serialization/builtins/PrimitiveSerializers.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+@file:Suppress("DEPRECATION_ERROR")
+
+package kotlinx.serialization.builtins
+
+import kotlinx.serialization.*
+
+/**
+ * Returns serializer for [String] with descriptor of [PrimitiveKind.STRING] kind.
```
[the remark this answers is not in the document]
qwwdfsad: It allows people to explore documentation further by following links

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -228,9 +230,12 @@ class ArrayListSerializer<E>(element: KSerializer<E>) : ListLikeSerializer<E, Li
     override fun ArrayList<E>.insert(index: Int, element: E) { add(index, element) }
 }
 
-class LinkedHashSetSerializer<E>(eSerializer: KSerializer<E>) : ListLikeSerializer<E, Set<E>, LinkedHashSet<E>>(eSerializer) {
-    override val descriptor = LinkedHashSetClassDesc(eSerializer.descriptor)
+@InternalSerializationApi // Use kotlinx.serialization.builtins.SetSerializer() instead
+internal class LinkedHashSetSerializer<E>(
```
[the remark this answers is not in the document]
qwwdfsad: We don't have `FRAMEWORK` tests apparently, so reverted this

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -159,27 +159,73 @@ public constructor(
...
-    companion object : StringFormat {
-        @Suppress("DEPRECATION")
+    /**
+     * The default instance of [Json] in the form of companion object.
+     */
+    @UnstableDefault
+    public companion object Default : StringFormat {
+        private const val message =
+            "Top-level JSON instances are deprecated for removal in the favour of user-configured one. " +
+                    "You can either use a Json top-level object or configure your own instance  via 'Json {}' builder-like constructor"
```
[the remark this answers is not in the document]
qwwdfsad: Nothing

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
```diff
@@ -143,18 +144,18 @@ public object JsonObjectSerializer : KSerializer<JsonObject> {
...
             "kotlinx.serialization.json.JsonObject",
-            StringSerializer.descriptor,
+            String.serializer().descriptor,
             JsonElementSerializer.descriptor
         )
 
     override fun serialize(encoder: Encoder, value: JsonObject) {
         verify(encoder)
-        LinkedHashMapSerializer(StringSerializer, JsonElementSerializer).serialize(encoder, value.content)
+        LinkedHashMapSerializer(String.serializer(), JsonElementSerializer).serialize(encoder, value.content)
```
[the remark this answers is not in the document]
qwwdfsad: What for?

runtime/commonTest/src/kotlinx/serialization/TuplesTest.kt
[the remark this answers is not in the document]
qwwdfsad: `Default` is named so for the same reasons it is done in `kotlin.Random`:
  -- Documentation is *much* prettier on the site
  --  Stacktraces are better

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -228,9 +230,12 @@ class ArrayListSerializer<E>(element: KSerializer<E>) : ListLikeSerializer<E, Li
     override fun ArrayList<E>.insert(index: Int, element: E) { add(index, element) }
 }
 
-class LinkedHashSetSerializer<E>(eSerializer: KSerializer<E>) : ListLikeSerializer<E, Set<E>, LinkedHashSet<E>>(eSerializer) {
-    override val descriptor = LinkedHashSetClassDesc(eSerializer.descriptor)
+@InternalSerializationApi // Use kotlinx.serialization.builtins.SetSerializer() instead
```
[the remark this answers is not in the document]
qwwdfsad: Only as the next step in a follow-up PR.
We have a lot of cross-module usages of these classes and I don't really want to suppress deprecation error

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
```diff
@@ -94,7 +102,8 @@ public class PairSerializer<K, V>(
 }
 
 
-// todo: move from internal package and add documentation
+@InternalSerializationApi
+@Deprecated(message = deprecationMessage, level = DeprecationLevel.HIDDEN)
```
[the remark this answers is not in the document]
qwwdfsad: It will, and it will be fixed with `alt + enter -> import`.
Unfortunately, it is not possible to deprecate it with `ERROR`: `LowPriorityInOverloadResolution` is inapplicable for classes and constructors, so deprecated symbol always wins in resolution if you have two import.

runtime/commonMain/src/kotlinx/serialization/builtins/PrimitiveSerializers.kt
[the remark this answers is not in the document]
sandwwraith: Then it makes sense to reference `SerialDescriptor` on 'descriptor' word

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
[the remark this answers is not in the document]
sandwwraith: You can start with warning

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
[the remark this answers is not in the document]
sandwwraith: I can see that this is an 'unresolved reference' until you explicitly have `import kotlinx.serialization.internal.TripleSerializer` (e.g. alt+enter would not work if one has a star import of internal)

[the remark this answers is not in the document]
sandwwraith: If we move out everything from `internal`, users won't need that star import anymore; therefore, it's nothing wrong in overload resolution

runtime/commonTest/src/kotlinx/serialization/builtins/KeyValueSerializersTest.kt
sandwwraith: So you've overrided `equals` here only for testing purposes?

qwwdfsad: yes

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -0,0 +1,73 @@
...
+    message = "Deprecated in the favour of the same extension from builtins package",
+    replaceWith = ReplaceWith("set", imports = ["kotlinx.serialization.builtins.set"])
+)
+@LowPriorityInOverloadResolution
+val <T> KSerializer<T>.set: KSerializer<Set<T>>
+    get() = LinkedHashSetSerializer(this)
+
+@Deprecated(
+    level = DeprecationLevel.ERROR, message = "Deprecated in the favour of the same extension from builtins package",
+    replaceWith = ReplaceWith("MapSerializer(this.first, this.second)", imports = ["kotlinx.serialization.builtins.map"])
```
sandwwraith:
```suggestion
    replaceWith = ReplaceWith("map", imports = ["kotlinx.serialization.builtins.map"])
```

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -200,7 +200,35 @@ public fun SerialDescriptorBuilder.setDescriptor(typeDescriptor: SerialDescripto
...
-    return HashSetClassDesc(serializer<T>().descriptor)
+    return setDescriptor(serializer<T>().descriptor)
+}
+
+/**
+ * Creates a [List] out of a child descriptors retrieved via [SerialDescriptor.getElementDescriptor].
+ *
+ * Size of a list is equal to [SerialDescriptor.elementsCount].
+ */
+public fun SerialDescriptor.elementDescriptors(): List<SerialDescriptor> {
```
sandwwraith: Not sure why these functions should belong to this file

runtime/commonMain/src/kotlinx/serialization/builtins/AbstractDecoder.kt
```diff
@@ -0,0 +1,73 @@
...
+
+package kotlinx.serialization.builtins
+
+import kotlinx.serialization.*
+import kotlinx.serialization.modules.*
+
+/**
+ * A skeleton implementation of both [Decoder] and [CompositeDecoder] that can be used
+ * for simple formats and for testability purpose.
+ * Most of the `decode*` methods have default implementation that delegates `decodeValue(value: Any) as TargetType`.
```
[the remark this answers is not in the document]
sandwwraith: At least add 'see `Decoder` documentation for information about particular methods'

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
sandwwraith: Otherwise, 'same extension' line and imports are incorrect.

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -0,0 +1,73 @@
...
+)
+val <K, V> Pair<KSerializer<K>, KSerializer<V>>.map: KSerializer<Map<K, V>>
```
sandwwraith: Btw, you'll need to update https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/runtime_usage.md#obtaining-serializers section.

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -268,10 +278,12 @@ class LinkedHashMapSerializer<K, V>(kSerializer: KSerializer<K>, vSerializer: KS
     override fun LinkedHashMap<K, V>.insertKeyValuePair(index: Int, key: K, value: V) = set(key, value)
 }
 
-class HashMapSerializer<K, V>(kSerializer: KSerializer<K>, vSerializer: KSerializer<V>) :
-    MapLikeSerializer<K, V, Map<K, V>, HashMap<K, V>>(kSerializer, vSerializer) {
-    override val descriptor = HashMapClassDesc(kSerializer.descriptor, vSerializer.descriptor)
+@InternalSerializationApi // Use kotlinx.serialization.builtins.MapSerializer() instead
+class HashMapSerializer<K, V>(
```
sandwwraith: Add `public` everywhere (in this or following PR)

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
```diff
@@ -78,7 +85,8 @@ public class MapEntrySerializer<K, V>(
     override fun toResult(key: K, value: V): Map.Entry<K, V> = MapEntry(key, value)
 }
 
-// todo: move from internal package and add documentation
+@InternalSerializationApi
+@Deprecated(message = deprecationMessage, level = DeprecationLevel.HIDDEN)
```
sandwwraith: Forgot `ReplaceWith`?

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -159,11 +159,21 @@ public constructor(
...
-        @Suppress("DEPRECATION")
+    /**
+     * The default instance of [Json] in the form of companion object.
+     */
+    @UnstableDefault
+    public companion object Default : StringFormat {
+        private const val message =
+            "Top-level JSON instances are deprecated for removal in the favour of user-configured one. " +
+                    "You can either use a Json top-level object, configure your own instance  via 'Json {}' builder-like constructor, " +
+                    "'Json(JsonConfiguration)' constructor or by tweaking stable configuration 'Json(JsonConfiguration.Stable.copy(prettyPrint = true))'"
```
sandwwraith: `Stable` configuration should [probably] be deprecated and/or renamed to `PreRelease`, because once we're 1.0, `Default` should be stable and preferred for new users.

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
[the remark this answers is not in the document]
sandwwraith: Because you've replaced `ArrayListSerializer` with `ListSerializer`

runtime/commonTest/src/kotlinx/serialization/builtins/KeyValueSerializersTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        override fun hashCode(): Int {
+            var result = key?.hashCode() ?: 0
+            result = 31 * result + (value?.hashCode() ?: 0)
+            return result
+        }
+    }
+
+//    @Test // ignore it until bug in JsonParser is fixed
```
sandwwraith: it's better to create corresponding issue and mention that issue

runtime/jvmTest/src/kotlinx/serialization/features/InternalInheritanceTest.kt
```diff
@@ -56,66 +56,65 @@ class InternalInheritanceTest {
     fun testStringify() {
         assertEquals(
                 "{parent:42,rootOptional:rootOptional,parent2:42,derived:derived,bodyDerived:body,parent3:42,lastDerived:optional}",
-                Json.unquoted.stringify(C(42))
+                Json { unquotedPrint = true }.stringify(C(42))
```
sandwwraith: `unquotedLenient` fit here, too

runtime/jvmTest/src/kotlinx/serialization/features/ResolvingTest.kt
```diff
@@ -70,7 +71,7 @@ class ResolvingTest {
             override fun getActualTypeArguments(): Array<Type> = arrayOf(String::class.java)
         }
         val serial = serializerByTypeToken(token)
-        val s = Json.unquoted.stringify(serial, myArr)
+        val s = Json { unquotedPrint = true }.stringify(serial, myArr)
```
sandwwraith: extract `Json` instance to field?

runtime/commonMain/src/kotlinx/serialization/Shorthands.kt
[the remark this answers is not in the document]
pdvrieze: While I wasn't aware of this one I have implemented exactly the same function

runtime/commonMain/src/kotlinx/serialization/builtins/AbstractDecoder.kt
```diff
@@ -0,0 +1,73 @@
...
+ */
+public abstract class AbstractDecoder : Decoder, CompositeDecoder {
```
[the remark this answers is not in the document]
pdvrieze: More significantly, merging both can be confusing to implement. Perhaps interface default methods would be the best solution.

runtime/commonMain/src/kotlinx/serialization/builtins/CollectionSerializers.kt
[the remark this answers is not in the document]
pdvrieze: At least make sure that it is valid to write your own version that decodes to your own? BloomFilterMap?

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
qwwdfsad: I'd rather deprecate it without replacement.
I was just tempted to remove `Shorthands.kt` at all, this file was the most relevant one

runtime/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
qwwdfsad: In the follow up with interfaces

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
qwwdfsad: This one is hidden

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
qwwdfsad: Let's fix it separately

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -0,0 +1,73 @@
...
+    get() {
+        @Suppress("UNCHECKED_CAST")
+        return if (descriptor.isNullable) (this as KSerializer<T?>) else NullableSerializer(this)
+    }
+
+@Deprecated(level = DeprecationLevel.ERROR,
+    message = "Deprecated in the favour of the same extension from builtins package",
+    replaceWith = ReplaceWith("list", imports = ["kotlinx.serialization.builtins.list"]))
+@LowPriorityInOverloadResolution
+val <T> KSerializer<T>.list: KSerializer<List<T>>
```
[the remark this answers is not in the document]
qwwdfsad: I am still not sure, so let's say it's postponed

runtime/commonMain/src/kotlinx/serialization/internal/Tuples.kt
sandwwraith: It still be marked as deprecated on fq-name import

docs/runtime_usage.md
```diff
@@ -29,16 +30,17 @@ val dataSerializer: KSerializer<Data> = Data.serializer()
...
 ```
 
-Built-in types, like `Int`, and standard collections doesn't have that method. You can use corresponding serializers from the library:
+Built-in types, like `Int`, and standard collections doesn't have that method and instead are available via 
+factories and companion extensions:
 ```kotlin
-val intSerializer: KSerializer<Int>  = IntSerializer // object
-val intListSerializer: KSerializer<List<Int>> = ArrayListSerializer(IntSerializer) // generic, requires instantiation
+val intSerializer: KSerializer<Int>  = Int.serializer()
+val intListSerializer: KSerializer<List<Int>> = ListSerializer(IntSerializer) // generic, requires instantiation
```
sandwwraith:
```suggestion
val intListSerializer: KSerializer<List<Int>> = ListSerializer(Int.serializer()) // generic, requires instantiation
```

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -371,19 +372,18 @@ public class Cbor(
         private const val HEADER_ARRAY: Int = 0b100_00000
         private const val HEADER_MAP: Int = 0b101_00000
 
+        @Deprecated(
+            message = "Deprecated for removal in the favour of user-defined instances or Cbor companion object",
+            level = DeprecationLevel.ERROR,
+            replaceWith = ReplaceWith("Cbor")
```
sandwwraith: Wouldn't application of such replacement result in `Cbor.plain` -> `Cbor.Cbor`?

runtime/commonMain/src/kotlinx/serialization/builtins/AbstractEncoder.kt
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public open fun encodeValue(value: Any): Unit =
+        throw SerializationException("Non-serializable ${value::class} is not supported by ${this::class} encoder")
+
+    override fun encodeNull() {
+        throw SerializationException("'null' is not supported by default")
+    }
+
+    override fun encodeUnit() {
+        val unitDescriptor = UnitSerializer().descriptor
```
sandwwraith: `UnitSerializer().serialize(this)`

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
```diff
@@ -260,8 +286,13 @@ public class FloatArrayBuilder internal constructor(
  * Encode elements one-by-one, as regular list,
  * unless format's Encoder/Decoder have special handling for this serializer.
  */
+@Deprecated(
+    message = "Deprecated in the favour of DoubleArraySerializer() factory",
+    level = DeprecationLevel.ERROR,
```
sandwwraith: would it make sense to deprecate `xArrayBuilder` with `HIDDEN` as well?

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
sandwwraith: But they are literally shorthands. And file itself was not doing any harm at all

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
qwwdfsad: It won't, I've checked it before the commit (and actually replaced it in tests)

runtime/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
qwwdfsad: Yeah, works for me

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
qwwdfsad: I'd like to avoid "util" files in the core package similarly to how Java doesn't have `FooUtil` in the JDK.
Such files don't help with API exploration and just look like a pile of unrelated methods

sandwwraith: In contrary to Java, file names are not exposed in Kotlin neither used as namespaces.

qwwdfsad: I mean, the problem is with the semantics of that, not with the filename.
Java doesn't have "semantically *Util" as well

PR #713, opened 2020-02-18
Arrays support

### LINKED ISSUE #710: Support array serializers in KType-fucntions familiy

qwwdfsad:
Fixes #710

### DIFF
--- a/runtime/jsMain/src/kotlinx/serialization/Serialization.kt
+++ b/runtime/jsMain/src/kotlinx/serialization/Serialization.kt
@@ -40,3 +40,20 @@ internal actual fun <T : Any> KClass<T>.simpleName(): String? = simpleName
 internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vararg args: KSerializer<Any?>): KSerializer<T>? {
     throw NotImplementedError("This method is not supported for Kotlin/JS yet. Please provide serializer explicitly.")
 }
+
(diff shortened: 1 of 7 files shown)


runtime/jsMain/src/kotlinx/serialization/Serialization.kt
```diff
@@ -40,3 +40,20 @@ internal actual fun <T : Any> KClass<T>.simpleName(): String? = simpleName
 internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vararg args: KSerializer<Any?>): KSerializer<T>? {
     throw NotImplementedError("This method is not supported for Kotlin/JS yet. Please provide serializer explicitly.")
 }
+
+internal actual fun isReferenceArray(type: KType, rootClass: KClass<Any>): Boolean {
+    val typeParameters = type.arguments
+    if (typeParameters.size != 1) return false
+    val parameter = typeParameters.single()
+    // Fun fact -- star projections pass this check
+    val variance = parameter.variance ?: error("Star projections are forbidden: $type")
```
qwwdfsad: This one is fixed in master (-> 1.4).
It is also possible to compare against `Array::class` in 1.4, so this hack is temporary

sandwwraith: better add it as a code comment then, with smth like `//fixme: remove in 1.4`

PR #714, opened 2020-02-18
Cleanup ConfigParser

sandwwraith:
move to proper package to be consistent with other formats,
add public modifiers, update copyright


sandwwraith:
I'll revise documentation altogether with Cbor and Protobuf

PR #715, opened 2020-02-19
Document contracts for JsonInput/JsonOutput

### LINKED ISSUE #624: Add more restrictions to JsonInput/JsonOutput
Currently, it is possible to call (public and "stable") `decodeJson` in an arbitrary place during serialization. The problem is that it is actually not possible to call this method in an arbitrary place of code.

We should carefully document, test and protect against such behaviour

qwwdfsad:
No code changes are required, because it is impossible to make Streaming* and Tree* versions completely consistent in terms of observable behaviour, so instead of deterministic mode, only Streaming* version throws an exception during contracts violation

Fixes #624

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
@@ -61,7 +61,7 @@ public interface JsonOutput: Encoder, CompositeEncoder {
      * fun serialize(encoder: Encoder, value: Holder) {
      *     // Completely okay, the whole Holder object is read
      *     val jsonObject = JsonObject(...) // build a JsonObject from Holder
-     *     (encoder asJsonOutput).encodeJson(jsonObject) // Write it
+     *     (encoder as JsonOutput).encodeJson(jsonObject) // Write it
      * }
      *
      * // Incorrect Holder serialize method
--- a/runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
@@ -61,7 +61,7 @@ public interface JsonOutput: Encoder, CompositeEncoder {
      * fun serialize(encoder: Encoder, value: Holder) {
      *     // Completely okay, the whole Holder object is read
      *     val jsonObject = JsonObject(...) // build a JsonObject from Holder
-     *     encoder.asJsonInput().encodeJson(jsonObject) // Write it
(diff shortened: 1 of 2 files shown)


runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
```diff
@@ -47,7 +49,32 @@ public interface JsonOutput: Encoder, CompositeEncoder {
...
+     * and may produce an invalid JSON result.
+     * For example:
+     * ```
+     * class Holder(val value: Int, val list: List<Int>())
+     *
+     * // Holder serialize method
+     * fun serialize(encoder: Encoder, value: Holder) {
+     *     // Completely okay, the whole Holder object is read
+     *     val jsonObject = JsonObject(...) // build a JsonObject from Holder
+     *     encoder.asJsonInput().encodeJson(jsonObject) // Write it
```
sandwwraith: `asJsonOutput`

Btw, these functions are internal, so maybe we shouldn't give an example which does not compile?

sandwwraith:
Also, `StreamingJson` implementation currently does not throw any exceptions, AFAIR

runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
qwwdfsad: Well, they matched accidentally. For the sake of simplicity,  I just used it as a pseudo-code example. Fixed it to avoid confusion

sandwwraith:
```suggestion
     *     (encoder asJsonOutput).encodeJson(jsonObject) // Write it
```

runtime/commonMain/src/kotlinx/serialization/json/JsonOutput.kt
```diff
@@ -47,7 +49,32 @@ public interface JsonOutput: Encoder, CompositeEncoder {
...
+     * and may produce an invalid JSON result.
+     * For example:
+     * ```
+     * class Holder(val value: Int, val list: List<Int>())
+     *
+     * // Holder serialize method
+     * fun serialize(encoder: Encoder, value: Holder) {
+     *     // Completely okay, the whole Holder object is read
+     *     val jsonObject = JsonObject(...) // build a JsonObject from Holder
+     *     (encoder asJsonOutput).encodeJson(jsonObject) // Write it
```
sandwwraith:
```suggestion
     *     (encoder as JsonOutput).encodeJson(jsonObject) // Write it
```

PR #718, opened 2020-02-20
Json literals

qwwdfsad:


runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -7,7 +7,7 @@ package kotlinx.serialization.json.internal
 import kotlinx.serialization.json.*
 import kotlinx.serialization.json.internal.EscapeCharMappings.ESCAPE_2_CHAR
 import kotlin.jvm.*
-import kotlin.native.concurrent.SharedImmutable
+import kotlin.native.concurrent.*
```
sandwwraith: Side note: maybe we should rename `JsonReader` to `JsonLexer` because `JsonReader` can be confused with existent `JsonParser`

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -139,7 +139,27 @@ internal class JsonReader(private val source: String) {
     }
 
     fun takeString(): String {
-        if (tokenClass != TC_OTHER && tokenClass != TC_STRING) fail("Expected string or non-null literal", tokenPosition)
+        if (tokenClass != TC_OTHER && tokenClass != TC_STRING) fail(
+            "Expected string or non-null literal. $lenientHint",
```
sandwwraith: `takeString()` function is mainly called with `isLenient = true`, so '$lenientHint' is kinda misleading here.

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -269,14 +289,16 @@ internal class JsonReader(private val source: String) {
                 TC_END_LIST -> {
                     if (tokenStack.last() != TC_BEGIN_LIST) throw JsonDecodingException(
                         currentPosition,
-                        "found ] instead of }"
+                        "found ] instead of }",
+                        source
```
sandwwraith: Same concern about printing whole source

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -142,15 +149,39 @@ internal class StreamingJsonInput internal constructor(
...
-    override fun decodeBoolean(): Boolean = reader.takeString().run { if (configuration.strictMode) toBooleanStrict() else toBoolean() }
+    override fun decodeBoolean(): Boolean {
+        /*
+         * We prohibit non true/false boolean literals at all as it is considered way too error-prone,
+         * but allow quoted literal in relaxed mode for booleans.
+         */
+        return if (configuration.isLenient) {
+            reader.takeString().toBooleanStrict()
+        } else {
+            val string = reader.takeBooleanStringUnquoted()
```
sandwwraith: you can return `reader.takeBooleanStringUnquoted().toBooleanStrict()` right away without `val string`

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -89,17 +89,36 @@ private sealed class AbstractJsonTreeInput(
         return
     }
 
-    override fun decodeTaggedBoolean(tag: String): Boolean = getValue(tag).boolean
+    override fun decodeTaggedBoolean(tag: String): Boolean {
+        val value = getValue(tag)
+        if (!json.configuration.isLenient) {
+            val literal = value as JsonLiteral
+            if (literal.isString) throw JsonException(
+                "Boolean literal for ley '$tag' should be unquoted. $lenientHint"
```
sandwwraith:
```suggestion
                "Boolean literal for key '$tag' should be unquoted. $lenientHint"
```

runtime/commonTest/src/kotlinx/serialization/features/PrimitiveArraySerializersTest.kt
```diff
@@ -9,59 +9,55 @@ import kotlinx.serialization.json.Json
...
-class PrimitiveArraySerializersTest {
     @Test
     fun testCanBeSerialized() = assertStringFormAndRestored(
         """{"arr":[1,2,3],"arr2":[1,2],"arr3":[true,false],"arr4":["a","b","c"],"arr5":[NaN,0.1,-0.25],"arr6":[1,2,3],"arr7":[1,2,3],"arr8":[1.25,2.25,3.25]}""",
         A(byteArrayOf(1, 2, 3)),
         A.serializer(),
-        Json.nonstrict,
-        printResult = false
-    )
+        Json.nonstrict)
```
sandwwraith: closing brace on newline

runtime/commonTest/src/kotlinx/serialization/json/JsonMapKeysTest.kt
```diff
@@ -37,19 +37,19 @@ class JsonMapKeysTest : JsonTestBase() {
                 streaming
             )
         }
-        assertTrue(e.message?.contains("can't be used in json as map key") == true, "wrong exception text")
+        assertTrue(e.message?.contains("can't be used in JSON as a key in the map") == true)
```
sandwwraith: Why you've erased assertion message?

runtime/commonTest/src/kotlinx/serialization/json/LenientTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    fun testUnquotedKey() = parametrizedTest {
+        val json = """{"i":1, "l":2, b:true, "s":"string"}"""
+        assertFailsWith<JsonDecodingException> { strict.parse(Holder.serializer(), json, it) }
+        assertEquals(value, lenient.parse(Holder.serializer(), json, it))
+    }
+
+    @Test
+    fun testUnquotedArray() = parametrizedTest {
```
sandwwraith: maybe `testUnquotedStringInArray`? because arrays per se should be unquoted in json

runtime/commonTest/src/kotlinx/serialization/json/LenientTest.kt
```diff
@@ -0,0 +1,68 @@
...
+    fun testQuotedLong() = parametrizedTest {
+        val json = """{"i":1, "l":"2", "b":true, "s":"string"}"""
+        assertEquals(value, strict.parse(Holder.serializer(), json, it))
+        assertEquals(value, lenient.parse(Holder.serializer(), json, it))
+    }
+
+    @Test
+    fun testQuotedBoolean() = parametrizedTest {
+        val json = """{"i":1, "l":2, "b":"true", "s":"string"}"""
+        assertFailsWith<JsonException> { strict.parse(Holder.serializer(), json, it) }
```
sandwwraith: Shouldn't it be `assertFailsWith<JsonDecodingException>` in all these tests?

integration-test/src/commonMain/kotlin/sample/Helpers.kt
```diff
@@ -1,138 +0,0 @@
-/*
- * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
- */
-
-package sample
-
-import kotlinx.serialization.*
-
-class KeyValueOutput(val sb: StringBuilder) : ElementValueEncoder() {
```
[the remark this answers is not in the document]
qwwdfsad: Nope, they are just unused

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -6,37 +6,62 @@ package kotlinx.serialization.json
...
-    @JvmField internal val unquoted: Boolean = false,
-    @JvmField internal val allowStructuredMapKeys: Boolean = false,
-    @JvmField internal val prettyPrint: Boolean = false,
-    @JvmField internal val indent: String = defaultIndent,
-    @JvmField internal val useArrayPolymorphism: Boolean = false,
-    @JvmField internal val classDiscriminator: String = defaultDiscriminator,
+    internal val encodeDefaults: Boolean = true,
+    internal val ignoreUnknownKeys: Boolean = false,
+    internal val isLenient: Boolean = false,
+    private val serializeSpecialFloatingPointValues: Boolean = false,
```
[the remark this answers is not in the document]
qwwdfsad: Well, it's clear and properly reflects what it does. Bonus points, it's not expected to be used frequently. Fixed `isLenient`

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
```diff
@@ -124,13 +127,16 @@ public object JsonLiteralSerializer : KSerializer<JsonLiteral> {
...
         }
-
         encoder.encodeString(value.content)
     }
 
     override fun deserialize(decoder: Decoder): JsonLiteral {
         verify(decoder)
-        return JsonLiteral(decoder.decodeString())
+        val result = (decoder as JsonInput).decodeJson()
+        if (result !is JsonLiteral) {
```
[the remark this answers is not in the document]
qwwdfsad: I'd fix 704 separately in the follow-up commit

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: well, it's internal, so shouldn't be a big problem to keep it that way or to rename.
I'm fine with both

runtime/commonTest/src/kotlinx/serialization/json/JsonMapKeysTest.kt
qwwdfsad: Because it creates visual noise and is meaningless anyway

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -6,37 +6,60 @@ package kotlinx.serialization.json
...
  *
  * Options list:
  * * [encodeDefaults] specifies whether default values are encoded.
- * * [strictMode] enables strict mode, which prohibits unknown keys in the JSON and non-numerical values in floating point numbers.
- * * [unquoted] specifies whether keys and values should be quoted.
+ *
+ * * [ignoreUnknownKeys] ignores encounters of unknown properties in the input JSON.
+ *
+ * * [isLenient] removes JSON specification restriction (RFC-4627) and makes parser
+ *   more liberal to the malformed input. In lenient modes quoted string literals,
```
sandwwraith:
```suggestion
 *   more liberal to the malformed input. In lenient mode quoted string literals,
```

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -6,37 +6,60 @@ package kotlinx.serialization.json
...
+ *   and unquoted string literals are allowed.
+ *
+ * * [serializeSpecialFloatingPointValues] removes JSON specification restriction on
+ *   special floating-point values such as `NaN` and `Infinity` and enables their
+ *   serialization. When enabling it, please assure that the receiving party will be
```
sandwwraith: Maybe `ensure` instead of `assure`?

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -72,13 +96,24 @@ public data class JsonConfiguration @UnstableDefault constructor(
...
         )
     }
 }
+
+@Deprecated(level = DeprecationLevel.ERROR, message = "This named parameters were deprecated." +
+        "Intstead of 'unquoted' please use 'unquotedPrint'. 'strictMode' was splitted into" +
+        "'ignoreUnknownKeys', 'isLenient' and 'serializeSpecialFloatingPointValues'")
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+@kotlin.internal.LowPriorityInOverloadResolution
+fun JsonConfiguration(strictMode: Boolean = true, unquoted: Boolean = false) {
```
sandwwraith: `public`

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
```diff
@@ -182,9 +188,13 @@ public object JsonArraySerializer : KSerializer<JsonArray> {
 }
 
 private fun verify(encoder: Encoder) {
-    if (encoder !is JsonOutput) error("Json element serializer can be used only by Json format")
+    require(encoder is JsonOutput) {
+        "Json element serializer can be used only by Json format, had $encoder"
```
sandwwraith: `encoder::class` may be slightly better

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -7,21 +7,57 @@ package kotlinx.serialization.json
 import kotlinx.serialization.*
 
 /**
- * Generic exception indicating a problem with JSON operations.
+ * Generic exception indicating a problem with JSON serialization.
```
sandwwraith:
```suggestion
 * Generic exception indicating a problem with JSON serialization and deserialization.
```

sandwwraith: Btw, this is not _100%_ true since this exception is also thrown when `JsonElement` mismatches expected type (e.g. `element.jsonObject` call when `element is JsonArray`)

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -7,21 +7,57 @@ package kotlinx.serialization.json
...
+
+internal fun InvalidFloatingPoint(value: Number, key: String, type: String, output: String) = JsonEncodingException(
+    "'$value' with key '$key' is not a valid $type as per JSON specification. " +
+            "You can enable 'serializeSpecialFloatingPointValues' property to serialize such values.\n" +
+            "Current output: ${output.minify()}"
+)
+
+internal fun UnknownKeyException(key: String, input: String) = JsonDecodingException(
+    -1,
+    "JSON encountered unknown key: '$key'. You can enable 'ignoreUnknownKeys' property to ignore unknown keys. JSON input: $input")
```
sandwwraith: `input.minify()`?

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
qwwdfsad: well, it can be considered as the problem with deserialization :)

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
qwwdfsad: For this one, I decided not to, it contains only the most nested JsonObject and should be short enough. WDYT?

sandwwraith: I've got it, but what if this most nested object consists of 1000 known fields and 1 unknown

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
sandwwraith: It's more of a transformation...

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -6,37 +6,60 @@ package kotlinx.serialization.json
...
  *
  * Options list:
  * * [encodeDefaults] specifies whether default values are encoded.
- * * [strictMode] enables strict mode, which prohibits unknown keys in the JSON and non-numerical values in floating point numbers.
- * * [unquoted] specifies whether keys and values should be quoted.
+ *
+ * * [ignoreUnknownKeys] ignores encounters of unknown properties in the input JSON.
+ *
+ * * [isLenient] removes JSON specification restriction (RFC-4627) and makes parser
+ *   more liberal to the malformed input. In lenient mode quoted string literals,
```
sandwwraith:
```suggestion
 *   more liberal to the malformed input. In lenient mode quoted boolean literals,
```

runtime/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -6,37 +6,60 @@ package kotlinx.serialization.json
...
+ *   and unquoted string literals are allowed.
+ *
+ * * [serializeSpecialFloatingPointValues] removes JSON specification restriction on
+ *   special floating-point values such as `NaN` and `Infinity` and enables their
+ *   serialization. When enabling it, please ensure that the receiving party will be
+ *   able to parse these special value.
```
sandwwraith:
```suggestion
 *   able to parse these special values.
```

(or `this special value`)

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
```diff
@@ -67,8 +67,9 @@ public object JsonPrimitiveSerializer : KSerializer<JsonPrimitive> {
 
     override fun deserialize(decoder: Decoder): JsonPrimitive {
         verify(decoder)
-        return if (decoder.decodeNotNullMark()) JsonPrimitive(decoder.decodeString())
-        else decoder.decodeSerializableValue(JsonNullSerializer)
+        val result = (decoder as JsonInput).decodeJson()
```
sandwwraith: WDYT of using already existing `.asJsonInput()` method?

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
```diff
@@ -67,8 +67,9 @@ public object JsonPrimitiveSerializer : KSerializer<JsonPrimitive> {
...
+        if (result !is JsonPrimitive) throw JsonDecodingException(-1, "Unexpected JSON element, expected JsonPrimitive, had ${result::class}", result.toString())
```
sandwwraith: and `result.primitive` ext val. It does not contain `result.toString()` currently, but we can add this (still we should mind that `result.toString()` can be rather big JSON)

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -7,21 +7,57 @@ package kotlinx.serialization.json
...
+
+internal fun InvalidKeyKindException(keyDescriptor: SerialDescriptor) = JsonEncodingException(
+    "Value of type '${keyDescriptor.serialName}' can't be used in JSON as a key in the map. " +
+            "It should have either primitive or enum kind, but its kind is '${keyDescriptor.kind}.'\n" +
+            "You can convert such maps to arrays [key1, value1, key2, value2,...] using 'allowStructuredMapKeys' property in JsonConfiguration"
+)
+
+private fun String.minify(offset: Int = -1): String {
+    if (offset < 200) return this
```
sandwwraith: Should this condition go _after_ the on with `offset == -1`? -1 is less than 200 and therefore `if (offset == -1)` never executed. Or did you want to go with `if (this.length < 200)` ?

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
```diff
@@ -207,16 +225,14 @@ internal class JsonReader(private val source: String) {
...
-            if (currentPosition >= length) fail("Unexpected EOF", currentPosition)
             if (source[currentPosition] == STRING_ESC) {
                 appendRange(source, lastPosition, currentPosition)
                 val newPosition = appendEsc(source, currentPosition + 1)
                 currentPosition = newPosition
                 lastPosition = newPosition
-            } else {
-                currentPosition++
+            } else if (++currentPosition >= source.length) {
+                fail("EOF", currentPosition)
```
sandwwraith:
```suggestion
                fail("Unexpected EOF", currentPosition)
```

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -104,23 +104,30 @@ internal class StreamingJsonInput internal constructor(
...
+            val index = descriptor.getElementIndex(key)
             if (index != CompositeDecoder.UNKNOWN_NAME) {
                 return index
             }
 
-            if (configuration.strictMode) reader.fail("Encountered an unknown key $key")
-            else reader.skipElement()
+            if (!configuration.ignoreUnknownKeys) {
+                reader.fail(
+                    "Encountered an unknown key '$key'. You can enable 'ignoreUnknownKeys' property" +
```
sandwwraith:
```suggestion
                    "Encountered an unknown key '$key'. You can enable 'JsonConfiguration.ignoreUnknownKeys' property" +
```

or

```suggestion
                    "Encountered an unknown key '$key'. You can enable 'ignoreUnknownKeys' property in JsonConfiguration" +
```

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -142,15 +149,47 @@ internal class StreamingJsonInput internal constructor(
...
+        } else {
+            reader.takeStringQuoted()
+        }
+    }
+
+    private inline fun <T> String.parse(type: String, block: String.() -> T): T {
+        try {
+            return block()
+        } catch (e: Throwable) {
+            reader.fail("Failed to parse $type")
```
sandwwraith: It would be nice to leave 'causedBy' on platforms where it is supported (JVM).

runtime/commonMain/src/kotlinx/serialization/json/internal/JsonReader.kt
qwwdfsad: This change is deliberate, the resulting text will be: `kotlinx.serialization.json.JsonDecodingException: Unexpected JSON token at offset 9: EOF`

runtime/commonMain/src/kotlinx/serialization/json/JsonElementSerializer.kt
qwwdfsad: Good. Though I am not sure whether it should throw SE instead of ISE. It's not "some serialization error", it's an explicit violation of API contract.
Don't mind if I change the type of the exception to ISE?

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
qwwdfsad: Yeah, matching the length of the string is much better

runtime/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
qwwdfsad: Good point, though such situation is improbable :)
Added minify to keep it consistent

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
qwwdfsad: It is generally a good point, but in this particular case (and in its `Tree` counterpart) cause will just duplicate the stack trace and won't really add any meaningful information.
The name of the primitive type and its values as part of the JSON input string already covers it

PR #723, opened 2020-02-28
Serializer interfaces

qwwdfsad:
I've decided not to introduce `CollectionLikeSerializer`: we don't have a compelling use-case (for now) and I'd like to avoid adding a deadweight to the core API when it's not really necessary

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
@@ -140,7 +140,6 @@ public class ProtoBuf(
         override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) = when {
             // encode maps as collection of map entries, not merged collection of key-values
             serializer is MapLikeSerializer<*, *, *, *> -> {
-                // TODO it probably requires some common interface exposure
                 val serializer = (serializer as MapLikeSerializer<Any?, Any?, T, *>)
                 val mapEntrySerial = MapEntrySerializer(serializer.keySerializer, serializer.valueSerializer)
                 SetSerializer(mapEntrySerial).serialize(this, (value as Map<*, *>).entries)
@@ -548,7 +547,6 @@ public class ProtoBuf(
         val reader = ProtobufReader(ProtobufDecoder(stream))
         return reader.decode(deserializer)
     }
-
 }
 
 private fun Short.toLittleEndian(): Short = (((this.toInt() and 0xff) shl 8) or ((this.toInt() and 0xffff) ushr 8)).toShort()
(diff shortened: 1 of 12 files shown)


sandwwraith:
Note that we _may_ need `CollectionLikeSerializer` for this operation: https://github.com/kotlin/kotlinx.serialization/blob/64f04e0648d4a603486d827f7150225b58db96c4/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt#L291

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -281,8 +285,8 @@ public class ProtoBuf(
         @Suppress("UNCHECKED_CAST")
         override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T = when {
             // encode maps as collection of map entries, not merged collection of key-values
-            deserializer is MapLikeSerializer<*, *, *, *> -> {
-                val serializer = (deserializer as MapLikeSerializer<Any?, Any?, T, *>)
+            deserializer is MapLikeSerializer<*, *, *> -> {
```
sandwwraith: Hm, it actually seems that approach with `.toMap` is faulty: if one have custom serializer with `StructureKind.MAP`, this method will still return `Map` instance — i.e. in your example with StringPairSerializer, we would be able to serialize `StringPair` with `toMap` function, **but** in deserialize we get an instance of `Map` instead of `StringPair`.

runtime/commonMain/src/kotlinx/serialization/MapLikeSerializer.kt
```diff
@@ -0,0 +1,65 @@
...
+
+package kotlinx.serialization
+
+/**
+ * Optional interface that represents [serializer][KSerializer] for map-like structures.
+ * [T] is a type of target type for the serializer,
+ * [K] and [V] type parameters represents types of the key and value.
+ *
+ * This interface is a marker that is recommended to implement for any serializer of [StructureKind.MAP] kind.
+ * Default formats, as well as endorsed ones, recognize serializers of [MapLikeSerializer] type and
```
sandwwraith:
```suggestion
 * Default formats, as well as endorsed ones, may recognize serializers of [MapLikeSerializer] type and
```

Recognising is not mandatory

sandwwraith:
Btw, WDYT of introducing `@AdvancedSerializationApi` annotation and moving some things from `@Internal` there?

pdvrieze:
The only thing I would say is that listlike also applies to sets and other (Mutable) collection interfaces.

qwwdfsad:
After starring to this problem (and your comments) for a while, I'm starting to realize these interfaces are, indeed, way too complex and fragile.
Looks like using the existing approach is a way to go for a while

PR #724, opened 2020-03-02
Use typeOf-based API where possible

qwwdfsad:


runtime/commonMain/src/kotlinx/serialization/SerialImplicits.kt
```diff
@@ -4,22 +4,21 @@
...
+ * Marks declaration that obtain serializer implicitly using experimental [typeOf] function and/or reflection.
  *
- * This approach is discouraged in general because it has several drawbacks, including:
- * - Reflection is not available on Kotlin/Native and is very limited on Kotlin/JS.
- * - Reflection won't infer correct serializers for generic classes, like collections.
- * - SerialModule may not be available, since it is bound to particular format, not serializer.
- * - Such reflection calls are usually slow.
+ * These declarations have the following limitations:
+ * - [typeOf] API does not work properly on K/JS with user-defined serializers.
+ * - Reflection can't infer correct serializers for generic classes, like collections.
```
sandwwraith: Most of people don't know what exactly 'typeOf API' means and whether it would help them with generics. Nor they know whether 'typeOf' or regular reflection would be used in particular case: Now we have functions annotated with same `@ImplicitReflectionSerializer` and some of them work with generics (but not on JS), and some of them don't (`X::class.serializer()`). 
I think we need either to rephrase this documentation completely or leave things as is and wait until we can replace _all_ functions to functions with typeof and removing this annotation (leaving it on `X::class.serializer()` or deprecating the latter)

qwwdfsad: I'd like to avoid leaving things as is, it is much better to have everything that is `inline <reified T>` to work properly (hint: this change doesn't break existing code, but enables new usages such as `json.stringify<Box<List<Int>>>`), so users (at least JVM ones) can rely on that.

I'd rather make `X::class.serializer` function internal with a suggestion to use `serializer<T>` instead, remove `@ImplicitReflectionSerializer` from `ContextSerializer` and adjust its documentation to reflect the fact it doesn't support generics (yet). WDYT?

qwwdfsad:
(postponed)

PR #726, opened 2020-03-02
Make SerialModule sealed class instead of interface

### LINKED ISSUE #448: Polymorphic serializer, ignore unregistered objects when deserializing list of polymorphic objects

### LINKED ISSUE #697: Creating a custom `SerialModule` for Json seems useless
**Describe the bug**
When creating my own `SerialModule` implementation to pass along as context, I have to overwrite several methods with names such as `getPolymorphic` -- which to me implies that these methods will be called/invoked at runtime if I use my context and something tries to (de-)serialize polymorphic values.

In reality however, this does not happen because the `Json` constructor [invokes a `plus` operation](https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/commonMain/src/kotlinx/serialization/json/Json.kt#L63) and the [modules are combined](https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleExtensions.kt#L39) via [`dumpTo`](https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/commonMain/src/kotlinx/serialization/modules/SerialModuleBuilders.kt#L78) which only supports statically determined seralizer mappings.

**To Reproduce**
Minimal-ish example to show my point:
```kotlin
import kotlinx.serialization.KSerializer
import kotlinx.serialization.SerialName
import kotlinx.serialization.modules.EmptyModule
import kotlinx.serialization.modules.SerialModule
import kotlin.reflect.KClass

@Serializable
sealed class Data {
    abstract val id: Int

    @Serializable
    @SerialName("something")
    data class Something(val message: String, override val id: Int) : Data()
    @Serializable
    @SerialName("something_else")
    data class SomethingElse(val number: Int, override val id: Int) : Data()
}

object MyModule : SerialModule by EmptyModule {
    override fun <T : Any> getPolymorphic(baseClass: KClass<T>, serializedClassName: String): KSerializer<out T>? {
        println("do something dynamic depending on $serializedClassName!")
        return null
    }
}

fun main() {
    val parser = Json(context = MyModule)
    val serial = """{"id": 42, "message":"Hello, World", "type": "something"}"""

    val parsed = parser.parse(Data.serializer(), serial)
    println(parsed)
}
```

**Expected behavior**
I expect the line `do something dynamic depending on something!` to be printed when executing. This does not happen -- in fact, nothing gets printed at all before the final `println(parsed)` statement.

I've taken a look at the code and it seems that the `dumpTo` method uses a `SerialModuleCollector` which is backed by a default `SerialModuleImpl`. This is unfortunate because it statically collects all pre-registered serialisers into one big map and then allows no dynamic invocation at runtime because it only uses `Map#get`.

It also defies the intuition that overriding any method other than `dumpTo` in the `SerialModule` interface would achieve anything, which makes me wonder why these methods are there in the first place.

For more context, please see my description and code sample at #448. I have a `sealed class` hierarchy of `extension` objects that are relevant for me, but I also need to parse "unknown" extensions (that were created by other people) into some form of "fallback extension". That is only possible, however, if I can *dynamically* intercept the serialiser lookup at runtime.

Thanks for your help and your amazing library! :)

**Environment**
 - Kotlin version: 1.3.61
 - Library version: 0.14.0
 - Kotlin platforms: JVM only
 - Gradle version: 6.1.1
 - IDE version (if bug is related to the IDE) not related to IDE

qwwdfsad:
* As explained in #697, it appears that concatenating two modules as "behaviours" (as opposed to "bag of serializers") is way too dangerous for the end users and implies a lot of inconsistent behaviours and performance hits. So the most rational solution here is to prohibit such concatenations by closing the way to implement your own SerialModule
    * In order to still solve "fallback serializer" use-case, new 'default' DSL is introduced along with its support in PolymorphicSerializer

Fixes #697
Fixes #448


runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -59,6 +61,21 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
...
 
+    /**
+     * Registers serializer provider that will be invoked if no polymorphic serializer is present.
+     * [defaultSerializerProvider] can be stateful and lookup a serializer for the missing type dynamically.
+     *
+     * Typically, if the class is not registered in advance, it is not possible to know the structure of the unknown
+     * type and have a precise serializer, so the default serializer has limited capabilities.
+     * To have a structural access to the unknown data, it is recommended to use [JsonTransformingSerializer] subclass.
+     */
+    public fun default(defaultSerializerProvider: (className: String) -> DeserializationStrategy<out Base>?) {
```
qwwdfsad: I've decided not to mention that it is only applicable for deserialization (it should be more or less obvious?) and express it in the return type instead

runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
```diff
@@ -27,22 +27,54 @@ class PolymorphismTest {
...
-        val s = json.stringify(PolymorphicSerializer(PolyDerived::class), obj)
+        val s = json.stringify(PolymorphicSerializer(PolyDerived::class), obj, useStreaming)
         assertEquals("[kotlinx.serialization.PolyDerived,{id:1,s:b}]", s)
     }
+
+    object PolyDefaultSerializer : JsonTransformingSerializer<PolyDefault>(PolyDefault.serializer(), "foo") {
+        override fun readTransform(element: JsonElement): JsonElement {
+            return json {
+                "json" to element
+                "id" to 42
```
qwwdfsad: This one is not really **that** convenient, especially the fact that I should specify `PolyDefault.serializer()`.
Maybe we can improve it even more

runtime/commonMain/src/kotlinx/serialization/internal/AbstractPolymorphicSerializer.kt
```diff
@@ -81,7 +81,8 @@ public abstract class AbstractPolymorphicSerializer<T : Any> internal constructo
     public open fun findPolymorphicSerializer(
         decoder: CompositeDecoder,
         klassName: String
-    ): KSerializer<out T> = decoder.context.getPolymorphic(baseClass, klassName)
+    ): DeserializationStrategy<out T> = decoder.context.getPolymorphic(baseClass, klassName)
```
qwwdfsad: It is not supported for sealed classes explicitly, WDYT?

runtime/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -59,6 +61,21 @@ public class PolymorphicModuleBuilder<Base : Any> internal constructor(
     @ImplicitReflectionSerializer
     public inline fun <reified T : Base> subclass(): Unit = addSubclass(T::class, serializer())
 
+    /**
+     * Registers serializer provider that will be invoked if no polymorphic serializer is present.
+     * [defaultSerializerProvider] can be stateful and lookup a serializer for the missing type dynamically.
+     *
+     * Typically, if the class is not registered in advance, it is not possible to know the structure of the unknown
+     * type and have a precise serializer, so the default serializer has limited capabilities.
+     * To have a structural access to the unknown data, it is recommended to use [JsonTransformingSerializer] subclass.
```
sandwwraith: I believe it's worthy to also mention `JsonParametricSerializer` here

runtime/commonMain/src/kotlinx/serialization/modules/SerialModule.kt
```diff
@@ -1,60 +1,140 @@
...
      */
-    public fun <T : Any> getPolymorphic(baseClass: KClass<T>, serializedClassName: String): KSerializer<out T>?
+    public abstract fun <T : Any> getPolymorphic(baseClass: KClass<T>, serializedClassName: String): KSerializer<out T>?
+
+    /**
+     * Returns a default polymorphic serializer registered for a [serializedClassName] in the scope of [baseClass].
+     * In contrast with [getPolymorphic], module user registers a factory function for default serializers that can
+     * provide a serializer even for an unknown in advance [serializedClassName].
+     */
+    public abstract fun <T : Any> getDefaultPolymorphic(
```
sandwwraith: Why do we need an extra method instead of regular `getPolymorphic(baseClass, serializedClassName)`?

runtime/commonMain/src/kotlinx/serialization/internal/AbstractPolymorphicSerializer.kt
sandwwraith: It should be supported: sealed class may have abstract inheritor, for whom we want to use 'default' serializer strategy

pdvrieze:
> I'm a little bit concerned because we block the way to add some user-defined meaningful logic to `getContextual` and `getPolymorphic(baseClass, value)`, but probably no one ever tried to do that.
> 
> Also, I do not see how it fixes #448 . Example with `PolyDefault` shows how we can read properties of unknown inheritor to a json tree (which is also extremely useful btw and worth documenting in samples), but now how we can actually **skip** it, if it is contained, say, in a list.

While I'd like the ability to do this in a user defined way it is already possible by having the parent serializer (the serializer that owns the polymorphic element) be custom - of course this is not global, then you could override formats). More significantly the current way that it appears to work is incorrect. I'd be all for sealing at this point, perhaps later having a way to inject custom behaviour.

qwwdfsad:
>Also, I do not see how it fixes #448 

It fixes it in a way that issue can be closed:

1) We provide API for "error object" fallback for any unknown type
2) We provide both json transforming and json parametric serializers to skip some elements if it really required.

And we definitely *won't* provide a built-in mechanism to skip unknown elements from the list

runtime/commonMain/src/kotlinx/serialization/modules/SerialModule.kt
```diff
@@ -1,60 +1,124 @@
...
+    }
+
+    override fun <T : Any> getPolymorphic(baseClass: KClass<in T>, serializedClassName: String): DeserializationStrategy<out T>? {
+        // Any subtypes
+        val standardPolymorphic =
+            if (baseClass == Any::class) StandardSubtypesOfAny.getDefaultDeserializer(serializedClassName)
+            else null
+        if (standardPolymorphic != null) return standardPolymorphic as KSerializer<out T>
+        // Registered
+        val registered = polyBase2NamedSerializers[baseClass]?.get(serializedClassName) as? KSerializer<out T>
```
sandwwraith: Shouldn't we place custom registered serializers before the standard any subtypes, as in `getPolymorphic(baseClass, value)`?

qwwdfsad: It was here before my change, but it indeed looks like a bug. Fixed

PR #738, opened 2020-03-04
Fix Gradle example rendering

lukaszkalnik:
With recent version updates to README apparently the Gradle examples got a little broken.
The closing backticks were removed from one of the examples, leading to improper rendering of the remainder of the paragraph.


sandwwraith:
Thank you!

PR #750, opened 2020-03-10
Move LongAsStringSerializer to builtins

sandwwraith:


runtime/commonMain/src/kotlinx/serialization/internal/LongAsStringSerializer.kt
```diff
@@ -4,17 +4,9 @@
...
-        encoder.encodeString(value.toString())
-    }
-
-    override fun deserialize(decoder: Decoder): Long {
-        return decoder.decodeString().toLong()
-    }
-}
+@Deprecated(
+    "Moved to kotlinx.serialization.builtins package",
+    ReplaceWith("LongAsStringSerializer", "kotlinx.serialization.builtins.LongAsStringSerializer"),
```
qwwdfsad: This replacement does not work, unfortunately

sandwwraith: it works on `import kotlinx.serialization.internal.LongAsStringSerializer`, but not on usage. At least something

PR #758, opened 2020-03-13
Enable library author's mode and adjust visibilities

sandwwraith:
Also add some missing parts of documentation

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
@@ -28,6 +28,7 @@ import kotlinx.serialization.internal.*
  * }
  * ```
  */
+@Suppress("FunctionName")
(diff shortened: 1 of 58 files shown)


runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -28,6 +28,7 @@ import kotlinx.serialization.internal.*
  * }
  * ```
  */
+@Suppress("FunctionName")
```
qwwdfsad: I'd rather suggest you to change the regexp of the inspection

sandwwraith: It's Kotlin's default inspection, not ours. We may create a ticket and discuss it, since constructor-like functions are widely used (by us at least), but it is not clear how to distinguish constructor function from regular one.

PR #760, opened 2020-03-17
Type parameters in descriptors

### LINKED ISSUE #734: Public API for declaring descriptors for classes with type parameters
Also, revisit hc/equals in the current builder

sandwwraith:
Fixes #734


runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -80,19 +88,32 @@ public open class PluginGeneratedSerialDescriptor(
...
     override fun toString(): String {
-        return indices.entries.joinToString(", ", "$serialName(", ")") { it.key + ": " + getElementDescriptor(it.value).serialName }
+        return indices.entries.joinToString(", ", "$serialName(", ")") {
+            it.key + ": " + getElementDescriptor(it.value).serialName
+        }
     }
 }
+
+// this function is for comparing user-defined and plugin-generated descriptors.
+// Do we really need it?
```
sandwwraith: It's unclear whether we should allow comparing different implementations of SerialDescriptor, but if we do so, equality should be symmetric.

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -80,19 +88,32 @@ public open class PluginGeneratedSerialDescriptor(
...
+@Suppress("DEPRECATION_ERROR")
+internal fun SerialDescriptor.typeParameters(): List<SerialDescriptor> = when (this) {
```
sandwwraith: I've tried not to add `public val typeParameterDescriptors` to the interface, but this may lead to the problem where some implementations are not equal or not symmetrically equal (e.g. ListLikeDesc and SerialDescImpl).

Perhaps, this needs to be public with a name like `SerialDescriptor.defaultEquals(other)`, so it can be used in user-defined and our own implementations consistently. However, such a decision requires adding `public val typeParameterDescriptors` to the interface.

runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
```diff
@@ -19,16 +19,16 @@ package kotlinx.serialization
  * children elements and additional metadata.
  *
  * * [serialName] uniquely identifies descriptor (and the corresponding serializer) for non-generic types.
- *   For generic types, the actual type substitution  is omitted from the string representation and the name
- *   identifies the family of the serializers without type substitutions.
+ *   For generic types, the actual type substitution is omitted from the string representation and the name
+ *   identifies the family of the serializers without type substitutions. However, type substitution is accounted
+ *   in [equals] and [hashCode] operations, meaning that descriptors of generic classes with the same name, but different type
```
sandwwraith: Such behaviour may leave user with question 'how can I look onto actual type substitution' ? See discussion about `public val typeParameterDescriptors`.

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -229,6 +245,7 @@ private class SerialDescriptorImpl(
         if (this === other) return true
         if (other !is SerialDescriptor) return false
         if (serialName != other.serialName) return false
```
qwwdfsad: Our resolution was the following:
1) Compare identities
2) Compare serial names
3)  Cache and compare hashcodes 
4) Compare type parameters
5) Compare element names
6) Compare element kinds

Also, please reflect these parameters in hashcode computation as well

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -203,10 +218,11 @@ public inline fun <reified T> SerialDescriptorBuilder.setDescriptor(): SerialDes
     return setDescriptor(serializer<T>().descriptor)
 }
 
-private class SerialDescriptorImpl(
+internal class SerialDescriptorImpl(
     override val serialName: String,
     override val kind: SerialKind,
     override val elementsCount: Int,
+    val typeParameters: List<SerialDescriptor>,
```
qwwdfsad: [nit] `@JvmField`

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -80,19 +88,32 @@ public open class PluginGeneratedSerialDescriptor(
         @Suppress("DEPRECATION_ERROR")
         if (other !is SerialDescriptor) return false
         if (serialName != other.serialName) return false
-        // TODO compare only serial names
-        if (elementDescriptors() != other.elementDescriptors()) return false
+        if (typeParameterDescriptors != other.typeParameters()) return false
```
qwwdfsad: Same here

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
```diff
@@ -9,6 +9,7 @@ import kotlinx.serialization.*
 @InternalSerializationApi
 public interface GeneratedSerializer<T> : KSerializer<T> {
     fun childSerializers(): Array<KSerializer<*>>
+    fun typeParametersSerializers(): Array<KSerializer<*>> = arrayOf()
```
qwwdfsad: Let's avoid default implementation for auto-generated code, it is not really needed

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
qwwdfsad: I'd say we should not. Asymmetric comparison bugs are really subtle and hard to catch, so we should do our best to avoid it. We don't really want to allow user-defined serializers that are equal to  e.g. `HashSetSerializer`

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
qwwdfsad: `internal` is fine for now

sandwwraith: If different implementations of `SerialDescriptor` are considered to be non-equal to each other, this function is not needed at all

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
sandwwraith: Why not? It is kinda useful and simplifies things a little.

Btw, we can put here smth like `EMPTY_ARRAY` instead to avoid allocations

gradle.properties
```diff
@@ -18,7 +18,7 @@ benchmarks_version=0.1.7-dev-14
 jackson_version=2.10.0.pr1
 native.deploy=false
 
-kotlin.js.compiler=both
+#kotlin.js.compiler=both
```
qwwdfsad: Don't forget to uncomment it before the merge

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -225,16 +241,12 @@ private class SerialDescriptorImpl(
...
-        if (serialName != other.serialName) return false
-        return true
-    }
+    override fun equals(other: Any?): Boolean =
+        equalsImpl(other) { otherDescriptor: SerialDescriptorImpl -> typeParameters == otherDescriptor.typeParameters }
 
-    override fun hashCode(): Int {
-        return serialName.hashCode()
-    }
+    private val _hashCode: Int by lazy { hashCodeImpl(typeParameters) }
```
qwwdfsad: 1) Please always declare `val`s in the beginning of the class, not in the middle of it.
2) For the idempotent computations, please prefer non-synchronized approach with zero-check similarly to `java.lang.String.hashCode`. It has a smaller footprint and better performance characteristics

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -59,11 +61,17 @@ public open class PluginGeneratedSerialDescriptor(
 
     override fun getElementDescriptor(index: Int): SerialDescriptor {
         return generatedSerializer?.childSerializers()?.get(index)?.descriptor
-        ?: throw IndexOutOfBoundsException("$serialName descriptor has only $elementsCount elements, index: $index")
+                ?: throw IndexOutOfBoundsException("$serialName descriptor has only $elementsCount elements, index: $index")
+    }
+
+    internal val typeParameterDescriptors: List<SerialDescriptor> by lazy {
```
qwwdfsad: 1) `private`
2) It doesn't have to be lazy
3) `val` in the middle of the class
4) Better make it an array

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -75,24 +83,44 @@ public open class PluginGeneratedSerialDescriptor(
...
+        typeParameterDescriptors == otherDescriptor.typeParameterDescriptors
     }
 
-    override fun hashCode(): Int {
-        var result = serialName.hashCode()
-        // TODO hashcode only for serial name
-        result = 31 * result + elementDescriptors().hashCode()
-        return result
-    }
+    private val _hashCode: Int by lazy { hashCodeImpl(typeParameterDescriptors) }
```
qwwdfsad: same here

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -75,24 +83,44 @@ public open class PluginGeneratedSerialDescriptor(
...
+
+    override fun hashCode(): Int = _hashCode
 
+    // todo: should type parameters be in serial name?
```
qwwdfsad: it shouldn't :)

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
qwwdfsad: Was thinking about excessive bytecode, but I was wrong.
Yeah, it would be nice to have `EMPTY_ARRAY`

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.serialization.internal.*
 
 /**
  * Builder for [SerialDescriptor].
- * The resulting descriptor will be uniquely identified by the given [serialName],
+ * The resulting descriptor will be uniquely identified by the given [serialName] and [typeParameters],
```
qwwdfsad: by the given name, type parameters and elements structure (in a non-recursive manner)

qwwdfsad: Also, it is worth to mention in `SerialDescriptor` documentation to give a hint on how to implement/compare/use SD interface

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
sandwwraith: Unfortunately, we can't save bytecode here. It would be either `DefaultInpls.typeParamSerializers()` or just generated `return arrayOf()`

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
sandwwraith: Non-synchronized approach is a good idea; however, I have a feeling that such an approach may fail in Native, if such a descriptor would be transferred between workers and frozen. Note that standard `lazy` is freeze-aware and serializers which are `object`s (w/o type params) are already frozen by default (https://github.com/JetBrains/kotlin-native/issues/1817).

It is also possible to simply not to cache hashcode if `this.isFrozen == true`, but I'm not sure that this is a great idea.

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
sandwwraith: It has to be lazy because of native freezing (see my previous comment) and recursiveness: simply `generatedSerializer?.typeParametersSerializers()?.map { it.descriptor }` does not work if `descriptor` is the same descriptor being created (because it is not written to the field yet). Yeah, I know, recursive generics are not used often, but still.

It is also possible to convert it to the property with getter; it shouldn't hit the performance much, probably.

sandwwraith: Why the array is better here?

runtime/commonMain/src/kotlinx/serialization/SerialDescriptorBuilder.kt
qwwdfsad: Oh, okay. Let'sdrop a comment then, maybe at some point, we'll introduce some platform-specific class for that

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
qwwdfsad: Array simply has a lesser footprint

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
```diff
@@ -6,10 +6,17 @@ package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
 
+private val EMPTY_SERIALIZER_ARRAY: Array<KSerializer<*>> = arrayOf()
+
+/**
+ * An interface for a [KSerializer] instance generated by the compiler plugin.
+ *
+ * Should not be inherited or used directly.
```
qwwdfsad, line 14: [nit] interfaces are implemented, not inherited

runtime/commonMain/src/kotlinx/serialization/internal/PluginHelperInterfaces.kt
```diff
@@ -6,10 +6,17 @@ package kotlinx.serialization.internal
 
 import kotlinx.serialization.*
 
+private val EMPTY_SERIALIZER_ARRAY: Array<KSerializer<*>> = arrayOf()
```
qwwdfsad, line 9: Should be `@SharedImmuteable`?

runtime/commonMain/src/kotlinx/serialization/internal/Util.kt
```diff
@@ -88,6 +88,15 @@ internal fun SerialDescriptor.cachedSerialNames(): Set<String> {
     return result
 }
 
+private val EMPTY_DESCRIPTOR_ARRAY: Array<SerialDescriptor> = arrayOf()
```
qwwdfsad, line 91: Should be `@SharedImmuteable`?

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
sandwwraith: I've added test for recursive generics and it passes even without lazy. However, JS test fail completely due to initialization order problem

sandwwraith: Because for some reason, in JS descriptor is created before actual assigment of type parameter serializers to fields.

sandwwraith: I'll leave `lazy` for now and try to fix #789 later

PR #767, opened 2020-03-20
Protobuf performance improvement

### LINKED ISSUE #216: Protobuf Speed
We were looking into using this library for our data transfer for a new project due to it supporting both JSON and Protobuf and since it's noticeably faster than Jackson, that is, it is for JSON anyway. I was surprised to find that your protobuf serialization/deserialization is much slower than your JSON (1.5-2x). With any other library I've used it's always very much the opposite. We were hoping to use Protobuf largely for performance purposes, but it's significantly slower by about the same margin regardless of data types.

I understand that this is still a fairly early get it working state. Is this something you're planning on working on once you have a stable release/api or is it perhaps something that might be looked into earlier?

qwwdfsad:
With these changes, ProtoBuf on basic usages became twice as fast as JSON and x2-x5 faster than its previous version

Fixes #216


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -230,48 +235,84 @@ public class ProtoBuf(
...
-    private open inner class ProtobufReader(val decoder: ProtobufDecoder) : TaggedDecoder<ProtoDesc>() {
+    private open inner class ProtobufReader(@JvmField val decoder: ProtobufDecoder) : TaggedDecoder<ProtoDesc>() {
         override val context: SerialModule
             get() = this@ProtoBuf.context
 
-        private val indexByTag: MutableMap<Int, Int> = mutableMapOf()
-        private fun findIndexByTag(desc: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean = false): Int =
-            (0 until desc.elementsCount).firstOrNull {
-                extractParameters(
+        private var indexCache: IntArray? = null
```
sandwwraith: WDYT of using something like https://github.com/Kotlin/kotlinx.serialization/pull/597/files#diff-bcaa7635dcd4f1d128805c3c282f8253R17 ? 

Currently, it looks like that this cache would be built each time `ProtobufReader` is created, basically making it useless, e.g. in `List<X>` cache for X is built for every element, since in `beginStructure` ProtobufReader is re-created.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -230,48 +235,84 @@ public class ProtoBuf(
...
+        private fun getIndexByTag(descriptor: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean): Int {
+            if (serialId < 32) {
+                // Fast path
+                var cache = indexCache
+                if (cache == null) {
+                    cache = IntArray(32) { -2 }
+                    indexCache = cache
+                }
+                if (cache[serialId] == -2) {
+                    cache[serialId] = findIndexByTag(descriptor, serialId, zeroBasedDefault)
```
sandwwraith: We also can try to fill cache for all serial-ids, not by one element each time

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -404,66 +458,71 @@ public class ProtoBuf(
...
             assertWireType(i32)
-            val ans = readIntLittleEndian()
+            val result = Float.fromBits(readIntLittleEndian())
             readTag()
-            return Float.fromBits(ans)
+            return result
         }
 
         private fun readIntLittleEndian(): Int {
+            // TODO this could be optimized by extracting method to the IS
```
sandwwraith: Is this functionality going to be in kotlinx-io?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
```diff
@@ -29,18 +37,48 @@ public enum class ProtoNumberType {
...
+    return ProtoNumberType.FIXED
+}
 
-internal inline fun <reified A: Annotation> SerialDescriptor.findAnnotation(elementIndex: Int): A? {
-    return getElementAnnotations(elementIndex).find { it is A } as A?
+internal fun SerialDescriptor.extractParameters(index: Int): ProtoDesc {
+    val annotations = getElementAnnotations(index)
+    var protoId: Int = index + 1
+    var format: ProtoNumberType = ProtoNumberType.DEFAULT
+    for (i in annotations.indices) { // Allocation-friendly loop
```
sandwwraith: why it is more allocation-friendly than `for annotation in annotations`?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
```diff
@@ -29,18 +37,48 @@ public enum class ProtoNumberType {
...
+@Suppress("NOTHING_TO_INLINE")
+internal inline fun ProtoDesc(protoId: Int, type: ProtoNumberType): ProtoDesc {
+    return type.signature or protoId.toLong()
 }
 
-public class ProtobufDecodingException(message: String) : SerializationException(message)
+internal inline val ProtoDesc.protoId: Int get() = (this and Int.MAX_VALUE.toLong()).toInt()
+
+internal val ProtoDesc.numberType: ProtoNumberType get() {
+    if (ProtoNumberType.DEFAULT.equalTo(this)) return ProtoNumberType.DEFAULT
```
sandwwraith: This probably should be more performant when expressed as bitshift with `when(mask)` instead of several if-elses

runtime/commonMain/src/kotlinx/io/Streams.kt
```diff
@@ -1,49 +1,252 @@
...
+        return readVarint32SlowPath()
+    }
+
+    override fun readVarint64(eofAllowed: Boolean): Long {
+        if (position == array.size) {
+            if (eofAllowed) return -1
+            else error("Unexpected EOF")
+        }
+
+        // Fast-path: single and two byte values
```
sandwwraith: Are there any text commentaries explaining this algorithm?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
qwwdfsad: It doesn't allocate an iterator

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
qwwdfsad: This one is likely to

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -230,48 +235,84 @@ public class ProtoBuf(
...
 
-        private val indexByTag: MutableMap<Int, Int> = mutableMapOf()
-        private fun findIndexByTag(desc: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean = false): Int =
-            (0 until desc.elementsCount).firstOrNull {
-                extractParameters(
+        private var indexCache: IntArray? = null
+        private var sparseIndexCache: MutableMap<Int, Int>? = null
+
+        private fun getIndexByTag(descriptor: SerialDescriptor, serialId: Int, zeroBasedDefault: Boolean): Int {
+            if (serialId < 32) {
```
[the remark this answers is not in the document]
qwwdfsad: Just a reasonably small constant for the number of fields

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
qwwdfsad: Current cache approach is **still** better than the previous quadratic algorithm.
I'll fix this one in follow-up work, I am not ready to do one more performance analysis for collections right now.

runtime/commonMain/src/kotlinx/io/Streams.kt
qwwdfsad: Added small explanation

PR #787, opened 2020-04-05
Fix JsonTransformingSerializer documentation

Dominaezzz:
Delegate methods docs' referred to the wrong caller.


sandwwraith:
Oh, what an overlook. Thanks for spotting it!

PR #788, opened 2020-04-06
Nibble at a hole in the config parser

x-rat4-x:


formats/config/src/main/kotlin/kotlinx/serialization/config/ConfigParser.kt
```diff
@@ -21,10 +21,10 @@ public class ConfigParser(
     public inline fun <reified T : Any> parse(conf: Config): T = parse(conf, context.getContextualOrDefault(T::class))
 
     public fun <T> parse(conf: Config, deserializer: DeserializationStrategy<T>): T =
-        ConfigReader(conf).decode(deserializer)
+        ConfigDecoder(conf).decode(deserializer)
 
 
-    private abstract inner class ConfigConverter<T> : TaggedDecoder<T>() {
+    public abstract inner class ConfigConverter<T> : TaggedDecoder<T>() {
```
sandwwraith, line 27: This class is not intended for public use, all interactions are designed to be performed via `ConfigParser().parse`. Why do you need it to be public?

x-rat4-x:
Perhaps new commit is more clear.

formats/config/src/main/kotlin/kotlinx/serialization/config/ConfigParser.kt
```diff
@@ -21,7 +21,9 @@ public class ConfigParser(
     public inline fun <reified T : Any> parse(conf: Config): T = parse(conf, context.getContextualOrDefault(T::class))
 
     public fun <T> parse(conf: Config, deserializer: DeserializationStrategy<T>): T =
-        ConfigReader(conf).decode(deserializer)
+        ConfigDecoder(conf).decode(deserializer)
+
+    public interface ConfigInput : Decoder, CompositeDecoder {}
```
sandwwraith, line 26: Am I right that you need this interface to perform format-specific operations in a same manner as with `JsonInput`? Can you please tell me more about your use-case? This interface does not have any useful methods that can be used during `deserialize()`. 

Also, pay attention that it is recommended to have documentation for every public interface.

x-rat4-x:
The use case is custom serializers that dispatch by `Encoder`.

x-rat4-x:
It seems stuck on serialization

PR #790, opened 2020-04-06
Support retreiving object's serializers via serializer<T>()

sandwwraith:
on Native

#656


sandwwraith:
Test also require plugin update

PR #794, opened 2020-04-08
Rework JSON DSL

### LINKED ISSUE #418: JsonArrayBuilder's unaryPlus extensions don't work for Number?

### LINKED ISSUE #627: `JsonObjectBuilder.to` method shadows `to` function from stdlib
Recently I encountered such a problem. Imagine we have a function like follow:

```
fun prepareRequest(alias: String): JsonObject {
    return json {
        "alias" to alias
    }
}
```

Then I did small refactoring:

```
fun prepareRequest(alias: Alias): JsonObject {
    return json {
        "alias" to alias
        // Fix:
        // "alias" to alias.name
    }
}
```

And suddenly method `to` became https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to.html which result is just lost. The compiler is silent.

Possibly it is worth to rename `JsonObjectBuilder.to` extension method or make json object construction more explicit.

qwwdfsad:
Fixes #418
Fixes #627


altavir:
I completely agree with decision to deprecate "pretty" DSL, espcially `to` infix. It is a source of a lot of errors. In my project I've replaced `to` by `put` infix, and it works much better. Also it is possible to use following syntax:
```kotlin
infix fun String.put(builderAction: JsonObjectBuilder.() -> Unit) = addJson(this,builderAction)
``` 

It allows to write:
```kotlin
json{
  "child" put {
    "subChild" put "fff
  }
}
```

runtime/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -1,101 +1,248 @@
...
-    internal val content: MutableMap<String, JsonElement> = linkedMapOf()
+    internal fun build(): JsonObject = JsonObject(content)
 
     /**
      * Adds given [value] to the current [JsonObject] with [this] as a key.
      */
+    @Deprecated(message = infixToDeprecated, replaceWith = ReplaceWith("add(this, value)"))
     public infix fun String.to(value: JsonElement) {
-        require(content[this] == null) {"Key $this is already registered in builder"}
+        require(content[this] == null) { "Key $this is already registered in builder" }
```
sandwwraith, line 125: This shows the difference between old and new behavior: old prohibits overwriting, new allows it. I'm not sure which one is better, but probably changing it without notice is not good either.

runtime/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -1,101 +1,248 @@
...
+     */
+    public fun addJson(builderAction: JsonObjectBuilder.() -> Unit) {
+        content += buildJson(builderAction)
+    }
+
+    /**
+     * Add the [JSON][JsonArray] produced by the [builderAction] function to a resulting array.
+     */
+    public fun addArray(builderAction: JsonArrayBuilder.() -> Unit) {
+        content.add(buildJsonArray(builderAction))
```
sandwwraith, line 177: Any particular reasons for using `add` here instead of `+=` like in other methods?

qwwdfsad: Yes, `JsonArray` implements list, making `content += jsonArray` ambigous

runtime/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
qwwdfsad: I am fine with both, decided not to check it to be consistent with `buildMap`

sandwwraith:
Changes are fine. WDYT of adding `jsonOf` function and making JsonPbjectBuilder implement mutable map?

Issue #796, opened 2020-04-11
.serializer() method returns null in companion obeject


Vlad8161:
```
@Polymorphic
@Serializable
sealed class SomeClass {

    @Serializable
    @SerialName("p1")
    data class SomeClass1(val a: Int) : SomeClass()

    @Serializable
    @SerialName("p2")
    data class SomeClass2(
        val a: Int,
        val b: Int
    ) : SomeClass()

    companion object {
        val serializerModule = SerializersModule {
            polymorphic(SomeClass::class) {
                SomeClass1::class with SomeClass1.serializer()
                SomeClass2::class with SomeClass2.serializer()
            }
        }
    }
}
```

When I'm trying to instantiate SomeClass1
```
val someClass = SomeClass1(1)
```

I have an exception:
```
java.lang.ExceptionInInitializerError
	at test.TestSerialization.error(TestSerialization.kt:17)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
[...]

PR #798, opened 2020-04-12
Add fail-fast check for InputStream.readExactNBytes

Him188:
ProtoBuf.kt:358
```kotlin
fun nextObject(): ByteArray {
    assertWireType(SIZE_DELIMITED)
    val len = decode32()
    check(len >= 0)
    val ans = inp.readExactNBytes(len)
    readTag()
    return ans
}
```
`len` can be very big if the input is not in valid protobuf format.
Then `readExactNBytes` will allocate a huge ByteArray, say 200MiB.

This problem real happened in my application and caused OOM.

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/internal/Util.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/Util.kt
@@ -9,6 +9,8 @@ import kotlinx.serialization.*
 
 @InternalSerializationApi
 public fun InputStream.readExactNBytes(bytes: Int): ByteArray {
+    val available = this.available()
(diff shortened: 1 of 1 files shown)


sandwwraith:
Hi, thanks for noticing. 

This place was refactored in `dev` branch, so open PR against it if you don't mind (note that this function was moved to the protobuf-specific module, but you can still find it by name). Or we can apply your suggestion ourselves (cc @qwwdfsad )

sandwwraith:
Also it will be probably affected by #778

Him188:
Thank you, it will be easier for yourselves to fix this issue. Feel free to close this.

qwwdfsad:
Thanks, superseded by #796 800

Thanks for the spotting and for the fix!

PR #810, opened 2020-04-20
Do not remove type discriminator in unsafe manner during polymorphic …

### LINKED ISSUE #807: java.util.NoSuchElementException when deserializing twice from the same JsonElement
Steps to reproduce:
```
@Serializable
sealed class Foo {
    @Serializable
    data class Bar(val a: Int) : Foo()
}
@Test
fun testFoo() {
    val json = Json.toJson(Foo.Bar(1))
    println(Json.fromJson<Foo>(json))
    println(Json.fromJson<Foo>(json))
}
```

qwwdfsad:
…JSON deserializer, properly filter out it in unknown keys check instead

Fixes #807

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
@@ -21,6 +21,14 @@ internal fun <T> Json.readJson(element: JsonElement, deserializer: Deserializati
     return input.decode(deserializer)
 }
 
+internal fun <T> Json.readPolymorphicJson(
+    discriminator: String,
+    element: JsonObject,
+    deserializer: DeserializationStrategy<T>
+): T {
+    return JsonTreeInput(this, element, discriminator, deserializer.descriptor).decode(deserializer)
+}
+
 private sealed class AbstractJsonTreeInput(
     override val json: Json,
     open val value: JsonElement
@@ -140,7 +148,12 @@ private class JsonPrimitiveInput(json: Json, override val value: JsonPrimitive) 
     }
 }
 
-private open class JsonTreeInput(json: Json, override val value: JsonObject) : AbstractJsonTreeInput(json, value) {
+private open class JsonTreeInput(
+    json: Json,
+    override val value: JsonObject,
+    private val polyDiscriminator: String? = null,
+    private val polyDescriptor: SerialDescriptor? = null
+) : AbstractJsonTreeInput(json, value) {
     private var position = 0
 
     override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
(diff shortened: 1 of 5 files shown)


qwwdfsad:
The commit fails because of problems in K/N. PTAL

runtime/commonMain/src/kotlinx/serialization/json/internal/TreeJsonInput.kt
```diff
@@ -155,12 +168,23 @@ private open class JsonTreeInput(json: Json, override val value: JsonObject) : A
 
     override fun currentElement(tag: String): JsonElement = value.getValue(tag)
 
+    override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
+        /*
+         * For polymorphic serialization we'd like to avoid excessive decoder creating in
+         * beginStructure to properly preserve 'polyDiscriminator' field and filter it out.
+         */
+        if (descriptor == polyDescriptor) return this
```
sandwwraith, line 176: Can we use `===` here?

runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
```diff
@@ -48,6 +48,7 @@ class PolymorphismTest : JsonTestBase() {
 
     object PolyDefaultSerializer : JsonTransformingSerializer<PolyDefault>(PolyDefault.serializer(), "foo") {
         override fun readTransform(element: JsonElement): JsonElement {
+            (element.jsonObject.content as MutableMap).remove("type")
```
sandwwraith, line 51: I'd like to avoid such hacks in test code, because it may be an example for someone.

runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonDeserializePolymorphicTwiceTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    @Serializable
+    sealed class Foo {
+        @Serializable
+        data class Bar(val a: Int) : Foo()
+    }
+
+    @Test
+    fun testDeserializeTwice() {
+        val json = Json.toJson(Foo.Bar(1) as Foo)
```
sandwwraith, line 21: it seems like sealed classes do not have special `AssociatedObject`, so this lookup won't work. I'll address it as a separate issue

sandwwraith: #812

runtime/commonTest/src/kotlinx/serialization/features/PolymorphismTest.kt
sandwwraith: Seems you forgot to delete the line after fix?

runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonDeserializePolymorphicTwiceTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @Serializable
+    sealed class Foo {
+        @Serializable
+        data class Bar(val a: Int) : Foo()
+    }
+
+    @Test
+    fun testDeserializeTwice() = noNative { // #812
+        val json = Json.toJson(Foo.Bar(1) as Foo)
```
sandwwraith, line 22: I'd rather use `Json.toJson(Foo.serializer(), Foo.Bar(1))` instead of ignoring test on Native completely

Issue #806, opened 2020-04-16
Field with emptyList isn't decoded by ProtoBuf


SerVB:
**To Reproduce**
If I do:

```kotlin
@Serializable
data class A(val myList: List<String>)

object Main {

    @JvmStatic
    fun main(vararg args: String) {
        val protobuf = ProtoBuf()

        val a = A(emptyList())

        val encoded = protobuf.dump(A.serializer(), a)
        val decoded = protobuf.load(A.serializer(), encoded)

        println(decoded == a)
    }
}
```

I get the following in console:

```
Exception in thread "main" kotlinx.serialization.MissingFieldException: Field 'myList' is required, but it was missing
	at A.<init>(Main.kt)
	at A$$serializer.deserialize(Main.kt)
	at A$$serializer.deserialize(Main.kt:59)
	at kotlinx.serialization.protobuf.ProtoBuf$ProtobufReader.decodeSerializableValue(ProtoBuf.kt:290)
	at kotlinx.serialization.DecodingKt.decode(Decoding.kt:521)
	at kotlinx.serialization.protobuf.ProtoBuf.load(ProtoBuf.kt:548)
	at Main.main(Main.kt:71)
```

**Expected behavior**

It should output "true" to the console.

**Workaround**

As workaround, I use `data class A(val myList: List<String> = emptyList())`.

**Environment**
 - Kotlin version: 1.3.71
 - Library version: 0.20.0
 - Kotlin platforms: JVM (JS is also checked with similar code)

SerVB:
By the way, I've just seen the following in the [docs](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/runtime_usage.md#protobuf):

> Because format spec says that if the list is empty, there will be no elements in the stream with such tag, you must explicitly mark any field of list type with default `= emptyList()`. Same for maps.

I think it should be done automatically.

qwwdfsad:
The compiler generates code in a general form for an abstract format and does not implies anything protobuf-specific. Adding an empty collection instead of a list is not a generally-applicable change.
Maybe we can make this change in the protobuf format itself, but we are not sure yet

SerVB:
I think possibility of runtime exceptions is a good enough reason to do something with this.

qwwdfsad:
The current solution is to fail-fast and to add an exception message that suggests adding an empty default value when attempting to serialize such value. It should prevent most of the bugs during the initial testing.

[...]

PR #811, opened 2020-04-21
Fail-fast for empty collections in Prorobuf. Add a meaningful excepti…

qwwdfsad:
…on that suggests how to fix this issue.

This is important in order to prevent situation when our implementation is capable of serializing something, but not deserializing it back

Addresses #806

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
@@ -223,12 +223,77 @@ public class ProtoBuf(
 
         override fun SerialDescriptor.getTag(index: Int) = extractParameters(index)
 
-        override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) = when {
-            serializer is MapLikeSerializer<*, *, *, *> -> {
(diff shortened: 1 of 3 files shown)


sandwwraith:
I doubt that this is the solution we want, because:

1. It adds some overhead each time we encounter a collection
2. It blocks use-case when protobuf is written to be read by another framework (e.g. grpc/ipc)
3. It does not help at all when deserializer reads protobuf produced by another framework.

A more helpful solution is to catch `MissingFieldException` and perform analogous operations there, unless proper 'rollback' mechanism is implemented.

qwwdfsad:
Let me clarify it.

1) Overhead (size check) is really negligible compared to the required work in order to serialize collection.
2) It's unlikely that someone is using kotlinx.serialization-based classes in order to serialize them to Protobuf that is consumed by **schemaful** non-kotlinx.serialization-consumer. 
Even if so, we can provide a temporary configuration flag to ignore it in future releases
3) Again, this is unlikely the case. And it won't get worse here anyway.


>A more helpful solution is to catch MissingFieldException ...

It is not. Current behaviour breaks our own invariant "we do not serialize what we can't deserialize". And this behaviour can be encountered *too late* when a lot of data is already serialized, providing a really frustrating user-experience.

qwwdfsad:
Won't fix, we'll actually support this functionality

PR #820, opened 2020-04-27
Treat unmatched JSON type during element deserialization of decoding …

### LINKED ISSUE #816: deserialization: IllegalStateException when null occurs in JSON input in place of an expected non-null object
`SerializationException` should be thrown instead.

Example: try to parse this JSON:

```
{"foo":null}
```

for deserializing this class:

```
class MyClass(val foo: SomeObject)
```

This exception is thrown:

```
    java.lang.IllegalStateException: Expected class kotlinx.serialization.json.JsonObject but found class kotlinx.serialization.json.JsonNull
        at kotlinx.serialization.json.internal.AbstractJsonTreeInput.beginStructure(TreeJsonInput.kt:224)
```

qwwdfsad:
…exception instead of ISE to maintain decoding invariant

Fixes #816

### DIFF
--- a/runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonPolymorphismExceptionTest.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonPolymorphismExceptionTest.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.json.polymorphic
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import kotlinx.serialization.modules.*
+import kotlin.test.*
+
+class JsonPolymorphismExceptionTest : JsonTestBase() {
+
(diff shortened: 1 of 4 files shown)


qwwdfsad:
One more native problem, PTAL

sandwwraith:
It's the same one, I think

runtime/commonTest/src/kotlinx/serialization/json/polymorphic/JsonPolymorphismExceptionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    @Test
+    fun testDecodingException() = parametrizedTest { useStreaming ->
+        val serialModule = SerializersModule {
+            polymorphic(Base::class) {
+                subclass<Derived>()
+            }
+        }
+
+        assertFailsWith<JsonDecodingException> {
```
sandwwraith, line 32: Don't you want to check the exception message?

qwwdfsad: I usually avoid it

PR #843, opened 2020-05-21
1.4-M2 & HMPP fixes

h0tk3y:


README.md
```diff
@@ -224,7 +222,6 @@ sourceSets {
     jvmMain {
         dependencies {
             implementation kotlin('stdlib-jdk8')
```
sandwwraith, line 224: Can JVM-only users still use `kotlinx-serialization-runtime` artifact or they have to migrate to `kotlinx-serialization-runtime-jvm`?

gradle.properties
```diff
@@ -5,20 +5,21 @@
...
 junit_version=4.12
 infra_version=0.1.0-dev-47
-bintray_version=1.8.4-jetbrains-5
-benchmarks_version=0.1.7-dev-14
 jackson_version=2.10.0.pr1
 dokka_version=0.10.1
 native.deploy=false
 
+kotlin.mpp.enableGranularSourceSetsMetadata=true
+kotlin.mpp.enableCompatibilityMetadataVariant=true
```
sandwwraith, line 21: What does this flag do?

README.md
sandwwraith: Ok, judging by `publishPlatformArtifactsInRootModule` the answer is yes

dsavvinov: Yes, exactly

gradle.properties
h0tk3y: `enableCompatibilityMetadataVariant` is a special mode for libraries published with HMPP that adds an old-style `*.kotlin_metadata` variant that can be consumed by non-HMPP users. It is only useful for migration period until we enable HMPP by default.

dsavvinov: In HMPP we've changed binary format of artifact published from common code. Previously it was a `.jar`-file with `.kotlin_metadata` inside, now it's a `.klib` without IR (well, essentially it's still an archive with `.kotlin_metadata` inside, but with different extension and layout)

This means that non-HMPP clients won't be able to consume that artifact. At the moment,  HMPP is not enabled by default, hence we need to publish a compatibility variant with old-style metadata .jar for anyone who hasn't enabled HMPP manually. 

If we enable HMPP by default in 1.4, then, most probably, we'll be able to drop this legacy-variant.

dsavvinov:
> And I assume you have already tested teamcity uploading to bintray?

Yes, we've deployed HMPP libs to bintray multiple times via teamcity during development/testing. If you have some specific case in mind, let us know, we'll re-check it

gradle/publish-mpp-root-module-in-platform.gradle
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+
+/** Publish the platform JAR and POM so that consumers who depend on this module and can't read Gradle module
+ metadata can still get the platform artifact and transitive dependencies from the POM: */
+project.ext.publishPlatformArtifactsInRootModule = { platformPublication ->
```
qwwdfsad, line 8: Please leave a comment to https://youtrack.jetbrains.com/issue/KMM-237#focus=streamItem-27-4115233.0-0 here

dsavvinov: Done!

Issue #842, opened 2020-05-19
Support encoding and decoding CBOR Byte Strings (major type 2)


Chris-Corea:
**What is your use-case and why do you need this feature?**

**Use Case:** I need to send a CBOR payload to an external device, who's firmware expects certain data to be in the `Byte String` format.

As I can't modify the firmware to accept an array of bytes as opposed to a byte string, I need this library to support encoding CBOR's `Byte String` type.

**Describe the solution you'd like**

As described by the _Major Types_ section in [CBOR RFC-7049](https://tools.ietf.org/html/rfc7049#section-2.1), CBOR includes a major type specifically for byte strings. In instances where a service or receiver expects a byte string, a user of this library's existing CBOR serialization implementation would be unable to send a byte string as the [`Cbor` class](https://github.com/Kotlin/kotlinx.serialization/blob/fa134d9cdf06b7656bd67dcc0c151ff9e4cb7e87/formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt#L370-L373) only includes support for the `Text`, `Negative Integer`, `Array` and `Map` major types.

Ideally, as a user of the library, it would be great to be able to either 

1. Wrap my `ByteArray` in a type that has its own serializer (could probably be the same as the `ByteArraySerializer`) and a descriptor that defaults to a CBOR specific `SerialKind` (`BYTE_STRING`); or

2. Use a `ByteStringSerializer()` to provide to my own custom serializer; or

3. Create my own serializer within my class and instruct the encoder to explicitly encode a provided `ByteArray` as a ByteString.

The `Cbor` class would need to include the byte string header type (`0b010_00000`) in its `Default` companion object and update the `CborWriter`, `CborEncoder`, `CborReader`, and `CborDecoder` to handle that type appropriately.

twyatt:
I tried taking the approach outlined in https://github.com/Kotlin/kotlinx.serialization/pull/846#issuecomment-637760591 (whereas encoding/decoding of `ByteArray` can be controlled via annotation): https://github.com/Kotlin/kotlinx.serialization/compare/dev...twyatt:CBOR/ByteString

[...]

PR #846, opened 2020-05-28
CBOR ByteString serialization

Chris-Corea:
I recently opened Issue #842 to request CBOR ByteString support. I went ahead and took a stab at implementing bytestring serialization for the CBOR format while (hopefully) conforming to project's code structure and paradigms.

## Changes Proposed

This change allows for the encoding of fixed-length `ByteStrings` and the decoding of definite-length or indefinite-length `ByteStrings`.

Encoding and Decoding "rules" can be found in [Section 2.1 of the CBOR RFC](https://tools.ietf.org/html/rfc7049#section-2.1) (major type 2). 

Information about treating indefinite-length byte strings can be found in [Section 2.2.2 of the CBOR RFC](https://tools.ietf.org/html/rfc7049#section-2.2.2).

**Note**: ByteStrings who's lengths don't fit in 4 bytes are considered unsupported with these changes. This matches [FasterXML/Jackson's current behavior](https://github.com/FasterXML/jackson-dataformats-binary/blob/bc6318a7fc6867bd2d700a4f138a634e031c0915/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2904-L2935).


x-rat4-x:
It seems strange that public format-agnostic `Encoder/Decoder` expose internal format-specific `ByteString`

Chris-Corea:
> It seems strange that public format-agnostic `Encoder/Decoder` expose internal format-specific `ByteString`

Yeah, that's a great catch. Considering that `ByteString` would be format-specific, I'll move it to the CBOR package and update the `Encoder/Decoder`s. I'll push an update soon.

sandwwraith:
Hi, 
thanks for your suggestions. I have a few notes about the proposed design:

1. Introducing new SerialKind, `VALUE_TYPE` seems unnecessary -- it is not used anywhere in the new code (except ByteString, but it is not used further in cbor itself). Also, I do not quite understand what problem this new kind solves. In general, we try to establish the correspondence between Kotlin concepts and kinds, but Kotlin does not have classes that are 'value types'. And I strongly advocate against baking something format-related into core API. VALUE_TYPE is format-related, because it may be crucial for some formats to actually receive begin/endStructure calls. Calling `encodeXxx` several times for some formats may lead to unexpected results.

2. Do we really need to differentiate ByteString vs ByteArray? While format-special classes indeed exist in serialization (JsonElement), they represent a special container (e.g Map<Stirng, Any> would be less typesafe and Any is not sealed clas) and special requirements (e.g. parser lookahead). For an arbitrary set of bytes, the container already exists: ByteArray. The 'byte string' thing is just an internal format representation of a bunch of bytes.

To conclude, I think it's better to avoid baking 'special ByteString' into a library. What can be done instead:  adding special handling for byte arrays in cbor format so they are encoded as byte strings. This is something that is done in protobuf: https://github.com/Kotlin/kotlinx.serialization/blob/938a35f4922d0dfad389fdd73bc163d0772245d9/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt#L230 -- byte arrays there are encoded as protobuf's byte arrays. Profits of such a design are:
1. No special class needed; class remains serializable even after format change.
2. No special CborEncoder/CborDecoder and new serial kinds are needed.
3. If someone wants to encode custom class as a byte string, they still can convert it to byte array and encode with `ByteArraySerializer` -- format would still recognize this.

Intercepting `encodeSerializableValue` is a recommended approach to add format-specific support for a certain class, so I suggest you should start with overriding this method in `CborEncoder`/`CborDecoder`.

Chris-Corea:
Thanks for the feedback. Let me preface this with saying: I'm happy to rework this to follow your suggestion of intercepting `encodeSerializableValue`. Here's just a few thoughts I had while working through this and where my hesitation to intercept `encodeSerializableValue` comes from. 

I agree that the preferred approach should be to utilize (or repurpose) an existing construct to handle something like a CBOR ByteString. Before I started with my implementation I went through to see if there was something similar already done for the other formats and noticed Protobuf's `ByteArray` handling. My first stab at this involved doing basically that same thing for CBOR, but I found that this approach makes it so that a consumer of the library would no longer be able to encode a `ByteArray` as an array of bytes. 

My thinking was/is that the CBOR format should support encoding an array regardless of its element type as well as support encoding bytestrings; which, in this case, basically boils down to support for both an array of bytes and bytestrings since they are both valid CBOR. I didn't want to make assumptions of the data a consumer could potentially send or receive so I thought handling arrays and bytestrings separately would make the most sense. 

I think it's also worth mentioning that overriding `encodeSerializableValue` works so well for Protobuf because there is no notion of an array of bytes in Protobuf (as far as I know). Whereas in CBOR (and as I mentioned before), encoding a `ByteArray` as an array of bytes _or_ a ByteString is totally valid.

> Introducing new SerialKind, `VALUE_TYPE` seems unnecessary -- it is not used anywhere in the new code (except ByteString, but it is not used further in cbor itself). Also, I do not quite understand what problem this new kind solves.

My goal was to create a `SerialKind` that could be used to identify a structure that didn't make use of begin/endStructure calls (like bytestrings). I tried to make this clear with the doc comment I wrote above `VALUE_TYPE`. It didn't seem like there was a viable `SerialKind` option that I could use to represent such a structure. But given your comment, I think I misunderstood what these were meant to represent and see now that I could have used `SerialKind.OBJECT`.

> Do we really need to differentiate ByteString vs ByteArray?

Yeah, another really good question, and if we go with the Protobuf approach of intercepting `encodeSerializableValue`, then no, I don't think we would need to draw a distinction between the two. However, if my point above about the CBOR format supporting array serialization regardless of the element's type rings true, then I think we do need to draw a distinction to enable users of the library to make the choice of how they want to serialize some bytes.

twyatt:
@sandwwraith to put what @Chris-Corea said another way:

CBOR RFC defines [6 major types](https://tools.ietf.org/html/rfc7049#section-2.1), 2 of which make sense to represent as Kotlin's `ByteArray`:

- **Major type 2**: a byte string.
- **Major type 4**: an array of data items
    - _This is the current encoding used by `kotlinx.serialization` for `ByteArray`s_

The dilemma then is how to communicate to `kotinx.serialization which CBOR major type to encode a `ByteArray` as, for example:

```kotlin
@Serializable
data class Data(
    val a: ByteArray, // CBOR Major type 2
    val b: ByteArray // CBOR Major type 4
)
```

Instead of defining a new type (e.g. `ByteString`) would it be possible (and make sense to) have an annotation to differentiate the desired CBOR encoding? For example:

```kotlin
@Serializable
data class Data(
    @Cbor(type = 2) val a: ByteArray,
    @Cbor(type = 4) val b: ByteArray
)
```

sandwwraith:
Sorry for the late response; I see, there's a point in differentiating between ByteString and ByteArray, because CBOR supports both. Are there any real examples of usage of this difference?

Anyway, from Kotlin's point of view, there's no difference in ByteString and ByteArray - they're both a sequence of bytes, so still, no new SerialKind is needed.

I also like the idea of controlling CBOR's output via a special annotation -- this approach seems to give benefits of both sides: one still controls what data type is used in cbor encoding, and it is still possible to use the same class with other formats.

twyatt:
@sandwwraith continuing discussion in #842 where hopefully we can get some guidance for alternate approaches.

Chris-Corea:
Closing in favor of the [ByteString annotation method](https://github.com/JuulLabs/kotlinx.serialization/pull/1)

PR #851, opened 2020-06-03
Add IDEA/Github integration to make all issue references navigable

qwwdfsad:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 .DS_Store
 
 # IntelliJ files
-.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,7 +1,8 @@
 .DS_Store
 
 # IntelliJ files
-.idea/
+**/.idea/*
+!/.idea/vcs.xml
```
sandwwraith, line 5: You can add a comment that it is required for GitHub integration

PR #852, opened 2020-06-03
Fix license file to match kotlinx.coroutines

benasher44:
The Apache 2.0 license wasn't properly filled out for this project.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


benasher44:
Nvm realizing this is the license explaining how to apply it to header files :facepalm:

JakeWharton:
I was about to jump in! This is my pet peeve. Funny enough the main Kotlin repo gets it wrong. I tried arguing that this was the template and you're not actually supposed to fill it out there but they didn't agree despite the fact that no project in the actual Apache foundation changes this file.

Issue #90, opened 2018-02-22
Default value for unknown Enum value


cosminstefanxp:
In order to be forward-compatible and support certain scenarios, it would be useful to have a way to specify a default value for a deserialized `Enum` field, if an unknown value is encountered. That way, if a new type is defined and being returned by remote APIs, apps already in production will work properly in the intermediate period before migration. 

Example:
```kotlin
enum class Status { ACTIVE, INACTIVE, UNKNOWN }
class Test(val status: Status = Status.UNKNOWN)
```
would return a `Test(UNKNOWN)` when provided with the following json: `{'status': 'PENDING'}`

simonovvasiliy:
Good use case, Jackson can do it and we rely its feature

avently:
@sandwwraith 
A year is gone. Any news related to enums?

sandwwraith:
@avently Not yet

Larsmyrup:
This would be really useful. Does anyone know if this will be implemented?

santoshastagi:
This is important for us too. Any other ways to do this?

ManuelDCR:
This would be a really appreciated feature!
Any news @sandwwraith ?

arberg:
This is just to help others who need to solve this issue here and now: Bolinfest showed us how to do create a custom serializer [here on issue #31 ](https://github.com/Kotlin/kotlinx.serialization/issues/31#issuecomment-488572539)

igrmk:
For those who need a solution here is a complete answer. For simplicity reasons I used `toString` instead of `@SerialName` however.
https://github.com/igrmk/smsq/blob/25e9dd454fb131d93503e8471f71b9e9c1d280c4/android/app/src/main/java/com/github/igrmk/smsq/entities/DeliveryResult.kt

azabost:
If you are interested in even better forward compatibility of enums, I would recommend using a wrapper like this: https://github.com/bright/codified. I have been using it for a quite long time and it works beautifully. Feel free to let me know what you think about it :blush:

tagy:
Enabling `coerceInputValues` and setting default enum values in you class was easiest way to solve this at the moment I think.

arberg:
> For those who need a solution here is a complete answer. https://github.com/igrmk/smsq/blob/532e3468266c8fc319c8ecca11be76f7026949a7/android/app/src/main/java/com/github/igrmk/smsq/entities/DeliveryResult.kt

[...]

PR #853, opened 2020-06-04
Introduce 'treatNullAsMissing' flag

### LINKED ISSUE #246: Use default when passed value is null
When deserializing a json and we get a null value passed down, an annotation/setting should be allowed to instead use the default value. The following scenario will fail if deserialize because it is looking for a non-null value

```
{ 
"name": "john",
"number": null
}

data class Person(val name: String = "", val number: String ="")
```
If we had an annotation that specified to use default if null, the the following would deserialize correctly using the default value.
```
{ 
"name": "john",
"number": null
}

data class Person(val name: String = "", @UseDefault val number: String ="")
```

If this exists then please point me to the documentation :) Thanks.

sandwwraith:
Treat `null` literals in JSON as a missing value indicator, allowing parsing `"x":null` to a non-nullable `x: Boolean = false`

It also allows treating unknown enum values as missing ones, so this flag may need a better name; 'treatNullOrUnknownAsMissing' sounds long and cumbersome.

Closes #246 and #90 (only for json, though)

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -279,7 +279,7 @@ public class JsonBuilder {
     public var prettyPrint: Boolean = false
     public var unquotedPrint: Boolean = false
     public var indent: String = "    "
-    public var treatNullAsMissing: Boolean = false
+    public var coerceInputValues: Boolean = false
     public var useArrayPolymorphism: Boolean = false
     public var classDiscriminator: String = "type"
     public var serialModule: SerialModule = EmptyModule
@@ -294,7 +294,7 @@ public class JsonBuilder {
             prettyPrint,
             unquotedPrint,
             indent,
-            treatNullAsMissing,
+            coerceInputValues,
             useArrayPolymorphism,
             classDiscriminator
         )
--- a/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -279,6 +279,7 @@ public class JsonBuilder {
     public var prettyPrint: Boolean = false
     public var unquotedPrint: Boolean = false
     public var indent: String = "    "
+    public var treatNullAsMissing: Boolean = false
     public var useArrayPolymorphism: Boolean = false
     public var classDiscriminator: String = "type"
     public var serialModule: SerialModule = EmptyModule
@@ -293,6 +294,7 @@ public class JsonBuilder {
             prettyPrint,
             unquotedPrint,
             indent,
+            treatNullAsMissing,
             useArrayPolymorphism,
             classDiscriminator
         )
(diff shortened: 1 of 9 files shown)


runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -279,6 +279,7 @@ public class JsonBuilder {
     public var prettyPrint: Boolean = false
     public var unquotedPrint: Boolean = false
     public var indent: String = "    "
+    public var treatNullAsMissing: Boolean = false
```
qwwdfsad, line 282: please rename it to `coerceInputValues` as we've discussed

runtime/commonTest/src/kotlinx/serialization/json/JsonUseDefaultOnNullTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    private inline fun <T> doTest(inputs: List<String>, expected: T, serializer: KSerializer<T>) {
+        for (input in inputs) {
+            parametrizedTest(json) {
+                assertEquals(expected, parse(serializer, input), "Failed on input: $input")
+            }
+        }
+    }
+
+    @Test
+    fun testBoolean() = doTest(
```
qwwdfsad, line 37: Please consider giving tests meaningful names for the sake of better readability

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -108,6 +109,19 @@ internal class StreamingJsonInput internal constructor(
         }
     }
 
+    private fun treatAsMissing(descriptor: SerialDescriptor, index: Int): Boolean {
+        if (!configuration.treatNullAsMissing) return false
```
qwwdfsad, line 113: I'd suggest extracting this check to the caller to help the inliner

runtime/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonInput.kt
```diff
@@ -108,6 +109,19 @@ internal class StreamingJsonInput internal constructor(
...
+        val elementDescriptor = descriptor.getElementDescriptor(index)
+        if (reader.tokenClass == TC_NULL && !elementDescriptor.isNullable) return true // null for non-nullable
+        if (elementDescriptor.kind == UnionKind.ENUM_KIND) {
+            val enumValue = reader.peekString(configuration.isLenient)
+                    ?: return false // if value is not a string, decodeEnum() will throw correct exception
+            val enumIndex = elementDescriptor.getElementIndex(enumValue)
+            if (enumIndex == UNKNOWN_NAME) return true
```
qwwdfsad, line 120: `return enumIndex == UNKNOWN_NAME` is a bit more straightforward

sandwwraith: I'd prefer to stick with the current version because it's easier to spot places where the function returns true, and the only place where it returns false is at the end. But I can change this if you insist.

Issue #862, opened 2020-06-09
Android Native Targets Feature Request


senagbe:
**What is your use-case and why do you need this feature?**
We are trying to share code between multiple platforms. The common code is buildable for desktop and ios mobile native but not natively for Android.

We wish to be able to build an Android library which also has uses kotlinx.serialization in its native code.

**Describe the solution you'd like**
Add Android Targets for Arm32/Arm64 and X86/X86_64

senagbe:
@sandwwraith is there anything stopping us from submitting a PR for this?

senagbe:
@qwwdfsad where can I find the tests and TeamCity mentioned here? I'm happy to put some work in to make sure the tests pass https://github.com/Kotlin/kotlinx.serialization/pull/871#discussion_r444076074

qwwdfsad:
Thanks for your enthusiasm @senagbe!

Unfortunately, adding a corresponding target and tweaking TeamCity configuration is not enough.
It also requires dedicated TeamCity agents (whether bare-metal or properly virtualized) that is capable of running arm32.

Configuring such agents requires serious effort from our build team (in order to make them consistent, cloneable, observable), and, frankly speaking, their resources are seriously limited right now, especially in the face of upcoming Kotlin 1.4 release. Long story short, _currently_ we neither have an infrastructure to test and release arm binaries **on a regular basis** nor resources to establish this infrastructure. We'll return to this problem right after Kotlin 1.4 post-release activities.

For now, the easiest solution will be to publish these binaries on your own side, either in your organization bintray on directly in maven local of the target machine (if you are going to build your native app right there). 
It can be achieved by adjusting serialization version in `gradle.proprerties` and adding the corresponding target to `native_mpp.gradle`. Feel free to ask any questions if you are struggling with building it locally.

senagbe:
@qwwdfsad thanks for this, please let m know what the status of this is as we are post 1.4.0. I'm interested to know what kind of resources are required and if there is any way I can help.

VVictorWang:
@qwwdfsad Hello. I also need this feature urgently. I'm wondering is there any plan for this feature? Thanks a lot.

TheTreek:
Has there been any updates on this? It's been quite a while.

sandwwraith:
These targets will be added in 1.5.1

PR #871, opened 2020-06-22
Implement some FRs

sandwwraith:
* Android NDK targets
* Rename UnitSerializer to Unit.serializer

### DIFF
--- a/gradle/native_mpp.gradle
+++ b/gradle/native_mpp.gradle
@@ -45,6 +45,12 @@ kotlin {
 
             // WASM
             addTarget(presets.wasm32)
+
+            // Android NDK
+            addTarget(presets.androidNativeArm32)
+            addTarget(presets.androidNativeArm64)
+            addTarget(presets.androidNativeX86)
+            addTarget(presets.androidNativeX64)
         }
     }
 
--- a/gradle/native_mpp.gradle
+++ b/gradle/native_mpp.gradle
@@ -45,12 +45,6 @@ kotlin {
 
             // WASM
             addTarget(presets.wasm32)
-
-            // Android NDK
-            addTarget(presets.androidNativeArm32)
-            addTarget(presets.androidNativeArm64)
-            addTarget(presets.androidNativeX86)
-            addTarget(presets.androidNativeX64)
         }
     }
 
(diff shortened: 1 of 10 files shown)


runtime/commonMain/src/kotlinx/serialization/builtins/PrimitiveSerializers.kt
```diff
@@ -99,10 +99,20 @@ public fun Boolean.Companion.serializer(): KSerializer<Boolean> = kotlinx.serial
...
 /**
- * Returns serializer for [UnitSerializer] with [descriptor][SerialDescriptor] of [StructureKind.OBJECT] kind.
+ * Returns serializer for [Unit] with [descriptor][SerialDescriptor] of [StructureKind.OBJECT] kind.
  */
+@Deprecated(
+    message = "Deprecated in the favour of Unit.serializer() extension",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("Unit.serializer()", imports = ["kotlinx.serialization.builtins.serializer"])
+)
 public fun UnitSerializer(): KSerializer<Unit> = kotlinx.serialization.internal.UnitSerializer
```
qwwdfsad, line 109: Please move this declaration to Migration.kt to keep them in one pleace

gradle/native_mpp.gradle
```diff
@@ -45,6 +45,12 @@ kotlin {
 
             // WASM
             addTarget(presets.wasm32)
+
+            // Android NDK
+            addTarget(presets.androidNativeArm32)
+            addTarget(presets.androidNativeArm64)
+            addTarget(presets.androidNativeX86)
+            addTarget(presets.androidNativeX64)
```
qwwdfsad, line 53: Have you checked that all these targets are being tested on our TeamCity, can be depend upon and are generally valid?

sandwwraith: It builds locally, and it builds on TeamCity. Not sure if we have any dedicated Gradle tools to test any targets except the host and iOS ones.

sandwwraith:
@senagbe You can checkout branch from this PR and build it with `./gradlew publishToMavenLocal -Pnative.deploy=true`. Android targets should be available in maven local after that.

runtime/commonMain/src/kotlinx/serialization/builtins/PrimitiveSerializers.kt
sandwwraith: There's no Migration.kt in builtins package

qwwdfsad:
Please merge with a sqaush

senagbe:
@sandwwraith I see #862 this has been reverted. Did it fail tests? How can I help?
Also, should we re-open the issue?

sandwwraith:
@senagbe Unfortunately, there is no testing infrastructure for android native targets at this moment. Since our goal is to provide a quality library, publishing artifacts that have not been tested is a bad practice. You still can build and use them if they work for you, without particular guarantees.

senagbe:
Thanks @sandwwraith I'm going to look at how to add tests for this when I have some bandwidth. I'd expect them to be essentially identical to the Linux tests.

PR #872, opened 2020-06-23
[API stabilization] Hide public IO classes, copy them to ProtoBuf and…

qwwdfsad:
… CBOR, remove redundant symbols and make internal in order to reduce public API shape


formats/cbor/commonMain/src/kotlinx/io/Streams.kt
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.io
```
sandwwraith, line 5: Maybe get rid of this package too? Place them in kotlinx.serialization.internal or kotlinx.serialization.cbor(.internal)

sandwwraith:
lgtm

PR #873, opened 2020-06-23
[API stabilization] Serial format renames

qwwdfsad:
* Native tests are moved because they were replicating common tests
* Companion migrations to JSON are not added because of our plans to make companion implement JSON
* .md documentation is not touched as it's being rewritten anyway
* Benchmarks are not changed, will do later


runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -149,5 +159,85 @@ public fun <T : Any> SerialModule.getContextualOrDefault(value: T): KSerializer<
...
+
+@Deprecated(
+    "This method was renamed to decodeFromString during serialization 1.0 stabilization",
+    ReplaceWith("decodeFromString<T>(string)"), DeprecationLevel.ERROR
+)
+public fun <T : Any> StringFormat.parse(string: String): T = noImpl()
+
+@Deprecated(
+    "This method was renamed to encodeToString during serialization 1.0 stabilization",
+    ReplaceWith("encodeToString<T>(serializer, value)"), DeprecationLevel.ERROR
```
sandwwraith, line 187: Type argument seems unnecessary for all two-arguments invocations. Does IDEA automatically remove it?

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -108,23 +110,31 @@ public annotation class SerialId @Deprecated(
...
 public inline fun <reified T : Any> StringFormat.parseList(objects: String): List<T> =
-    parse(context.getContextualOrDefault<T>().list, objects)
+    decodeFromString(context.getContextualOrDefault<T>().list, objects)
 
-@Deprecated(level = DeprecationLevel.WARNING, message = "Use default parse overload instead", replaceWith = ReplaceWith("parse(map)"))
-public inline fun <reified K : Any, reified V : Any> StringFormat.parseMap(map: String): Map<K, V>
-        = parse(MapSerializer(context.getContextualOrDefault<K>(), context.getContextualOrDefault<V>()), map)
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "Use default parse overload instead",
```
sandwwraith, line 117: Update overload text so it does not mention obsolete parse method

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -108,23 +110,31 @@ public annotation class SerialId @Deprecated(
...
+    replaceWith = ReplaceWith("parse(map)")
+)
+public inline fun <reified K : Any, reified V : Any> StringFormat.parseMap(map: String): Map<K, V> =
+    decodeFromString(MapSerializer(context.getContextualOrDefault<K>(), context.getContextualOrDefault<V>()), map)
 
 // ERROR migrations that affect **only** users that called these functions with named parameters
 
 @LowPriorityInOverloadResolution
 @Deprecated(level = DeprecationLevel.ERROR, message = "Use default stringify overload instead", replaceWith = ReplaceWith("stringify(objects)"))
```
sandwwraith, line 126: Same about stringify

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
qwwdfsad: This particular replacement doesn't work without explicit type argument for some reason

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -19,7 +19,7 @@ import kotlin.reflect.*
  * or for converting objects to [JsonElement] back and forth.
  *
  * This is the only serial format which has first-class [JsonElement] support.
- * Any serializable class can be serialized to or from [JsonElement] with [Json.fromJson] and [Json.toJson] respectively or
+ * Any serializable class can be serialized to or from [JsonElement] with [Json.decodeFromJsonElement] and [Json.encodeToJsonElement] respectively or
```
dlenskiSB, line 22: The current docs _don't mention that this was changed_ in 1.7.3: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-from-json-element.html

And there appears to be **no way to see docs for older versions of the library** from the web-based docs viewer: https://discuss.kotlinlang.org/t/language-docs-for-previous-kotlin-versions/28800 

… which combines to make the Kotlin developer experience very painful for those of us who can't use the latest-version-of-everything.

Python really gets these documentation and versioning issues right in a way that Kotlin doesn't: https://discuss.kotlinlang.org/t/language-docs-for-previous-kotlin-versions/28800/2

sandwwraith: This is a 2020 PR with pre-1.0 names. Can you please clarify what exactly was changed in 1.7.3?

dlenskiSB: Apologies, I completely misread the tags and the _name_ change apparently occurred far earlier 🤦🏻

PR #875, opened 2020-06-26
[API stabilization] Make JSON sealed class, implement it in its companion and internal im…

qwwdfsad:
…plementation. Replace public constructors with factories, extract reified methods into extensions.

Notes: annotations on constructors are not preserved, I expect to tweak them when working on our DSL plan

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -1,7 +1,7 @@
 /*
  * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-@file:Suppress("EXPERIMENTAL_API_USAGE")
+@file:OptIn(UnstableDefault::class)
 
 package kotlinx.serialization.json
 
@@ -54,7 +54,16 @@ public sealed class Json(internal val configuration: JsonConfiguration) : String
      * @throws [JsonException] if the given value can not be encoded to JSON
      * @throws [SerializationException] if the given value cannot be serialized
      */
-    public abstract override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String
+    public final override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String {
+        val result = StringBuilder()
+        val encoder = StreamingJsonEncoder(
+            result, this,
+            WriteMode.OBJ,
+            arrayOfNulls(WriteMode.values().size)
+        )
+        encoder.encode(serializer, value)
+        return result.toString()
+    }
 
     /**
      * Deserializes the given JSON [string] into a value of type [T] using the given [deserializer].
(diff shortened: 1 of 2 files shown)


runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -140,29 +93,10 @@ public constructor(
...
-
-    /**
-     * The default instance of [Json] in the form of companion object. Configured with [JsonConfiguration.Default].
-     */
-    @UnstableDefault
-    public companion object Default : StringFormat {
-
-        private val jsonInstance = Json(JsonConfiguration.Default)
+    public companion object Default : Json(JsonConfiguration.Default) {
+        private val jsonInstance = Json(JsonConfiguration.Default, EmptyModule)
```
sandwwraith, line 99: Why do we have to use delegation here? I think we can move all implementation from JsonImpl to Json, because the only difference between Default and JsonImpl is configuration, which can be passed to `Json()` constructor.

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -173,48 +107,48 @@ public constructor(
...
-        /**
-         * @see Json.decodeFromJsonElement
-         */
-        public inline fun <reified T : Any> decodeFromJsonElement(tree: JsonElement): T =
-            decodeFromJsonElement(context.getContextualOrDefault(), tree)
-    }
+/**
+ * DSL-like constructor for [Json].
+ */
+public fun Json(block: JsonBuilder.() -> Unit): Json = JsonImpl(JsonBuilder().apply { block() })
```
sandwwraith, line 129: Constructor refactoring is postponed to next PR?

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -276,3 +210,47 @@ private val defaultJsonModule = serializersModuleOf(
 )
 
 internal const val lenientHint = "Use 'JsonConfiguration.isLenient = true' to accept non-compliant JSON"
+
+
+internal class JsonImpl(
```
sandwwraith, line 215: imo, `JsonImpl` in stacktraces looks very enterprise-ish and not nice to users, but I don't know how can we solve this.

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -1,12 +1,13 @@
 /*
  * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
+@file:Suppress("EXPERIMENTAL_API_USAGE")
```
sandwwraith, line 4: What's experimental in here?

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
qwwdfsad: Indeed

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
qwwdfsad: `@UnstableDefault`. It will be cleaned out along with ctor changes

sandwwraith: Why not `@OptIn(UnstableDefault)` then?

qwwdfsad: This one was suggested by the quickfix first :)

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
qwwdfsad: Good point, thanks for spotting

Issue #832, opened 2020-05-06
Make JsonConfiguration itself @Serializable


hfhbd:
**What is your use-case and why do you need this feature?**
I would like to store the JsonConfiguration itself in a Json file and read them at startup. This allows to use different behavior with a simple change in the config files.

eg having a prod configuration and a debug configuration, which use a less strict Json De-/Encoder or uses pretty print.

Of course, the JsonConfiguration for reading the json configs must be static defined and cannot be defined in the config file.
 
**Describe the solution you'd like**
Add @Serializable to JsonConfiguration

Current workaorund:
Implement a custom serializer, which I did for my project

sandwwraith:
Closing for the reason described in https://github.com/Kotlin/kotlinx.serialization/pull/877#pullrequestreview-441582372

hfhbd:
Its fine, my implementation was only adding 1 attribute :) But I would like to have the option changing the json behaviour with a config file, even with the new API.

sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/pull/877#issuecomment-1081826204

sandwwraith:
Not planned because it is still not possible to create `Json` from `JsonConfiguration` and `JsonConfiguration` constructor is intended to be internal.

PR #877, opened 2020-06-29
Made JsonConfiguration itself @Serializable (#832)

hfhbd:
This is the PR to fix (#832). Maybe there should be more tests...


sandwwraith:
@qwwdfsad My bad. I've marked corresponding FR as 'up for grabs' before the decision on hiding JsonConfiguration was made.

hfhbd:
Hey, JsonConfiguration is public again. So how about making JsonConfiguration `@Serializable` (again)? I still would like to config the serializer itself via different files (prod/dev) and would like to remove my custom encoder.

qwwdfsad:
We are looking into that, this is indeed worth doing, though we are not sure about its contract yet

PR #883, opened 2020-07-02
[API stabilization] Document core annotations, increase retention lev…

qwwdfsad:
…el to binary to enable potential future work


runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
 )
 
 /**
- * Instructs plugin to turn this class into serializer for specified class [forClass].
+ * Instructs the serialization plugin to turn this class into serializer for specified class [forClass].
  * However, it would not be used automatically. To apply it on particular class or property,
  * use [Serializable] or [UseSerializers], or [ContextualSerialization] with runtime registration.
  */
 @Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.BINARY)
```
qwwdfsad, line 82: This annotation is not documented intentionally, let's do it later along with extensive test suite

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
 import kotlin.reflect.*
 
 /**
- * Sets specific serializer for class, property or type argument.
- * When argument is omitted, plugin will generate default implementation inside the class.
+ * The main entry point to the serialization process.
+ * Applying [Serializable] to the Kotlin class instructs the serialization plugin to automatically generate [KSerializer]
```
sandwwraith, line 14:
```suggestion
 * Applying [Serializable] to the Kotlin class instructs the serialization plugin to automatically generate implementation of [KSerializer]
```

imo, 'generate [interface]' sounds clumsy

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ * for the current class, that can be used to serialize and deserialize the class.
+ * The generated serializer can be accessed with `.serializer()` extension method that is generated by the plugin as well.
```
sandwwraith, line 16: better mention that this is a companion/type extension method because you do not have to have a class instance.

Also, isn't it better to use 'function' instead of 'method'?

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ * class MyAnotherData(...)
+ *
+ * MyAnotherData.serializer() // <- returns MyAnotherDataCustomSerializer
+ * ```
+ *
+ * For annotated properties, specifying [with] parameter is mandatory and can be used to override
+ * serializer on the use-site without affecting the rest of the usages:
+ * ```
+ * @Serializable // By default is serialized as 3 byte components
+ * class RgbPixel(val red: Byte, val green: Byte, val blue: Byte)
```
sandwwraith, line 37: A common mistake that i've made in my talk once: RGB values are from 0 to 255, so `Byte` is not enough (unless you want to fiddle with sign)

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ * Box.serializer(Int.serializer()) // Returns serializer for Box<Int>
+ * Box.serializer(Box.serializer(Int.serializer())) // Returns serializer for Box<Box<Int>>
+ * ```
+ *
+ * ### Implementation details
+ *
+ * In order to generate `serializer` method that is not a method on the particular instance, the class should have a companion object, either named or unnamed.
+ * Companion object is generated by the plugin if it is not declared, effectively exposing both companion and `serializer()` method to class ABI.
+ * If companion object already exists, only `serializer` method will be generated.
+ * If `serializer` method already exists, it will be matched against the desired signature `serializer(vararg type params): KSerializer<T>`.
```
sandwwraith, line 63: `serializer(vararg type params)` -- we do not have such signature. generated serializer() function has exactly N KSerializer parameters, where N is number of type parameters.

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ * On successful match, the existing method will be used as [KSerializer] source instead of auto-generated one.
```
sandwwraith, line 64: But it won't be used in generated code. If you override serializer(), the compiler still inserts to other classes serializer which is specified in `@Serializable(with=...)`

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
-    val forClass: KClass<*> // what class to create serializer for
+    val forClass: KClass<*> //  target class to create serializer for
 )
 
 /**
- * Overrides name visible to the runtime part of serialization framework
+ * Overrides the name of the current element in the [SerialDescriptor].
+ *
+ * By default, [SerialDescriptor.serialName] and [SerialDescriptor.getElementName]
+ * are matched with fully-qualified name of the target class and the name of the property respectively.
```
sandwwraith, line 91:
```suggestion
 * are filled with fully-qualified name of the target class and the name of the property respectively.
```

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
 @Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.BINARY)
 public annotation class Serializer(
-    val forClass: KClass<*> // what class to create serializer for
+    val forClass: KClass<*> //  target class to create serializer for
 )
 
 /**
- * Overrides name visible to the runtime part of serialization framework
+ * Overrides the name of the current element in the [SerialDescriptor].
```
sandwwraith, line 88: It's not clear what 'current element' refers to (unless one carefully reads documentation to Encoder and SerialDescriptor). Better simply use 'class or property'

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ *     println(descriptor.getElementName(0))
+ * }
+ *
+ * // Prints "foo.Foo" and "myInt"
+ * printNames(Foo.serializer().descriptor)
+ *
+ * // Prints "Bar" and "int"
+ * printNames(Bar.serializer().descriptor)
+ * ```
+ * Element names and serial names are used by text-based serial formats in order to encode the name of the class or
```
sandwwraith, line 115: Better to put this in the beginning -- this is what most of the users want to know, introspecting descriptors is an advanced usage pattern.

Also, you may add code with Json to example.

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -10,60 +10,156 @@ import kotlinx.serialization.json.*
...
+ * Meta-annotation that commands the compiler plugin to handle the annotation as serialization-specific.
+ * Serialization-specific annotations are preserved in the [SerialDescriptor] and can be retrieved
+ * during serialization process with [SerialDescriptor.getElementAnnotations]:
+ *
+ * ```
+ * /**
+ *  * Specifies ProtoBuf unique id for a field in the protobuf message, assigned to a Kotlin property
+ *  */
+ * @SerialInfo
+ * @Target(AnnotationTarget.PROPERTY)
```
sandwwraith, line 147: This example is not useful unless it's shown how to retrieve it.

I suggest removing example completely, since declaring serialinfo annotations is mainly for format authors. For them, we'll eventually have an advanced guide on custom formats.

runtime/commonMain/src/kotlinx/serialization/Annotations.kt
qwwdfsad: Right now the following snippet just crashes the compiler:
```
    @Serializable
    class Foo(val a: Int) {

        companion object {
            fun serializer(): KSerializer<Foo> = error("Whoa!")
        }
    }

    @Serializable
    class Baz(val f: Foo)
```

I'll remove this paragraph

PR #892, opened 2020-07-08
[API stabilization] Builtin package stabilization

qwwdfsad:
* Deprecate redundant extensions
    * Merge all serializers into single file
    * Extract AbstractDecoder and AbstractEncoder to a separate package


runtime/commonMain/src/kotlinx/serialization/builtins/BuiltinsMigrations.kt
```diff
@@ -0,0 +1,31 @@
...
+    replaceWith = ReplaceWith("Unit.serializer()", imports = ["kotlinx.serialization.builtins.serializer"])
+)
+public fun UnitSerializer(): KSerializer<Unit> = Unit.serializer()
+
+@Deprecated(
+    message = "Deprecated during serialization 1.0 API stabilization",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("ListSerializer(this)")
+)
+public val <T> KSerializer<T>.list: KSerializer<List<T>>
```
sandwwraith, line 22: We can leave this function internal since it's heavily used in our tests

qwwdfsad: I'd prefer to use the same code style we recommend to our users

PR #899, opened 2020-07-10
[API stabilization] SerialDescriptor

qwwdfsad:
* Restrructure SerialKind
    * Replace methods that returned lists with vals that return iterable
    * Get rid of unnecessary allocation

### DIFF
--- a/integration-test/src/commonTest/kotlin/sample/BasicTypesSerializationTest.kt
+++ b/integration-test/src/commonTest/kotlin/sample/BasicTypesSerializationTest.kt
@@ -138,7 +138,7 @@ class BasicTypesSerializationTest {
             val name = inp.nextUntil(':', '}')
             if (name.isEmpty())
                 return CompositeDecoder.DECODE_DONE
-            val index = desc.getElementIndexOrThrow(name)
+            val index = desc.getElementIndex(name)
             inp.expect(':')
             return index
         }
--- a/runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
@@ -18,7 +18,7 @@ package kotlinx.serialization
  * Serial descriptor is identified by its [name][serialName] and consists of kind, potentially empty set of
  * children elements and additional metadata.
  *
- * * [serialName] uniquely identifies descriptor (and the corresponding serializer) for non-generic types.
+ * * [serialName] uniquely identifies the descriptor (and the corresponding serializer) for non-generic types.
  *   For generic types, the actual type substitution is omitted from the string representation and the name
  *   identifies the family of the serializers without type substitutions. However, type substitution is accounted
  *   in [equals] and [hashCode] operations, meaning that descriptors of generic classes with the same name, but different type
@@ -61,8 +61,8 @@ package kotlinx.serialization
  * number of actual children values. Valid indices range is not known statically
  * and implementations of descriptor should provide consistent and unbounded names and indices.
  *
- * In practice, it means that for regular classes it is allowed to invoke `getElement*(index)` methods
- * with index within `0 until elementsCount` range and element at the particular index corresponds to the
+ * In practice, for regular classes it is allowed to invoke `getElement*(index)` methods
+ * with an index within `0 until elementsCount` range and element at the particular index corresponds to the
  * serializable property at the given position.
  * For collections and maps, index parameter for `getElement*(index)` methods is effectively bound
  * by the maximal number of collection/map elements.
@@ -135,7 +135,7 @@ public interface SerialDescriptor {
      * Formats use serial kind to add and parse serializer-agnostic metadata to the result.
      *
      * For example, JSON format wraps [classes][StructureKind.CLASS] and [StructureKind.MAP] into
-     * brackets, while ProtoBuf just serialize these types in a separate ways.
+     * brackets, while ProtoBuf just serialize these types in separate ways.
      *
      * Kind should be consistent with the implementation, for example, if it is a [primitive][PrimitiveKind],
      * then its elements count should be zero and vice versa.
@@ -162,15 +162,15 @@ public interface SerialDescriptor {
     public val elementsCount: Int
 
     /**
-     * Returns a serial annotations of the associated class.
-     * Serial annotations can be used to specify an additional
-     * metadata that may be used during serialization, for example a [serial id][SerialId].
+     * Returns serial annotations of the associated class.
+     * Serial annotations can be used to specify an additional metadata that may be used during serialization.
+     * Only annotations marked with [SerialInfo] are added to the resulting list.
      */
     public val annotations: List<Annotation> get() = emptyList()
 
     /**
-     * Returns a _positional_ name of the child at the given [index].
-     * Positional name usually represents a corresponding property name in the class, associated with
+     * Returns a positional name of the child at the given [index].
+     * Positional name represents a corresponding property name in the class, associated with
      * the current descriptor.
      *
      * @throws IndexOutOfBoundsException for an illegal [index] values.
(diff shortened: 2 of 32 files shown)


integration-test/src/commonTest/kotlin/sample/BasicTypesSerializationTest.kt
```diff
@@ -138,7 +138,7 @@ class BasicTypesSerializationTest {
             val name = inp.nextUntil(':', '}')
             if (name.isEmpty())
                 return READ_DONE
-            val index = desc.getElementIndexOrThrow(name)
+            val index = desc.getElementIndex(name)
```
sandwwraith, line 141: why not `orThrow`? at least add manual `check(index != UNKNOWN_NAME)`

runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
```diff
@@ -117,7 +117,7 @@ package kotlinx.serialization
  * }
  * ```
  *
- * For a classes that are represented as a single primitive value, [PrimitiveDescriptor] builder function can be used instead.
+ * For classes representing a single primitive value, [PrimitiveDescriptor] builder function can be used instead.
```
sandwwraith, line 120: I'd rather stick with a previous wording --- a value represents a class, not vice versa. Class can be serialized to many different forms of values with different serializers

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -26,6 +26,40 @@ import kotlinx.serialization.modules.*
...
+
+    /**
+     * Structure kind for singleton objects defined with `object` keyword.
+     * By default, objects are serialized as empty structures without any state and their identity is preserved
+     * across serialization within the same process, so you always have the same instance of the object.
+     *
+     * Empty structure is represented as a call to [Encoder.beginStructure] with the following [CompositeEncoder.endStructure]
+     * without any intermediate encodings.
+     */
+    public object OBJECT : SerialKind()
```
sandwwraith, line 38: AFAIR, we shouldn't have moved `OBJECT` out of structure kinds. The '*MOVE* objects to SerialKind' line in quip was referring to ENUM and CONTEXTUAL `object`s.

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -259,3 +266,20 @@ public sealed class PolymorphicKind : SerialKind() {
     public object OPEN : PolymorphicKind()
 }
 
+@Deprecated(
+    level = DeprecationLevel.ERROR,
+    message = "This object is added only as temporary migration mechanism"
```
sandwwraith, line 271: `UnionKind` is not a temporary migration mechanism, it's literally a deprecated object that existed

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -28,19 +28,20 @@ public class EnumDescriptor(
...
 
     override fun toString(): String {
-        return elementNames().joinToString(", ", "$serialName(", ")")
+        return elementNames.joinToString(", ", "$serialName(", ")")
     }
 
     override fun hashCode(): Int {
         var result = serialName.hashCode()
-        result = 31 * result + elementNames().hashCode()
+        val elementsHashCode = elementNames.fold(1) { hash, element -> 31 * hash + element.hashCode() }
```
sandwwraith, line 43: Why different hashing algorithm?

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -120,8 +120,10 @@ internal inline fun <reified SD : SerialDescriptor> SD.equalsImpl(
 internal inline fun SerialDescriptor.hashCodeImpl(typeParams: Array<SerialDescriptor>): Int {
     var result = serialName.hashCode()
     result = 31 * result + typeParams.contentHashCode()
-    val elementDescriptors = elementDescriptors()
-    result = 31 * result + elementDescriptors.map { it.serialName }.hashCode()
-    result = 31 * result + elementDescriptors.map { it.kind }.hashCode()
+    val elementDescriptors = elementDescriptors
+    val namesHash = elementDescriptors.fold(1) { hash, element -> 31 * hash + element.serialName.hashCode() }
```
sandwwraith, line 124: if you're so eager to remove allocation in `map` (which looks like a micro-optimization to me, since the result is cached anyway), you can use `.asSequence`. Or write `.hashByKey(key: (T) -> K): Int` function to hide `fold` as an implementation detail and reduce amount of copy-pasted code

integration-test/src/commonTest/kotlin/sample/BasicTypesSerializationTest.kt
qwwdfsad: Because that's a test and it will fail anyway, was following the path of the least resistance

runtime/commonMain/src/kotlinx/serialization/SerialDescriptor.kt
qwwdfsad: Sure, don't mind

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
qwwdfsad: It is the same hashing algorithm, it's just a handrolled version of `ArrayList.hashCode`.

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
qwwdfsad: `asSequence().hashCode()` invokes identity hashcode.

>Or write .hashByKey(key: (T) -> K): Int

I prefer not to introduce unnecessary abstractions unless there is real need

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
qwwdfsad: Goot catch, thanks

runtime/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
sandwwraith: It really improves readability imo

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -26,10 +26,37 @@ import kotlinx.serialization.modules.*
...
+     */
+    public object CONTEXTUAL : SerialKind()
+
     override fun toString(): String {
         // KNPE should never happen, because SerialKind is sealed and all inheritors are non-anonymous
         return this::class.simpleName!!
     }
+
+    // Provide a stable hashcode for objects
+    override fun hashCode(): Int = toString().hashCode()
```
sandwwraith, line 59: Why it's not stable given that all objects have identity?

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
```diff
@@ -259,3 +255,20 @@ public sealed class PolymorphicKind : SerialKind() {
...
+        replaceWith = ReplaceWith("SerialKind.ENUM")
+    )
+    public val ENUM_KIND: SerialKind get() = error("Should not be called")
+
+    @Deprecated(
+        "Was moved to the top-level serial kind during 1.0 API stabilization", level = DeprecationLevel.ERROR,
+        replaceWith = ReplaceWith("SerialKind.CONTEXTUAL")
+    )
+    public val CONTEXTUAL: SerialKind get() = error("Should not be called")
+}
```
sandwwraith, line 274: new line

runtime/commonMain/src/kotlinx/serialization/SerialKinds.kt
qwwdfsad: Because it uses `identityHashcode` that differs from run to run

PR #911, opened 2020-07-15
Descriptors

qwwdfsad:


runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -271,6 +272,34 @@ public fun <T : Any?> Encoder.encode(strategy: SerializationStrategy<T>, value:
 ) // TODO make internal when migrations are removed
 public fun <T : Any> Encoder.encode(obj: T): Unit = noImpl()
 
+@Deprecated(
+    "This method was renamed to buildClassSerialDescriptor during serialization 1.0 API stabilization",
+    ReplaceWith("buildClassSerialDescriptor(serialName, *typeParameters, builderAction)"), DeprecationLevel.ERROR
```
sandwwraith, line 277: shouldn't we specify `imports` here?

runtime/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -8,6 +8,9 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.builtins.*
+import kotlinx.serialization.builtins.MapEntrySerializer
```
sandwwraith, line 11: Redundant imports?

qwwdfsad: No, to avoid migration clashing

PR #912, opened 2020-07-15
[API stabilization] ProtoBuf rework

qwwdfsad:
* Transition the the sealed class + companion + builder scheme
    * ProtoId renamed to ProtoNumber and ProtoNumberType to ProtoIntegerType to be better aligned with ProtoBuf specification
    * All internal implementation details are hidden from the public package
    * Hide ProtobufDecodingException from public API


qwwdfsad:
It seems that a lot of files were added and removed, but in fact, they were moved to `internal` with the following changes:

* `inner` modifier replaced with `proto: ProtoBuf` parameter in each constructor
* some `@JvmField` properties were marked with `protected` as an additional safety net
* constants are moved to the top-level
* internal methods are moved to `Helpers.kt`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -115,480 +113,63 @@ import kotlin.jvm.*
...
+ * and adjusted with [builderAction].
+ */
+public fun ProtoBuf(from: ProtoBuf = ProtoBuf, builderAction: ProtoBufBuilder.() -> Unit): ProtoBuf {
+    val b = ProtoBufBuilder(from)
+    b.builderAction()
+    return ProtoBufImpl(b.encodeDefaults, b.serializersModule)
+}
+
+public class ProtoBufBuilder internal constructor(proto: ProtoBuf) {
+    public var encodeDefaults: Boolean = proto.encodeDefaults
```
qwwdfsad, line 152: Will restructure the whole documentation in the follow-up

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoBuf.kt
```diff
@@ -39,10 +35,10 @@ import kotlin.jvm.*
  * By default, protobuf fields ids are being assigned to Kotlin properties in incremental order, i.e.
  * the first property in the class has id 1, the second has id 2, and so forth.
  * If you need a more stable order (e.g. to avoid breaking changes when reordering properties),
- * provide custom ids using [ProtoId] annotation.
+ * provide custom ids using [ProtoNumber] annotation.
  *
  * By default, all numbers are encoded using [varint][https://developers.google.com/protocol-buffers/docs/encoding#varints]
```
sandwwraith, line 40:
```suggestion
 * By default, all integer numbers are encoded using [varint][https://developers.google.com/protocol-buffers/docs/encoding#varints]
```

PR #914, opened 2020-07-16
Hocon unification

qwwdfsad:


formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/PackageMigrations.kt
```diff
@@ -4,12 +4,12 @@
...
 
-
 private const val pkg = "kotlinx.serialization.config"
 private const val message = "Moved to $pkg package"
 
+@Suppress("DEPRECATION_ERROR")
 @Deprecated(message, ReplaceWith("ConfigParserConfiguration", pkg), level = DeprecationLevel.ERROR)
 typealias ConfigParserConfiguration = kotlinx.serialization.config.ConfigParserConfiguration
 
 @Deprecated(message, ReplaceWith("ConfigParser", pkg), level = DeprecationLevel.ERROR)
```
EdwarDDay, line 14: I think the `ReplaceWith` expression needs to be updated as well.
```suggestion
@Deprecated(message, ReplaceWith("Hocon", pkg), level = DeprecationLevel.ERROR)
```

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/ConfigMigrations.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@Deprecated(
+    "ConfigParser was renamed to Hocon during serialization 1.0 API stabilization",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("Hocon")
+)
+typealias ConfigParser = Hocon
+
+@Deprecated(
+    "ConfigParserConfiguration eas deprecated during serialization 1.0 API stabilization, please use direct constructor parameters instead",
```
EdwarDDay, line 18: Small typo
```suggestion
    "ConfigParserConfiguration was deprecated during serialization 1.0 API stabilization, please use direct constructor parameters instead",
```

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/PackageMigrations.kt
sandwwraith: Indeed, these migrations aid can be updated to their final variants

PR #919, opened 2020-07-20
Packages relocation

sandwwraith:
No migrations for now, because star importing both `kotlinx.serialization.*` and `kotlinx.serializartion.encoding.*` causes 'Unresolved reference' on conflict

Also added manifest information and bumped version

### DIFF
--- a/runtime/build.gradle
+++ b/runtime/build.gradle
@@ -21,6 +21,19 @@ kotlin {
     }
 }
 
+/*
+ These manifest values help kotlinx.serialization compiler plugin determine if it is compatible with a given runtime library.
+ Plugin reads them during compilation.
+
+ Implementation-Version is used to determine whether runtime library supports a given plugin feature (e.g. inline classes serialization
+ in Kotlin 1.x may require runtime library version 1.y to work).
+ Compiler plugin may enable or disable features by looking on Implementation-Version.
+
+ Require-Kotlin-Version is used to determine whether runtime library with new features can work with old compilers.
+ In ideal case, its value should always be 1.4, but some refactorings (e.g. adding a method to the Encoder interface)
+ may unexpectedly break old compilers, so it is left out as a safety net. Compiler plugins, starting from 1.4 are instructed
+ to reject runtime if runtime's Require-Kotlin-Version is greater then the current compiler.
+ */
 tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
     manifest {
         attributes(
--- a/runtime/build.gradle
+++ b/runtime/build.gradle
@@ -20,3 +20,12 @@ kotlin {
         }
     }
 }
+
+tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
+    manifest {
+        attributes(
+                "Implementation-Version": version,
+                "Require-Kotlin-Version": "1.4-RC",
+        )
+    }
+}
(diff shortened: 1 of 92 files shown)


runtime/build.gradle
```diff
@@ -20,3 +20,12 @@ kotlin {
         }
     }
 }
+
+tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
```
qwwdfsad, line 24: Please drop a **detailed** comment on what is this, how it is used and why it's important.
Otherwise, its purpose can be lost in a year or so

runtime/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
```diff
@@ -21,7 +23,7 @@ import kotlin.reflect.*
  * To pass it to encoder and decoder, refer to particular [SerialFormat]'s documentation.
  */
 @OptIn(UnsafeSerializationApi::class)
-public class ContextSerializer<T : Any>(
+public class ContextualSerializer<T : Any>(
```
qwwdfsad, line 26: Please fix documentation (search by `ContextualSerializer`) as well

runtime/commonTest/src/kotlinx/serialization/TuplesTest.kt
```diff
@@ -43,7 +44,7 @@ class TuplesTest : JsonTestBase() {
         val desc = PairWrapper.serializer().descriptor.getElementDescriptor(0)
         assertEquals(desc.serialName, "kotlin.Pair")
         assertEquals(
-            desc.elementDescriptors.map(SerialDescriptor::kind),
+            desc.elementDescriptors.map(kotlinx.serialization.descriptors.SerialDescriptor::kind),
```
qwwdfsad, line 47: FQNS are not necessary

runtime/commonMain/src/kotlinx/serialization/descriptors/SerialKinds.kt
```diff
@@ -2,10 +2,10 @@
  * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package kotlinx.serialization
+package kotlinx.serialization.descriptors
```
qwwdfsad, line 5: Shouldn't we rename this file to `SerialKind.kt`?

runtime/commonTest/src/kotlinx/serialization/TuplesTest.kt
sandwwraith: I think this is inserted by the 'move' refactoring. Looks like an issue in it with function references

runtime/build.gradle
qwwdfsad: Also, it would be nice to have an integration test that verifies the version is present in the manifest

runtime/commonMain/src/kotlinx/serialization/descriptors/SerialKinds.kt
sandwwraith: Not so sure about that. The root level class is SerialKind, but we have many kinds in it. Also, it does not have any extensions, so `SerialKindsKt` is not present in the ABI.

sandwwraith:
Merged manually to `dev`

PR #920, opened 2020-07-20
Retreive serializer for non-serializable enum on JVM

### LINKED ISSUE #913: Enum serializer is not found by its class
**Describe the bug**

Enum serializer is not found by its class

**To Reproduce**

```kotlin
enum class Status { OPEN, CLOSED }

fun main() {
    println(Json.encodeToString(Status.OPEN))
}
```

**Expected behavior**

Prints `"OPEN"`

**Actual**

```
Exception in thread "main" kotlinx.serialization.SerializationException: Serializer for class 'Status' is not found. Mark the class as @Serializable or provide the serializer explicitly.
```

qwwdfsad:
On the rest of the platforms this feature is blocked by KT-14743 (or plugin intrinsics). Anyway, providing the best possible user-experience only on the biggest platform is better than provide nothing, but consistently

Fixes #913

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
@@ -47,10 +47,9 @@ internal class EnumDescriptor(
     }
 }
 
-// Used for enums that are not explicitly serializable
-@Deprecated(level = DeprecationLevel.ERROR, message = "For plugin-generated code, " +
-        "should not be used directly. For the custom serializers please report your use-case to project issues, so proper public API could be introduced instead")
-public class EnumSerializer<T : Enum<T>>(
+// Used for enums that are not explicitly serializable by the plugin
+@PublishedApi
+internal class EnumSerializer<T : Enum<T>>(
     serialName: String,
     private val values: Array<T>
 ) : KSerializer<T> {
@@ -78,4 +77,6 @@ public class EnumSerializer<T : Enum<T>>(
         }
         return values[index]
     }
+
+    override fun toString(): String = "kotlinx.serialization.internal.EnumSerializer<${descriptor.serialName}>"
 }
--- a/runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
@@ -9,7 +9,7 @@ import kotlinx.serialization.internal.EnumSerializer
 import kotlin.test.*
 
 @Suppress("TestFunctionName")
-inline fun <reified E : Enum<E>> EnumSerializer(serialName: String): EnumSerializer<E> =
+internal inline fun <reified E : Enum<E>> EnumSerializer(serialName: String): EnumSerializer<E> =
     EnumSerializer(serialName, enumValues())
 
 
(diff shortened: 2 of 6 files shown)


runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -77,4 +76,6 @@ public class EnumSerializer<T : Enum<T>>(
         }
         return values[index]
     }
+
+    override fun toString(): String = "kotlinx.serialization.internal.EnumSerializer<${descriptor.serialName}>"
```
sandwwraith, line 80: We usually do not have `toString()` on serializers, why is it needed?

runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
```diff
@@ -9,7 +9,7 @@ import kotlinx.serialization.internal.EnumSerializer
 import kotlin.test.*
 
 @Suppress("TestFunctionName")
-inline fun <reified E : Enum<E>> EnumSerializer(serialName: String): EnumSerializer<E> =
+internal inline fun <reified E : Enum<E>> EnumSerializer(serialName: String): EnumSerializer<E> =
```
sandwwraith, line 12: it is function in test sources anyway

runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -43,18 +43,31 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
...
     }
+
+    if (default != null) return default
+    // Fallback: enum without @Serializable annotation as last resort
+    return createEnumSerializer()
+}
+
+@Suppress("UNCHECKED_CAST")
+private fun <T : Any> KClass<T>.createEnumSerializer(): KSerializer<T>? {
+    if (!Enum::class.java.isAssignableFrom(java)) return null
```
sandwwraith, line 60:
```suggestion
    if (!Enum::class.java.isAssignableFrom(this.java)) return null
```

slightly more readable

runtime/jvmTest/src/kotlinx/serialization/features/SerializerByTypeTest.kt
```diff
@@ -183,4 +183,12 @@ class SerializerByTypeTest {
         val superType = base::class.java.genericSuperclass!!
         return (superType as ParameterizedType).actualTypeArguments.first()!!
     }
+
+    enum class Foo
+
+    @Test
+    fun testNonSerializableEnum() {
+        val serializer = serializer<Foo>()
+        assertTrue { serializer.descriptor.kind is SerialKind.ENUM }
```
sandwwraith, line 192:
```suggestion
        assertTrue(serializer.descriptor.kind is SerialKind.ENUM)
```

no need for lambda, actually

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
qwwdfsad: For better debugger experience when debugging formats/serializers

runtime/commonTest/src/kotlinx/serialization/test/TestHelpers.kt
qwwdfsad: it exposes internal `EnumSerializer` and should be internal

runtime/jvmTest/src/kotlinx/serialization/features/SerializerByTypeTest.kt
qwwdfsad: Not sure what's the difference here, but okay

runtime/commonMain/src/kotlinx/serialization/internal/Enums.kt
sandwwraith: AFAIR, we once agreed that we should delegate `KSerializer.toString()` to `descriptor.toString()`, and this function doesn't follow this decision. It also doesn't help much until all serializers will have .toString(), so I suggest removing it for now

PR #924, opened 2020-07-22
Make Properties consistent with Json format and serialize enums as st…

### LINKED ISSUE #818: Inconsistent serialisation of enums between Json and Properties parsers
**Describe the bug**

I'm trying to deserialize a `Map` to an object (in JVM). The `Map` comes from a Firestore DB which I populated from Kotlin/JS using `kotlinx.serialization.json.Json`.

The data inside the DB contains enums, which by default are serialised as `String`s in `Json`, but the `kotlinx.serialization.Properties` deserialiser seems to expect them as `Integer`s because it throws the error:

```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
```

I'm not using any custom `KSerializer`s for enums, just the defaults in each case.

**To Reproduce**

```kotlin
fun main(args: Array<String>) {
    val values = mapOf("id" to "asfqweertsdsdg", "enumValue" to "VAL_1")

    val result = Properties.loadNullable(TestClass.serializer(), values)
}

@Serializable
data class TestClass(
    val id: String,
    val enumValue: TestEnum
)

enum class TestEnum { VAL_1, VAL_2 }
```

I also tried making `TestEnum` explicitly serializable:

```kotlin
@Serializable
enum class TestEnum {
    @SerialName("VAL_1")
    VAL_1,
   @SerialName("VAL_2")
    VAL_2
}
```

but that also failed.

If I change the map to:

```kotlin
mapOf("id" to "asfqweertsdsdg", "enumValue" to 0)
```

then the example above works fine.

**Expected behavior**

`Json` and `Properties` should use the same default way to serialise and deserialise enums.

**Environment**

 - Kotlin version: 1.3.72
 - Library version: 0.20.0
 - Kotlin platforms: JVM and JS
 - Gradle version: 5.6.4

qwwdfsad:
…ring literals instead of integer constants.

It seems to be legacy from initial implementation and cannot be changed later after 1.0 in a safe manner without additional flags, but for (java.lang.)Properties serializing enums as strings is the most desirable behaviour

Fixes #818

### DIFF
--- a/formats/properties/commonMain/src/kotlinx/serialization/Properties.kt
+++ b/formats/properties/commonMain/src/kotlinx/serialization/Properties.kt
@@ -63,8 +63,8 @@ public class Properties(override val serializersModule: SerializersModule = Empt
             // ignore nulls in output
         }
 
-        override fun encodeTaggedEnum(tag: String, enumDescription: SerialDescriptor, ordinal: Int) {
-            map[tag] = enumDescription.getElementName(ordinal)
+        override fun encodeTaggedEnum(tag: String, enumDescriptor: SerialDescriptor, ordinal: Int) {
+            map[tag] = enumDescriptor.getElementName(ordinal)
         }
     }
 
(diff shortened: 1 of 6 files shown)


formats/properties/commonMain/src/kotlinx/serialization/Properties.kt
```diff
@@ -62,6 +62,10 @@ public class Properties(override val serializersModule: SerializersModule = Empt
         override fun encodeTaggedNull(tag: String) {
             // ignore nulls in output
         }
+
+        override fun encodeTaggedEnum(tag: String, enumDescription: SerialDescriptor, ordinal: Int) {
```
sandwwraith, line 66: It looks like parameter name should be `enumDescriptor` instead of `enumDescription`. Seems like an oversight in `TaggedEncoder`, pls rename it, if you don't mind

PR #930, opened 2020-07-23
Json polymorphism

qwwdfsad:


runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -112,11 +112,11 @@ internal class SealedClassSerializer<T : Any>(
             }.mapValues { it.value.value }
     }
 
-    override fun findPolymorphicSerializer(decoder: CompositeDecoder, klassName: String): DeserializationStrategy<out T> {
-        return serialName2Serializer[klassName] ?: super.findPolymorphicSerializer(decoder, klassName)
+    override fun findPolymorphicSerializerOrNull(decoder: CompositeDecoder, klassName: String?): DeserializationStrategy<out T>? {
```
sandwwraith, line 115: It's a bit strange that `OrNull` method is a member, and regular method is an extension. WDYT renaming them into `findXXX` (member) and `findXXXOrThrow` (ext) ?

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -112,11 +112,11 @@ internal class SealedClassSerializer<T : Any>(
...
+        return serialName2Serializer[klassName] ?: super.findPolymorphicSerializerOrNull(decoder, klassName)
     }
 
-    override fun findPolymorphicSerializer(encoder: Encoder, value: T): SerializationStrategy<T> {
-        return (class2Serializer[value::class] ?: super.findPolymorphicSerializer(encoder, value)).cast()
+    override fun findPolymorphicSerializerOrNull(encoder: Encoder, value: T): SerializationStrategy<T>? {
```
sandwwraith, line 119: Also, maybe I oversaw something, but why do we have to make this method non-throwing at all? Only to support alternative exception text in JSON? Looks like an overkill.

However, the nullable result is more future-proof, e.g. for formats/serializers that imply some default behavior regardless of the module. Not sure we have to encourage such design.

runtime/commonMain/src/kotlinx/serialization/json/internal/PolymorphismValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+        if (!useArrayPolymorphism) {
+            // Collisions with "type" can happen only for JSON polymorphism
+            checkDiscriminatorCollisions(descriptor, actualClass)
+        }
+    }
+
+    private fun checkKind(descriptor: SerialDescriptor, actualClass: KClass<*>) {
+        val kind = descriptor.kind
+        if (kind is PolymorphicKind || kind == SerialKind.CONTEXTUAL) {
+            throw IllegalArgumentException("$kind kind in serializer for $actualClass is not supported as subclass for polymorphic serialization")
```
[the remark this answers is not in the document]
qwwdfsad: Great idea

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
qwwdfsad: To allow other formats (e.g. it's easy to see XML wants to use it) leveraging it.
We already rely on it, it's still internal, so doesn't seem to be a problem anyway

runtime/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -56,7 +64,14 @@ internal fun <T> JsonDecoder.decodeSerializableValuePolymorphic(deserializer: De
...
     val discriminator = json.configuration.classDiscriminator
     val type = jsonTree[discriminator]?.jsonPrimitive?.content
-        ?: throw JsonDecodingException(-1, "Missing polymorphic discriminator $discriminator", jsonTree.toString())
-    val actualSerializer = deserializer.findPolymorphicSerializer(this, type).cast<T>()
+    val actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)?.cast<T>()
+        ?: throwSerializerNotFound(type, jsonTree)
     return json.readPolymorphicJson(discriminator, jsonTree, actualSerializer)
 }
+
+private fun throwSerializerNotFound(type: String?, jsonTree: JsonObject): Nothing {
```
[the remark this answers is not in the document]
qwwdfsad: Just to provide more context for easier debugging. This one prints surrounding JSON as well

runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
qwwdfsad: It will become inconsistent with Kotlin code style and stdlib in particular. We have `x: T` that throws and `xOrNull: T?` for nullable version

runtime/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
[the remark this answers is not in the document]
antonianemi: @qwwdfsad i am currently using this api and i got this exception, can you help me out figuring out what do i need, i am not sure what is the role of the "type" which in my case is coming null after a post. thank you

PR #931, opened 2020-07-23
More approachable exception messages during reflective lookup and sea…

### LINKED ISSUE #925: Very obscure error when not all subclasses of a sealed class are serializable
Run this code:

```kotlin

@Serializable
sealed class Repository {
    abstract val name: String
}
            
// THIS IS NOT @Serializable
class SimpleRepository(override val name: String) : Repository()

@Serializable
class OwnedRepository(override val name: String, val owner: String) : Repository()

fun main() {
    val data: Repository = OwnedRepository("kotlinx.coroutines", "kotlin")
    println(Json.encodeToString(data))
}  
```

It results with a very obscure error:

```
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at kotlinx.serialization.internal.PlatformKt.constructSerializerForGivenTypeArgs(Platform.kt:40)
	at kotlinx.serialization.internal.PlatformKt.compiledSerializerImpl(Platform.kt:23)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializerOrNull(Serializers.kt:109)
	at example.examplePoly04.Example_poly_04Kt.main(example-poly-04.kt:23)
	at example.examplePoly04.Example_poly_04Kt.main(example-poly-04.kt)
Caused by: java.lang.IllegalArgumentException: Arrays of classes and serializers must have the same length, got arrays: [class example.examplePoly04.SimpleRepository, class example.examplePoly04.OwnedRepository], [example.examplePoly04.OwnedRepository$$serializer@281e3708]
Please ensure that @Serializable annotation is present on each sealed subclass
	at kotlinx.serialization.SealedClassSerializer.<init>(SealedSerializer.kt:96)
	at example.examplePoly04.Repository$Companion.serializer(example-poly-04.kt)
	... 10 more
```

Expected: It should either work or, at least, produce a human-friendly error message.

qwwdfsad:
…led serialer specifically

Fixes #925

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
@@ -93,11 +93,10 @@ internal class SealedClassSerializer<T : Any>(
     private val serialName2Serializer: Map<String, KSerializer<out T>>
 
     init {
-        require(subclasses.size == subclassSerializers.size) {
-            "Arrays of classes and serializers must have the same length," +
-                    " got arrays: ${subclasses.contentToString()}, ${subclassSerializers.contentToString()}\n" +
-                    "Please ensure that @Serializable annotation is present on each sealed subclass"
+        if (subclasses.size != subclassSerializers.size) {
+            throw IllegalArgumentException("All subclasses of sealed class ${baseClass.simpleName} should be marked @Serializable")
         }
+
         class2Serializer = subclasses.zip(subclassSerializers).toMap()
         serialName2Serializer = class2Serializer.entries.groupingBy { it.value.descriptor.serialName }
             .aggregate<Map.Entry<KClass<out T>, KSerializer<out T>>, String, Map.Entry<KClass<*>, KSerializer<out T>>>
--- a/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -30,16 +30,8 @@ internal actual fun <T : Any, E : T?> ArrayList<E>.toNativeArrayImpl(eClass: KCl
 internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vararg args: KSerializer<Any?>): KSerializer<T>? {
     val jClass = this.java
     // Search for serializer defined on companion object.
-    val companion =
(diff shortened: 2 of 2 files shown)


runtime/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -93,11 +93,25 @@ internal class SealedClassSerializer<T : Any>(
...
 
     init {
-        require(subclasses.size == subclassSerializers.size) {
-            "Arrays of classes and serializers must have the same length," +
-                    " got arrays: ${subclasses.contentToString()}, ${subclassSerializers.contentToString()}\n" +
-                    "Please ensure that @Serializable annotation is present on each sealed subclass"
+
+        if (subclasses.size != subclassSerializers.size) {
+            // Remove prefix is required for JVM to trim "class " from KClass.toString
+            val allClasses = subclasses.map { it.toString().substringAfter("class ") }
```
sandwwraith, line 99: Did you know that when kotlin-reflect is not in a classpath, `KClass.toString()` looks like this: 
```
class java.lang.String (Kotlin reflection is not available)
```
So you may want to cut out 'reflection is not available' too.
But I think it's better to use `simpleName()`. 

Anyway, what's wrong with original message? I think most of the users can see what classes are missing `@Serializable` annotation because they're all in the same file, so  `Please ensure that @Serializable annotation is present on each sealed subclass / All subclasses of sealed class $bc should be marked @Serializable` is more than enough. Such guessing logic is vague and may give strange results in many cases (e.g. custom serializer that does not contain $$ in classname).

Also, please use `bc.simpleName()` (due to mentioned kotlin-reflect note)

qwwdfsad: Agree, it should be enough. 
Was thinking about providing best-effort guess, but it really seems to be too much

PR #937, opened 2020-07-27
[API stabilization] Properly handle nullable type parameters in KSeri…

### LINKED ISSUE #824: KSerializer.serialize() of nullable type not invoked for null values
In the project's code there is a nullable value. 
Magic not null values need to be deserialized to null.
And nulls need to be serialized to non-null magic values.

While I'm able to deserialize to nulls the serialization does not work. `serialize()` method is not fired and null is passed to serialized form.

**To Reproduce**
Serializer code:
```kotlin
internal object LocalTimeSerializer : KSerializer<LocalTime?> {

    override val descriptor = PrimitiveDescriptor("LocalTime?", PrimitiveKind.INT)

    override fun deserialize(decoder: Decoder): LocalTime? = when (val seconds = decoder.decodeInt()) {
       -1 -> null
        else -> LocalTime.ofSecondOfDay(seconds.toLong())
    }

    override fun serialize(encoder: Encoder, value: LocalTime?) {
        when (value) {
            null -> encoder.encodeInt(-1) //this line is never reached despite that nulls exist in serialized lists
            else -> encoder.encodeInt(value.toSecondOfDay())
        }
    }
}
```

Usage inside serializable class:
```kotlin
val times: List<@Serializable(with = LocalTimeSerializer::class) LocalTime?>,
```

**Expected behavior**
`serialize()` method is invoked for nulls.

**Environment**
 - Kotlin version: 1.3.72
 - Library version: 0.20.0
 - Kotlin platforms: JVM
 - Gradle version: 6.3+

qwwdfsad:
…alizer

Effectively enabling custom serialization and deserialization of nulls and non-nulls by properly checking whether descriptor is nullable

Fixes #824

### DIFF
--- a/runtime/commonTest/src/kotlinx/serialization/SerializerForNullableTypeTest.kt
+++ b/runtime/commonTest/src/kotlinx/serialization/SerializerForNullableTypeTest.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.json.*
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+public class SerializerForNullableTypeTest : JsonTestBase() {
+
+    // Nullable boxes
+    @Serializable(with = StringHolderSerializer::class)
+    data class StringHolder(val s: String)
+
+    @Serializer(forClass = StringHolder::class)
+    object StringHolderSerializer : KSerializer<StringHolder?> {
+        override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("SHS", PrimitiveKind.STRING).nullable
+
+        override fun serialize(encoder: Encoder, value: StringHolder?) {
+            if (value == null) encoder.encodeString("nullable")
+            else encoder.encodeString("non-nullable")
+        }
+
+        override fun deserialize(decoder: Decoder): StringHolder? {
+            if (decoder.decodeNotNullMark()) {
+                return StringHolder("non-null: " + decoder.decodeString())
+            }
+            decoder.decodeNull()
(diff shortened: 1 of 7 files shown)


qwwdfsad:
PR fails on Native but it seems to be a plugin bug. Please take a look at it as well

sandwwraith:
I dag a little into the problem and found out that the problem is not in the plugin, but in this optimization: https://github.com/Kotlin/kotlinx.serialization/blob/c228861ffce393ba3942a2df067721f8f05af03d/runtime/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt#L62 that calls `serializerOrNull` before `serializer(KType)`. Given that this API is not very stable, it may, actually, crash instead of returning null -- that what happens in Native. However, a simple addition of `try/catch` is not enough -- JS IR problems are much bigger. In runtime, it fails not in `serializerOrNull`, but later, given the fact that `arrayOf()[i]` not fails on JS immediately, but silently returns `undefined` which is stored as a serializer for `T` inside `GenericBox.serializer(vararg KSerializer)`. We should either remove this optimization or make it JVM-specific. Another possible route is to rewrite `compiledSerializerImpl` so it does not use `constructSerializerForGivenTypeArgs()` on JS.

runtime/commonTest/src/kotlinx/serialization/SerializerForNullableTypeTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        if (isJsLegacy()) return
+        val data = GenericBox<StringHolder?>(null)
+        val json = Json.encodeToString(data)
+        assertEquals("""{"value":"nullable"}""", Json.encodeToString(data))
+        assertEquals(GenericBox(StringHolder("non-null: nullable")), Json.decodeFromString(json))
+    }
+
+    @Test
+    fun testGenericNullableBoxFromNull() {
+        if (isJsLegacy()) return // BE bug
```
sandwwraith, line 125:
```suggestion
        if (isJsLegacy()) return
```

PR #938, opened 2020-07-27
Remaster properties

sandwwraith:
Refactor Properties:

- move to correct package
- renamed load/store
- remove *Nullable overloads

Do not write/read the size of a collection into Properties' map.

Note that this introduces the same problem as with protobuf: we can't tell whether a map was empty or property with map type was missing, so maps should have `=emptyMap()` default value.


formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * A [Properties] instance that can be used as default and does not have any [SerializersModule] installed.
+     */
+    public companion object Default : Properties(EmptySerializersModule, null)
+}
+
+private class PropertiesImpl(serializersModule: SerializersModule) : Properties(serializersModule, null)
+
+/**
+ * Create an instance of [Properties] with a given [module].
```
qwwdfsad, line 144: Create**s**

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -0,0 +1,181 @@
...
+ */
+public fun Properties(module: SerializersModule): Properties = PropertiesImpl(module)
+
+/**
+ * A reified version of [encodeToMap].
```
qwwdfsad, line 149: Let's refrain from "reified version of X" in the documentation and make it more self-contained. 
E.g. like this: https://github.com/Kotlin/kotlinx.serialization/blob/dev/runtime/commonMain/src/kotlinx/serialization/SerialFormat.kt#L65

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -0,0 +1,181 @@
...
+                    // if map does not contain key we look for, then indices in collection have ended
+                    break
+                }
+            }
+            return CompositeDecoder.DECODE_DONE
+        }
+    }
+
+    /**
+     * Stores properties from given [value] to a map and returns this map.
```
qwwdfsad, line 104: Wording should be changed from store/load to encode/decode (here and below)

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * @param serializersModule A [SerializersModule] which should contain registered serializers
+ * for [Contextual] and [Polymorphic] serialization, if you have any.
+ */
+public sealed class Properties(
+    override val serializersModule: SerializersModule,
+    ctorMarker: Nothing?
+) : SerialFormat {
+
+    private inner class OutMapper : NamedValueEncoder() {
```
qwwdfsad, line 47: [nit] I am a fan of separating implementation to a different file, so users will see **only** public API, but I do not insist

sandwwraith: Given that these implementations are 30-40 lines long, I do not think this is really a problem to leave them in the same file

PR #947, opened 2020-07-29
Add CBOR ignoreUnknownKeys option

### LINKED ISSUE #935: Provide ignoreUnknownKeys for CBOR decoding

If a remote endpoint is updated to add field(s) then client will fail.

This becomes particularly cumbersome with Android where users may continue to run an older version of an app for some time. Ideally, the remote endpoint that the Android app communicates with could add new fields to a CBOR payload and updated apps could read the new field (to support new functionality) while older versions of the app just silently ignore the new fields.

For example, if we had the following definition in a client:

```kotlin
@Serializable
data class Example(
    val a: String
)
```

If remote endpoint was updated to add the field `b`:

```diff
  BF           # map(*)
     61        # text(1)
        61     # "a"
     63        # text(3)
        313233 # "123"
+    61        # text(1)
+       62     # "b"
+    63        # text(3)
+       393837 # "987"
     FF        # primitive(*)
```

Then clients will fail with:

```java
kotlinx.serialization.SerializationException: Example does not contain element with name 'b'
```

JSON deserialization provides the `ignoreUnknownKeys` option which allows the encoded JSON to contain fields that are absent in the Kotlin class definition.

Ideally, CBOR decoding could also be configured with `ignoreUnknownKeys` (or similar) option.

twyatt:
Add support for ignoring elements in encoded CBOR that are not present in the Kotlin class representation.

Closes #935

## How it works

When an unknown element is encountered, `decoder.skipElement()` is invoked until a known element is encountered.

Skipping of an element begins at the first byte of the element's value.

Bytes are processed to determine element type (and corresponding length), to ultimately
determine how many bytes can be skipped. The general process is demonstrated by the following pseudo-code:

_`lengthStack` represents stack of "number of elements" at each depth._

```
fun skipElement() {
  do {
    when (element) {
      0xFF -> lengthStack.removeLast(); prune() // removes last `indefinite` in stack
      type is indefinite -> lengthStack.add(indefinite)
      type is collection -> lengthStack.add(collection.length) // e.g. array, list, map
      type is primitive -> skip(primitive.length); prune()
    }
  } while (lengthStack.isNotEmpty())
}

fun prune() {
  for (length in lengthStack[last]..lengthStack[0]) { // iterate over `lengthStack` values in reverse order
    if (length is indefinite) return // stop "pruning"
    length--
    if (length == 0) pop() // remove last `length` from stack
  }
}
```

## Examples

The following examples show the hex notation of CBOR encoded payloads with the cooresponding `lengthStack` _after_ processing the bytes in that row.

### Array of objects

The encoded CBOR `82A16161636B656BA16161636B656B` would undergo the following `skipElement` (decoding) process:

| 1    | 2    | 3           | CBOR          | `lengthStack` | Action(s) Taken |
|------|------|-------------|---------------|---------------|-------------|
| `82` |      |             | array(2)      | `[2]`         | pushed `2` (child count of `82`: `A1`, `A1`) on stack |
|      | `A1` |             | map(1)        | `[2, 2]`      | pushed `2` (child count of `A1`: "a", "kek") on stack |
|      |      | `61 61`     | text(1) "a"   | `[2, 1]`      | jumped over "a", decremented last stack item |
|      |      | `63 6B656B` | text(3) "kek" | `[1]`         | jumped over "kek", no more children of `A1`, decrement stack items |
|      | `A1` |             | map(1)        | `[1, 2]`      | pushed `2` (child count of `A1`: "a", "kek") on stack |
|      |      | `61 61`     | text(1) "a"   | `[1, 1]`      | jumped over "a", decremented last stack item |
|      |      | `63 6B656B` | text(3) "kek" | `[]`          | jumped over "kek", no more children of `A1`, decremented stack items |

### Array with nested indefinite length array

The encoded CBOR `829FA16161636B656BFF6162` would undergo the following `skipElement` (decoding) process:

| 1    | 2       | 3    | 4           | CBOR          | `lengthStack` | Action(s) Taken |
|------|---------|------|-------------|---------------|---------------|--------|
| `82` |         |      |             | array(2)      | `[2]`         | pushed `2` (child count of `82`: `array(*)`, "b") on stack |
|      | `9F`    |      |             | array(*)      | `[2, -1]`     | pushed _indefinite_ on stack |
|      |         | `A1` |             | map(1)        | `[2, -1, 2]`  | pushed `2` (child count of `A1`: "a", "kek") on stack |
|      |         |      | `61 61`     | text(1) "a"   | `[2, -1, 1]`  | jumped over "a", decremented last stack item |
|      |         |      | `63 6B656B` | text(3) "kek" | `[2, -1]`     | jumped over "kek", no more children of `A1`, decrement stack items (from end to start; stopped at `-1`) |
|      |         | `FF` |             | primitive(*)  | `[1]`         | encountered break, no more children of `9F`, decrement stack items |
|      | `61 62` |      |             | text(1) "b"   | `[]`          | jumped over "b", no more children of `82`, decremented stack items |


formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
```diff
@@ -104,4 +108,343 @@ class CborReaderTest {
             )
         }
     }
+
+    @Test
+    fun testIgnoreUnknownKeysFailsWhenKotlinClassIsMissingKeys() {
```
twyatt, line 113: This test attempts to decode same CBOR encoded data as `testDecodeComplicatedObject` but attempts to decode as `Simple` rather than `TypesUmbrella`.

Validates that although `ignoreUnknownKeys` permits properties to be absent in the Kotlin class definition, the inverse (of requiring that the CBOR encoded data have all the expected fields) is still enforced.

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
```diff
@@ -104,4 +108,338 @@ class CborReaderTest {
...
+         *    42                        # bytes(2)
+         *       CAFE                   # "\xCA\xFE"
+         *    61                        # text(1)
+         *       64                     # "d"
+         *    6B                        # text(11)
+         *       48656C6C6F20776F726C64 # "Hello world"
+         *    61                        # text(1)
+         *       65                     # "e"
+         *    02                        # unsigned(2)
+         */
```
twyatt, line 164: Are comments describing the CBOR encoded data helpful or just noise?

_A developer who later encounters this code could themselves paste the `withDecoder` `input` into [cbor.me](http://cbor.me/), so having it as a comment feels somewhat superfluous, although it can be rather convenient having it quickly available for reference. Downside being that it's a comment that needs to be maintained if any changes to `input` are made._

twyatt:
Pulled in latest `dev` (and resolved conflicts).
Added documentation.
Did not run/commit `./gradlew knit` (to reduce change noise in PR) but did run `./gradlew knit check` on my local machine and it compiled/passed.

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
```diff
@@ -104,4 +108,338 @@ class CborReaderTest {
             )
         }
     }
+
+    @Test
+    fun testIgnoreUnknownKeysFailsWhenKotlinClassIsMissingKeys() {
+        // with maps & lists of indefinite length
+        assertEquals(
+            expected = "Field 'a' is required, but it was missing",
```
twyatt, line 116: Asserting against the `Exception`'s message text which is fragile (if `SerializationException` message language changes it fails this unrelated test).

Is simply asserting against the exception type sufficient? In other words, should I remove the outer `assertEquals` and just keep the `assertFailsWith`?

qwwdfsad:
Small update: we are busy with releasing Kotlin 1.4.0 and release candidate of serialization right now.
I'll return to your PR right after the release (~middle of this week)

twyatt:
@qwwdfsad let me know if there is anything I can do to simplify this PR, or anything I can otherwise do to help with the review process?

Totally understand if the JetBrains team is super busy with other priorities. Absolutely appreciate the tools that the JetBrains team creates; thanks again!

sandwwraith:
@twyatt Sorry for not keeping in touch. I see you've rebased the branch, this should be enough of help. I'll try to return to you soon; I think this is an important feature for 1.0, so we'll try to put it there

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -428,6 +448,64 @@ internal class CborDecoder(private val input: ByteArrayInput) {
...
+                        1 -> lengthStack.removeAt(i)
+                        else -> {
+                            lengthStack[i] = lengthStack[i] - 1
+                            break
+                        }
+                    }
+                }
+            }
+
+            if (readByte() == -1 && lengthStack.isNotEmpty()) error("EOF with length stack $lengthStack")
```
sandwwraith, line 505: Same here about exceptions

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
sandwwraith: We have `assertFailsWithMessage` utility function in json tests — I think you can copy paste it to cbor test utils, too. We usually find it sufficient because exception texts rarely change

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
sandwwraith: I think they are more helpful than harmful. Again, test data changes relatively rarely.

sandwwraith (changes requested):
Thank you for the PR and explanations — they were really helpful. I'd left some comments, mainly about exceptions, but the rest of PR is fine.

sandwwraith:
The build check has failed — seems you have to run `./gradlew knit` to re-generate documentation tests

twyatt:
@sandwwraith thanks for the review. I'll run `knit` when I address the comments/feedback. Thanks again!

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
```diff
@@ -33,6 +33,10 @@ internal class ByteArrayInput(private var array: ByteArray) {
         return copied
     }
 
+    fun skip(length: Int) {
+        require(length >= 0) { "Cannot skip $length bytes" }
```
qwwdfsad, line 37: Either length cannot be negative here or `SerializationException` should be thrown as well

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -233,10 +233,29 @@ internal open class CborReader(private val cbor: Cbor, protected val decoder: Cb
...
 
     override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
-        if (!finiteMode && decoder.isEnd() || (finiteMode && readProperties >= size)) return CompositeDecoder.DECODE_DONE
-        val elemName = decoder.nextString()
-        readProperties++
-        val index = descriptor.getElementIndexOrThrow(elemName)
+        val index = if (cbor.ignoreUnknownKeys) {
+            val knownIndex: Int
+            while (true) {
+                if (isDone()) return CompositeDecoder.DECODE_DONE
```
qwwdfsad, line 239: This check could be lifted to the very first line of the method to simplify the control flow and make it slightly easier to read

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -428,6 +448,64 @@ internal class CborDecoder(private val input: ByteArrayInput) {
         return result
     }
 
+    fun skipElement() {
```
qwwdfsad, line 451: This method is quite huge and hard to read.
It would be nice if you could extract a few subroutines (e.g. one from `if (prune) { ... }` can be replaced with `if (prune) prune()`, `length` extraction and so on) to make it as straightforward as possible

twyatt: > This method is quite huge and hard to read.

Agreed.

> `if (prune) { ... }` can be replaced with `if (prune) prune()`

I was thinking that too, but I worried that a `prune()` function is very specific to element skipping and a `prune()` function in the `CborDecoder` class feels like the `CborDecoder` is taking on too many responsibilities? Would it warrant a separate `private` class that is responsible for all this element skipping logic (to help isolate the element skipping functions from other CBOR decoding functionality)? Or is that overkill?

qwwdfsad: IMO It's a bit too much, let's stick with single `CborDecoder ` for now.

docs/formats.md
```diff
@@ -167,6 +167,55 @@ BF               # map(*)
    FF            # primitive(*)
 ```
 
+### Ignoring unknown keys
+
+CBOR format is often used to communicate with [IoT] devices where new properties could be added as a part of a device's
+API evolution. By default, unknown keys encountered during deserialization produces an error.
```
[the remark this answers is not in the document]
twyatt: Copied from JSON documentation ( https://github.com/Kotlin/kotlinx.serialization/blame/7e526a8079a03a25f42667e62b9ecba2fc4b4fd8/docs/json.md#L126).

Should I also make the change there?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -428,6 +448,64 @@ internal class CborDecoder(private val input: ByteArrayInput) {
...
+        val lengthStack = mutableListOf<Int>()
```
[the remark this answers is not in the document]
twyatt: Is there a multiplatform-friendly dynamically sizing stack I could use in its place? Or would I have to roll my own?

If needed I could try to build a basic stack implementation to avoid the auto-boxing but I'm not confident my implementation would be optimal.

twyatt:
Did a first pass of addressing comments, I'll try to finish up on the remaining comments tomorrow or this weekend.

docs/formats.md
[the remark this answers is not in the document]
sandwwraith: Seems so

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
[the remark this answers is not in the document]
sandwwraith: Well, there's an `IntArrayBuilder` for deserializing IntArrays, but it is hidden for now. I think we can go with MutableList<Int> until we have a clear benchmarks that show if this is a real problem. @qwwdfsad WDYT?

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
twyatt: Summarized this PR comment as a code comment on the test function.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
twyatt: Unfortunately `decoder.skipElement()` (line 248) will skip the current value element and we'll need to check `isDone()` on the next loop to keep the previous short-circuit behavior (when beginning to read the next element value).

I tried lifting `isDone()` line (just to be sure) and it does in fact cause a number of tests to fail.

Would it be better to remove some of the code that is duplicated in the top-most `if` branches by having it always enter the `while` loop?:

```kotlin
override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
    val index: Int

    while (true) {
        if (isDone()) return CompositeDecoder.DECODE_DONE
        val elemName = decoder.nextString()
        readProperties++

        if (cbor.ignoreUnknownKeys) {
            val elementIndex = descriptor.getElementIndex(elemName)
            if (elementIndex == CompositeDecoder.UNKNOWN_NAME) {
                decoder.skipElement()
            } else {
                index = elementIndex
                break
            }
        } else {
            index = descriptor.getElementIndexOrThrow(elemName)
            break
        }
    }

    decodeByteArrayAsByteString = descriptor.isByteString(index)
    return index
}
```

---

I agree that the control flow is a bit difficult to follow in this function, so any other suggestions/strategies to refactor and simplify it a bit are appreciated.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -32,6 +32,9 @@ private const val HEADER_NEGATIVE: Byte = 0b001_00000
 private const val HEADER_ARRAY: Int = 0b100_00000
 private const val HEADER_MAP: Int = 0b101_00000
 
+/** Value to represent an indefinite length CBOR item within a "length stack". */
+private const val LENGTH_STACK_INDEFINITE = -1
```
twyatt, line 36: Created constant to avoid confusion with `-1` returned by `readByte` (which is used in `skipElement`).

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
```diff
@@ -104,4 +108,481 @@ class CborReaderTest {
...
+         *       61                            # text(1)
+         *          32                         # "2"
+         *       02                            # unsigned(2)
+         *       61                            # text(1)
+         *          33                         # "3"
+         *       03                            # unsigned(3)
+         *       FF                            # primitive(*)
+         *    61                               # text(1)
+         *       65                            # "e"
+         *    00                               # unsigned(0)
```
twyatt, line 407: This trailing data is so that we have something to validate that we skipped the last element (also appears in a few other tests).

Would it be better to add an `isEof` function to `CborDecoder`?
I could use the `isEof` in [`skipElement` function](https://github.com/twyatt/kotlinx.serialization/blob/abb0d5aac1d50bd384215b7f72fb1d9db4169823/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt#L488), but other than that I had trouble justifying adding a function simply for tests?

twyatt (commented):
I believe I've addressed or responded to all the comments/feedback.

Additionally, added tests for non-happy-paths and executed/committed `./gradlew knit`.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
twyatt: Good call, I've dropped the negative length check as the reading of the length (in `elementLength()` function) only reads headers that would **not** result in `readNumber` returning a negative number.

docs/formats.md
```diff
@@ -167,6 +168,55 @@ BF               # map(*)
    FF            # primitive(*)
 ```
 
+### Ignoring unknown keys
```
twyatt, line 171: Should I move this section above the directly proceeding section (**Byte arrays and CBOR data types**) since it uses the same example CBOR data so would flow a bit nicer for the reader?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -428,6 +451,88 @@ internal class CborDecoder(private val input: ByteArrayInput) {
...
+                val length = elementLength()
+                if (header == HEADER_ARRAY || header == HEADER_MAP) {
+                    if (length > 0) lengthStack.add(length)
+                } else {
+                    input.skip(length)
+                    prune(lengthStack)
+                }
+            }
+
+            if (readByte() == -1 && lengthStack.isNotEmpty()) throw SerializationException("EOF while skipping element")
```
sandwwraith, line 488:
```suggestion
            if (readByte() == -1 && lengthStack.isNotEmpty()) throw CborDecodingException("Unexpected EOF while skipping element")
```

This error usually caused by a malformed input, not a general problems with serialization framework

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -428,6 +451,88 @@ internal class CborDecoder(private val input: ByteArrayInput) {
...
+     * For primitive (finite length) elements (e.g. unsigned integer, text string), their length is read and
+     * corresponding number of bytes are skipped.
+     *
+     * For elements that contain children (e.g. array, map), the child count is read and added to a "length stack"
+     * (which represents the "number of elements" at each depth of the CBOR data structure). When a child element has
+     * been skipped, the "length stack" is [pruned][prune]. For indefinite length elements, a special marker is added to
+     * the "length stack" which is only popped from the "length stack" when a CBOR [break][isEnd] is encountered.
+     */
+    fun skipElement() {
+        if (curByte == -1) throw SerializationException("Missing element value")
```
sandwwraith, line 467: This message is kinda vague. Isn't this also a `CborDecodingException("Unexpected EOF while skipping element")` case?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +573,22 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
     getElementDescriptor(index) == ByteArraySerializer().descriptor &&
         getElementAnnotations(index).find { it is ByteString } != null
+
+/**
+ * Removes an item from the top of the [lengthStack], cascading the removal if the item represents the last item (i.e. a
+ * length value of `1`) at its stack depth.
+ *
+ * For example, pruning a [lengthStack] of `[3, 2, 1, 1]` would result in `[3, 1]`.
+ */
+private fun prune(lengthStack: MutableList<Int>) {
```
sandwwraith, line 583: Since this method is tied to a particular logic and algorithm (in contrary to `ByteArray.flatten()`), I'd suggest to move it inside decoder next to `skipElement` and `isIndefinite`

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborReaderTest.kt
sandwwraith: Since decoder does not have any public api, I think it is ok to add `isEof` to it. Moreover, cases when we skip last element with and without trailing data seem different to me, so it is important to test both.

sandwwraith (changes requested):

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
twyatt: 🤦  I totally missed the `CborDecodingException` constructor that takes a `message` (and doesn't require `curByte`).

twyatt:
Pulled in the latest `dev` but am no longer able to confirm everything is working:

```
./gradlew clean knit check

...

> Task :dokkaHtmlMultiModule FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dokkaHtmlMultiModule'.
> java.nio.file.Files.writeString(Ljava/nio/file/Path;Ljava/lang/CharSequence;[Ljava/nio/file/OpenOption;)Ljava/nio/file/Path;
```

I get the same failure when attempting the same command on `dev` (@ d3d2dca), so could just be something broken on my local machine setup w/ the new Dokka. 🤷

sandwwraith:
@twyatt What version of JDK do you use? Apparently this method is available only since Java 11.

sandwwraith (approved):
I ran the check for you with JDK 11 — it passes

sandwwraith:
Thank you for your contribution, hope it'll help a lot of users. It was a pleasure working with you!

PR #948, opened 2020-07-29
Serializer lookup enchancements

### LINKED ISSUE #802: typeOf serializer lookup should check contextual serializers

This feature is useful to support serializer lookup at runtime. My specific use-case is to implement a Ktor serializer that respects contextual serializers. I'm working with linked data, which means the deserialized data types should hold a reference to the HttpClient so they can resolve links. I've implemented a contextual serializer that injects the configured HttpClient, but to actually use the serializer (and support collection types), requires additional work.

Add a function `SerialModule.serializer(type: KType)`, replace:

https://github.com/Kotlin/kotlinx.serialization/blob/cf892556abc91d45bdb141e6992231b6c186b4fc/runtime/commonMain/src/kotlinx/serialization/SerializerResolving.kt#L57

With 

```kotlin
typeArguments.isEmpty() -> getContextualOrDefault(rootClass)
```

### LINKED ISSUE #803: serializerByTypeToken should support contextual serializers

Similar to #802, this would add support for contextual serializers when interoperating with Java APIs that provide `Type`. This would specifically help resolve this issue https://github.com/JakeWharton/retrofit2-kotlinx-serialization-converter/issues/18.

Implement `SerialModule.serializerByTypeToken(type: Type)`. Replace:

https://github.com/Kotlin/kotlinx.serialization/blob/cf892556abc91d45bdb141e6992231b6c186b4fc/runtime/jvmMain/src/kotlinx/serialization/JvmResolving.kt#L56

With

```kotlin
requireNotNull<KSerializer<out Any>>(getContextualOrDefault(type.kotlin as KClass<Any>)) as KSerializer<Any>
```

I'm still not sure how to find a registered contextual serializer for built-in generic types. So if someone had registered a contextual serializer for `List<Foo>`, this function would fail.

qwwdfsad:
* Lookup contextual serializer recursively in getcontextualOrDefault to give consistent experience when contextual type is nested in generic type argument
    * Provide getContextualOrDefault(javaType) for the same purpose
    * Get rid of UnsafeSerializationApi

Fixed #802
Fixes #803

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/Migrations.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/Migrations.kt
@@ -113,7 +113,7 @@ public annotation class SerialId @Deprecated(
 
 @Deprecated(level = DeprecationLevel.WARNING, message = "Use default parse overload instead", replaceWith = ReplaceWith("parse(objects)"))
 public inline fun <reified T : Any> StringFormat.parseList(objects: String): List<T> =
-    decodeFromString(ListSerializer(serializersModule.getContextualOrDefault<T>()), objects)
+    decodeFromString(ListSerializer(serializersModule.serializer<T>()), objects)
 
 @Deprecated(
     level = DeprecationLevel.WARNING,
@@ -121,7 +121,7 @@ public inline fun <reified T : Any> StringFormat.parseList(objects: String): Lis
     replaceWith = ReplaceWith("decodeFromString(map)")
 )
 public inline fun <reified K : Any, reified V : Any> StringFormat.parseMap(map: String): Map<K, V> =
-    decodeFromString(MapSerializer(serializersModule.getContextualOrDefault<K>(), serializersModule.getContextualOrDefault<V>()), map)
+    decodeFromString(MapSerializer(serializersModule.serializer<K>(), serializersModule.serializer<V>()), map)
 
 // ERROR migrations that affect **only** users that called these functions with named parameters
 
@@ -132,7 +132,7 @@ public inline fun <reified K : Any, reified V : Any> StringFormat.parseMap(map: 
     replaceWith = ReplaceWith("encodeToString(objects)")
 )
 public inline fun <reified T : Any> StringFormat.stringify(objects: List<T>): String =
-    encodeToString(ListSerializer(serializersModule.getContextualOrDefault<T>()), objects)
+    encodeToString(ListSerializer(serializersModule.serializer<T>()), objects)
 
 @LowPriorityInOverloadResolution
 @Deprecated(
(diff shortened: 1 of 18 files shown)


runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -141,28 +141,31 @@ public inline fun <reified T : Any> StringFormat.stringify(objects: List<T>): St
...
     message = "This method is deprecated for removal. Please use reified getContextualOrDefault<T>() instead",
-    replaceWith = ReplaceWith("getContextualOrDefault<T>()")
+    replaceWith = ReplaceWith("serializer<T>()")
 )
 public fun <T : Any> SerializersModule.getContextualOrDefault(value: T): KSerializer<T> =
     getContextual(value::class)?.cast() ?: value::class.serializer().cast()
 
+@Deprecated(
+    level = DeprecationLevel.ERROR,
+    message = "This method is deprecated for removal. Please use reified getContextualOrDefault<T>() instead",
```
sandwwraith, line 164:
```suggestion
    message = "This method is deprecated for removal. Please use reified serializer<T>() instead",
```

But since this method wasn't released yet, maybe it makes sense not to include migration?

runtime/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -92,7 +105,7 @@ public fun serializer(type: KType): KSerializer<Any?> {
  *
  * @throws SerializationException if serializer can't be found.
  */
-@UnsafeSerializationApi
+@InternalSerializationApi
```
sandwwraith, line 108: Not sure why this methods should be internal instead of unsafe. There is still possibility that one does not have `KType`/`j.l.r.Type`, but have `KClass`

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
qwwdfsad: Good idea

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -141,24 +141,20 @@ public inline fun <reified T : Any> StringFormat.stringify(objects: List<T>): St
...
 public inline fun <reified K : Any, reified V : Any> StringFormat.stringify(map: Map<K, V>): String =
-    encodeToString(MapSerializer(serializersModule.getContextualOrDefault<K>(), serializersModule.getContextualOrDefault<V>()), map)
+    encodeToString(MapSerializer(serializersModule.serializer<K>(), serializersModule.serializer<V>()), map)
 
-@ImplicitReflectionSerializer
-@OptIn(UnsafeSerializationApi::class)
 @Deprecated(
-    level = DeprecationLevel.WARNING,
+    level = DeprecationLevel.ERROR,
     message = "This method is deprecated for removal. Please use reified getContextualOrDefault<T>() instead",
```
sandwwraith, line 148: Are you sure this is the correct message given `ReplaceWith("getContextual(klass) ?: klass.serializer()")` ?

runtime/commonMain/src/kotlinx/serialization/Migrations.kt
```diff
@@ -141,24 +141,20 @@ public inline fun <reified T : Any> StringFormat.stringify(objects: List<T>): St
...
 )
 public fun <T : Any> SerializersModule.getContextualOrDefault(klass: KClass<T>): KSerializer<T> =
     getContextual(klass) ?: klass.serializer()
 
-@ImplicitReflectionSerializer
-@OptIn(UnsafeSerializationApi::class)
 @Deprecated(
-    level = DeprecationLevel.WARNING,
+    level = DeprecationLevel.ERROR,
     message = "This method is deprecated for removal. Please use reified getContextualOrDefault<T>() instead",
```
sandwwraith, line 156:
```suggestion
    message = "This method is deprecated for removal. Please use reified serializer<T>() instead",
```

sandwwraith:
Please double-check deprecation messages and replacements

PR #953, opened 2020-07-30
Renamed kotlinx-serialization-runtime to kotlinx-serialization-core

sandwwraith:
I don't want to change the actual folder name because that would mean a lot of changed files

No documentation was changed because readme will be new anyway.

Also bumped version to 1.0-M2

### DIFF
--- a/gradle/dokka.gradle
+++ b/gradle/dokka.gradle
@@ -9,10 +9,10 @@ dokka {
     outputFormat = 'html'
     outputDirectory = "$buildDir/dokka"
 
-    subProjects = ["kotlinx-serialization-runtime",
+    subProjects = ["kotlinx-serialization-core",
                    "kotlinx-serialization-cbor",
                    "kotlinx-serialization-properties",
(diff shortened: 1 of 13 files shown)


gradle/dokka.gradle
```diff
@@ -9,10 +9,10 @@ dokka {
     outputFormat = 'html'
     outputDirectory = "$buildDir/dokka"
 
-    subProjects = ["kotlinx-serialization-runtime",
+    subProjects = ["kotlinx-serialization-core",
                    "kotlinx-serialization-cbor",
                    "kotlinx-serialization-properties",
-//                   "kotlinx-serialization-runtime-configparser",  todo: fix publishing of jvm-only docs
+//                   "kotlinx-serialization-hocon",  todo: fix publishing of jvm-only docs
```
qwwdfsad, line 15: Do you plan to fix it here or later?

sandwwraith: later

PR #960, opened 2020-08-05
[API stabilization] Experimental API

qwwdfsad:
Didn't opt-in into experimental annotations on a project level to simplify reviewing and see where it leaks.
Will do after your initial review


runtime/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -419,8 +434,11 @@ public interface CompositeEncoder {
...
  * Begins a structure, encodes it using the given [block] and ends it.
  */
-public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, crossinline block: CompositeEncoder.() -> Unit) {
+public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, block: CompositeEncoder.() -> Unit) {
     val composite = beginStructure(descriptor)
-    composite.block()
-    composite.endStructure(descriptor)
+    try {
+        composite.block()
+    } finally {
```
sandwwraith, line 441: Not sure how much benefit it will bring since we do not give any exception-safety guarantees on encoder/decoder internal state anyway

runtime/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -184,8 +185,8 @@ public class JsonBuilder internal constructor(conf: JsonConf) {
 
     /**
      * Specifies indent string to use with [prettyPrint] mode.
-     * TODO specify whitespaces
      */
+    @ExperimentalSerializationApi
```
sandwwraith, line 189: It would be nice to drop a line with explanation why it's experimental

sandwwraith:
By the way, maybe we should use _beta_ instead of _experimental_ according to new Kotlin terminology? https://github.com/JetBrains/kotlin-web-site/pull/1792/files

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -23,6 +23,7 @@ import kotlinx.serialization.modules.*
  * @param serializersModule A [SerializersModule] which should contain registered serializers
  * for [Contextual] and [Polymorphic] serialization, if you have any.
  */
+@ExperimentalSerializationApi
```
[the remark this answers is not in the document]
qwwdfsad: Marking the receiver should be enough, but let's play it safe and apply your suggestion

runtime/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
qwwdfsad: This is done mostly for convenience, e.g. take a look at `AbstractPolymorphicSerializer#deserialize`

runtime/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -168,6 +173,7 @@ public inline fun <reified T> setSerialDescriptor(): SerialDescriptor {
  * Returns new serial descriptor for the same type with [isNullable][SerialDescriptor.isNullable]
  * property set to `true`.
  */
+@ExperimentalSerializationApi
```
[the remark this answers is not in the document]
qwwdfsad: Agree

runtime/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -121,13 +121,15 @@ public fun serialDescriptor(type: KType): SerialDescriptor = serializer(type).de
 /**
  * Creates a descriptor for the type `List<T>` where `T` is the type associated with [elementDescriptor].
  */
+@ExperimentalSerializationApi
```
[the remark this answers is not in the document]
qwwdfsad: Yes, we'll probably drop experimentality here around 1.1.0

runtime/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -28,6 +28,7 @@ import kotlin.reflect.*
  *     // intField is deliberately ignored by serializer -- not present in the descriptor as well
  *     element<Long>("_longField") // longField is named as _longField
  *     element("stringField", listDescriptor<String>())
```
sandwwraith, line 30: Maybe `descriptor<List<String>>` then? because `listDescriptor` is experimental

PR #965, opened 2020-08-11
Remove WASM target

qwwdfsad:
This target is not properly supported at the language level, is beign actively reworked and was never tested, thus it is not really applicable for stable release

### DIFF
--- a/gradle/native_mpp.gradle
+++ b/gradle/native_mpp.gradle
@@ -42,9 +42,6 @@ kotlin {
             // Windows
             addTarget(presets.mingwX64)
             addTarget(presets.mingwX86)
-
-            // WASM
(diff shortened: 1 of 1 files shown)


twyatt:
Apologies if this is a silly question (as I'm not super familiar with Kotlin multiplatform) but would Javascript still be a supported platform with the removal of the WASM target?

qwwdfsad:
@twyatt JS is still supported as a stable target, nothing is going to happen with that.
 WASM is (at this moment, this will change in the future) an experimental target of Kotlin/Native compiler and WASM as a standard is not yet ready for high-level languages (GC proposal is not done, no notion of exceptions etc.)

PR #972, opened 2020-08-13
Add compatibility guide

sandwwraith:
This doc, unlike kotlinx.coroutines, does not use knit. You may convert it to knit if you want


qwwdfsad:
Merged manually

Issue #989, opened 2020-08-19
Deprecation warnings with new lines break Kotlin/Native generated objective c


bsimon36:
**Describe the bug**
Upon upgrading to kotlin 1.4.0 / serialization 1.0.0-RC, building our Xcode project (which depends on our Kotlin library via the native.cocoapods plugin) fails because of the new lines in the deprecation message for JsonConfiguration. The new lines seem to make their way into the generated objective-c, which is then invalid because string literals can't span multiple lines. The error is `Expected expression` with a warning of `Missing terminating '"'` at the same line. The objective-c looks like this:

@interface ClientJsonConfiguration : ClientBase
- (instancetype)initWithEncodeDefaults:(BOOL)encodeDefaults ignoreUnknownKeys:(BOOL)ignoreUnknownKeys isLenient:(BOOL)isLenient serializeSpecialFloatingPointValues:(BOOL)serializeSpecialFloatingPointValues allowStructuredMapKeys:(BOOL)allowStructuredMapKeys prettyPrint:(BOOL)prettyPrint unquotedPrint:(BOOL)unquotedPrint indent:(NSString *)indent coerceInputValues:(BOOL)coerceInputValues useArrayPolymorphism:(BOOL)useArrayPolymorphism classDiscriminator:(NSString *)classDiscriminator __attribute__((swift_name("init(encodeDefaults:ignoreUnknownKeys:isLenient:serializeSpecialFloatingPointValues:allowStructuredMapKeys:prettyPrint:unquotedPrint:indent:coerceInputValues:useArrayPolymorphism:classDiscriminator:)"))) __attribute__((objc_designated_initializer)) __attribute__((unavailable("This class is deprecated for removal during serialization 1.0 API stabilization.
For configuring Json instances, the corresponding builder function can be used instead, e.g. instead of'Json(JsonConfiguration.Stable.copy(isLenient = true))' 'Json { isLenient = true }' should be used.
Instead of storing JsonConfiguration instances of the code, Json instances can be used directly:'Json(MyJsonConfiguration.copy(prettyPrint = true))' can be replaced with 'Json(from = MyApplicationJson) { prettyPrint = true }'"))); 

**To Reproduce**
Create a project using kotlin 1.4.0, depending on kotlinx.serialization 1.0.0-RC and link it to an Xcode project as described here: https://kotlinlang.org/docs/reference/native/cocoapods.html. Then try to build the project.

**Expected behavior**
No error, valid objective-c generated.

[...]

PR #990, opened 2020-08-19
Remove new lines in deprecation warnings that break native

bsimon36:
Fixes issue #989 

This is arguably just a Kotlin/Native bug, but they don't seem to be working on it: https://youtrack.jetbrains.com/issue/KT-39206


qwwdfsad:
Thanks!

bsimon36:
any idea when an updated version will be released with this fix?

PR #1007, opened 2020-08-24
DateAsStringSerializer → DateAsLongSerializer

madhead:
It seems like we are talking about `DateAsLongSerializer`, not `DateAsStringSerializer` here.


qwwdfsad (approved):
Thanks!

Issue #704, opened 2020-02-14
Wrapping any exception in deserialization process


krupt:
Example model: 
```
@Serializable
data class Model(
    val id: Long
)
```

```
json.parse("""
{
    "id": ""
}
""")
```
throws java.lang.NumberFormatException: For input string: ""

```
json.parse("""
{
    "id": "
}
""")
```
throws java.lang.StringIndexOutOfBoundsException: String index out of range: 99

I think the better is catching any exception in deserialization process and wrapping it into SerializationException or a new one (for example: JsonParseException).
What do you think?

PR #1011, opened 2020-08-25
StringIndexOutOfBoundsException when parsing empty string literal at the end of the input

ErikP0:
This PR fixes a bug where an unfinished string literal at the end of the input throws a `StringIndexOutOfBoundsException` instead of a `JsonDecodingException`.

```
Json(JsonConfiguration.Stable).parse("\"")
```
throws `StringIndexOutOfBoundsException`. Similar to #704, this should be a generic decoding exception instead.


sandwwraith (approved):

sandwwraith:
Good catch, thank you

PR #1013, opened 2020-08-26
Alignment with Android API <24

### LINKED ISSUE #1010: NoSuchMethodError: No virtual method getDeclaredAnnotation on Android API level <24
**Describe the bug**
```Kotlin
java.lang.NoSuchMethodError: No virtual method getDeclaredAnnotation(Ljava/lang/Class;)Ljava/lang/annotation/Annotation; in class Ljava/lang/Class; or its super classes (declaration of 'java.lang.Class' appears in /system/framework/core-libart.jar)
        at kotlinx.serialization.internal.PlatformKt.polymorphicSerializer(Platform.kt:70)
        at kotlinx.serialization.internal.PlatformKt.constructSerializerForGivenTypeArgs(Platform.kt:53)
        at kotlinx.serialization.internal.PlatformKt.compiledSerializerImpl(Platform.kt:23)
        at kotlinx.serialization.SerializersKt__SerializersKt.serializerOrNull(Serializers.kt:129)
        at kotlinx.serialization.SerializersKt__SerializersKt.serializerByKTypeImpl$SerializersKt__SerializersKt(Serializers.kt:65)
        at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:51)
        at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
        at kotlinx.serialization.SerializersKt__SerializersKt.builtinSerializerOrNull$SerializersKt__SerializersKt(Serializers.kt:76)
        at kotlinx.serialization.SerializersKt__SerializersKt.serializerByKTypeImpl$SerializersKt__SerializersKt(Serializers.kt:66)
        at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:51)
        at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
        at com.example.myapplication.MainActivity.onCreate(MainActivity.kt:17)
```

[Related issue with resolution](https://github.com/FasterXML/jackson-module-kotlin/issues/176)

**To Reproduce**
On Android API level <24:
```Kotlin
Json.decodeFromString<List<Int>>("[0,1,2,3]")
```

**Environment**
 - Kotlin version: 1.4.0
 - Library version: 1.0.0-RC
 - Kotlin platforms: JVM (Android)
 - Gradle version: 6.1.1

tom5079:
resolves #1010

### DIFF
--- a/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -53,16 +53,12 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
     return polymorphicSerializer()
 }
 
-@Suppress("UNCHECKED_CAST")
-private fun <A: Annotation, T: Any> Class<T>.getDeclaredAnnotation(annotationClass: KClass<A>) =
-    declaredAnnotations.firstOrNull { it.annotationClass == annotationClass } as? A
-
 private fun <T: Any> Class<T>.isNotAnnotated(): Boolean {
     /*
      * For annotated enums search serializer directly (or do not search at all?)
      */
-    return getDeclaredAnnotation(Serializable::class) == null &&
-            getDeclaredAnnotation(Polymorphic::class) == null
+    return getAnnotation(Serializable::class.java) == null &&
+            getAnnotation(Polymorphic::class.java) == null
 }
 
 private fun <T: Any> KClass<T>.polymorphicSerializer(): KSerializer<T>? {
@@ -71,10 +67,10 @@ private fun <T: Any> KClass<T>.polymorphicSerializer(): KSerializer<T>? {
      * annotations.
      */
     val jClass = java
-    if (jClass.getDeclaredAnnotation(Polymorphic::class) != null) {
+    if (jClass.getAnnotation(Polymorphic::class.java) != null) {
         return PolymorphicSerializer(this)
     }
-    val serializable = jClass.getDeclaredAnnotation(Serializable::class)
+    val serializable = jClass.getAnnotation(Serializable::class.java)
     if (serializable != null && serializable.with == PolymorphicSerializer::class) {
         return PolymorphicSerializer(this)
     }
@@ -87,7 +83,7 @@ private fun <T: Any> KClass<T>.interfaceSerializer(): KSerializer<T>? {
      * Check if it has no annotations or `@Serializable(with = PolymorphicSerializer::class)`,
      * otherwise bailout.
      */
-    val serializable = java.getDeclaredAnnotation(Serializable::class)
+    val serializable = java.getAnnotation(Serializable::class.java)
     if (serializable == null || serializable.with == PolymorphicSerializer::class) {
         return PolymorphicSerializer(this)
     }
--- a/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -53,12 +53,16 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
     return polymorphicSerializer()
 }
 
+@Suppress("UNCHECKED_CAST")
+private fun <A: Annotation, T: Any> Class<T>.getDeclaredAnnotation(annotationClass: KClass<A>) =
+    declaredAnnotations.firstOrNull { it.annotationClass == annotationClass } as? A
+
 private fun <T: Any> Class<T>.isNotAnnotated(): Boolean {
     /*
      * For annotated enums search serializer directly (or do not search at all?)
      */
-    return getDeclaredAnnotation(Serializable::class.java) == null &&
-            getDeclaredAnnotation(Polymorphic::class.java) == null
+    return getDeclaredAnnotation(Serializable::class) == null &&
+            getDeclaredAnnotation(Polymorphic::class) == null
 }
 
 private fun <T: Any> KClass<T>.polymorphicSerializer(): KSerializer<T>? {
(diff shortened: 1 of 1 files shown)


runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -53,12 +53,16 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
     return polymorphicSerializer()
 }
 
+@Suppress("UNCHECKED_CAST")
+private fun <A: Annotation, T: Any> Class<T>.getDeclaredAnnotation(annotationClass: KClass<A>) =
+    declaredAnnotations.firstOrNull { it.annotationClass == annotationClass } as? A
```
sandwwraith, line 58: I think there is no need to iterate through all the possible annotations, because it may be quite a slow process. Moreover, [`getDeclaredAnnotations()` method](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html#getDeclaredAnnotations--) is also marked as 'since Java 1.8', so it does not solve the original problem.

I think what should be done here is to replace `getDeclaredAnnotation` with regular [`getAnnotation`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html#getAnnotation-java.lang.Class-). 

The only difference between them, according to https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html , is that `getDeclaredAnnotation` ignores `@Inherited` annotations from superclasses. Since no annotation in kotlinx.serialization is inheritable, there should be no difference.

sandwwraith (changes requested):

runtime/jvmMain/src/kotlinx/serialization/internal/Platform.kt
tom5079: The problem is that android does not support Java 8 fully.
The reason behind the change is that [`getDeclaredAnnotations`](https://developer.android.com/reference/java/lang/reflect/AnnotatedElement#getDeclaredAnnotations()) is onle available for Android 24 and above while [`getDeclaredAnnotation`](https://developer.android.com/reference/java/lang/reflect/AnnotatedElement#getDeclaredAnnotation(java.lang.Class%3CT%3E)`) is available on all android versions.

The performance shouldn't be an issue too, because if you take a look at the implementation of `getDeclaredAnnotation`, it internally calls `getDeclaredAnnotations` so there's no deifference.

[Android](https://android.googlesource.com/platform/libcore/+/refs/tags/android-7.0.0_r1/ojluni/src/main/java/java/lang/Class.java#2395)
```java
@Override public Annotation[] getAnnotations() {
    HashMap<Class<?>, Annotation> map = new HashMap<Class<?>, Annotation>();
    for (Annotation declaredAnnotation : getDeclaredAnnotations()) {
        map.put(declaredAnnotation.annotationType(), declaredAnnotation);
    }
    for (Class<?> sup = getSuperclass(); sup != null; sup = sup.getSuperclass()) {
        for (Annotation declaredAnnotation : sup.getDeclaredAnnotations()) {
            Class<? extends Annotation> clazz = declaredAnnotation.annotationType();
            if (!map.containsKey(clazz) && clazz.isDeclaredAnnotationPresent(Inherited.class)) {
                map.put(clazz, declaredAnnotation);
            }
        }
    }
    Collection<Annotation> coll = map.values();
    return coll.toArray(new Annotation[coll.size()]);
}

```

[OpenJDK 8](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/Class.java#l3344)
```java
public <A extends Annotation> A getDeclaredAnnotation(Class<A> annotationClass) {
    Objects.requireNonNull(annotationClass);

    return (A) annotationData().declaredAnnotations.get(annotationClass);
}
```

tom5079: The problem is that android does not support Java 8 fully.
The reason behind the change is that [`getDeclaredAnnotations`](https://developer.android.com/reference/java/lang/reflect/AnnotatedElement#getDeclaredAnnotations()) is onle available for Android 24 and above while [`getDeclaredAnnotation`](https://developer.android.com/reference/java/lang/reflect/AnnotatedElement#getDeclaredAnnotation(java.lang.Class%3CT%3E)`) is available on all android versions.

The performance shouldn't be an issue too, because if you take a look at the implementation of `getDeclaredAnnotation`, it internally calls `getDeclaredAnnotations` so there's no deifference.

[OpenJDK 8](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/Class.java#l3344)
```java
public <A extends Annotation> A getDeclaredAnnotation(Class<A> annotationClass) {
    Objects.requireNonNull(annotationClass);

    return (A) annotationData().declaredAnnotations.get(annotationClass);
}
```

sandwwraith: Indeed, I misread the description: `getDeclaredAnnotations()` is available in Java 1.5 and Android API 1. However, aforementioned `annotationData().declaredAnnotations` is a map, so while `getDeclaredAnnotations()` allocates an additional array and puts map contents to it, `getDeclaredAnnotation` simply calls `Map.get`. So I'd still rather stick with `getAnnotation` instead of `getDeclaredAnnotations`.

tom5079: Hmm, I don't think the array of `Annotation` will take up that much memory even if there was several hundred annotation in a class to be honest

sandwwraith: It's more about unnecessary allocation/deallocation and GC work

tom5079: oops, I thought you're talking about `getDeclaredAnnotation` 😅 
the names are confusing, aren't they?
Also didn't notice that `getAnnotation` existed since API level 1.

Fixed.

sandwwraith:
Thank you!

PR #1031, opened 2020-09-02
Update default polymorphic serialization guide section

### LINKED ISSUE #1025: Can't round trip JSON from example polymorphic serialization
**Describe the bug**

The below link documents how to serialize and deserialize with a generic fallback element.  However it seems you are only able to deserialize.  When trying to serialize it gives an error because it can't use the "type" field on the "default" serializer.  The documentation doesn't indicate this would happen or suggest any work-around for creating JSON/Model combinations that can be round tripped

https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#default-polymorphic-type-handler-for-deserialization

**To Reproduce**

```import kotlinx.serialization.SerialName
import kotlinx.serialization.modules.SerializersModule
import kotlinx.serialization.modules.polymorphic
import kotlinx.serialization.modules.subclass
import kotlin.test.Test
import kotlin.test.assertEquals

@Serializable
abstract class Project {
    abstract val name: String
}

@Serializable
data class BasicProject(override val name: String, val type: String): Project()

@Serializable
@SerialName("OwnedProject")
data class OwnedProject(override val name: String, val owner: String) : Project()

val module = SerializersModule {
    polymorphic(Project::class) {
        subclass(OwnedProject::class)
        default { BasicProject.serializer() }
    }
}

val format = Json { serializersModule = module }

        const val jsonString = """
        [
            {"type":"unknown","name":"example"},
            {"type":"OwnedProject","name":"kotlinx.serialization","owner":"kotlin"} 
        ]
    """
class TestModule {
    @Test
    fun serialize() {

        val decoded = format.decodeFromString<List<Project>>(jsonString)

        assertEquals(jsonString, format.encodeToString(decoded))
    }
}
```

**Expected behavior**

Expected that this would at worst give an error for mismatched JSON, but it doesn't output JSON to even compare

**Environment**
 - Kotlin version: 1.4.0
 - Library version: 1.4.0
 - Kotlin platforms: JS
 - Gradle version: 6.6.1
 - IDE version (if bug is related to the IDE) [e.g. IntellijIDEA 2019.1, Android Studio 3.4]
 - Other relevant context [e.g. OS version, JRE version, ... ]

qwwdfsad:
Fixes #1025

### DIFF
--- a/docs/polymorphism.md
+++ b/docs/polymorphism.md
@@ -855,6 +855,11 @@ Notice, how `BasicProject` had also captured the specified type key in its `type
 
 <!--- TEST -->
 
+We used a plugin-generated serializer as a default serializer, implying that 
+the structure of the "unknown" data is known in advance. In a real-world API it's rarely the case and for
+that purpose a custom, less-structured serializer is requried. You will see the example of such serializer in the future section
+[Maintaining JSON custom attributes](json.md#maintaining-json-custom-attributes)
+
 ---
 
 The next chapter covers [JSON features](json.md).
--- a/docs/polymorphism.md
+++ b/docs/polymorphism.md
@@ -856,9 +856,9 @@ Notice, how `BasicProject` had also captured the specified type key in its `type
 <!--- TEST -->
 
 We used a plugin-generated serializer as a default serializer, implying that 
-the structure of the "unknown" data is known in advance. In a real-world API it's rarely the case and for
-that purpose a custom, less-structured serializer is requried. You will see the example of such serializer in the future section
-[Maintaining JSON custom attributes](json.md#maintaining-json-custom-attributes)
+the structure of the "unknown" data is known in advance. In a real-world API it's rarely the case.
+For that purpose a custom, less-structured serializer is needed. You will see the example of such serializer in the future section
+on [Maintaining custom JSON attributes](json.md#maintaining-custom-json-attributes).
 
 ---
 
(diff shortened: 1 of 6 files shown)


docs/polymorphism.md
```diff
@@ -34,6 +34,13 @@ import kotlinx.serialization.*
 import kotlinx.serialization.json.*
 -->
 
+<!--- INCLUDE .*-poly-18.*
```
elizarov, line 37: It's fragile (will need to fixed during any renumbering). Don't put the file name here. Just write INCLUDE directive without params to include this piece of code into the next example. And place it after the opening header of the section to help future editors of this text/code finding it.

docs/polymorphism.md
```diff
@@ -803,25 +810,49 @@ abstract class Project {
     abstract val name: String
 }
 
-@Serializable
-data class BasicProject(override val name: String, val type: String): Project()
-
 @Serializable
 @SerialName("OwnedProject")
 data class OwnedProject(override val name: String, val owner: String) : Project()
+
```
elizarov, line 816: I'd suggest to move it to a separate section named "Maintaining custom attributes" in the chapter on "JSON elements". That is, give a simpler example here, explain its disadvantage, and give a link to another section. Otherwise, people reading the guide top-to-bottom will not be able to understand what's going on here.

elizarov (changes requested):

elizarov (approved):
👍 Great. Just made some minor tweaks. Merging.

PR #1038, opened 2020-09-03
Support deserialization of top-level nullable types

### LINKED ISSUE #1035: java.util.NoSuchElementException: List is empty when decodeFromJsonElement with nullable type
**Describe the bug**
java.util.NoSuchElementException is raised when decodeFromJsonElement with nullable type is called

**To Reproduce**
```kotlin
@Serializable
data class A(val a: Int)

@Test
fun test() {
    val a = A(42)
    
    val json = Json.encodeToString(a)
    val jsonElement = Json.parseToJsonElement(json)
    
    print(Json.decodeFromString<A?>(json))               // A(a=42)
    print(Json.decodeFromJsonElement<A?>(jsonElement))   // java.util.NoSuchElementException
}
```

**Expected behavior**
Successful decode

**Environment**
 - Kotlin version: 1.4.0
 - Library version: 1.0.0-RC
 - Kotlin platforms: JVM
 - Gradle version: 6.3

qwwdfsad:
Also, fix the same for Hocon, but do not generalize it too frequently.

Fixes #1035

### DIFF
--- a/runtime/commonMain/src/kotlinx/serialization/internal/Tagged.kt
+++ b/runtime/commonMain/src/kotlinx/serialization/internal/Tagged.kt
@@ -10,9 +10,10 @@ import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
 import kotlinx.serialization.encoding.updateModeDeprecated
 import kotlinx.serialization.modules.*
+
 /*
  * These classes are intended to be used only within the kotlinx.serialization.
- * They neither do have stable API, not internal invariants and are changed without any warnings.
+ * They neither do have stable API, nor internal invariants and are changed without any warnings.
  */
 @InternalSerializationApi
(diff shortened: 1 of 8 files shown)


runtime/commonMain/src/kotlinx/serialization/internal/Tagged.kt
```diff
@@ -214,7 +215,13 @@ public abstract class TaggedDecoder<Tag : Any?> : Decoder,
 
     // ---- Implementation of low-level API ----
 
-    final override fun decodeNotNullMark(): Boolean = decodeTaggedNotNullMark(currentTag)
+    // TODO this method should be overridden by any sane format that supports top-level nulls
```
qwwdfsad, line 218: I am very happy we've decided to make this class internal and do not provide any stability :)

sandwwraith (approved):

PR #1054, opened 2020-09-08
Make all serialization exceptions eligible for coroutines exception recovery

### LINKED ISSUE #1047: Make all exceptions stacktrace recovery friendly
It would be nice to have proper stacktrace when using this lib in conjunction with coroutines https://github.com/Kotlin/kotlinx.coroutines/blob/1.3.9/docs/debugging.md#stacktrace-recovery-machinery

All exceptions should implement [CopyableThrowable](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-copyable-throwable/index.html) interface.

elizarov:
Fixes #1047

### DIFF
--- a/formats/cbor/build.gradle
+++ b/formats/cbor/build.gradle
@@ -25,6 +25,7 @@ kotlin {
                 implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jackson_version
                 implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: jackson_version
                 implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-cbor', version: jackson_version
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version"
             }
         }
     }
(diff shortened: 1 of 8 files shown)


formats/cbor/build.gradle
```diff
@@ -25,6 +25,7 @@ kotlin {
                 implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jackson_version
                 implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: jackson_version
                 implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-cbor', version: jackson_version
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version"
```
sandwwraith, line 28: The `protobuf` module also has its own `ProtobufDecodingException`, although it features only `message: String?` constructor, so it does not require changes. Should we add a test for it too?

sandwwraith:
@elizarov seems your new tests are failing during Gradle build (they're working in IDE though)

sandwwraith (approved):

PR #1057, opened 2020-09-08
fixes build error, added spaces on example guides

### LINKED ISSUE #1056: Cannot build the library
**Due to some inconsistencies knitCheck fails the build.**

**To Reproduce**
on master
```
./gradlew clean build
```

**Expected behavior**
build should be successful.

**Environment**
 - Kotlin version: Whatever code set in code
 - Library version: Latest code on master
 - Kotlin platforms: JVM
 - Gradle version: Whatever set in code
 - IDE version: Using terminal
 - Other relevant context: Using MacOS latest stable.

**Possible Cause**
#1052 without updating example serializers.

guness:


sandwwraith:
Indeed, thanks

PR #1064, opened 2020-09-12
Add community CBOR format from L-Briand

L-Briand:
I just finished my cbor serializer/deserializer and wants to share it with community :slightly_smiling_face: [link to repo](https://github.com/L-Briand/obor)

### DIFF
--- a/formats/README.md
+++ b/formats/README.md
@@ -70,3 +70,11 @@ not be available in other formats such as JSON.
 * Platform: JVM only
 
 Allows serialization and deserialization of objects to and from [YAML](http://yaml.org).
+
+### CBOR
+
+* GitHub repo: [L-Briand/obor](https://github.com/L-Briand/obor)
+* Artifact ID: `net.orandja.obor:obor`
+* Platform: JVM, Android
+
+Allow serialization and deserialization of objects to and from [CBOR](https://cbor.io/). This codec can be used to read and write from Java InputStream and OutputStream.


qwwdfsad (approved):
Cool, thanks for your work!
I'll merge it soon.

Could you please elaborate do you have to maintain your own implementation of CBOR if we are going to provide integration with Java streams?
It would be nice if we could avoid duplicate work :)

L-Briand:
I am alone on this project. My goal was to understand how serialization worked with kotlin on a format i like. I was a bit lost when looking a the sources on the kotlinx cbor project as its a bit messy. I have done a serializer with rust a while back and have based my work on it.

I'll continue to develop it in my spare time to add some more functionalities and because i just like it :) If you want to take some of my work i clearly do not mind. The repo is in BSD 3 clause licence.

Will you add supports on streams with kotlinx-io ? Will it not take a lot of time to do it ?

qwwdfsad:
>If you want to take some of my work i clearly do not mind. The repo is in BSD 3 clause licence

Great, many thanks!

>Will you add supports on streams with kotlinx-io?

For a start, without kotlinx-io, but with Java streams and, probably, with Okio. ETA is version 1.1.0. It's required not only for CBOR, but for JSON and ProtoBuf as well

>Is there someone who leads on kotlinx-cbor project

Not at this point. kotlinx-serialization-cbor is experimental. We do care about its users, we accept pull-requests (e.g. recently added `@ByteString` was contributes externally), fix bugs, but we don't have a strong commitment to stabilize it in the nearest future.

PR #1070, opened 2020-09-14
RC cleanups

qwwdfsad:


formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicObjectSerializer.kt
```diff
@@ -71,6 +71,7 @@ private class DynamicObjectEncoder(
         OBJ, MAP, LIST
     }
 
+    @OptIn(ExperimentalSerializationApi::class)
```
sandwwraith, line 74: We have `file:OptIn`. Why it's not enough?

sandwwraith (approved):

formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicObjectSerializer.kt
qwwdfsad: Because IDE support is buggy

PR #1075, opened 2020-09-16
Minor fix on serializers module docs

f-lombardo:
I think that actually the `contextual` fun needs the class to deserialize to as is 1st parameter

### DIFF
--- a/docs/serializers.md
+++ b/docs/serializers.md
@@ -884,7 +884,7 @@ class this serializer is defined for is fetched automatically via the `reified` 
 
 ```kotlin
 private val module = SerializersModule { 
-    contextual(DateAsLongSerializer)
+    contextual(Date::class, DateAsLongSerializer)
 }
 ``` 
 


elizarov:
This code is actually being tested. See here https://github.com/Kotlin/kotlinx.serialization/blob/master/guide/example/example-serializer-18.kt
It is not the class that is being registered, but an instance of the serializer object.

Anyway, if you change it, you'll need to also rebuild the examples and tests with `./gradlew knit`, make sure it passes all the tests, and commit the changes together with PR.

f-lombardo:
> 
> 
> This code is actually being tested. See here https://github.com/Kotlin/kotlinx.serialization/blob/master/guide/example/example-serializer-18.kt
> It is not the class that is being registered, but an instance of the serializer object.
> 
> Anyway, if you change it, you'll need to also rebuild the examples and tests with `./gradlew knit`, make sure it passes all the tests, and commit the changes together with PR.

Oh, I see: the problem is in the wrong import. In my conde I wrote:
`import kotlinx.serialization.modules.SerializersModule`

While in the example is:
`import kotlinx.serialization.modules.*`

Isn't this API a bit weird, since the IDEA autocomplete automatically leads to the import I made?

Thank you

elizarov:
I've tried to reproduce this problem, but I could not. I've turned off star import and tried writing this code. It is red, but pressing "Alt+ENTER" adds the appropriate `import kotlinx.serialization.modules.contextual` for me, making it all correct. If you can figure out how to reproduce this problem in IDEA, then can you, please, submit an issue to https://kotl.in/issue

f-lombardo:
OK, I will try.
Do you think that adding a note in the docs that warns about making the right import would be useful?
Moreover, would it be a good idea to mention `./gradlew knit` in  `CONTRIBUTING.md`?
Thank you

elizarov:
> Moreover, would it be a good idea to mention `./gradlew knit` in `CONTRIBUTING.md`?

Done. #1076

PR #1079, opened 2020-09-16
Fix "docs needed" tagged issues

qwwdfsad:
I'll squash & merge them into a single one after the review

### DIFF
--- a/core/jsMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/core/jsMain/src/kotlinx/serialization/internal/Platform.kt
@@ -28,7 +28,7 @@ internal actual fun KClass<*>.platformSpecificSerializerNotRegistered(): Nothing
     throw SerializationException(
         "Serializer for class '${simpleName}' is not found.\n" +
                 "Mark the class as @Serializable or provide the serializer explicitly.\n" +
-                "On K/JS explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
+                "On Kotlin/JS explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
     )
 }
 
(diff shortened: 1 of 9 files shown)


core/jsMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -24,6 +24,14 @@ internal actual fun <T : Any, E : T?> ArrayList<E>.toNativeArrayImpl(eClass: KCl
 
 internal actual fun Any.isInstanceOf(kclass: KClass<*>): Boolean = kclass.isInstance(this)
 
+internal actual fun KClass<*>.platformSpecificSerializerNotRegistered(): Nothing {
+    throw SerializationException(
+        "Serializer for class '${simpleName}' is not found.\n" +
+                "Mark the class as @Serializable or provide the serializer explicitly.\n" +
+                "On K/JS explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
```
sandwwraith, line 31: Let's use official naming
```suggestion
                "On Kotlin/JS explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
```

core/nativeMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -17,6 +17,14 @@ internal actual inline fun BooleanArray.getChecked(index: Int): Boolean {
     return get(index)
 }
 
+internal actual fun KClass<*>.platformSpecificSerializerNotRegistered(): Nothing {
+    throw SerializationException(
+        "Serializer for class '${simpleName}' is not found.\n" +
+                "Mark the class as @Serializable or provide the serializer explicitly.\n" +
+                "On K/N explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
```
sandwwraith, line 24:
```suggestion
                "On Kotlin/Native explicitly declared serializer should be used for interfaces and enums without @Serializable annotation"
```

sandwwraith (approved):

PR #1084, opened 2020-09-17
Change default value of `encodeDefaults` config flag

sandwwraith:


qwwdfsad (changes requested):
Please fix our guide tests

docs/basic-serialization.md
```diff
@@ -437,10 +437,10 @@ fun main() {
 
 > You can get the full code [here](../guide/example/example-classes-09.kt).
 
-It produces the following output, which has `language` property, even though its value is equal to the default one.
+It produces the following output, which does not have `language` property, because its value is equal to the default one.
```
qwwdfsad, line 440: Please add a short sentence explaining the rationale of this behaviour

qwwdfsad (changes requested):
Our arguments regarding encode defaults are also applicable to Protobuf, Cbor, Hocon etc. Please change it everywhere

docs/basic-serialization.md
qwwdfsad: Friendly reminder

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -25,6 +25,7 @@ import kotlinx.serialization.modules.*
  * from corresponding Kotlin objects. However, other 3rd-party parsers (e.g. `jackson-dataformat-cbor`) may not accept such maps.
  *
  * @param encodeDefaults specifies whether default values of Kotlin properties are encoded.
+ *                       False by default; meaning that properties with values equal to defaults would be elided.
```
qwwdfsad, line 28: would be -> will be

qwwdfsad (approved):
Otherwise looks good

docs/basic-serialization.md
sandwwraith: Yeah thanks

v4dkou:
> It produces the following output, which does not have `language` property, because its value is equal to the default one.

@sandwwraith Why is that a problem that the value is equal to the default one?

Almost submitted this as a bug, because this behaviour messes things up when transferring objects from Kotlin to JS context as plain JSONs. An object has a value in Kotlin, but then it loses it in JS (unless the flag is specifically set to true), which is **really** unexpected.

PR #1089, opened 2020-09-19
Do not provide default implementation for serializersModule in Abstra…

qwwdfsad:
…ctEncoder and AbstractDecoder

This default is not reasonable, never wanted and potentially error-prone


formats/json/commonTest/src/kotlinx/serialization/UnknownElementIndexTest.kt
```diff
@@ -14,6 +15,8 @@ class UnknownElementIndexTest {
     data class Holder(val c: Choices)
 
     class MalformedReader : AbstractDecoder() {
+        override val serializersModule: SerializersModule =EmptySerializersModule
```
sandwwraith, line 18:
```suggestion
        override val serializersModule: SerializersModule = EmptySerializersModule
```

sandwwraith (approved):

PR #1096, opened 2020-09-23
Add handling of the project property that allows to disable cross-compilation

sandwwraith:
without removing targets from module file.

This is needed to reduce amount of time wasted during release process (linux targets are built and uploaded three times, because they're available on all hosts)

### DIFF
--- a/gradle/native-targets.gradle
+++ b/gradle/native-targets.gradle
@@ -2,12 +2,70 @@
  * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
+/**
+ * Specifies what subset of Native targets to build
+ *
+ * ALL — all possible for compilation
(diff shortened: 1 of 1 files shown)


gradle/native-targets.gradle
```diff
@@ -2,24 +2,78 @@
...
 project.ext.nativeMainSets = []
 project.ext.nativeTestSets = []
 
+/**
+ * Disables compilation but leaves the target in .module file
+ */
+def disableCompilation(targets) {
+    configure(targets) {
+        compilations.all {
+            cinterops.all { project.tasks[interopProcessingTaskName].enabled = false }
```
qwwdfsad, line 46: We don't have cinterop, do we?

qwwdfsad (approved):

gradle/native-targets.gradle
sandwwraith: We don't. Although I've decided to left this snippet 'as-is' so it would not differ much with ktor's one

PR #1106, opened 2020-09-29
Add decoding map in the root of HOCON config Fixes

### LINKED ISSUE #1046: Hcon: error while trying to decode MAP
**Describe the bug**

**bug.conf**
```
value1 {
    name = "name1",
    dir = "dir1"
    fileExtension = ".xlsx"
}

value2 {
    name = "name2",
    dir = "dir2"
    fileExtension = ".xlsx"
}

value3 {
    name = "name3",
    dir = "dir3"
    fileExtension = ".xlsx"
}
```

**code**
```
@Serializable
data class TestConfig(
    val name: String,
    val dir: String = "",
    val fileExtension: String = ".xlsx",
)

@ExperimentalSerializationApi
fun main() {
    val bugConfig = Files.newBufferedReader(Path.of("./bug.conf")).use(ConfigFactory::parseReader)
    val bugMap = Server.hoconParser.decodeFromConfig<Map<String, TestConfig>>(bugConfig)
}
```
**Exception**
```
Exception in thread "main" java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.last(_Collections.kt:380)
	at kotlinx.serialization.internal.TaggedDecoder.getCurrentTag(Tagged.kt:308)
	at kotlinx.serialization.hocon.Hocon$ConfigReader.beginStructure(Hocon.kt:138)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.merge(CollectionSerializers.kt:29)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.deserialize(CollectionSerializers.kt:43)
	at kotlinx.serialization.encoding.Decoder$DefaultImpls.decodeSerializableValue(Decoding.kt:243)
	at kotlinx.serialization.internal.TaggedDecoder.decodeSerializableValue(Tagged.kt:179)
	at kotlinx.serialization.hocon.Hocon.decodeFromConfig(Hocon.kt:34)
	at TestKt.main(test.kt:194)
	at TestKt.main(test.kt)

```
error in code https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt#L138:
```
descriptor.kind == StructureKind.MAP -> MapConfigReader(conf.getObject(currentTag)) // currentTag NULL, tagStack: size 0
```

**Environment**
 - Kotlin version: 1.4.0
 - Library version: 1.0.0-RC
 - Kotlin platforms: JVM
 - Gradle version: 6.6.1

shanshin:


formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconMapTest.kt
```diff
@@ -0,0 +1,47 @@
...
+
+class HoconRootMapTest {
+    @Serializable
+    data class CompositeValue(
+            val a: String,
+            val b: Int
+    )
+
+    @Test
+    fun `config with root map`() {
```
qwwdfsad, line 38: Backticked tests are legacy from the initial versions. We stick to `testFoo` naming convention.

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconMapTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val config = ConfigFactory.parseString(CONFIG_FOR_ROOT_MAP)
+        val hocon = Hocon {}
```
qwwdfsad, line 40: You can use `Hocon` object: `Hocon.decodeFromConfig<...>(...)`

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconMapTest.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package kotlinx.serialization.hocon
+
+import com.typesafe.config.ConfigFactory
+import kotlinx.serialization.Serializable
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+private const val CONFIG_FOR_ROOT_MAP = """
```
qwwdfsad, line 12: Please embed it into the test class to have a single classfile. Don't insist tho

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -141,7 +141,9 @@ public sealed class Hocon(
             when {
                 descriptor.kind.listLike -> ListConfigReader(conf.getList(currentTag))
                 descriptor.kind.objLike -> if (ind > -1) ConfigReader(conf.getConfig(currentTag)) else this
-                descriptor.kind == StructureKind.MAP -> MapConfigReader(conf.getObject(currentTag))
+                descriptor.kind == StructureKind.MAP ->
+                    //if current tag is null - map in the root of config
+                    MapConfigReader(if (currentTagOrNull != null) conf.getObject(currentTag) else conf.root())
```
qwwdfsad, line 146: This fix indeed fixes reported problem, but while we are here, I'd suggest to fix all similar problems consistently and thoughtfully.

What else here worth investigating:
- Top-level lists. If they are not supported in HOCON, then clear exception message
- Top-level primitives. Ditto as with lists
- Nullability. Test `Map<..>?` both when it's null (if it's supported) and when it's not. Same for lists/primitives

qwwdfsad (changes requested):

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
shanshin: 1. root-level lists supported in HOCON spec https://github.com/lightbend/config/blob/master/HOCON.md#syntax but typesafe config library support only subset of HOCON syntax https://github.com/lightbend/config/issues/460. So library not decode root-level lists
2. root-level primitives are not supported by HOCON and typesafe lib
3. root-level Map<>? supported but have unexpected behavior: if decode empty object "{}" not-null map reads as empty and nullable map reads as null. There is no way to define "null" config because primitives are illegal

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -141,7 +141,9 @@ public sealed class Hocon(
             when {
                 descriptor.kind.listLike -> ListConfigReader(conf.getList(currentTag))
                 descriptor.kind.objLike -> if (ind > -1) ConfigReader(conf.getConfig(currentTag)) else this
-                descriptor.kind == StructureKind.MAP -> MapConfigReader(conf.getObject(currentTag))
+                descriptor.kind == StructureKind.MAP ->
+                    //if current tag is null - map in the root of config
```
sandwwraith, line 145:
```suggestion
                    // if current tag is null - map in the root of config
```

IMO, it's more readable. Most comments in codebase a written with this style (although capitalization differs, we probably need standard on that too)

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    fun testEmptyObjectDecode() {
+        val config = ConfigFactory.parseString(configWithEmptyObject)
+        //non-null map decoded from empty object as empty map
+        val map = Hocon.decodeFromConfig<Map<String, String>>(config)
+        assertTrue(map.isEmpty())
+
+        //nullable map decoded from empty object as null - not obvious
```
sandwwraith, line 57: I think this behavior is actually ok: given that `Hocon.decodeFromConfig<ComplexValue?>(emptyConfig)` returns null (or I really hope it does so, needs testing), this behavior is more consisntent that it may looks like at first sight.

Anyway, we can change it if we really want. Since original `config` is Java library, it seems that it cannot distinguish between nullable and non-null types, so this part is likely caused by our parser.

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        val nullableMap = Hocon.decodeFromConfig<Map<String, String>?>(config)
+        assertNull(nullableMap)
+
+        //root-level list in config not supported but nullable list can be decoded from empty object
+        val nullableList = Hocon.decodeFromConfig<List<String>?>(config)
+        assertNull(nullableList)
```
sandwwraith, line 63: It seems there's no need for separate variable and you can join lines: `assertNull(Hocon.decodeFromConfig<List<String>?>(config))`

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun testErrors() {
+        //because com.typesafe:config lib not support list in root we can't decode non-null list
+        val config = ConfigFactory.parseString(configWithEmptyObject)
+        var error: Exception? = null
+        try {
+            Hocon.decodeFromConfig<List<String>>(config)
+        } catch (e: NoSuchElementException) {
+            error = e
+        }
+        assertNotNull(error)
```
sandwwraith, line 77: It seems to me that this construction can be replaced with `assertFailsWith` function

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
shanshin: Disagree. In case of static-size types (e.q. some custom ComplexValue) it's better to return null.

But as user with variable-size structure (Map) I expect size of map will be equals with JSON object properties count. In config object exists, so it's surprise for me - function behavior changed after changing return type to nullable, not changing of config string.
For example Jackson does as I said
objectMapper.readValue("{}", Map::class.java)  returns empty map and it is expected - I see begin object symbol and end object symbol - so object exists but not contains any properties.

Returning null is not a bug but some users can be confused this behavior.

sandwwraith: I've tested json and
```kotlin
        println(Json.decodeFromString<Map<String, String>>(emptyJson))
        println(Json.decodeFromString<Map<String, String>?>(emptyJson))
```

both give empty map as a result. And decoding even nullable user type from "{}" gives an error.

So we'd indeed better change that, if it's possible

shanshin: Change HOCON decoder behavior? In this PR?

sandwwraith: Yes

sandwwraith: For this case (nullable map)

sandwwraith:
Follow-up: #1157

PR #1111, opened 2020-09-30
Get rid of deprecated API and replacements

qwwdfsad:


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -163,7 +163,7 @@ internal open class ProtobufDecoder(
 
     override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T = decodeSerializableValue(deserializer, null)
 
-    @Suppress("UNCHECKED_CAST")
+    @Suppress("UNCHECKED_CAST", "INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```
qwwdfsad, line 166: Not sure about this one, but seems to be ok

qwwdfsad:
Build is failing because we're waiting for the fix in new Dokka

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
```diff
@@ -543,14 +496,22 @@ public interface CompositeDecoder {
...
+    @kotlin.internal.LowPriorityInOverloadResolution
+    @Deprecated(decodeMethodDeprecated, level = DeprecationLevel.ERROR)
+    public fun <T : Any?> decodeSerializableElement(
+        descriptor: SerialDescriptor,
+        i: Int, // renamed from index to be called even with LowPriority
+        deserializer: DeserializationStrategy<T>
+    ): T = decodeSerializableElement(descriptor, i, deserializer, null)
 
     @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE", "DeprecatedCallableAddReplaceWith")
     @kotlin.internal.LowPriorityInOverloadResolution
```
qwwdfsad, line 517: Compiler _still_ uses these two deprecated declarations. It should be fixed in 1.4.20

sandwwraith: Can we deprecate them as hidden though?

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
```diff
@@ -256,16 +242,10 @@ public interface Decoder {
...
-    @Deprecated(updateMethodDeprecated, level = DeprecationLevel.ERROR)
-    public fun <T> updateSerializableValue(deserializer: DeserializationStrategy<T>, old: T): T {
-        return decodeSerializableValue(deserializer)
-    }
-
     // Not documented
     @OptIn(ExperimentalSerializationApi::class)
     @Suppress("DeprecatedCallableAddReplaceWith")
-    @Deprecated(updateMethodDeprecated, level = DeprecationLevel.ERROR)
+    @Deprecated("", level = DeprecationLevel.HIDDEN)
```
sandwwraith, line 248: Why HIDDEN instead of deletion?

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
sandwwraith: Actually, I've checked plugin code and it seems to me that plugin uses 4-argument functions. How you determined that these are still in use?

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -82,7 +81,7 @@ public sealed class ListLikeSerializer<Element, Collection, Builder>(
     }
 }
 
-@InternalSerializationApi // Not intended for public use at all
+@InternalSerializationApi
```
sandwwraith, line 84: Why not `@PublishedApi internal class` too?

core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
```diff
@@ -74,170 +74,58 @@ private fun checkName(serialName: String) {
...
-public object UnitSerializer : KSerializer<Unit> by ObjectSerializer("kotlin.Unit", Unit)
+internal object UnitSerializer : KSerializer<Unit> by ObjectSerializer("kotlin.Unit", Unit)
 
-@Deprecated(
-    message = "Deprecated in the favour of Boolean.serializer() extension",
-    level = DeprecationLevel.ERROR,
-    replaceWith = ReplaceWith("Boolean.serializer()", imports = ["kotlinx.serialization.builtins.serializer"])
-)
-public object BooleanSerializer : KSerializer<Boolean> {
+internal object BooleanSerializer : KSerializer<Boolean> {
```
sandwwraith, line 79: For the additional reassurance, lets mark them with `@PublishedApi` too — they can be used in plugin-generated code

core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
```diff
@@ -7,6 +7,7 @@
 
 package kotlinx.serialization
 
+import kotlinx.serialization.builtins.MapEntrySerializer
```
sandwwraith, line 10: unnecessary?

formats/json/jvmTest/src/kotlinx/serialization/features/JsonUpdateCustomTest.kt
```diff
@@ -1,53 +0,0 @@
...
-    }
-
-    @Serializable
-    data class WrappedMap<T>(val mp: Map<String, T>)
-
-    val json = Json
-
-    @Test
-    fun canUpdateMap() {
-        val parsed = json.decodeFromString(WrappedMap.serializer(Int.serializer()), """{"mp": { "x" : 23, "x" : 42, "y": 4 }}""")
```
sandwwraith, line 44: these tests now actually check that json uses the last encountered value, so we'd better rename them and put to proper place

sandwwraith (changes requested):

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
qwwdfsad: If I remove them, tests starts to fail with NSME

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
qwwdfsad: See above

core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
qwwdfsad: Sure

core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
qwwdfsad: Nope, triggers ambiguity between factory and internal class constructor

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
qwwdfsad: Used in ProtoBuf, I've left a comment

formats/json/jvmTest/src/kotlinx/serialization/features/JsonUpdateCustomTest.kt
qwwdfsad: Done

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
sandwwraith: Ok, I've figured it out: compiler generates wrong code only for JVM and only for code path where `decodeSequentially = true`. Seems I was using there 3-arg function because, well, `oldValue` does not make sense for sequential reading. So we indeed have to leave this functions hidden if we want sequential reading working (for me it looks like a better solution instead of breaking it)

sandwwraith (approved):
pls revert my last commit

sandwwraith:
and rebase to actual `dev`

PR #1112, opened 2020-10-01
Fix deserialization of float and double types in CBOR

shanshin:
* add support of half-precision floating-point values for double and float decoders
  * double decoder now can read float values
  * write tests for boundary and other values

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
@@ -17,6 +17,7 @@ private const val FALSE = 0xf4
 private const val TRUE = 0xf5
 private const val NULL = 0xf6
 
+private const val NEXT_HALF = 0xf9
 private const val NEXT_FLOAT = 0xfa
 private const val NEXT_DOUBLE = 0xfb
 
@@ -34,6 +35,15 @@ private const val HEADER_MAP: Int = 0b101_00000
 
 /** Value to represent an indefinite length CBOR item within a "length stack". */
 private const val LENGTH_STACK_INDEFINITE = -1
+
+private const val HALF_PRECISION_EXPONENT_BIAS = 15
+private const val HALF_PRECISION_MAX_EXPONENT = 0x1f
+private const val HALF_PRECISION_MAX_MANTISSA = 0x3ff
+
+private const val SINGLE_PRECISION_EXPONENT_BIAS = 127
+private const val SINGLE_PRECISION_MAX_EXPONENT = 0xFF
+
+private const val SINGLE_PRECISION_NORMALIZE_BASE = 0.5f
 
 // Differs from List only in start byte
 private class CborMapWriter(cbor: Cbor, encoder: CborEncoder) : CborListWriter(cbor, encoder) {
@@ -422,15 +432,22 @@ internal class CborDecoder(private val input: ByteArrayInput) {
     }
 
     fun nextFloat(): Float {
-        if (curByte != NEXT_FLOAT) throw CborDecodingException("float header", curByte)
-        val res = Float.fromBits(readInt())
+        val res = when (curByte) {
+            NEXT_FLOAT -> Float.fromBits(readInt())
+            NEXT_HALF -> floatFromHalfBits(readShort())
+            else -> throw CborDecodingException("float header", curByte)
+        }
         readByte()
         return res
     }
 
     fun nextDouble(): Double {
-        if (curByte != NEXT_DOUBLE) throw CborDecodingException("double header", curByte)
-        val res = Double.fromBits(readLong())
+        val res = when (curByte) {
+            NEXT_DOUBLE -> Double.fromBits(readLong())
+            NEXT_FLOAT -> Float.fromBits(readInt()).toDouble()
+            NEXT_HALF -> floatFromHalfBits(readShort()).toDouble()
+            else -> throw CborDecodingException("double header", curByte)
+        }
         readByte()
         return res
     }
@@ -442,6 +459,12 @@ internal class CborDecoder(private val input: ByteArrayInput) {
             result = (result shl 8) or byte.toLong()
         }
         return result
+    }
+
+    private fun readShort(): Short {
+        val highByte = input.read()
+        val lowByte = input.read()
+        return (highByte shl 8 or lowByte).toShort()
     }
 
     private fun readInt(): Int {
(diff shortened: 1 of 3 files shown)


formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
+
+            mant = mant and 0x3ff //max half mantis value
+        }
+    } else {
+        //normalized value
+        exp = (halfExp + 1008) //double exponent bias - half exponent bias = 1023 - 15
+        mant = halfMant
+    }
+    return (sign shl 63) or (exp shl 52) or (mant shl 42)
+}
```
twyatt, line 519: The [example code provided in the CBOR RFC](https://tools.ietf.org/html/rfc7049#appendix-D) seems a bit simpler and doesn't have a loop. Are there specific advantages to this algorithm as apposed to the one listed in the RFC?

sandwwraith: Not sure RFC's one work with denormalized numbers. Btw, sources of AOSP (https://android.googlesource.com/platform/libcore/+/master/luni/src/main/java/libcore/util/FP16.java#578) are using magic constant instead of a loop.

formats/cbor/jvmTest/src/kotlinx/serialization/cbor/CborCompatibilityTest.kt
```diff
@@ -76,4 +79,42 @@ class CborCompatibilityTest {
...
+        compare(FloatData(Float.MAX_VALUE), FloatData.serializer())
+        compare(FloatData(Float.MIN_VALUE), FloatData.serializer())
+        compare(FloatData(0.0f), FloatData.serializer())
+        compare(FloatData(-0.0f), FloatData.serializer())
+        compare(FloatData(-1.0f), FloatData.serializer())
+        compare(FloatData(1.0f), FloatData.serializer())
+        compare(FloatData(123.56f), FloatData.serializer())
+        compare(FloatData(123.0f), FloatData.serializer())
+        //minimal denormalized value in half-precision
+        compare(FloatData(5.9604645E-8f), FloatData.serializer())
```
sandwwraith, line 97: Am I right that the library we are testing our implementation with always encodes numbers as half-precision if this is possible to do so?

sandwwraith (changes requested):
Sorry, I haven't dug enough into the differences of HFP algorithms to determine which one is better. I can return to them later if you wish. Right now, the only questionable part for me in current implementation is the loop, but it is not something critical.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -397,15 +398,22 @@ internal class CborDecoder(private val input: ByteArrayInput) {
     }
 
     fun nextFloat(): Float {
-        if (curByte != NEXT_FLOAT) throw CborDecodingException("float header", curByte)
-        val res = Float.fromBits(readInt())
+        val res: Float = when(curByte) {
```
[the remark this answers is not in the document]
qwwdfsad: Also, an indent is required

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -397,15 +398,22 @@ internal class CborDecoder(private val input: ByteArrayInput) {
...
+            else -> throw CborDecodingException("float header", curByte)
+        }
         readByte()
         return res
     }
 
     fun nextDouble(): Double {
-        if (curByte != NEXT_DOUBLE) throw CborDecodingException("double header", curByte)
-        val res = Double.fromBits(readLong())
+        val res: Double = when(curByte) {
```
qwwdfsad, line 411: Same here, indent

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
 private fun SerialDescriptor.isByteString(index: Int): Boolean =
     getElementDescriptor(index) == ByteArraySerializer().descriptor &&
         getElementAnnotations(index).find { it is ByteString } != null
+
+private fun halfToDouble(bits: Short): Long {
```
qwwdfsad, line 485: It also would be nice to drop a link to the specification here

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
+    val sign = (bits.toLong() shr 15 and 0x0001)
+    val halfExp = bits.toLong() shr 10 and 0x1f //max half exponent value
+    val halfMant = bits.toLong() and 0x3ff //max half mantis value
+
+    var exp: Long
+    var mant: Long
+
+    if (halfExp == 0x1fL) {//max half exponent value
+        //if mantissa maximal - value is NaN or Infinity
```
qwwdfsad, line 494: code style: typically we leave an indent after `//`

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
+
+private fun halfToDouble(bits: Short): Long {
+    val sign = (bits.toLong() shr 15 and 0x0001)
+    val halfExp = bits.toLong() shr 10 and 0x1f //max half exponent value
+    val halfMant = bits.toLong() and 0x3ff //max half mantis value
+
+    var exp: Long
+    var mant: Long
+
+    if (halfExp == 0x1fL) {//max half exponent value
```
qwwdfsad, line 493: for more than one `if-else` `when` is preferable

formats/cbor/jvmTest/src/kotlinx/serialization/cbor/CborCompatibilityTest.kt
shanshin: Library tries to encode lossless by smallest format.
So double value can be encoded as double (as is), float (single precision) or half precision. Float can be sent as is or as half precision.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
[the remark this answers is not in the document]
sandwwraith: You mean space? `when(x)` -> `when (x)`

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
+            }
+            mant = mant and 0x3ff //max half mantis value
+        }
+    } else {
+        //normalized value
+        exp = (halfExp + 112) //float exponent bias - half exponent bias = 127 - 15
+        mant = halfMant
+    }
+    return (sign shl 31) or (exp shl 23) or (mant shl 13)
+}
```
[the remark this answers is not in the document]
shanshin: Thanks

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
shanshin: Specification of what? IEEE-754?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -467,3 +481,74 @@ private fun Iterable<ByteArray>.flatten(): ByteArray {
...
+    }
+    return (sign shl 63) or (exp shl 52) or (mant shl 42)
+}
+
+private fun halfToFloat(bits: Short): Int {
+    val sign = (bits.toInt() shr 15 and 0x0001)
+    val halfExp = bits.toInt() shr 10 and 0x1f //max half exponent value
+    val halfMant = bits.toInt() and 0x3ff //max half mantis value
+
+    var exp: Int
```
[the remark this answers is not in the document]
shanshin: Drop function halfToFloat to avoid duplication. Now used halfToDouble().toFloat().

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
shanshin: Examples in RFC use function ldexp(man, pow) which returns  (man * 2^pow). There is no same function in Kotlin so alternative is to work with bits by yourself.

shanshin: "(https://android.googlesource.com/platform/libcore/+/master/luni/src/main/java/libcore/util/FP16.java#578" - it is a good trick with 0.5 float value!
FPU normalize value faster then shifts in a loop.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
sandwwraith: To corresponding appendix of CBOR rfc, I presume

formats/cbor/jvmTest/src/kotlinx/serialization/cbor/CborCompatibilityTest.kt
```diff
@@ -76,4 +79,42 @@ class CborCompatibilityTest {
...
+        compare(FloatData(123.0f), FloatData.serializer())
+        // minimal denormalized value in half-precision
+        compare(FloatData(5.9604645E-8f), FloatData.serializer())
+        // maximal denormalized value in half-precision
+        compare(FloatData(0.000060975552f), FloatData.serializer())
+    }
+
+    @Test
+    fun testDouble() {
+        compare(DoubleData(Double.NaN), DoubleData.serializer())
```
sandwwraith, line 104: nit: you can define local function `fun compareDouble(d: Dobule) = compare(DoubleData(d), DoubleData.serializer())` to duplicate less code

sandwwraith (approved):

PR #1117, opened 2020-10-05
Support nullable types at top-level for JsonElement decoding

### LINKED ISSUE #1102: Serializing nullable types to json thorws exception
**Describe the bug**
Attempting to serialize nullable variable throws `NoSuchElementException: List is empty`
**To Reproduce**
```kotiln
import kotlinx.serialization.json.encodeToJsonElement

@Serializable
class A {}

fun main() {
    val a: A? = A()
    println(Json.encodeToJsonElement(a))
}
```

**Expected behavior**
Output of `{}`, same as what happens if type of `a` is changed to `A`.
**Actual behavior**
```
Exception in thread "main" java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.last(_Collections.kt:380)
	at kotlinx.serialization.internal.TaggedEncoder.getCurrentTag(Tagged.kt:154)
	at kotlinx.serialization.internal.TaggedEncoder.encodeNotNullMark(Tagged.kt:60)
	at kotlinx.serialization.internal.NullableSerializer.serialize(NullableSerializer.kt:22)
	at kotlinx.serialization.json.internal.AbstractJsonTreeEncoder.encodeSerializableValue(TreeJsonEncoder.kt:202)
	at kotlinx.serialization.json.internal.TreeJsonEncoderKt.writeJson(TreeJsonEncoder.kt:20)
	at kotlinx.serialization.json.Json.encodeToJsonElement(Json.kt:95)
```
**Environment**
 - Kotlin version: 1.4.10
 - Library version: 1.0.0-RC2
 - Kotlin platforms: JVM
 - Gradle version: 5.5.1

qwwdfsad:
Fixes #1102

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/Tagged.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Tagged.kt
@@ -158,7 +158,7 @@ public abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
         tagStack.add(name)
     }
 
-    private fun popTag(): Tag =
+    protected fun popTag(): Tag =
         if (tagStack.isNotEmpty())
             tagStack.removeAt(tagStack.lastIndex)
         else
--- a/core/commonMain/src/kotlinx/serialization/internal/Tagged.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Tagged.kt
@@ -31,7 +31,6 @@ public abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
     protected open fun encodeTaggedValue(tag: Tag, value: Any): Unit =
         throw SerializationException("Non-serializable ${value::class} is not supported by ${this::class} encoder")
 
-    protected open fun encodeTaggedNotNullMark(tag: Tag) {}
     protected open fun encodeTaggedNull(tag: Tag): Unit = throw SerializationException("null is not supported")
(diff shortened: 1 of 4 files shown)


core/commonMain/src/kotlinx/serialization/internal/Tagged.kt
```diff
@@ -159,7 +158,7 @@ public abstract class TaggedEncoder<Tag : Any?> : Encoder, CompositeEncoder {
         tagStack.add(name)
     }
 
-    protected fun popTag(): Tag =
+    private fun popTag(): Tag =
```
sandwwraith, line 161: AFAIR, this method is `protected` because one of the formats authors asked us about it (XML or YAML, I don't remember actually). I know that we do not intend to support these interfaces for such a usage, however, we do not have viable alternative for now besides 'copy-pasting all TaggedEncoder class to your format'. So let's check if this is really true

sandwwraith: Seems there's BSON for now: https://github.com/agcom/bson/blob/bb58e4259518257de212388d83f07935417f67ca/serialization/src/main/kotlin/com/github/agcom/bson/serialization/decoders/BsonTreeInput.kt#L66

qwwdfsad: Sure, let's return that

sandwwraith (approved):

PR #1122, opened 2020-10-08
Add object-based serialization and deserialization of polimorphic types in JSON on JS platform

shanshin:
* implement JSON serialization (deserialization) of polymorphic type from (to) dynamic with discriminator
 * add tests on object-based and array-based polimorphism encoding and decoding on JS
 * fix array-based encodind polymorphism to dynamic type on JS

### DIFF
--- a/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
+++ b/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
@@ -58,6 +58,11 @@ private class DynamicObjectEncoder(
     private lateinit var currentDescriptor: SerialDescriptor
     private var currentElementIsMapKey = false
 
+    /**
+     * Flag of usage polymorphism with discriminator attribute
+     */
+    private var writePolymorphic = false
+
     private object NoOutputMark
 
     class Node(val writeMode: WriteMode, val jsObject: dynamic) {
@@ -73,11 +78,12 @@ private class DynamicObjectEncoder(
         current.index = index
         currentDescriptor = descriptor
 
-        if (current.writeMode == WriteMode.MAP) {
-            currentElementIsMapKey = current.index % 2 == 0
-        } else {
-            currentName = descriptor.getElementName(index)
+        when {
+            current.writeMode == WriteMode.MAP -> currentElementIsMapKey = current.index % 2 == 0
+            current.writeMode == WriteMode.LIST && descriptor.kind is PolymorphicKind -> currentName = index.toString()
+            else -> currentName = descriptor.getElementName(index)
         }
+
         return true
     }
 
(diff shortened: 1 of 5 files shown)


formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
```diff
@@ -58,6 +58,8 @@ private class DynamicObjectEncoder(
     private lateinit var currentDescriptor: SerialDescriptor
     private var currentElementIsMapKey = false
 
+    private var objectPolymorphism = false
```
sandwwraith, line 61: I'd suggest to rename this flag to `writePolymorphic` as it is named in other implementation because it is doing basically the same thing.

formats/json/jsTest/src/kotlinx/serialization/json/DynamicPolymorphismTest.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    @Serializable
+    data class AnyWrapper(@Polymorphic val any: Any)
+
+    @Serializable
+    @SerialName("string_wrapper")
+    data class StringWrapper(val text: String)
+
+    private val arrayJson = Json {
+        this.useArrayPolymorphism = true
```
sandwwraith, line 37: `this` seems to be redundant in this DSL block

formats/json/jsTest/src/kotlinx/serialization/json/DynamicPolymorphismTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private val objectJson = Json {
+        this.useArrayPolymorphism = false
```
sandwwraith, line 41: and here

sandwwraith (approved):

formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
shanshin: OK, but I think name writePolymorphic is incorrect - it is only true when the discriminator attribute is used

ankushg:
I know this has already been merged, but does serializing to/from `dynamic` actually have to be part of the `json` format? Or should it perhaps be separate?

Asking mainly because I have a project that _only_ needs `dynamic` handling, and it might be overkill to pull in all the machinery behind "real" JSON support

sandwwraith:
@ankushg Well, since JSON is a JavaScript Object Notation, javascript's dynamic objects can be mapped on it really well. It allows us to provide first-class support for features like object-based polymorphism and interopability with JsonElement, so it's an important part.

ankushg:
@sandwwraith definitely! My question was really if it was possible to put this in `core` instead of `json`, since `dynamic` is really just part of the JS language itself (as opposed to  parsing JSON-formatted `String`s or `JSONObject`s, which are more external)

PR #1132, opened 2020-10-10
Remove kotlin-stdlib from README.md

martinbonnin:
Remove kotlin-stdlib from README.md as the kotlin plugin adds it to the dependencies automatically since Kotlin `1.4`


qwwdfsad:
Thanks!

PR #898, opened 2020-07-10
Add CBOR major type 2 (byte string) support

twyatt:
Adds support for CBOR major type 2 (byte string).

Per [RFC 7049 2.1. Major Types](https://tools.ietf.org/html/rfc7049#section-2.1), when describing **Major type 3: a text string**:

> The format of this type is identical to that of byte strings (major type 2), that is, as with major type 2, the length gives the number of bytes.

As such, this byte string implementation piggybacks on the existing text string support, so both share decoding logic. Additionally, this PR builds on #846 which added support for decoding indefinite length byte strings. Since **text string** and **byte string** share decoding logic, both gain the indefinite length decoding support.

Closes #842.

PR #1139, opened 2020-10-13
Fix ByteString annotation detection when ByteArray is nullable

twyatt:
Fixes bug where `@ByteString` annotation is not detected when `ByteArray` property is marked **nullable**.
_(was an oversight in #898)_

Added tests to validate fix.
As expected, tests fail without fix applied.


sandwwraith (approved):
Great, thank you!

PR #1153, opened 2020-10-19
Optimize check for missing fields in deserialization

shanshin:


core/commonMain/src/kotlinx/serialization/internal/PluginUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+@InternalSerializationApi
+public fun throwMissingFieldException(seen: Int, goldenMask: Int, descriptor: SerialDescriptor) {
+    val missingFields = mutableListOf<String>()
+    var seenBits = seen
+    var goldenMaskBits = goldenMask
+    for (i in 0 until 32) {
+        if ((goldenMaskBits and 1 != 0) && (seenBits and 1 == 0)) {
+            missingFields += descriptor.getElementName(i)
+        }
+        seenBits = seenBits shr 1
```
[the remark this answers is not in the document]
sandwwraith: Or, you can test numbers with `1 shl i` instead

core/commonMain/src/kotlinx/serialization/internal/PluginUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+
+@OptIn(ExperimentalSerializationApi::class)
+@InternalSerializationApi
+public fun throwMissingFieldException(seen: Int, goldenMask: Int, descriptor: SerialDescriptor) {
+    val missingFields = mutableListOf<String>()
+    var seenBits = seen
+    var goldenMaskBits = goldenMask
+    for (i in 0 until 32) {
+        if ((goldenMaskBits and 1 != 0) && (seenBits and 1 == 0)) {
```
sandwwraith, line 15: It's also possible to get all reauired-and-missing bits set via `!(!goldenMaskBits or seenBits)` to replace two conditions in `if` with one. Not sure if this microoptimization has real benefit though.

sandwwraith (changes requested):
Do you have any tests for that? I know they would fail on buildserver and pass only locally, but  that's ok for now (we can depend on -dev- version of compiler later)

core/commonMain/src/kotlinx/serialization/internal/PluginUtils.kt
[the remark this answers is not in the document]
shanshin: Because of "and 1" operation  "shr 1" works fine here

core/commonMain/src/kotlinx/serialization/internal/PluginUtils.kt
shanshin: ok, thanks

shanshin:
> Do you have any tests for that? I know they would fail on buildserver and pass only locally, but that's ok for now (we can depend on -dev- version of compiler later)

Now optimization turned off in compiler plugin to prevent fails. Anyway, the plugin should be improved

core/commonMain/src/kotlinx/serialization/internal/PluginUtils.kt
shanshin: So it can be simplified to (goldenMaskBits and !seenBits)

core/commonMain/src/kotlinx/serialization/SerializationException.kt
```diff
@@ -62,8 +62,11 @@ public open class SerializationException : IllegalArgumentException {
  * Thrown when [KSerializer] did not receive property from [Decoder], and this property was not optional.
  */
 @PublishedApi
-internal class MissingFieldException(fieldName: String) : // TODO: add (message, cause) ctor after 1.4.20 for coroutines stacktrace recovery
-    SerializationException("Field '$fieldName' is required, but it was missing")
+internal class MissingFieldException : SerializationException {
+    // TODO: add (message, cause) ctor after 1.4.20 for coroutines stacktrace recovery
+    constructor(fieldName: String) : super("Field '$fieldName' is required, but it was missing")
```
sandwwraith, line 67: The error in build is caused by the fact that old plugins can't find primary constructor of this class (because now it has two secondary constructors). In 1.4.20 this problem is fixed, so you may try to bump kotlin version to 1.4.20-M2

core/commonMain/src/kotlinx/serialization/SerializationException.kt
```diff
@@ -62,8 +62,11 @@ public open class SerializationException : IllegalArgumentException {
...
+    constructor(fieldNames: List<String>, serialName: String) : super(if (fieldNames.size == 1) "Field '${fieldNames[0]}' is required, but it was missing" else "Fields $fieldNames are required for type with serial name '$serialName', but they were missing")
```
sandwwraith, line 68: The string for `fieldNames.size == 1` does not contain serial name

formats/json/jvmTest/src/kotlinx/serialization/JvmMissingFieldsExceptionTest.kt
```diff
@@ -0,0 +1,144 @@
...
+                subclass(ChildA::class)
+            }
+        }
+
+        assertFailsWithMessages(listOf("p2", "c3")) {
+            Json {
+                serializersModule = module
+                useArrayPolymorphism = false
+            }
+                    .decodeFromString<PolymorphicWrapper>("""{"nested": {"type": "a", "p1": 1, "c1": 11}}""")
```
sandwwraith, line 109: formatting

sandwwraith (approved):
LGTM (modulo comments)
Seems that we have to merge this somewhere around after 1.4.20 (so we can raise min version to 30)

integration-test/gradle.properties
```diff
@@ -2,7 +2,7 @@
 # Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-mainKotlinVersion=1.4.10
+mainKotlinVersion=1.4.30-M1
```
sandwwraith, line 5: Don't forget to change `mainLibVersion` to `1.1.0-SNAPSHOT` so your optimizations will be enabled

Issue #1152, opened 2020-10-19
Properties should support String encoding/decoding


sandwwraith:
**What is your use-case and why do you need this feature?**

Currently, Properties allow to parse `Map<String, Any>` object to a Kotlin class. Usually, such properties are loaded from configuration file. While we do not have multiplatform IO or files, we still can parse properties from a string contents of such file, since parsing line breaks/separators/etc is an additional amount of work.

Originally reported here: https://youtrack.jetbrains.com/issue/KT-42716

Note: it is relatively easy task though with common collection operations.

daniel-jasinski:
I think it would be a good intermediate step to provide API for parsing `Map<String,String>` before moving to full File/IO support.
At the moment I have to handle this case "manually".

RaphaelTarita:
Hi, here's the original creator of the YouTrack issue. Thank you very much for taking care of this. While work was actually done here, I was also working on a personal implementation of this functionality. However, I was not aware of the fact that there already exist ways of resolving the required primitive types of the model classes in the core lib. You may imagine that I had quite a struggle trying to implement the type resolving on my own. Anyway, I'm now somewhat relieved to see that I don't have to use my own version for this.

Apparently, the current issue is to get from the `Map<String, String>` to an actual properties format string (and vice versa). If it is any help, [here's what I've tried](https://gist.github.com/RaphaelTarita/748e02c06574b20c25ab96c87235096d); I've written this quite a while ago. But as far as I can tell, even this part has already been taken care of, which is wonderful.

Can we expect this feature to appear in a new version in the near future?

PS, a few notes on the gist:
- I tried to stick to the design of `kotlinx.serialization.json.Json.kt`
- I also tried to conform to the properties format spec from the [Java Properties class](https://docs.oracle.com/javase/10/docs/api/java/util/Properties.html#load(java.io.Reader))
- Due to the format description, I added support for the different accepted "flavors" of `.properties`, most notably the different key-value separators. There is also an option specifying the comment char, which is currently not in use.
- The gist has an earlier revision which just uses `java.util.Properties` to do the encoding/decoding.

hfhbd:
I would love to see this support:
1. to create a properties files from multiplatform targets, mostly (node) JS
2. the Java implementation always stores the current date as comment which is useless and annoying in tests

PR #1158, opened 2020-10-21
Add support for encoding/deconding Properties as String

daniel-jasinski:
Implementation for issue [1152](https://github.com/Kotlin/kotlinx.serialization/issues/1152)

### DIFF
--- a/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
+++ b/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
@@ -11,7 +11,6 @@ import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
-import kotlinx.serialization.properties.Properties.Default.toString
 
 /**
  * Transforms a [Serializable] class' properties into a single flat [Map] consisting of
--- a/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
+++ b/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
@@ -45,13 +45,15 @@ public sealed class Properties(
     ctorMarker: Nothing?
 ) : SerialFormat {
 
-    private inner class OutMapper : NamedValueEncoder() {
+    private abstract inner class OutMapper<Value : Any> : NamedValueEncoder() {
         override val serializersModule: SerializersModule = this@Properties.serializersModule
 
-        internal val map: MutableMap<String, Any> = mutableMapOf()
+        val map: MutableMap<String, Value> = mutableMapOf()
+
+        protected abstract fun encode(value: Any): Value
 
         override fun encodeTaggedValue(tag: String, value: Any) {
-            map[tag] = value
+            map[tag] = encode(value)
         }
 
         override fun encodeTaggedNull(tag: String) {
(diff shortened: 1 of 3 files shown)


daniel-jasinski:
@sandwwraith please take a look at this proposal for implementation.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -101,13 +113,48 @@ public sealed class Properties(
...
     public fun <T> encodeToMap(serializer: SerializationStrategy<T>, value: T): Map<String, Any> {
-        val m = OutMapper()
+        val m = OutAnyMapper()
+        m.encodeSerializableValue(serializer, value)
+        return m.map
+    }
+
+    /**
+     * Encodes properties from the given [value] to a map using the given [serializer].
+     * Converts all values to [String].
```
sandwwraith, line 152: I'd add that this conversion is performed with `toString` method

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -118,7 +165,18 @@ public sealed class Properties(
...
     @ExperimentalSerializationApi
     public fun <T> decodeFromMap(deserializer: DeserializationStrategy<T>, map: Map<String, Any>): T {
-        val m = InMapper(map, deserializer.descriptor)
+        val m = InAnyMapper(map, deserializer.descriptor)
+        return m.decodeSerializableValue(deserializer)
+    }
+
+    /**
+     * Decodes properties from the given [map] to a value of type [T] using the given [deserializer].
+     * All values are expected to be of type [String].
```
sandwwraith, line 174: `All values are expected to be of type [String]` — this phrase looks redundant, because this fact is already expressed by type `Map<String, String>`. 

I'd replace it with the phrase indicating that strings are then converted to target kotlin type using default conversions.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -146,6 +204,14 @@ public fun Properties(module: SerializersModule): Properties = PropertiesImpl(mo
 public inline fun <reified T> Properties.encodeToMap(value: T): Map<String, Any> =
     encodeToMap(serializersModule.serializer(), value)
 
+/**
+ * Encodes properties from given [value] to a map using serializer for reified type [T] and returns this map.
```
sandwwraith, line 208: This documentation does not mention strings or whatsoever, so it looks like it was based on the `encodeToMap` function, not the `encodeToStringMap`

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -154,6 +220,14 @@ public inline fun <reified T> Properties.encodeToMap(value: T): Map<String, Any>
 public inline fun <reified T> Properties.decodeFromMap(map: Map<String, Any>): T =
     decodeFromMap(serializersModule.serializer(), map)
 
+/**
+ * Decodes properties from given [map], assigns them to an object using serializer for reified type [T] and returns this object.
```
sandwwraith, line 224: Same here: no notion of string map

sandwwraith (approved):
Thanks for the great PR! I think we can merge this after all the comments are addressed

sandwwraith (changes requested):
.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
daniel-jasinski: Applied suggestion.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
daniel-jasinski: Applied suggestion.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
daniel-jasinski: Applied suggestion.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
daniel-jasinski: Applied suggestion.

daniel-jasinski:
Added changes according to the comments/suggestions. Please have a look.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -11,6 +11,7 @@ import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
+import kotlinx.serialization.properties.Properties.Default.toString
```
sandwwraith, line 14:
```suggestion

```

Looks like it got here accidentally by referencing toString method in documentation.

sandwwraith (approved):

sandwwraith:
Note that it does not close #1152 because it takes `Map<String, String>`, not a whole single `String`

PR #1159, opened 2020-10-22
Properly cache generated serializers in PluginGeneratedSerialDescript…

### LINKED ISSUE #1156: Poor performance compared to moshi (up to 37x) on Android device
KxSerialization works mach slower than Moshi. 

**Benchmarks**
I've measured performance using android instrumental tests on Xiaomi Mi box 3. Decoding string looks 1.9s using kxSerialization and 53ms using Moshi. It is huge difference in performance. I also tested it on Sony Bravia 4k 2015 and results are same.

Снимок экрана 2020-10-20 в 19 43 30

Снимок экрана 2020-10-21 в 12 25 47

**To Reproduce**
You can find benchmark test in following archive. I also add some middle size test strings and kxSerialization is slow for them too. 
[SerializationBenchmark.zip](https://github.com/Kotlin/kotlinx.serialization/files/5414536/SerializationBenchmark.zip)

Also after decoding objects Android log reports that GC actively trying to deallocate objects.

**Environment**
 - Kotlin version: 1.4.10
 - Library version: 1.0.0
 - Kotlin platforms: Android (JVM)
 - Gradle version: 6.5
 - Device: Xiaomi Mi box 3 and Sony Bravia 4k 2015

qwwdfsad:
…or in order to avoid allocation of nullable serializer and descriptor.

Also, slightly reduce bytecode size

Fixes #1156

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
@@ -1,3 +1,6 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
 package kotlinx.benchmarks.protobuf
 
 import kotlinx.serialization.Serializable
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
@@ -1,10 +1,6 @@
-/*
- * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
- */
+package kotlinx.benchmarks.protobuf
 
-package kotlinx.benchmarks
-
-import kotlinx.serialization.*
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.protobuf.*
 import org.openjdk.jmh.annotations.*
 import java.util.concurrent.*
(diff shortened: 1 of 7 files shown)


benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
```diff
@@ -1,10 +1,6 @@
-/*
- * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
- */
+package kotlinx.benchmarks.protobuf
 
-package kotlinx.benchmarks
-
-import kotlinx.serialization.*
+import kotlinx.serialization.Serializable
```
sandwwraith, line 3: Star import, copyright accidentally deleted (in all protobuf-related files)

core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -116,8 +118,8 @@ internal inline fun <reified SD : SerialDescriptor> SD.equalsImpl(
     return true
 }
 
-@Suppress("NOTHING_TO_INLINE")
-internal inline fun SerialDescriptor.hashCodeImpl(typeParams: Array<SerialDescriptor>): Int {
+@OptIn(ExperimentalSerializationApi::class)
+internal fun SerialDescriptor.hashCodeImpl(typeParams: Array<SerialDescriptor>): Int {
```
sandwwraith, line 122: Are you sure that we do not want to inline this?

sandwwraith (approved):

core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
qwwdfsad: Yes, it has no benefits, but bloats bytecode size for nothing

benchmark/src/jmh/kotlin/kotlinx/benchmarks/protobuf/ProtoBaseline.kt
qwwdfsad: Move refactoring -- the gift that keeps on giving

PR #1160, opened 2020-10-23
Add Pair and Triple to serializer resolving from Java type token

### LINKED ISSUE #1146: serializer(java.lang.reflect.Type) does not work with Pairs while serializer(KType) does.
**Describe the bug**
Reported in the slack. Looks like Triple also affected. The corresponding `when` branch exists in serializer(KType) but not in serializer(Type).

**To Reproduce**
```
// OK:
println(Json.decodeFromString(serializer<Pair<String, Int>>(), pairJson)) 

// Errror:
println(Json.decodeFromString(serializer(object : TypeToken<Pair<String, Int>>() {}.type), pairJson))
```

**Environment**
 - Kotlin version: 1.4.10
 - Library version: 1.0.0
 - Kotlin platforms: JVM

sandwwraith:
Add tests on Pair and Triple both for serializer(KType) and serializer(Type)

Fixes #1146

### DIFF
--- a/core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
+++ b/core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
@@ -7,8 +7,10 @@
 
 package kotlinx.serialization
 
-import kotlinx.serialization.builtins.MapEntrySerializer
 import kotlinx.serialization.builtins.*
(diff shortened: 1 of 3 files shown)


qwwdfsad (approved):
👍

PR #1164, opened 2020-10-26
Add serializerOrNull function for KType and Type arguments

### LINKED ISSUE #1163: Provide api for getting serialiser by KType when exists or null otherwise
Right now to check if an object has a serializer user need to write such code
```
    return try {
        module.serializer(responseType)
    } catch (_: SerializationException) {
        null
    }
``` 
This will require creating an exception that is unnecessary and expensive 

Provide a method to get serializer if it exists like 
```
public fun SerializersModule.serializerOrNull(type: KType): KSerializer<Any?>?
```

sandwwraith:
to avoid building framework logic on catching exceptions.

Fixes #1163

Didn't add orNull method for reified overloads because usually the type is specified there explicitly.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -8,9 +8,9 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.builtins.*
-import kotlinx.serialization.builtins.TripleSerializer
-import kotlinx.serialization.builtins.PairSerializer
 import kotlinx.serialization.builtins.MapEntrySerializer
+import kotlinx.serialization.builtins.PairSerializer
+import kotlinx.serialization.builtins.TripleSerializer
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 import kotlin.jvm.*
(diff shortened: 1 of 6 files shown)


core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -35,48 +35,64 @@ public inline fun <reified T> SerializersModule.serializer(): KSerializer<T> {
...
 @OptIn(ExperimentalSerializationApi::class)
-public fun SerializersModule.serializer(type: KType): KSerializer<Any?> {
-    val kclass = type.kclass()
-    val isNullable = type.isMarkedNullable
-    val builtin = serializerByKTypeImpl(type)
-    if (builtin != null) {
-        return builtin.nullable(isNullable).cast()
-    }
+public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
+    serializerByKTypeImpl(type, checkTypeArgs = true) ?: type.kclass().platformSpecificSerializerNotRegistered()
```
qwwdfsad, line 59: Here and everywhere: `checkTypeArgs` tells us how exactly it works internally, not its intent, so it is required to study the implementation in order to understand why is it here.
I suggest to rename it to better reflect the intent e.g. `nullableSerializer`/`nullWhenMissing`/`failOnMissingSerializer`

formats/json/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
```diff
@@ -207,6 +205,29 @@ class SerializersLookupTest : JsonTestBase() {
         assertEquals("42", json.encodeToString(42))
     }
 
+    class NonSerializable
+
+    class NonSerializableBox<T>(val boxed: T)
+
+    @Test
+    fun testLookupFail() {
```
qwwdfsad, line 213: It's also good to test that this function works for serializable classes :)

qwwdfsad (changes requested):

formats/json/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
sandwwraith: That's what the rest of test do, doesn't it?

qwwdfsad (approved):
Nice!

Issue #1143, opened 2020-10-14
JAR does not exist in BIntray or Manven central as described in the documentation


HunterSherms:
**Describe the bug**

Following [the documentation](https://github.com/Kotlin/kotlinx.serialization#maven) and including the dependency in maven like so:

```
<dependency>
    <groupId>org.jetbrains.kotlinx</groupId>
    <artifactId>kotlinx-serialization-json</artifactId>
    <version>1.0.0</version>
</dependency>
```

Maven is unable to find the dependency, with or without the bintray repository. After digging in a bit, visiting https://kotlin.bintray.com/kotlinx/org/jetbrains/kotlinx/kotlinx-serialization-json/1.0.0/ shows no JAR files available. There are, however, many sub modules that do appear to be available such as https://kotlin.bintray.com/kotlinx/org/jetbrains/kotlinx/kotlinx-serialization-json-jvm/1.0.0/

**To Reproduce**
Include `kotlinx-serialization-json` as a dependency in any project according to the readme documentation.

**Expected behavior**
MVN includes the dependency after following the documentation.

**Environment**
 - Kotlin version: latest
 - Library version: 1.0.0
 - Kotlin platforms: JVM
 - IDE version: IntelliJ latest

h0tk3y:
It should work with Maven, too, as we publish a POM in the `kotlinx-serialization-json` module which has a single dependency on the actual module that contains the JVM artifact:
```
<dependencies>
  <dependency>
    <groupId>org.jetbrains.kotlinx</groupId>
    <artifactId>kotlinx-serialization-json-jvm</artifactId>
    <version>1.0.0</version>
    <scope>compile</scope>
  </dependency>
</dependencies>
```
Let me try to reproduce and investigate this...

h0tk3y:
The default dependency declaration seems to be broken indeed when consumed from Maven (but not when consumed from Gradle versions that can't read Gradle module metadata).
As a temporary workaround, this version can be consumed with `<type>pom</type>` added to the dependency declaration.

HunterSherms:
I was unable to get it working using `<type>pom</type>` - what eventually worked for me was using the `                <artifactId>kotlinx-serialization-json-jvm</artifactId>` directly

sandwwraith:
@h0tk3y Is this a bug that can be fixed on our side or should we simply change maven instuctions?

qwwdfsad:
It will be fixed in 1.4.20, no documentation updates required

[...]

PR #1177, opened 2020-11-02
Add JVM IR target in build_snapshot_train mode

ALikhachev:


udalov:
As in https://github.com/ktorio/ktor/pull/2175#issuecomment-721058147, I suggest not to commit API dumps for JVM IR, but rather mute the tests if possible (with the option to unmute locally).

benchmark/build.gradle
```diff
@@ -8,32 +8,56 @@ apply plugin: "me.champeau.gradle.jmh"
...
-    compile 'com.google.code.gson:gson:2.8.1'
-    compile 'com.esotericsoftware:kryo:4.0.0'
+    implementation 'com.squareup.okio:okio:1.13.0'
+    implementation 'com.google.code.gson:gson:2.8.1'
+    implementation 'com.esotericsoftware:kryo:4.0.0'
 
-    compile project(':kotlinx-serialization-core')
-    compile project(':kotlinx-serialization-json')
-    compile project(':kotlinx-serialization-protobuf')
+    jvmImplementation project(':kotlinx-serialization-core')
```
sandwwraith, line 55: Why there is a need for separate configuration? I though benchmarks can consume the same class/jar files

build.gradle
```diff
@@ -163,11 +170,27 @@ subprojects {
 
     apply from: rootProject.file('gradle/teamcity.gradle')
     // Configure publishing for some artifacts
-    if (project.name != "benchmark" && project.name != "guide") {
+    if (project.name != "benchmark" && project.name != "guide" && !(project.name in multiplatformProjects)) {
```
sandwwraith, line 173: Why make this condition and move publishing setup to other gradle files?

formats/hocon/build.gradle
```diff
@@ -14,15 +14,17 @@
  *  limitations under the License.
  */
 
-apply plugin: 'kotlin'
+apply plugin: 'kotlin-multiplatform'
```
sandwwraith, line 17: Will it affect publishing? I'd rather try to avoid making this subproject multiplatform

sandwwraith: Also, it would be nice to set up source dirs so actual files should not be moved

sandwwraith (commented):
Can you briefly describe how changes will affect build process? As far as I can see, there will be new `compileJvmIr` and `jvmIrTest` tasks? Why the new dependency configuration is needed? Will jvm ir artifacts be published and how?

build.gradle
ALikhachev: There is some problem at the moment: user-defined variant attributes (JVM compiler type in our case) are not taken into account when publishing is set up before applying kotlin-multiplatform plugin

benchmark/build.gradle
ALikhachev: Adding of JVM IR target causes ambiguity in dependency resolution, so we need to mark dependency either with JVM or JVM IR

ALikhachev:
The main purpose of those changes is to ensure that JVM IR compiler is correct. We compile JVM sources using JVM IR compiler and run test classes against it's output the same as with the old compiler. JVM IR jar doesn't pretend to be published.

formats/hocon/build.gradle
ALikhachev: It shouldn't negatively affect publishing

benchmark/build.gradle
```diff
@@ -8,32 +8,56 @@ apply plugin: "me.champeau.gradle.jmh"
...
+    jvmImplementation project(':kotlinx-serialization-json')
+    jvmImplementation project(':kotlinx-serialization-protobuf')
+    jvmImplementation project(':kotlinx-serialization-json')
```
hfhbd, line 58: Why did you implement `-json` twice?

ALikhachev: This dependency was duplicated after rebase, and I didn't notice it. Thank you!

benchmark/build.gradle
sandwwraith: Can you please add the comment explaining this in details to configuration delcaration?

formats/hocon/build.gradle
sandwwraith: Well, there's now `kotlinx-serialization-hocon` and `kotlinx-serialization-hocon-jvm` artifacts. The first one depends on the second, so users shouldn't face issues unless they're using maven: #1143

sandwwraith (approved):

sandwwraith:
Note: due to hocon setup change, this PR also seems to fix #1189

benchmark/build.gradle
ALikhachev: We create configuration named 'jvmImplementation' [here](https://github.com/Kotlin/kotlinx.serialization/blob/7d39fd395249e314f842b3e17caa1ec6b8167a68/benchmark/build.gradle#L13), set JVM compiler attribute with value **jvm** and then declare dependencies to that configuration. Then we configure **jmhTask** to use this configuration. Do you mean add the comment to the code?

ALikhachev:
During private discussion with @qwwdfsad we've decided to try another way. I'll open new PR.

Issue #1171, opened 2020-10-29
Properties-format deserialization is bugged when names share initial character


TorRanfelt:
**Describe the bug**
Properties-format deserialization is bugged when names share initial character.

**To Reproduce**
```
@Serializable
class Foo {
@SerialName("0")
val bar: Bar
}

@Serializable
class Bar {
@SerialName("1")
val baz1: Int = 100

@SerialName("11")
val baz2: Int = 101
}
```

Properties file:
0.11=82

With this we are informed that 0.1 lacks a value.
If @SerialName("1") is change to @Serializable("A1") it works.

- I'm also informed that we have experienced that the default value magically has been changed with these kinds of SerialNames

**Environment**
 - Kotlin version: 1.4.10
 - Library version: 1.0.1
 - Kotlin platforms: JS
 - Gradle version: 6.0.1

TorRanfelt:
I have made a unit-test and a fix and will make a pull-request tomorrow.

PR #1183, opened 2020-11-05
Alllow shared prefix names

TorRanfelt:
Allow names with a prefix that equals another name in the same class.

Fixing https://github.com/Kotlin/kotlinx.serialization/issues/1171


sandwwraith (approved):
Thanks!

Issue #507, opened 2019-07-04
Global custom serializers


sksamuel:
**What is your use-case and why do you need this feature?**

If you have a data class with many "unsupported" types, but very common types in the JDK it is unweidly to use @ContextualSerialization everywhere - either on a file basic or a property basis.

For example,

`data class Foo(val decimal: BigDecimal, val id: UUID, val date: LocalDateTime)`

Three types, needing three @ContextualSerialization.

**Describe the solution you'd like**

Some way of registering global serializers.

sandwwraith:
We have several concerns regarding this:

1. Implicitness – it is very easy to forget and hard to find which serializer is used for `UUID` in a particular class. Easy to mess/mix things up.
2. Security – you may accidentally deserialize data that you didn't want to in some area of the program
3. Modules, incremental compilation – since serializers are compiled into the code by the plugin, changing global settings would require to recompile the whole module.

However, these problems may not concern small or mid-size applications. Probably this restriction would be relaxed in the future.

sksamuel:
Valid concerns, but here are a couple of points from the other side.

1. Common types like BigDecimal, UUID, javax dates, etc are not supported at the moment. It's tedious and error prone to annotate every instance of these. These are extremely common base JDK types - in my code base for example a domain type could have 10 of these out of 30 fields, across hundreds of domain objects. Financial domain objects will use BigDecimals everywhere.

2. The most popular serialisation frameworks - think Jackson, Gson for example - handle serialization of any types. Jackson for example will serialization everything. If this was causing security issues that couldn't be addressed then it would have materialized itself in these libraries.

JakeWharton:
The Gson maintainers think point 2 is a failure of the library and have not supported this feature in Moshi (aka Gson v3). It has _absolutely_ caused problems in practice.

sksamuel:
Ok I stand corrected on Gson. It's still a feature / bug (delete as appropriate) of Jackson though.

pdvrieze:
I'm not sure whether it is actually valid (or easy to do), but package level annotations would be the way to go. I know it is possible in Java using package-info.java, but I've got no idea whether Kotlin could use such annotations in the compiler.

sandwwraith:
AFAIR, Kotlin doesn't support package-level annotation. It recognizes annotations on `package-info.java`, but this requires enabling Java support in JVM target (hence, no MPP).

pdvrieze:
[...]

PR #1184, opened 2020-11-05
Jvm global serializers (UUID)

red-avtovo:
Just follow up on #507 
I'm not sure what's you thinking right now about this approach, but I decided to show the way how to make it under control, but let contributors extend platform standard types serializers
Looking forward to see your opinion

### DIFF
--- a/core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
+++ b/core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
@@ -11,9 +11,13 @@ import kotlinx.serialization.builtins.*
 import kotlinx.serialization.builtins.MapEntrySerializer
 import kotlinx.serialization.builtins.PairSerializer
 import kotlinx.serialization.builtins.TripleSerializer
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 import java.lang.reflect.*
+import java.util.*
 import kotlin.reflect.*
 
 /**
@@ -112,4 +116,11 @@ private fun SerializersModule.genericArraySerializer(type: GenericArrayType): KS
 @OptIn(ExperimentalSerializationApi::class)
 private fun <T: Any> SerializersModule.reflectiveOrContextual(kClass: KClass<T>): KSerializer<T> {
     return kClass.serializerOrNull() ?: getContextual(kClass) ?: kClass.serializerNotRegistered()
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+internal object UUIDSerializer: KSerializer<UUID> {
+    override val descriptor = PrimitiveSerialDescriptor("java.util.UUID", PrimitiveKind.STRING)
+    override fun serialize(encoder: Encoder, value: UUID) = encoder.encodeString(value.toString())
+    override fun deserialize(decoder: Decoder): UUID = UUID.fromString(decoder.decodeString())
 }
--- a/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -51,11 +51,19 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
         null
     }
     if (fromNamedCompanion != null) return fromNamedCompanion
+    //Check globally registered serializers
+    jvmGlobalSerializer(jClass)?.let { return it }
     // Check for polymorphic
     return polymorphicSerializer()
 }
 
-private fun <T: Any> Class<T>.isNotAnnotated(): Boolean {
+private fun <T> jvmGlobalSerializer(jClass: Class<T>): KSerializer<T>? =
+        when (jClass) {
+            UUID::class.java -> UUIDSerializer as? KSerializer<T>
+            else -> null
+        }
+
+private fun <T : Any> Class<T>.isNotAnnotated(): Boolean {
     /*
      * For annotated enums search serializer directly (or do not search at all?)
      */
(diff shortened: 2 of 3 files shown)


core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -51,11 +51,19 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
...
     // Check for polymorphic
     return polymorphicSerializer()
 }
 
-private fun <T: Any> Class<T>.isNotAnnotated(): Boolean {
+private fun <T> jvmGlobalSerializer(jClass: Class<T>): KSerializer<T>? =
+        when (jClass) {
+            UUID::class.java -> UUIDSerializer as? KSerializer<T>
+            else -> null
+        }
```
pdvrieze, line 64: This is not a scalable solution, more significantly it doesn't work in a way that keeps parity between standard and custom serializers. In addition, this code path is only for reflectively resolving serializers. In that case the serialmodule is available. The method is mainly for the code path with type arguments, which UUID doesn't have. As such it will result in surprising and inconsistent behaviour: `serializer(typeOf<UUID>())` will not return the serializer because it has no type arguments. The place that you would add the additional built-in would be in `Primitives.kt` line 69, `builtinSerializerOrNull`. But frankly the approach I would take would be to create a different method that uses serializerOrNull and then adds whatever you want to add as alternative. Maybe even better is to just use a `ContextSerializer` and then you create a module which adds all these defaults.

Of course kotlinx.serialization is primarily focused on statically (at compile time) resolving serializers and for that you need to announce the serializers to the compiler with the right annotations. It may be worthwhile to allow libraries to declare serializers in a way that they are automatically picked up by the compiler but that is a whole different thing.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -51,11 +51,19 @@ internal actual fun <T : Any> KClass<T>.constructSerializerForGivenTypeArgs(vara
...
+
+private fun <T : Any> Class<T>.isNotAnnotated(): Boolean {
```
pdvrieze, line 66: This has a spurious space after the `T`, invalid change

core/jvmTest/src/kotlinx/serialization/SerializeFlatTest.kt
```diff
@@ -297,4 +298,34 @@ class SerializeFlatTest() {
...
+    ): AbstractDecoder() {
+        override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
+            assertEquals(1, descriptor.elementsCount)
+            return 1
+        }
+        override fun decodeString(): String = expected.toString()
+    }
+
+    @Test
+    fun testUUID() {
```
pdvrieze, line 323: This method uses a reflective approach to resolve the serializer. Also add one that uses `UUID.serializer`, You will find that will not work currently.

pdvrieze (commented):
I think building a library of common custom serializers (especially for more frequently used JVM types) is worthwhile, however I don't think that it should be in the core module, rather in a support module that users can opt in to.

The way that the UUID serializer is hooked up is incorrect (in the wrong place, and not supporting compile time serializer selection). In addition it would very quickly break scalability and library parity.

core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
```diff
@@ -113,3 +117,10 @@ private fun SerializersModule.genericArraySerializer(type: GenericArrayType): KS
 private fun <T: Any> SerializersModule.reflectiveOrContextual(kClass: KClass<T>): KSerializer<T> {
     return kClass.serializerOrNull() ?: getContextual(kClass) ?: kClass.serializerNotRegistered()
 }
+
+@OptIn(ExperimentalSerializationApi::class)
+internal object UUIDSerializer: KSerializer<UUID> {
+    override val descriptor = PrimitiveSerialDescriptor("java.util.UUID", PrimitiveKind.STRING)
+    override fun serialize(encoder: Encoder, value: UUID) = encoder.encodeString(value.toString())
+    override fun deserialize(decoder: Decoder): UUID = UUID.fromString(decoder.decodeString())
+}
```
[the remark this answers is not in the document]
JakeWharton: I cannot thumbs-up this comment enough. Unless a type has an intrinsic representation in the underlying serialization format (like, for example, a string or a number or a list) providing a default means encoding an opinion on how a type is serialized that will not be universal.

As an existing example, the library has built-in support for pair and triple that I wish were not included. For pair, one may want to encode an array that is always two elements.

So I too strongly believe these should not be provided in the core. Especially given how trivial it is to create them yourself where the responsibility of making the decision about the serialized form.

(If you need a really, really terrible example of this, look at Gson's built-in support for serializing dates and note Moshi's intentional complete absence of that as a built-in feature and reliance on user-supplied adapters)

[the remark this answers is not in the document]
sandwwraith: Agree. At least, this one should be named as `UUIDAsStringSerializer`, but we can also imagine `Base64UUIDSerializer` or `UUIDAsByteArraySerializer`. Ultimately, users should choose what serializer to pick on their own.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
sandwwraith: Actually, it will work with `serializer<UUID>()` (as it shown in tests), because this method resolves all serializers. 

However, the bigger problem that this method used only for top-level reflective resolving (serializer<> function) and it does not affect static compile-time resolving. That is, any class that contains UUID (e.g. `@Serializable data class MyData(val uuid: UUID)`) will report compile-time exception that no serializer is found, require using `@UseSerializers/SerializersModule`, so original issue is kinda not solved.

sandwwraith (commented):
Thanks for your contribution and time! 
However, we do not have plans to include any pre-defined serializers for Java classes directly into the core module. There are several reasons to it: 

1. Core module should be kept lightweight — not all users need serializers for Java classes, especially users that use library in multiplatform context.
2. Opinionated serializers — as @jw and @pdvrieze correctly noticed, there are a lot of different approaches to serialize UUIDs, Dates, and so on.

We have a plan to provide a separate jvm-only module for serializers for java classes. However, there are still big design questions around that: should we provide several serializing strategies for one class out of the box for most popular use-cases? How these serializers should be connected to compile-time resolving, is `@UseSerializers` annotation enough? I think answers for these questions will be provided in future releases of kotlinx.serialization. For now, the most efficient way to serialize UUIDs/Dates is to copy-paste and tweak serializer to your particular needs.

Issue #1091, opened 2020-09-21
Provide @AlwaysEncode / @EncodeDefault annotation


sandwwraith:
**What is your use-case and why do you need this feature?**

kotlinx.serialization framework allows to drop properties from output when its values are equal to default ones (see `encodeDefault` flag). However, this is configurable per format and not per property. Sometimes, it may be convenient to have a default value in Kotlin class, but it also may be necessary to always serialize this value to JSON — for example, if this is a protocol/api version for some app.

See also: https://github.com/Kotlin/kotlinx.serialization/issues/58#issuecomment-608583101 

**Describe the solution you'd like**

Annotation with property target that allows to force encoding even if property value is equal to the default one.

1354092549:
For null-not-supported formats like ProtoBuf, it is needed to encode nullable value with `EncodeDefault = false`.
However, we may need to encode other default values like `foo : Int = 0`.
In this way, config per field is useful.

1354092549:
#### Describe the solution you'd like
Add a enum and a annotation
```kotlin
public enum EncodeDefaultMode{
    DEFAULT,
    ALWAYS,
    NEVER
}

@Target(AnnotationTarget.PROPERTY)
public annotation class EncodeDefault(public val mode: EncodeDefaultMode = EncodeDefaultMode.DEFAULT)
```
Then: 
- Change the compiler plugin, when `EncodeDefaultMode` is set to `ALWAYS` or `NEVER`, do not call `shouldEncodeElementDefault` but using a constant (**high-performance**, in my favor)
- **OR** change `shouldEncodeElementDefault` to check `EncodeDefaultMode` at runtime (low-performance, but more maintainable)

sandwwraith:
Can you, please, explain, why there is a need in this enum values (especially NEVER)? Original issue was just about `@AlwaysEncode` annotation without parameters. NEVER, judging by your PRs, simply checks equality with the default value and does not respect format settings, which may confuse users.

Also, default value of annotation parameter (DEFAULT) seems dull: when `@EncodeDefault` annotation is used with it, it does not change anything at all. Maybe DEFAULT shouldn't even be an option.

1354092549:
> NEVER does not respect format settings

If you think so, `ALWAYS` does not "respect" format settings too, since we set `encodeDefaults=false` but the field will also be encoded.

> Why NEVER is needed.

To provide a more precise control of optionality (per field), of course.
[...]

PR #1190, opened 2020-11-08
Provide @EncodeDefault annotation for fields

1354092549:
More details are described in Kotlin/kotlinx.serialization#1091.
Tests are included but CI will fail since it needs the new version of the compiler plugin.


sandwwraith:
Superseded by https://github.com/Kotlin/kotlinx.serialization/pull/1528

PR #1195, opened 2020-11-11
Improved test of supporting of nullable serializer in UserSerializers annotation

### LINKED ISSUE #984: @UseSerializers doesn't distinguish between nullable and non-nullable property when choosing KSerializer
**Describe the bug**
If you have a nullable serializer for a type and a non-nullable serializer for a type and you want to register both of them with `@UseSerializers`, the nullable one is not chosen for nullable properties.

**To Reproduce**
https://github.com/Kotlin/kotlinx.serialization/compare/master...edenman:nullable-and-nonnullable-serializer

**Expected behavior**
Choose the correct serializer, either by checking the SerialDescriptor or introspecting the type argument on the KSerializer class.

**Environment**
 - Kotlin version: 1.4.0
 - Library version: 1.0.0-RC
 - Kotlin platforms: common (tested on jvm)
 - Gradle version: 6.3

shanshin:
Fixes #984

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -38,7 +38,7 @@ tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
     manifest {
         attributes(
                 "Implementation-Version": version,
-                "Require-Kotlin-Version": "1.4-RC",
+                "Require-Kotlin-Version": "1.4.30-M1",
         )
     }
 }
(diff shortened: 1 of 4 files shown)


sandwwraith (approved):
I assume tests would start to pass when compiler fixes will be available?

core/build.gradle
```diff
@@ -38,7 +38,7 @@ tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
     manifest {
         attributes(
                 "Implementation-Version": version,
-                "Require-Kotlin-Version": "1.4-RC",
+                "Require-Kotlin-Version": "1.4.30-M1",
```
sandwwraith, line 41: NB: bug fixes do not require raising minimal kotlin version, but for this case, we wanted to do it anyway

PR #1197, opened 2020-11-12
Fix serializing property with custom accessors and backing field

### LINKED ISSUE #956: VerifyError when serializing property with backing field
**Describe the bug**
Invalid bytecode produced to serialize a class containing a field backed property with getter and setter specified.

Notably, all works as expected with one or the other present, just not both.

```
Exception in thread "main" java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    Test.<init>(ILkotlinx/serialization/internal/SerializationConstructorMarker;)V @12: putfield
  Reason:
    Type 'kotlinx/serialization/internal/SerializationConstructorMarker' (current frame, stack[1]) is not assignable to 'java/lang/String'
  Current Frame:
    bci: @12
    flags: { }
    locals: { 'Test', integer, 'kotlinx/serialization/internal/SerializationConstructorMarker' }
    stack: { 'Test', 'kotlinx/serialization/internal/SerializationConstructorMarker' }
  Bytecode:
    0x0000000: 2ab7 0016 1b04 7e99 000b 2a2c b500 0fa7
    0x0000010: 0009 2a12 18b5 000f b1                 
  Stackmap Table:
    full_frame(@18,{Object[#2],Integer,Object[#45]},{})
    same_frame(@24)

	at TestKt.main(Test.kt:18)
	at TestKt.main(Test.kt)
```

**To Reproduce**
```kotlin

@Serializable
class Test {
    private var test: String = ""
        set(value) {
            field = value
        }
        get() {
            return field
        }
}

fun main() {
    println(Json { }.encodeToString(Test.serializer(), Test()))
}
```

**Expected behavior**
Backing field serialized without error

**Environment**
 - Kotlin version: 1.4.0-rc
 - Library version: 1.0-M1-1.4.0-rc
 - Kotlin platforms: JVM
 - Gradle version: 6.5.1

shanshin:
Fixes #956

Added tests of serialization compiler plugin for properties with custom accessors

### DIFF
--- a/core/commonTest/src/kotlinx/serialization/CustomPropertyAccessorsTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/CustomPropertyAccessorsTest.kt
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.AbstractDecoder
+import kotlinx.serialization.encoding.CompositeDecoder
+import kotlinx.serialization.modules.EmptySerializersModule
+import kotlinx.serialization.modules.SerializersModule
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFails
+
+@Suppress("RedundantSetter", "RedundantGetter", "JoinDeclarationAndAssignment")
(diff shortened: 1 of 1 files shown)


sandwwraith:
It looks like tests are failing on IR plugin. Is it because the fixes aren't available in used compiler yet?

sandwwraith (approved):

shanshin:
> It looks like tests are failing on IR plugin. Is it because the fixes aren't available in used compiler yet?

Build fails because changes of compiler plugin not merged.

PR #1201, opened 2020-11-13
Do not try to end structure in encode/decode stucture extensions if a…

### LINKED ISSUE #1187: decodeStructure swallows exceptions
Exceptions thrown within the decodeStructure block are swallowed.

This code:

```    
decoder.decodeStructure(descriptor) {
        TODO()
    }
```

throws an JsonDecodingException:
Unexpected JSON token at offset 1: Expected '}'

I assume the original exception is hidden, because the decoder tries to finish decoding the JSON object.

I strongly prefer to have the the original exception thrown.

qwwdfsad:
…n exception has been thrown, so the original exception will be propagated

Fixes #1187

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
+++ b/core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
@@ -5,9 +5,9 @@
 package kotlinx.serialization.encoding
 
 import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.modules.*
-import kotlinx.serialization.builtins.*
 
 /**
  * Decoder is a core deserialization primitive that encapsulates the knowledge of the underlying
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
```diff
@@ -524,10 +524,15 @@ public inline fun <T> Decoder.decodeStructure(
     block: CompositeDecoder.() -> T
 ): T {
     val composite = beginStructure(descriptor)
+    var ex: Throwable? = null
```
sandwwraith, line 527: Wouldn't it be the same just to write

```kotlin
val composite = beginStructure(descriptor)
val result = composite.block()
composite.endStructure()
return result
```

without all try-catch logic?

qwwdfsad: It's going to break for a `return` statement.
And marking it with `crossinline` makes things less pretty

sandwwraith (approved):

PR #1212, opened 2020-11-19
Fixed a couple typos in the guide intro.

ianbrandt:


docs/serialization-guide.md
```diff
@@ -1,11 +1,11 @@
 # Kotlin Serialization Guide
 
-Kotlin Serialization is cross-platform and multi-format framework for data serialization &mdash;
-converting trees of objects to strings, byte arrays, or other _serial_ representations and back.
-Kotlin Serialization fully supports and enforces Kotlin type system, making sure only valid 
+Kotlin Serialization is a cross-platform and multi-format framework for data serialization&mdash;converting
```
sandwwraith, line 3: Em dash, in contradiction to hyphen, should have spaces on moth sides

docs/serialization-guide.md
```diff
@@ -1,11 +1,11 @@
...
-Kotlin Serialization is cross-platform and multi-format framework for data serialization &mdash;
-converting trees of objects to strings, byte arrays, or other _serial_ representations and back.
-Kotlin Serialization fully supports and enforces Kotlin type system, making sure only valid 
+Kotlin Serialization is a cross-platform and multi-format framework for data serialization&mdash;converting 
+trees of objects to strings, byte arrays, or other _serial_ representations and back.
+Kotlin Serialization fully supports and enforces the Kotlin type system, making sure only valid 
 objects can be deserialized. 
  
-Kotlin Serialization is not just a library. It is a compiler plugin that is bundled with the Kotlin
+Kotlin Serialization is not just a library. It's a compiler plugin that is bundled with the Kotlin
```
sandwwraith, line 8: Since this guide is more or less formal document, we tend to use more formal 'it is' version — check the rest of the guide

sandwwraith (changes requested):

docs/serialization-guide.md
ianbrandt: Thanks for the feedback. I've always known it to be the opposite for formal writing:

* https://en.wikipedia.org/wiki/Dash#Spacing_and_substitution
* https://www.thepunctuationguide.com/em-dash.html#spaces
* https://www.quickanddirtytips.com/education/grammar/em-dash-space

The En dash is sometimes spaced, when used as an Em dash:

* https://en.wikipedia.org/wiki/Dash#Spacing

If this guide is meant to be formal, as you say in your other comment, I would think the unspaced Em dash would be correct. Still, I can change it back if you'd like.

docs/serialization-guide.md
ianbrandt: Reverted.

docs/serialization-guide.md
sandwwraith: Ok then, thanks for clarifying that

sandwwraith (approved):

PR #1244, opened 2020-12-09
Experimental support for inline classes

sandwwraith:
* Support for `@Serializable inline class` with generated (requires 1.4.30 IR compiler) and custom serializers
* Standard serializers for UInt, ULong, UByte and UShort
* For both streaming and tree Json


formats/json/commonTest/src/kotlinx/serialization/features/inline/InlineClassesCompleteTest.kt
```diff
@@ -0,0 +1,133 @@
...
+
+package kotlinx.serialization.features.inline
+
+import kotlinx.serialization.*
+import kotlinx.serialization.Box
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+@Serializable
+inline class MyInt(val i: Int)
```
shanshin, line 14: Can these tests be reused later for other formats?

shanshin (approved):
It currently supports unsigned types only for JSON format.

formats/json/commonTest/src/kotlinx/serialization/features/inline/InlineClassesCompleteTest.kt
sandwwraith: In theory yes, in practice — currently there's no way to move test data to core test source, because json-common-test source set does not depend on core-common-test and can't see declaration in there

core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -378,12 +403,50 @@ public interface CompositeEncoder {
...
+     * Calling [Encoder.beginStructure] on returned instance is an undefined behavior.
+     *
+     * @see Encoder.encodeInline
+     */
+    @ExperimentalSerializationApi
+    public fun encodeInlineElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        inlineDescriptor: SerialDescriptor
+    ): Encoder?
```
qwwdfsad, line 448: Here as well

core/commonMain/src/kotlinx/serialization/internal/InlineClassDescriptor.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+
+@Suppress("Unused")
+@PublishedApi
+@OptIn(ExperimentalSerializationApi::class)
+internal class InlineClassDescriptor(name: String, generatedSerializer: GeneratedSerializer<*>) :
+    PluginGeneratedSerialDescriptor(name, generatedSerializer, 1) {
```
qwwdfsad, line 15: We are typically using 
```
internal class InlineClassDescriptor(
    name: String, generatedSerializer:
    GeneratedSerializer<*>
) : PluginGeneratedSerialDescriptor(name, generatedSerializer, 1) {
```
instead

core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
```diff
@@ -37,12 +37,12 @@ internal open class PluginGeneratedSerialDescriptor(
...
 
     // Lazy because of JS specific initialization order (#789)
-    private val typeParameterDescriptors: Array<SerialDescriptor> by lazy {
+    internal val typeParameterDescriptors: Array<SerialDescriptor> by lazy {
         generatedSerializer?.typeParametersSerializers()?.map { it.descriptor }.compactArray()
     }
 
     // Can be without synchronization but Native will likely break due to freezing
-    private val _hashCode: Int by lazy { hashCodeImpl(typeParameterDescriptors) }
+    internal val _hashCode: Int by lazy { hashCodeImpl(typeParameterDescriptors) }
```
qwwdfsad, line 45: It could be left private and `hashCode()` can be used instead

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -225,3 +228,16 @@ internal class StreamingJsonDecoder internal constructor(
         return enumDescriptor.getElementIndexOrThrow(decodeString())
     }
 }
+
+@OptIn(ExperimentalSerializationApi::class)
+@ExperimentalUnsignedTypes
+internal class JsonDecoderForUnsignedTypes(private val reader: JsonReader, json: Json) :
```
qwwdfsad, line 234: Same here

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -225,3 +228,16 @@ internal class StreamingJsonDecoder internal constructor(
...
+    AbstractDecoder() {
+    override val serializersModule: SerializersModule = json.serializersModule
+    override fun decodeElementIndex(descriptor: SerialDescriptor): Int = error("unsupported")
+
+    override fun decodeInt(): Int = reader.takeString().toUInt().toInt()
+    override fun decodeLong(): Long = reader.takeString().toULong().toLong()
+    override fun decodeByte(): Byte = reader.takeString().toUByte().toByte()
+    override fun decodeShort(): Short = reader.takeString().toUShort().toShort()
```
qwwdfsad, line 242: This code will throw `NumberFormatException` , it would be nice to throw more specific `SE` instead

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
```diff
@@ -130,6 +148,16 @@ internal class StreamingJsonEncoder(
         return true
     }
 
+    override fun encodeInline(inlineDescriptor: SerialDescriptor): Encoder {
+        return if (inlineDescriptor.isUnsignedNumber) StreamingJsonEncoder(
+            ComposerForUnsignedNumbers(
+                composer.sb,
+                composer.json
+            ), json, mode, null
```
qwwdfsad, line 156: Why not just passing `modeReuseCache` to keep the code as simple as it has been?

qwwdfsad (changes requested):

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
sandwwraith: Because of https://github.com/Kotlin/kotlinx.serialization/blob/dev/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt#L36 and `!==` in it:

Since we re-create `StreamingJsonEncoder` with different composer, it fails reference equality check and in result, cache is populated with encoder+ComposerForUnsigned numbers, which is incorrect

qwwdfsad: Ok, thanks!

core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
sandwwraith: You mean `super.hashCode()`? Yeah, why not

docs/inline-classes.md
```diff
@@ -0,0 +1,196 @@
...
+@Serializable(UIDSerializer::class)
+class UID(val uid: Int)
+```
+
+For some reason,`uid` type is `Int`, but we want it to be an unsigned integer in JSON. We can start writing the
+following custom serializer:
+
+```kotlin
+object UIDSerializer: KSerializer<UID> {
+  override val descriptor = UInt.serializer().descriptor
```
sandwwraith, line 142: Here we lack analogue of `PrimitiveClassDescriptor`

core/commonMain/src/kotlinx/serialization/internal/EmptyEncoder.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.modules.*
+
+/**
+ * Encoder that does not do any operations. Its main purpose is to ignore data instead of writing it.
+ */
+@ExperimentalSerializationApi
+internal object EmptyEncoder : AbstractEncoder() {
```
qwwdfsad, line 16: It is not empty, but rather `NoOpEncoder`

formats/json/commonTest/src/kotlinx/serialization/features/inline/InlineClassesTest.kt
```diff
@@ -0,0 +1,111 @@
...
+package kotlinx.serialization.features.inline
+
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+@Serializable
+data class SimpleContainerForUInt(val i: UInt)
```
qwwdfsad, line 17: These classes can be moved into `InlineClassesTest`, can't they?

formats/json/commonTest/src/kotlinx/serialization/features/inline/InlineClassesCompleteTest.kt
```diff
@@ -0,0 +1,133 @@
...
+    val nullableOverSerializable: OverSerializableNullable, // LIntData;
+    val nullableOverSerializableNullable: OverSerializableNullable?, // LOverSerializableNullable;
+    val withT: WithT<Int>, // LBox;
+    val withTNullable: WithT<Int>?, // LBox
+    val withNullableTNullable: WithT<Int?>?, // LBox;
+    val withTNullableTNullable: WithTNullable<Int?>? // LWithTNullable;
+)
+
+@Serializable
+data class WithGenerics(
```
qwwdfsad, line 48: Here the scope of the test classes can also be reduced to `InlineClassesCompleteTest`

formats/json/commonTest/src/kotlinx/serialization/features/inline/EncodeInlineElementTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.test.*
+import kotlin.test.Test
+
+@Serializable(WithUnsignedSerializer::class)
+data class WithUnsigned(val u: UInt)
+
+object WithUnsignedSerializer : KSerializer<WithUnsigned> {
```
qwwdfsad, line 18: Here as well

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt
```diff
@@ -98,6 +98,19 @@ private sealed class AbstractJsonTreeEncoder(
         putElement(tag, JsonPrimitive(value.toString()))
     }
 
+    @OptIn(ExperimentalUnsignedTypes::class)
+    override fun encodeTaggedInline(tag: String, inlineDescriptor: SerialDescriptor): Encoder {
+        return object : AbstractEncoder() {
+            override val serializersModule: SerializersModule = json.serializersModule
+            fun putUnquotedString(s: String) = putElement(tag, JsonLiteral(s, isString = false))
+
+            override fun encodeInt(value: Int) = putUnquotedString(value.toUInt().toString())
```
qwwdfsad, line 107:
```suggestion
            override val serializersModule: SerializersModule = json.serializersModule
            
            fun putUnquotedString(s: String) = putElement(tag, JsonLiteral(s, isString = false))
            override fun encodeInt(value: Int) = putUnquotedString(value.toUInt().toString())
```

core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -378,12 +401,48 @@ public interface CompositeEncoder {
...
+     * Calling [Encoder.beginStructure] on returned instance leads to an undefined behavior.
+     *
+     * @see Encoder.encodeInline
+     */
+    @ExperimentalSerializationApi
+    public fun encodeInlineElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        inlineDescriptor: SerialDescriptor
+    ): Encoder
```
qwwdfsad, line 444: I've studied all our implementations of this method (end `decodeInlineElement` as well) and have the following observation:

It seems the third parameter is redundant and is always equal to `descriptor.getElementDescriptor(index)`.
Also, it's pretty error-prone for external serializers, because e.g. for `class CustomlySerializedClass(val color: Color)` `descriptor.getElementDescriptor(index)` and `Color.serializer().descriptor` may point to **different** descriptors (e.g. "hex string" and "single integer").

Can we get rid of the third parameter?

qwwdfsad (changes requested):

formats/json/commonTest/src/kotlinx/serialization/features/inline/InlineClassesTest.kt
sandwwraith: 'Inline classes are only allowed on top level' says Kotlin compiler

qwwdfsad (approved):
Nice, great job!

PR #1246, opened 2020-12-10
Fix JsonObject.toString()

Karlatemp:
Before fix
```kotlin
fun main() {
  val jsonObject = JsonObject(
    mapOf("\\.", 1)
  )
  println(jsonObject) // {"\.": 1}
  println(Json.encodeToString(jsonObject)) // {"\\.": 1}
}
```


qwwdfsad (approved):
Great job, thanks!

PR #1255, opened 2020-12-14
Generator for .proto files based on serializable Kotlin classes

### LINKED ISSUE #34: Generator for .proto files based on serializable Kotlin classes
The idea is that you can have your Kotlin classes as your "master" code that conveniently declares your APIs and that you can analyze and refactor in IDE. Then, if you want your APIs to be used from other languages, you could generate a protobuf .proto files based on that and use them for other languages. 

The design shall have at least two layers. On lower layer we need a format-agnostic way to traverse structure of serializable classes, that is `ClassDesc` shall contain the corresponding methods to make full reflection on the serial representation of classes possible. On the next layer, the format-specific code would be used. In this case, there shall be a tool that takes a set serializable classes (with an ability to scan a whole folder of compiled classes) and outputs a proto file.

shanshin:
Resolves #34


formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,344 @@
...
+                "kind '${descriptor.kind}'." +
+                "As map key type in protobuf allowed only scalar type except for floating point types and bytes.")
+    }
+    return scalarTypeName(descriptor)
+}
+
+@ExperimentalSerializationApi
+private fun protobufMapValueType(descriptor: SerialDescriptor): String {
+    if (descriptor.isProtobufRepeated) {
+        throw ProtoSchemeGenerationException("List is not allowed as a map value type in protobuf!")
```
sandwwraith, line 323: I'd avoid 'screaming' error messages (with exclamation marks)

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,344 @@
...
+
+        try {
+            when {
+                childDescriptor.isProtobufNamedType -> generateNamedType(descriptor, childDescriptor, index, builder)
+                childDescriptor.isProtobufMap -> generateMapType(childDescriptor, builder)
+                childDescriptor.isProtobufRepeated -> generateListType(childDescriptor, builder)
+                else -> throw ProtoSchemeGenerationException("Unprocessed message field type with serial name " +
+                        "'${childDescriptor.serialName}' and kind '${childDescriptor.kind}'! Internal error.")
+            }
+        } catch (e: Exception) {
```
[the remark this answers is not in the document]
sandwwraith: Also, exception causes seems to be handled correctly only on JVM

formats/protobuf-generator/jvmTest/resources/FieldNumberClass.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto2";
+
+package kotlinx.serialization.protobuf.generator.scheme;
+
+// serial name: kotlinx.serialization.protobuf.generator.GenerationTest.FieldNumberClass
+message FieldNumberClass {
+  required int32 a = 1;
+  required int32 b = 5;
+  required int32 c = 3;
+}
```
sandwwraith, line 10: On `formats/protobuf-generator/jvmTest/resources/Generelized.proto`, github says 'empty file'

formats/protobuf-generator/jvmTest/resources/AbstractHolder.proto
```diff
@@ -0,0 +1,15 @@
...
+
+package kotlinx.serialization.protobuf.generator.scheme;
+
+// serial name: kotlinx.serialization.protobuf.generator.GenerationTest.AbstractHolder
+message AbstractHolder {
+  required Polymorphic_AbstractClass_ abs = 1;
+}
+
+// serial name: kotlinx.serialization.Polymorphic<AbstractClass>
+message Polymorphic_AbstractClass_ {
```
sandwwraith, line 11: It seems that such generator will produce new message for each polymorphic-base class. Is that what we intended?

formats/protobuf-generator/jvmTest/src/kotlinx/serialization/protobuf/generator/GenerationTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        assertProtoForType(OptionsClass::class, mapOf("java_package" to "api.proto", "java_outer_classname" to "Outer"))
+        assertProtoForType(ListClass::class)
+        assertProtoForType(MapClass::class)
+        assertProtoForType(OptionalClass::class)
+        assertProtoForType(ContextualHolder::class)
+        assertProtoForType(AbstractHolder::class)
+        assertProtoForType(SealedHolder::class)
+    }
+
+    private inline fun <reified T : Any> assertProtoForType(clazz: KClass<T>, options: Map<String, String> = emptyMap()) {
```
sandwwraith, line 131: I'd also add a function here that allows to re-generate proto files (and can be launched from IDE). Probably just `main` in test sources.

sandwwraith (changes requested):
Publication seems to fail on CI — can you check it?

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,344 @@
...
+    return result
+}
+
+
+@ExperimentalSerializationApi
+private fun addCustomTypeWithElements(descriptor: SerialDescriptor, all: MutableMap<String, SerialDescriptor>) {
+    when {
+        descriptor.isProtobufScalar -> return
+        descriptor.isProtobufStaticMessage ->
+            if (!all.containsKey(descriptor.serialName)) {
```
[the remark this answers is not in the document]
shanshin: else brunch isn't correct here. 
all.containsKey(descriptor.serialName) - means we are already added message type in generation queue and we no need to generate it twice.

Simple example:
```
@Serializable
data class A(val i: Int)
@Serializable
data class B(val a: A)
@Serializable
data class C(val a: A)
```
in this case descriptors for B and for C class contains descriptor for A as element. Enough to add it one time to generation queue.

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,344 @@
...
+                    addCustomTypeWithElements(childDescriptor, all)
+                }
+            }
+        descriptor.isProtobufEnum -> {
+            if (!all.containsKey(descriptor.serialName)) {
+                all[descriptor.serialName] = descriptor
+            }
+        }
+        descriptor.isProtobufRepeated -> addCustomTypeWithElements(descriptor.getElementDescriptor(0), all)
+        descriptor.isProtobufMap -> addCustomTypeWithElements(descriptor.getElementDescriptor(1), all)
```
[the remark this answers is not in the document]
shanshin: Protobuf supports only scalar type as a key of map - see check below.

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+        descriptor.isProtobufOpenMessage -> {
+            if (!all.containsKey(descriptor.serialName)) {
+                all[descriptor.serialName] = descriptor
+            }
+        }
+        descriptor.isProtobufContextualMessage -> return
+        else -> throw ProtoSchemeGenerationException("Unrecognized custom type with serial name "
+                + "'${descriptor.serialName}' and kind '${descriptor.kind}'! Internal error.")
```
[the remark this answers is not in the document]
shanshin: In this file "internal error" means that this error should not happen on any user data. If it occurred - сode not working after some refactoring/fixes/improvements.

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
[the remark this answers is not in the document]
shanshin: Rethrow exception is for adding information about wrong descriptor element.
I think it's a bad idea to pass names of descriptor, message and a field to several levels of functions

formats/protobuf-generator/jvmTest/resources/AbstractHolder.proto
shanshin: Ambiguous.
On the one hand we can reuse only one polymorphic message type. It gives shorter scheme.
On the other after protoc class generation the user will not be able to determine the type compatibility and can easily write data to the wrong type. Having an individual message, the user will be able to see the original polymorphic type and find it in the kotlin sources.

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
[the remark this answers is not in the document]
sandwwraith: Indeed, you're right. As a consequence, it seems we can't detect when user accidentally provided the same serial name for 2 different classes (unless we compare descriptors fully, maybe). 

Still, we have to check that no message names will collide after we cut out the package

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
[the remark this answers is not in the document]
sandwwraith: Then it may be an `AssertionError/IllegalStateException`

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,383 @@
...
+                            "'${childDescriptor.serialName}' and kind '${childDescriptor.kind}'"
+                )
+            }
+        } catch (e: IllegalArgumentException) {
+            throw IllegalArgumentException(
+                "An error occurred during value generation for field $fieldName of message ${serialDescriptor.protobufCustomTypeName} " +
+                        "(serial name ${serialDescriptor.serialName}): ${e.message}", e
+            )
+        } catch (e: Exception) {
+            throw IllegalStateException(
```
sandwwraith, line 160: I'd add a comment that this branch is for internal errors, and previous is for incorrect user input.

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,383 @@
...
+    builder.append(protobufRepeatedType(descriptor.getElementDescriptor(0)))
+}
+
+private val SerialDescriptor.isProtobufNamedType: Boolean
+    get() {
+        return isProtobufScalar || isProtobufCustomType
+    }
+
+private val SerialDescriptor.isProtobufCustomType: Boolean
+    get() {
```
sandwwraith, line 231: You can use `get() = ...` instead of creating block body

formats/protobuf-generator/jvmTest/src/kotlinx/serialization/protobuf/generator/GenerationTest.kt
```diff
@@ -0,0 +1,196 @@
...
+    ) {
+        val scheme = clazz.java.getResourceAsStream("/${clazz.simpleName}.proto").readBytes().toString(Charsets.UTF_8)
+        assertEquals(scheme, generateProto(listOf(serializer(typeOf<T>()).descriptor), TARGET_PACKAGE, options))
+    }
+
+}
+
+/*
+// Regenerate all proto file for tests.
+private fun main() {
```
sandwwraith, line 170: Are there any problems with `main()` being uncommented?

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
```diff
@@ -0,0 +1,383 @@
...
+    val customTypesByName = linkedMapOf<String, CustomTypeDeclaration>()
+    customTypeDescriptorsBySerialName.values.forEach {
+        val serialNameAsIdent = makeProtobufIdent(it.serialName.substringAfterLast('.', it.serialName))
+        var nameVariantNumber = 1
+        var name = serialNameAsIdent
+        while (customTypesByName.containsKey(name)) {
+            nameVariantNumber++
+            name = "${serialNameAsIdent}_$nameVariantNumber"
+        }
+        customTypesByName[name] = CustomTypeDeclaration(name, it)
```
sandwwraith, line 36: Have you written the test for this functionality on duplicated names?

sandwwraith (commented):
I'm still not sure if we should mangle incorrect serial names instead of throwing an exception

sandwwraith:
Also, please investigate why the build is failing

formats/protobuf-generator/commonMain/src/kotlinx.serialization.protobuf.generator/Generation.kt
shanshin: What kind of test are we talking about?
Now if two descriptors have the same `serialName` they are considered to be the same classes.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    builder.append("message ").append(messageName).appendLine(" {")
+
+    val usedNumbers: MutableSet<Int> = mutableSetOf()
+    val nestedTypes = mutableListOf<SerialDescriptor>()
+    for (index in 0 until messageDescriptor.elementsCount) {
+        val fieldName = messageDescriptor.getElementName(index)
+        if (!fieldName.isProtobufIdent) {
+            throw IllegalArgumentException("Invalid name of the field in protobuf scheme '$messageName' for class with serial name '${messageDescriptor.serialName}'")
```
sandwwraith, line 119:
```suggestion
            throw IllegalArgumentException("Invalid name of the field '$fieldName' in protobuf scheme '$messageName' for class with serial name '${messageDescriptor.serialName}'")
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+        }
+        else -> {
+            // enum or regular message
+            nestedTypes = listOf(fieldDescriptor)
+            fieldDescriptor.messageOrEnumName
+        }
+    }
+
+    if (fieldDescriptor.isNullable && messageDescriptor.isElementOptional(index)) {
+        builder.appendLine("  // WARNING: this field is nullable and has default value, it's impossible to unambiguously interpret an absence value.")
```
sandwwraith, line 185: This bigger message bears the same information that the following smaller one: default value is missing from schema. 't's impossible to unambiguously interpret an absence value' is a valid point, but does not tell user what happens because of this. Also most of the default values would be 'null' and 'null value does not support' may confuse users (because we actually support it in that case)

I'd rather remove this clause and join it with the following with simplified a message to e.g. 'Absence of this value is decoded to a default value declared in the Kotlin class and omitted from this schema'. Additional restrictions and null handling details should be described in the documentation.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+    messageDescriptor: SerialDescriptor,
+    index: Int,
+    builder: StringBuilder
+): List<SerialDescriptor> {
+    val mapDescriptor = messageDescriptor.getElementDescriptor(index)
+    val fieldName = messageDescriptor.getElementName(index)
+    val valueDescriptor = mapDescriptor.getElementDescriptor(1).let {
+        if (it.isProtobufCollection) {
+            if (it.isNullable) {
+                builder.appendLine("  // WARNING: null value is not supported for nested collections")
```
sandwwraith, line 210: This message can also have more explanations to the user, e.g. 'Nullable collections can not be represented in protobuf. Missing values would be treated as an empty collection'

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+            }
+            val wrapperName = "${messageDescriptor.messageOrEnumName}_${fieldName}"
+            buildClassSerialDescriptor(wrapperName) {
+                element("value", it)
+            }
+        } else {
+            if (it.isNullable) {
+                builder.appendLine("  // WARNING: null value is not supported for map value")
```
sandwwraith, line 218: Same here

sandwwraith (changes requested):
Great job! What's left to do is to tide up warning messages, finish documentation and come up with a good name

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
+@file:OptIn(ExperimentalSerializationApi::class)
+
+package kotlinx.serialization.protobuf.schema
+
+import kotlinx.serialization.ExperimentalSerializationApi
```
qwwdfsad, line 5: General note: a single file with public methods that are not widely used will pollute the global scope, let's avoid that.

The easiest way to do so is to create an object `ProtobufSchemaGenerator` and declare all methods within this object

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+
+/**
+ * Generate protocol buffers version 2 schema text for given serializable descriptors.
+ * All descriptors will be placed in one schema which can be saved in one `.proto` file.
+ * @param descriptors descriptors of serializable classes for which the scheme will be generated
+ * @param packageName protobuf common package for schema
+ * @param options protobuf options to include in schema
+ */
+@ExperimentalSerializationApi
+public fun generateProto2Schema(
```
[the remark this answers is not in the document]
qwwdfsad: I double this, let's stick to 'generateProtobufSchema'

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+        PrimitiveKind.STRING -> "string"
+    }
+}
+
+private fun removeLineBreaks(text: String): String {
+    return text.replace('\n', ' ').replace('\r', ' ')
+}
+
+@SharedImmutable
+private val IDENT_REGEX = Regex("[A-Za-z][A-Za-z0-9_]*")
```
qwwdfsad, line 408: `IDENT` looks like a misspelled `indent` :)
Could you please use its full name, `IDENTIFIER_REGEX`?

qwwdfsad: Applicable to other methods as well

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.protobuf.ProtoIntegerType
+import kotlinx.serialization.protobuf.ProtoNumber
+import kotlinx.serialization.protobuf.ProtoType
+import kotlin.native.concurrent.SharedImmutable
+
+/**
+ * Generate protocol buffers version 2 schema text for given serializable descriptors.
+ * All descriptors will be placed in one schema which can be saved in one `.proto` file.
```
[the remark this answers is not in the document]
qwwdfsad: Yes, please. 

It all can be written in the detail documentation of `ProtobufSchemaGenerator` object

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+}
+
+private fun generateMapType(
+    messageDescriptor: SerialDescriptor,
+    index: Int,
+    builder: StringBuilder
+): List<SerialDescriptor> {
+    val mapDescriptor = messageDescriptor.getElementDescriptor(index)
+    val fieldName = messageDescriptor.getElementName(index)
+    val valueDescriptor = mapDescriptor.getElementDescriptor(1).let {
```
qwwdfsad, line 207: This level of nesting indicates `let` abuse.

Please consider avoiding `let` in such complex scenarios

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+        if (it.isProtobufCollection) {
+            if (it.isNullable) {
+                builder.appendLine("  // WARNING: null value is not supported for nested collections")
+            }
+            val wrapperName = "${messageDescriptor.messageOrEnumName}_${fieldName}"
+            buildClassSerialDescriptor(wrapperName) {
+                element("value", it)
+            }
+        } else {
```
qwwdfsad, line 216: [nit] you can add a `return` in the end of first branch, then `else` (what's more important -- its nesting) won't be necessary

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+private fun generateEnum(
+    enum: SerialDescriptor,
+    builder: StringBuilder
+) {
+    val enumName = enum.messageOrEnumName
+    if (!enumName.isProtobufIdent) {
+        throw IllegalArgumentException("Invalid name for the message in protobuf scheme '$enumName'. Serial name of the class '${enum.serialName}'")
```
qwwdfsad, line 306: Here and everywhere: `scheme` -> `schema`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+    return text.replace('\n', ' ').replace('\r', ' ')
+}
+
+@SharedImmutable
+private val IDENT_REGEX = Regex("[A-Za-z][A-Za-z0-9_]*")
+
+private val String.isProtobufFullIdent: Boolean
+    get() = split('.').all { it.isProtobufIdent }
+
+private val String.isProtobufIdent: Boolean get() = matches(IDENT_REGEX)
```
qwwdfsad, line 413: This property is used to assert that the name **is** valid identifier and throw an exception otherwise.

It's a good indicator that you need something like `checkIsValidIdentifier(name)` instead

qwwdfsad (changes requested):

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
sandwwraith: We have only one public function here btw, so I doubt we need an object. 

Maybe move this method to a  `ProtoBuf` companion? Although it already has default methods for encoding/decoding....

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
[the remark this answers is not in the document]
sandwwraith: I don't like 'Schema' very much because one may think that this method returns a meaningful representation of the schema in Kotlin classes (like JsonElement), but it returns String. I'd add 'Text' or something else as a suffix to highlight that result of this method is useful for only one purpose — to write it to `.proto` file

formats/protobuf/jvmTest/resources/common/schema.proto
```diff
@@ -0,0 +1,259 @@
...
+  // WARNING: null value is not supported for list elements
+  repeated int32 value = 1;
+}
+
+// serial name 'NullableNestedCollections_legacyMap'
+message NullableNestedCollections_legacyMap {
+  // WARNING: This field is marked as nullable but it does not support null values
+  repeated int32 key = 1;
+  // WARNING: This field is marked as nullable but it does not support null values
+  repeated int32 value = 2;
```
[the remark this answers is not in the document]
shanshin: It is worth adding reading data from classes generated by protoc. For protoc to work, it is necessary that messages do not repeat in files - the easiest way to achieve this is by generating just one common proto file.
M.b. it should be implemented later.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
shanshin: This designation is taken from the official google documentation, maybe they wanted to confuse everyone :)

shanshin: This designation is taken from the official google documentation, maybe they wanted to confuse everyone :)

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
shanshin: A single method will throw an exception with a common and not very verbose message.

qwwdfsad: you can slightly modify it if you want to :)
Few options here: single line check-like check `checkIsValidIdentifier(name) { ...lazy error message... }` or `descriptor. checkIsValidIdentifier(name)`.

Don't really insist tho

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/Generation.kt
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    if (fieldDescriptor.isNullable && messageDescriptor.isElementOptional(index)) {
+        builder.appendLine("  // WARNING: this field is nullable and has default value, it's impossible to unambiguously interpret an absence value.")
+        builder.appendLine("  //   For this field null value does not support and absence value denotes as default value. Default value is not present in the schema.")
+    }
+    if (!fieldDescriptor.isNullable && messageDescriptor.isElementOptional(index)) {
+        builder.appendLine("  // WARNING: an absence value is decoded as a default value that is not present in the schema")
+    }
+    val optional = fieldDescriptor.isNullable || messageDescriptor.isElementOptional(index)
```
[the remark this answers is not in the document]
shanshin: It's implemented together with nested collections.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+
+/**
+ * Experimental generator of ProtoBuf schema that is compatible with [serializable][Serializable] Kotlin classes
+ * and data encoded and decoded by [ProtoBuf] format.
+ *
+ * The schema is generated based on provided [SerialDescriptor] and is compatible with proto2 schema definition.
+ * An arbitrary Kotlin class represent much wider object domain than the ProtoBuf specification, thus schema generator
+ * has the following list of restrictions:
+ *
+ *  * Serial name of the class and all its field should be a valid Proto [identifier](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec)
```
sandwwraith, line 17:
```suggestion
 *  * Serial name of the class and all its fields should be a valid Proto [identifier](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec)
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ *  * Nullable values are allowed only for Kotlin [nullable][SerialDescriptor.isNullable] types
+ *  * [SerialName] of the type uniquely identifies the proto message type and two or more fields with the same serial name
+ *    are considered to have the same type.
+ *  * Nested collections, e.g. `List<List<Int>>` are represented using the artificial wrapper message in order to distinguish
+ *    repeated fields boundaries.
+ *  * Default Kotlin values are nto representable in proto schema. A special commentary is generated for properties with default values.
```
sandwwraith, line 23:
```suggestion
 *  * Default Kotlin values are not representable in proto schema. A special commentary is generated for properties with default values.
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ */
+@ExperimentalSerializationApi
+public object ProtoBufSchemaGenerator {
+
+    /**
+     * Generate text of protocol buffers schema version 2 for the given [rootDescriptor].
+     * The resulting schema will contain all types referred by [rootDescriptor].
+     *
+     * [packageName] define common protobuf package for all messages and enum in the schema, it may contain `'a'`..`'z'`
+     * letters in upper and lower case, decimal digits, `'.'` or `'_'` chars, but started only from a letter and
```
sandwwraith, line 45: you mean 'must be started'?

sandwwraith (approved):

LouisCAD:
Several typos in the commit message of the "Schema generator improvements" commit @qwwdfsad FYI.

Didn't check code changes because I'm on mobile, and the app won't let me see the "large diff".

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ *  * Default Kotlin values are not representable in proto schema. A special commentary is generated for properties with default values.
+ *  * Empty nullable collections are not supported by the generated schema and will be prohibited in [ProtoBuf] as well
+ *    due to their ambiguous nature.
+ *
+ * Temporary limitation:
+ *  * [Contextual] data is represented as as `bytes` type
+ *  * [Polymorphic] data is represented as a artificial `KotlinxSerializationPolymorphic` message.
+ *
+ * Other types are mapped according to their specification: primitives as primitives, lists as 'repeated' fields and
+ * maps as 'repeated' map entries.
```
shanshin, line 32: For some cases, Map is a `map`, for others is a `repeated`, it depends on the type of key.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ *    repeated fields boundaries.
+ *  * Default Kotlin values are not representable in proto schema. A special commentary is generated for properties with default values.
+ *  * Empty nullable collections are not supported by the generated schema and will be prohibited in [ProtoBuf] as well
+ *    due to their ambiguous nature.
+ *
+ * Temporary limitation:
+ *  * [Contextual] data is represented as as `bytes` type
+ *  * [Polymorphic] data is represented as a artificial `KotlinxSerializationPolymorphic` message.
+ *
+ * Other types are mapped according to their specification: primitives as primitives, lists as 'repeated' fields and
```
shanshin, line 31: Perhaps you meant scalar types in protobuf? However, they also include the String that is not primitive in Kotlin.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ * Experimental generator of ProtoBuf schema that is compatible with [serializable][Serializable] Kotlin classes
+ * and data encoded and decoded by [ProtoBuf] format.
+ *
+ * The schema is generated based on provided [SerialDescriptor] and is compatible with proto2 schema definition.
+ * An arbitrary Kotlin class represent much wider object domain than the ProtoBuf specification, thus schema generator
+ * has the following list of restrictions:
+ *
+ *  * Serial name of the class and all its fields should be a valid Proto [identifier](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec)
+ *  * Nullable values are allowed only for Kotlin [nullable][SerialDescriptor.isNullable] types
+ *  * [SerialName] of the type uniquely identifies the proto message type and two or more fields with the same serial name
```
shanshin, line 19: Not just the same serial name, but the same extracted name.
e.g. fields with descriptors with serial names `com.example.Foo` and `org.foo.Foo?` will have the same message type `Foo` in the schema.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+ *    are considered to have the same type.
```
shanshin, line 20: It is also important that there is no duplicate extracted names in the passed `SerialDescriptor`s, since they are all considered classes with different structures.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,509 @@
...
+/**
+ * Experimental generator of ProtoBuf schema that is compatible with [serializable][Serializable] Kotlin classes
+ * and data encoded and decoded by [ProtoBuf] format.
+ *
+ * The schema is generated based on provided [SerialDescriptor] and is compatible with proto2 schema definition.
+ * An arbitrary Kotlin class represent much wider object domain than the ProtoBuf specification, thus schema generator
+ * has the following list of restrictions:
+ *
+ *  * Serial name of the class and all its fields should be a valid Proto [identifier](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec)
+ *  * Nullable values are allowed only for Kotlin [nullable][SerialDescriptor.isNullable] types
```
shanshin, line 18: null value allowed only for nullable, not optional, and not collection/array types.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
qwwdfsad: It's "primitive" in serialization tho

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
```diff
@@ -0,0 +1,511 @@
...
+/**
+ * Experimental generator of ProtoBuf schema that is compatible with [serializable][Serializable] Kotlin classes
+ * and data encoded and decoded by [ProtoBuf] format.
+ *
+ * The schema is generated based on provided [SerialDescriptor] and is compatible with proto2 schema definition.
+ * An arbitrary Kotlin class represent much wider object domain than the ProtoBuf specification, thus schema generator
+ * has the following list of restrictions:
+ *
+ *  * Serial name of the class and all its fields should be a valid Proto [identifier](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec)
+ *  * Nullable values are allowed only for Kotlin [nullable][SerialDescriptor.isNullable] types, but not [optional][SerialDescriptor.isElementOptional]
```
sandwwraith, line 18: It's the restriction of encoder, not schema generator, isn't it?

qwwdfsad: It's still worth mentioning it here unless we have our ProtoBuf doc rephrased

PR #1257, opened 2020-12-18
Treat Collection as ArrayList in serializer by type lookups

### LINKED ISSUE #1251: A compiler-generated serializer for a Collection property works, serializer<Collection<…>>() does not
**Describe the bug**

- A property of type `Collection<String>` is serialized as `[ … ]`.
- Serializing a value of type `Collection<String>` using `serializer<Collection<String>>()` fails with an error.

**To Reproduce**
```kt
@Serializable
class Foo(val value: Collection<String>)

fun main() {
    val foo = Foo(listOf("foo"))
    val bar: Collection<String> = listOf("bar")

    println(Json.encodeToString(foo)) // outputs {"value":["foo"]}
    println(Json.encodeToString(bar)) // Class 'SingletonList' is not registered for polymorphic serialization in the scope of 'Collection'.
}
```

**Expected behavior**
`Collection` should consistently be serialized as a list, independent of the actual implementation of `Collection`.

Should output:
```
{"value":["foo"]}
["bar"]
```

**Environment**
 - Kotlin version: 1.4.21
 - Library version: 1.0.1
 - Kotlin platforms: JVM
 - Gradle version: 6.5.1

sandwwraith:
Fixes #1251

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -8,6 +8,9 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.builtins.*
+import kotlinx.serialization.builtins.TripleSerializer
+import kotlinx.serialization.builtins.PairSerializer
+import kotlinx.serialization.builtins.MapEntrySerializer
 import kotlinx.serialization.internal.*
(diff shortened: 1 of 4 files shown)


qwwdfsad (approved):
Good catch

PR #1260, opened 2020-12-21
Remove deprecated declarations from runtime

sandwwraith:
- Remove deprecated functions from Decoder (1.4.30 plugin required)
- Adapt exception constructors for coroutines stacktrace recovery (1.4.20 plugin required)
- Bump kotlin & minkotlin versions


qwwdfsad (approved):
Good riddance

Issue #1276, opened 2021-01-02
Contextual serializers do not work for derived classes


martinraison:
**Describe the bug**

It seems contextual serializers only work for concrete classes currently.

**To Reproduce**

Example setup:

```kt
@Serializable
abstract class Message

@Serializable
class SimpleMessage(val body: String) : Message()

@Serializable
class Holder(@Contextual val message: Message)

// A custom serializer for Message
object MessageAsStringSerializer : KSerializer<Message> {
    override val descriptor: SerialDescriptor =
        PrimitiveSerialDescriptor("kotlinx.serialization.MessageAsStringSerializer", PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: Message) {
        // dummy serializer that assumes Message is always SimpleMessage
        check(value is SimpleMessage)
        encoder.encodeString(value.body)
    }

    override fun deserialize(decoder: Decoder): Message {
        return SimpleMessage(decoder.decodeString())
    }
}

val module = SerializersModule {
    contextual(MessageAsStringSerializer)
}
val format = Json { serializersModule = module }
```

Now when we try to do this

```kt
val data = Holder(SimpleMessage("hello"))
println(format.encodeToString(data))
```

we get the following error:

```
Class 'SimpleMessage' is not registered for polymorphic serialization in the scope of 'Message'.
Mark the base class as 'sealed' or register the serializer explicitly.
kotlinx.serialization.SerializationException: Class 'SimpleMessage' is not registered for polymorphic serialization in the scope of 'Message'.
Mark the base class as 'sealed' or register the serializer explicitly.
```

**Expected behavior**

In the scenario above, the `message` field of the `Holder` class should be serialized using `MessageAsStringSerializer`.

**Environment**
 - Kotlin version: 1.4.20
 - Library version: 1.0.1
 - Kotlin platforms: JVM
 - Gradle version: 6.7.1

shanshin:
In your example, you can try using an external serializer for the Message class:
```
@Serializable(with = MessageAsStringSerializer::class)
abstract class Message

@Serializable
class SimpleMessage(val body: String) : Message()

@Serializable
class Holder(val message: Message)
```
and encode instance without contextual serializer
```
val format = Json
val data = Holder(SimpleMessage("hello"))
println(format.encodeToString(data))
```

In this solution properties with `Message` type will serialize/deserialize by `MessageAsStringSerializer`

martinraison:
[...]

PR #1277, opened 2021-01-02
Contextual serialization for derived classes

martinraison:
Fixes: #1276 

The issue #1276 is happening because, in the implementation of `ContextualSerializer`, the serializer is looked up using the concrete class instead of the serializable class: https://github.com/Kotlin/kotlinx.serialization/blob/master/core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt#L56.

The reason for that is not clear to me. In this PR, I'm updating the implementation to do the lookup using the serializable class instead (which is actually simpler). This solves the issue, and all tests still pass.

 I also added the example from the issue as a test case.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
+++ b/core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
@@ -45,6 +45,9 @@ public class ContextualSerializer<T : Any>(
     private val typeParametersSerializers: Array<KSerializer<*>>
 ) : KSerializer<T> {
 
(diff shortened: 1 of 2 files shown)


sandwwraith:
Indeed, you're right — documentation says that contextual serializer uses statically known KClass, so `value::class` is an erroneous approach here (which, luckily, works the same way until we add inheritance).

Thanks for your contribution and time!

sandwwraith (approved):

martinraison:
That's great, thanks a lot for the review & merge!

PR #1294, opened 2021-01-18
Optimize the loop for writing varints in ProtoBuf

qwwdfsad:
It has no performance impact on small (~1-2 bytes) varints, but increases throughput of large (7-9 bytes) varints by up to 25%

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
@@ -8,7 +8,7 @@ import kotlinx.serialization.*
 
 internal class ByteArrayInput(private var array: ByteArray, private val endIndex: Int = array.size) {
     private var position: Int = 0
-    public val availableBytes: Int get() = endIndex - position
+    private val availableBytes: Int get() = endIndex - position
 
     fun slice(size: Int): ByteArrayInput {
         ensureEnoughBytes(size)
@@ -124,6 +124,17 @@ internal class ByteArrayInput(private var array: ByteArray, private val endIndex
 }
 
 internal class ByteArrayOutput {
+
+    private companion object {
+        /*
+         * Map number of leading zeroes -> varint size
+         * See the explanation in this blogpost: https://richardstartin.github.io/posts/dont-use-protobuf-for-telemetry
+         */
+        private val VAR_INT_LENGTHS = IntArray(65) {
+            (63 - it) / 7
+        }
+    }
+
     private var array: ByteArray = ByteArray(32)
     private var position: Int = 0
 
@@ -136,11 +147,11 @@ internal class ByteArrayOutput {
         array = newArray
     }
 
-    public fun size(): Int {
+    fun size(): Int {
         return position
     }
 
-    public fun toByteArray(): ByteArray {
+    fun toByteArray(): ByteArray {
         val newArray = ByteArray(position)
         array.copyInto(newArray, startIndex = 0, endIndex = this.position)
         return newArray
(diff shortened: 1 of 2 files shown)


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
```diff
@@ -189,42 +200,33 @@ internal class ByteArrayOutput {
...
+    }
+
+    fun encodeVarint64(value: Long) {
+        val length = varIntLength(value)
+        ensureCapacity(length + 1)
+        encodeVarint(value, length)
     }
 
-    private fun encodeVarint32SlowPath(value: Int) {
+    private inline fun encodeVarint(value: Long, length: Int) {
```
shanshin, line 219: It is unlikely that someone now is using a 32-bit CPU but these calculations can be slow on them.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
```diff
@@ -189,42 +200,33 @@ internal class ByteArrayOutput {
     }
 
     fun encodeVarint32(value: Int) {
-        ensureCapacity(5)
         // Fast-path: unrolled loop for single byte
+        ensureCapacity(5)
```
shanshin, line 204: Why now for `long` we ensure dynamic length but here always 5?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
```diff
@@ -189,42 +200,33 @@ internal class ByteArrayOutput {
...
-            return
-        }
-        encodeVarint32SlowPath(current)
+        val length = varIntLength(value.toLong())
+        encodeVarint(value.toLong(), length)
+    }
+
+    fun encodeVarint64(value: Long) {
+        val length = varIntLength(value)
+        ensureCapacity(length + 1)
```
shanshin, line 215: Suggest do +1 inside `varIntLength` or `VAR_INT_LENGTHS` init block to match the name.

shanshin (approved):

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
qwwdfsad: Because 32-bit varints are used on a regular basis for tags, it's always beneficial to allocate all the space.
Also, because tags are small, one-byte fast-path is used and we do not have to calculate varint length at all for such values

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
qwwdfsad: Could you please elaborate on what you mean here?

shanshin (approved):

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
qwwdfsad: We generally do not care about 32-bit platforms

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Streams.kt
qwwdfsad: Then it will slightly complicate the loop (have to do decrement), let's leave it as is

PR #1299, opened 2021-01-20
Throw SerializationException when encountering an invalid boolean in …

### LINKED ISSUE #1261: Deserializing invalid JSON does not always throw SerializationException
When deserializing JSON that does not fit the target classes, `SerializationException` should be thrown always. If the input JSON has the wrong type (e.g. `123` where a `Boolean` is expected), `IllegalStateException` is thrown instead (by `StreamingJsonDecoder`, which uses `toBooleanStrict`, for example).

This makes it difficult to distinguish invalid JSON from other errors, especially in a server environment, where you want to inform clients that their request is invalid (as opposed to some internal server error).

**To Reproduce**
```kotlin
@Serializable
data class SerializerBug(val booleanValue: Boolean)

fun main() {
    val json = Json {  }
    // throws java.lang.IllegalStateException: 123 does not represent a Boolean
    json.decodeFromString(SerializerBug.serializer(), """{"booleanValue": 123}""")
}
```

**Expected behavior**
`kotlinx.serialization.SerializationException` is thrown instead of generic `IllegalStateException`.

**Environment**
 - Kotlin version: 1.4.21
 - Library version: 1.0.1
 - Kotlin platforms: Tested on JVM, but this bug exists in common code
 - Gradle version: 6.7

qwwdfsad:
…JSON

Also, narrow down the type of swallowed exceptions

Fixes #1261

### DIFF
--- a/formats/json/commonTest/src/kotlinx/serialization/json/JsonParserFailureModesTest.kt
+++ b/formats/json/commonTest/src/kotlinx/serialization/json/JsonParserFailureModesTest.kt
@@ -64,4 +64,13 @@ class JsonParserFailureModesTest : JsonTestBase() {
             assertFailsWith<JsonDecodingException> { default.decodeFromString(Holder.serializer(), """{""", it) }
         }
     }
+
+    @Serializable
+    class BooleanHolder(val b: Boolean)
+
+    @Test
+    fun testBoolean() = parametrizedTest {
+        assertFailsWith<JsonDecodingException> { default.decodeFromString(BooleanHolder.serializer(), """{"b": fals}""", it) }
+        assertFailsWith<JsonDecodingException> { default.decodeFromString(BooleanHolder.serializer(), """{"b": 123}""", it) }
+    }
 }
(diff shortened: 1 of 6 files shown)


formats/json/commonTest/src/kotlinx/serialization/json/JsonParserFailureModesTest.kt
```diff
@@ -54,6 +54,14 @@ class JsonParserFailureModesTest : JsonTestBase() {
                     it
                 )
             }
+            // 9223372036854775807 is Long.MAX_VALUE
+            assertFailsWith<JsonDecodingException> {
+                default.decodeFromString(
+                    Holder.serializer(),
+                    """{"id":922337203685477580700}""",
```
sandwwraith, line 61: It's not very convenient to spot the problem here. Maybe replace this with `Long.MAX_VALUE.toString() + "00"`?

sandwwraith (approved):

formats/json/commonTest/src/kotlinx/serialization/json/JsonParserFailureModesTest.kt
qwwdfsad: Sure

PR #1301, opened 2021-01-20
Pr/1282

qwwdfsad:


gradle/configure-source-sets.gradle
```diff
@@ -87,4 +88,19 @@ kotlin {
     sourceSets.matching({ it.name.contains("Main") }).all { srcSet ->
         project.ext.set("kotlin.mpp.freeCompilerArgsForSourceSet.${srcSet.name}", "-Xexplicit-api=warning")
     }
+
+    def targetsWithoutTestRunners = ["linuxArm32Hfp", "linuxArm64", "mingwX86"]
+    configure(targets) {
+        // Configure additional binaries to run tests in the background
+        if (["macos", "linux", "mingw"].any { name.startsWith(it) && !targetsWithoutTestRunners.contains(name) }) {
+            binaries {
+                test("background", [nativeDebugBuild]) {
```
sandwwraith, line 97: > this problem couldn't be reproduced with nativeTest because K/N runtime ensures all globals are initialized before main entry point. But it isn't the case for external callers and Obj-C frameworks

Does this runner solve the issue?

qwwdfsad: No. This runner consistently check that serialization can be used from the background threads -- something we didn't have before

sandwwraith (approved):

PR #1302, opened 2021-01-21
Test case to demonstrate Protobuf enum deserialization bug

MariusVolkhart:


qwwdfsad:
Thanks for the reproduce. We've used another one, but providing a self-contained repro is always a good bit of help for us!

PR #1307, opened 2021-01-26
Fix typo in serializers.md

wplong11:
In the document, ISO 6801 is modified to ISO 8601.
The date and time format standard is ISO 8601, not ISO 6801.


sandwwraith:
> ISO 6801:1983
Rubber or plastics hoses — Determination of volumetric expansion

Indeed, not very related to date and time 😄 

Thank you!

PR #1322, opened 2021-02-04
Remove JCenter() from repositories where possible

sandwwraith:


README.md
```diff
@@ -105,7 +105,7 @@ Kotlin DSL:
 
 ```kotlin
 buildscript {
-    repositories { jcenter() }
+    repositories { mavenCentral() }
 
     dependencies {
         val kotlinVersion = "1.4.10"
```
shanshin, line 111: m.b. also actualize version of kotlin and replace usage of 1.0.1 library in README?

shanshin (approved):

README.md
qwwdfsad: It should have been done as part of 1.1.0-RC release

qwwdfsad (approved):

README.md
sandwwraith: Usually, we replace all versions in readme with the next one only on stable releases

PR #1325, opened 2021-02-05
Throw SerializationException instead of IllegalStateException in Enum…

### LINKED ISSUE #1303: Protobuf: Unknown enum values cause ArrayIndexOutOfBoundsException
**Describe the bug**

**To Reproduce**
See #1302 for a test case in the repo that produces the failure.

**Environment**
`master` at `b6eaad1974c6e544f230b0b4e6ccdf15e85ef209`
 - Kotlin platforms: JVM

qwwdfsad:
…Serializer

Fixes #1303


qwwdfsad:
There seems to be a bug in our JS codegen, @sandwwraith could you please take a look?

sandwwraith:
I'll see, but it looks like the problem is not codegen — both IR and legacy test fail

PR #1366, opened 2021-03-11
Add support for watchosX64 target

sbogolepov:


sandwwraith:
```
e: org.jetbrains.kotlin.konan.KonanExternalToolFailure: The /opt/teamcity-agent/system/.persistent_cache/konan/dependencies/clang-llvm-8.0.0-linux-x86-64/bin/clang++ command returned non-zero exit code: 127.
output:
/opt/teamcity-agent/system/.persistent_cache/konan/dependencies/clang-llvm-8.0.0-linux-x86-64/bin/clang++: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```

@sbogolepov does it mean that we have to have some new library installed on agents?

sbogolepov:
@sandwwraith it's a longstanding problem (https://youtrack.jetbrains.com/issue/KT-42693). Actually, it's other way around: agents with relatively new OSes don't have this library installed by default.

sandwwraith:
It turns out agent with missing ncurses library was not meant to be in libraries pool. Merging now

PR #1367, opened 2021-03-12
Fix numeric overflow in JsonLexer

EdwarDDay:
Before this fix `9223372036854775808` is read as `-9223372036854775808`.


qwwdfsad (approved):
Aha, sneaky `Long.MIN_VALUE`. Thanks for spotting!

PR #1370, opened 2021-03-13
Test case to demonstrate Protobuf enum deserialization bug

MariusVolkhart:
During protobuf deserialization, if an unknown enum value is encountered, and the enum is using @ProtoNumber, the deserializer throws an `ArrayIndexOutOfBoundsException` on the JVM instead of a `SerializationException`.

I'm on a limited bandwidth connection, so can't pull all the latest dependencies & am submitting this PR for remote-run before verifying this reproduces the issue.


MariusVolkhart:
Confirmed that this reproduces the bug.

qwwdfsad:
Thanks! We apparently have two code paths for enum deserialization and both contained the bug. Should be fixed now by #1373

PR #1371, opened 2021-03-14
Activate js legacy tests

EdwarDDay:
kotlin version is `1.4.30`, so these tests could be enabled


sandwwraith:
Good catch, thanks!

PR #1373, opened 2021-03-15
Throw SerializationException instead of AIOB when ProtoBug encounters…

### LINKED ISSUE #1303: Protobuf: Unknown enum values cause ArrayIndexOutOfBoundsException

qwwdfsad:
… unknown enum value

Fixes #1303

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
@@ -90,7 +90,10 @@ internal open class ProtobufDecoder(
             if (protoId == protoTag) return i
         }
 
-        return -1
+        throw ProtobufDecodingException(
+            "$protoTag is not among valid ${descriptor.serialName} enum indicies, " +
+                    "values size is ${desc.elementNames}"
+        )
     }
 
     override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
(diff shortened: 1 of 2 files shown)


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -90,7 +90,10 @@ internal open class ProtobufDecoder(
             if (protoId == protoTag) return i
         }
 
-        return -1
+        throw ProtobufDecodingException(
+            "$protoTag is not among valid ${descriptor.serialName} enum indicies, " +
+                    "values size is ${desc.elementNames}"
```
sandwwraith, line 95: Given that it is 'values size', maybe you meant `desc.elementsCount`? Anyway, count/size is not very helpful here because proto numbers can be arbitrary

sandwwraith (approved):

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
qwwdfsad: Good spot

PR #1375, opened 2021-03-15
Support for multiple alternative names during JSON parsing

### LINKED ISSUE #203: Can SerialName annotation support alternate?

sandwwraith:
* Configurable via `@JsonNames` annotation

Fixes #203

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/model/MacroTwitter.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/model/MacroTwitter.kt
@@ -11,11 +11,11 @@ data class MacroTwitterFeed(
 
 @Serializable
 data class MicroTwitterFeed(
-    val statuses: List<TwitterReducedStatus>
+    val statuses: List<TwitterTrimmedStatus>
 )
 
 @Serializable
-data class TwitterReducedStatus(
+data class TwitterTrimmedStatus(
     val metadata: Metadata,
     val created_at: String,
     val id: Long,
@@ -23,8 +23,8 @@ data class TwitterReducedStatus(
     val text: String,
     val source: String,
     val truncated: Boolean,
-    val user: TwitterReducedUser,
-    val retweeted_status: TwitterReducedStatus? = null,
+    val user: TwitterTrimmedUser,
+    val retweeted_status: TwitterTrimmedStatus? = null,
 )
 
 @Serializable
@@ -111,7 +111,7 @@ data class Metadata(
 )
 
 @Serializable
-data class TwitterReducedUser(
+data class TwitterTrimmedUser(
     val id: Long,
     val id_str: String,
     val name: String,
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/model/MacroTwitter.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/model/MacroTwitter.kt
@@ -10,6 +10,24 @@ data class MacroTwitterFeed(
 )
 
 @Serializable
+data class MicroTwitterFeed(
+    val statuses: List<TwitterReducedStatus>
+)
+
+@Serializable
+data class TwitterReducedStatus(
+    val metadata: Metadata,
+    val created_at: String,
+    val id: Long,
+    val id_str: String,
+    val text: String,
+    val source: String,
+    val truncated: Boolean,
+    val user: TwitterReducedUser,
+    val retweeted_status: TwitterReducedStatus? = null,
+)
+
+@Serializable
 data class TwitterStatus(
     val metadata: Metadata,
     val created_at: String,
@@ -90,6 +108,24 @@ data class Urls(val urls: List<Url>)
 data class Metadata(
     val result_type: String,
     val iso_language_code: String
+)
+
+@Serializable
+data class TwitterReducedUser(
+    val id: Long,
+    val id_str: String,
+    val name: String,
+    val screen_name: String,
+    val location: String,
+    val description: String,
+    val url: String?,
+    val entities: UserEntities,
+    val protected: Boolean,
+    val followers_count: Int,
+    val friends_count: Int,
+    val listed_count: Int,
+    val created_at: String,
+    val favourites_count: Int,
 )
 
 @Serializable
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,14 +154,14 @@ Project(name=kotlinx.serialization)
 
 ### Alternative Json names
 
-It's not a rare case when JSON fields got renamed due to a schema version change or something else.
+It's not a rare case when JSON fields got renamed due to a schema version change.
 Renaming JSON fields is available with [`@SerialName` annotation](basic-serialization.md#serial-field-names), but 
-such a renaming blocks ability to decode data with old name. 
+such a renaming blocks ability to decode data with an old name. 
 For the case when we want to support multiple JSON names for the one Kotlin property, there is a [JsonNames] annotation:
 
 ```kotlin
 @Serializable
-data class Project(@JsonNames(["title"]) val name: String)
+data class Project(@JsonNames("title") val name: String)
 
 fun main() {
   val project = Json.decodeFromString<Project>("""{"name":"kotlinx.serialization"}""")
--- a/docs/json.md
+++ b/docs/json.md
@@ -13,6 +13,7 @@ In this chapter we'll walk through various [Json] features.
   * [Pretty printing](#pretty-printing)
   * [Lenient parsing](#lenient-parsing)
   * [Ignoring unknown keys](#ignoring-unknown-keys)
+  * [Alternative Json names](#alternative-json-names)
   * [Coercing input values](#coercing-input-values)
   * [Encoding defaults](#encoding-defaults)
   * [Allowing structured map keys](#allowing-structured-map-keys)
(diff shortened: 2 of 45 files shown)


sandwwraith:
**TODO**: measure performance impact and decide whether this flag should be enabled by default (or should the feature be flagged at all)

formats/json/commonTest/src/kotlinx/serialization/json/JsonCustomSerializersTest.kt
```diff
@@ -13,6 +13,7 @@ import kotlinx.serialization.modules.*
 import kotlinx.serialization.test.*
 import kotlin.test.*
 
+@Ignore // todo: SerialDesc.hashCode works incorrectly for these guys
```
qwwdfsad, line 16: Are you going to address TODO in this PR?

formats/json/commonTest/src/kotlinx/serialization/features/JsonAlternativeNamesTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+@Serializable
+data class WithNames(@JsonNames(arrayOf("foo", "_foo")) val data: String)
```
qwwdfsad, line 15: Code style: classes are not within tests and tests not started with `test` prefix

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -101,6 +101,19 @@ internal open class StreamingJsonDecoder(
...
 
+    private fun SerialDescriptor.getJsonElementIndex(key: String): Int {
+        val index = this.getElementIndex(key)
+        if (!json.configuration.useAlternativeNames) return index
+        // Fast path, do not go through ConcurrentHashMap.get
+        // Note, it blocks ability to detect collisions between the primary name and alternate,
+        // but it eliminates a significant performance penalty (about -15% without this optimization)
+        if (index != UNKNOWN_NAME) return index
+        // slow path
+        val alternativeNamesMap =
```
qwwdfsad, line 112: We expect this map to be pretty small (1-4 entries).

So I see two more potential ways to do that lookup:
1) Do not cache anything, lookup annotation and then do a linear scan in the array.
2) Cache array of names from annotation and do a linear lookup

For our constraints one of these approaches might be slightly faster, could you please try it?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -101,6 +101,19 @@ internal open class StreamingJsonDecoder(
         }
     }
 
+    private fun SerialDescriptor.getJsonElementIndex(key: String): Int {
+        val index = this.getElementIndex(key)
+        if (!json.configuration.useAlternativeNames) return index
+        // Fast path, do not go through ConcurrentHashMap.get
+        // Note, it blocks ability to detect collisions between the primary name and alternate,
+        // but it eliminates a significant performance penalty (about -15% without this optimization)
+        if (index != UNKNOWN_NAME) return index
```
qwwdfsad, line 110: Performance nit: this if should go before `json.configuration.useAlternativeNames`

formats/json/commonMain/src/kotlinx/serialization/json/JsonNames.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    val builder = createMapForCache<String, Int>(elementsCount)
+    for (i in 0 until elementsCount) {
+        builder.putOrThrow(getElementName(i), i)
+        getElementAnnotations(i).filterIsInstance<JsonNames>().singleOrNull()?.names?.forEach { name ->
+            builder.putOrThrow(name, i)
+        }
+    }
+    return builder
```
qwwdfsad, line 47: Redundant memory pressure here, especially with `ignoreUnknownKeys` enabled.
Suggest replacing this line with `return if (builder.isEmpty()) emptyMap() else builder`

formats/json/commonMain/src/kotlinx/serialization/json/JsonNames.kt
```diff
@@ -0,0 +1,48 @@
...
+@ExperimentalSerializationApi
+public annotation class JsonNames(val names: Array<String>)
+
+@SharedImmutable
+internal val JsonAlternativeNamesKey = DescriptorSchemaCache.Key<Map<String, Int>>()
+
+@OptIn(ExperimentalSerializationApi::class)
+internal fun SerialDescriptor.buildAlternativeNamesMap(): Map<String, Int> {
+    fun MutableMap<String, Int>.putOrThrow(name: String, index: Int) {
+        check(name !in this) {
```
qwwdfsad, line 33: Shouldn't it be `SerializationException`?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonLexer.kt
```diff
@@ -422,15 +422,19 @@ internal class JsonLexer(private val source: String) {
         }
     }
 
-    fun skipElement() {
+    fun skipElement(allowLenientStrings: Boolean) {
```
qwwdfsad, line 425: Great catch with this function!

qwwdfsad (changes requested):

formats/json/commonTest/src/kotlinx/serialization/json/JsonCustomSerializersTest.kt
sandwwraith: I think I'll just disable useAlternativeNames flag for these tests

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
sandwwraith: Not sure if this would make any difference, since `if (!json.configuration.useAlternativeNames)` is a very predictable check, but ok

KevinBassaDevelopment:
> 
> 
> **TODO**: measure performance impact and decide whether this flag should be enabled by default (or should the feature be flagged at all)

Is there anyway I could help?
If extensive tests are needed that take more time, could the feature be added with default flag false, and then later on be included by default if the performance impact is negligible?

benchmark/src/jmh/kotlin/kotlinx/benchmarks/model/MacroTwitter.kt
```diff
@@ -92,6 +110,24 @@ data class Metadata(
     val iso_language_code: String
 )
 
+@Serializable
+data class TwitterReducedUser(
```
qwwdfsad, line 114: Probably the word you are looking for is "trimmed"

docs/json.md
```diff
@@ -151,6 +152,40 @@ Project(name=kotlinx.serialization)
 
 <!--- TEST -->
 
+### Alternative Json names
+
+It's not a rare case when JSON fields got renamed due to a schema version change or something else.
```
qwwdfsad, line 157:
```suggestion
It's not a rare case when JSON fields got renamed due to a schema version change.
```

docs/json.md
```diff
@@ -151,6 +152,40 @@ Project(name=kotlinx.serialization)
...
+Renaming JSON fields is available with [`@SerialName` annotation](basic-serialization.md#serial-field-names), but 
+such a renaming blocks ability to decode data with old name.
```
qwwdfsad, line 159:
```suggestion
such a renaming blocks ability to decode data with an old name. 
```

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -228,6 +232,14 @@ public class JsonBuilder internal constructor(configuration: JsonConfiguration)
      */
     public var allowSpecialFloatingPointValues: Boolean = configuration.allowSpecialFloatingPointValues
 
+    /**
+     * Switches whether Json instance make use of [JsonNames] annotation; enabled by default.
```
qwwdfsad:
```suggestion
     * Specifies whether Json instance makes use of [JsonNames] annotation; enabled by default.
```

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -228,6 +232,14 @@ public class JsonBuilder internal constructor(configuration: JsonConfiguration)
...
+     *
+     * Disabling this flag when one do not use [JsonNames] at all may sometimes result in better performance,
```
qwwdfsad, line 238:
```suggestion
     * Disabling this flag when one does not use [JsonNames] at all may sometimes result in better performance,
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -209,6 +209,20 @@ private open class JsonTreeDecoder(
...
+        val mainName = desc.getElementName(index)
+        if (!configuration.useAlternativeNames) return mainName
+        // Fast path, do not go through ConcurrentHashMap.get
+        // Note, it blocks ability to detect collisions between the primary name and alternate,
+        // but it eliminates a significant performance penalty (about -15% without this optimization)
+        if (mainName in value.keys) return mainName
+        // Slow path
+        val alternativeNamesMap =
+            json.schemaCache.getOrPut(desc, JsonAlternativeNamesKey, desc::buildAlternativeNamesMap)
+        val nameInObject = value.keys.find { it == mainName || alternativeNamesMap[it] == index }
```
qwwdfsad, line 222: `it == mainName` is redundant here, we've already checked that above

formats/json/commonTest/src/kotlinx/serialization/test/TestingFramework.kt
```diff
@@ -1,12 +1,12 @@
 /*
- * Copyright 2017-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
qwwdfsad, line 2: 2021 :)

qwwdfsad (approved):
Great job!

Please address a few minor stylistic comments and it's good to go

sandwwraith:
@KevinBassaDevelopment Thanks for suggestion! I've already did some performance evaluations, and as a result, we concluded that this flag can be safely set to 'true' by default.

Issue #195, opened 2018-08-15
Allow to always ignore nulls and empty collections when writing to JSON (regardless whether they are optional or not)


qwwdfsad:

sandwwraith:
#58 , isn't it?

qwwdfsad:
This one has JSON-only scope and doesn't affect optionality/default values

LouisCAD:
I can currently use the following for null values:
```kotlin
val json = Json(encodeDefaults = false)
```
That applies to all models though, no way to specify it on a per-property basis, and no support for empty collections.
Maybe the title should specify "per property" if that's the intention, or drop the reference to "nulls" otherwise?

anschnapp:
I guess...
```
val json = Json(encodeDefaults = false)
```
... is not for the exact same use case.
It would allow the ignore null values if they are the default value.
But it would also ignore all other default values which might not be null.

I'm currently searching for a solution where null values will not be part of the json, but I want to have other default values (like strings) be part of the json, so that they could be read from the consumer side.

FrancoSabadini:
Is there a workaround for this? (apart from `Json(encodeDefaults = false)`)

LouisCAD:
@FrancoSabadini Writing a custom serializer and applying it to the relevant properties. It's quite easy to do if you look at the doc, although it might be redundant if you have to apply it to many properties.

FrancoSabadini:
Thanks for the answer @LouisCAD. 

Yes, I have to add it to quite a few properties so it gets a bit annoying. Can you point me to the specific doc you are referring to? And what would this custom serializer look like?

matejdro:
`encodeDefaults` solution does not appear to apply to nulls:

```kotlin
@Serializable
data class Test(val value: String?)

fun main() {
    val json = Json(
        JsonConfiguration(encodeDefaults = false)
    )

    println(json.toJson(Test.serializer(), Test(null)))
}
```

prints out

```json
{"value":null}
```

Dominaezzz:
That needs to be,
```kotlin
@Serializable
data class Test(val value: String? = null)
```

matejdro:
Thanks, what about when passing in null explicitly?

Dominaezzz:
If you only pass null explicitly, then it's not a default, so it is encoded. But by telling `kotlinx.serialisation` that the default is `null`, it knows not to encode it when `encodeDefaults = false` is specified.

matejdro:
[...]

PR #1376, opened 2021-03-17
Don't encode nulls

Vacxe:
Allow to exclude object from encoded string if it's `null`

`encodeNulls` by default `true` to support backward compatibility 

See: #195

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -233,6 +233,12 @@ public class JsonBuilder internal constructor(conf: JsonConf) {
      */
     public var serializersModule: SerializersModule = conf.serializersModule
 
+    /**
+     * Encode values if it's `null`
+     * `true` by default.
+     */
+    public var encodeNulls: Boolean = conf.encodeNulls
+
     @OptIn(ExperimentalSerializationApi::class)
     internal fun build(): JsonConf {
         if (useArrayPolymorphism) require(classDiscriminator == defaultDiscriminator) {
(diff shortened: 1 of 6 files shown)


qwwdfsad (changes requested):
With the proposed change, the most important invariant of the serialization framework is violated: now we cannot deserialize back what we've just serialized.

The issue, unfortunately, is more complex than that and that's why it's not yet addressed. We are not ready to accept this PR and break our own invariant

Vacxe:
@qwwdfsad thank for quick reply. You have a reasonable arguments to close this PR however:

* We have about 2-3 additional major libraries for serialization such as Gson and Moshi which also popular for usage. [All of them provide different strategies for serialization](https://github.com/Vacxe/json-parsers-branchmark)

For example Moshi do not serialized nulls values

* Moshi

    * Input object: <BR>
    ```PrimitiveDataClass(name=Object 770591202, int=-822616830, double=0.8621779982333868, float=0.33106828, boolean=true, optional=null)```
    * Serialized `String` <BR>
    ```{"name":"Object 770591202","int":-822616830,"double":0.8621779982333868,"float":0.33106828,"boolean":true}}```
    * Deserialize object <BR>
    ```PrimitiveDataClass(name=Object 770591202, int=-822616830, double=0.8621779982333868, float=0.33106828, boolean=true, optional=Optional String))```
    

* From time to time serialization is used for network calls. And it can affect some client-server logic due migration from one library to this one.

* This feature under configuration flag and can be enabled if it's required.

* Removing null values (if client-server contract accepting this) can significantly reduce network call weight.

* Anyway, this is not a main strategy for serialization and can be applied if required.

---

I honestly understanding your path of light but in real world we have to many casts of shadows. 


Anyway I respecting your decision

qwwdfsad:
Thanks for your input! Your concerns are indeed valid, we are going to address this issue right after the next release (including deserialization of such data back as well)

Vacxe:
@qwwdfsad Cheers. Look forward to solving this problem.

PR #1377, opened 2021-03-17
Remove redundant visibility operators

Vacxe:
Just cleanup


qwwdfsad (changes requested):
Thanks, but we are considering making this class public in the next release, then these modifiers won't be redundant. Let's avoid potential ping-pong change here, especially when it's internal class

PR #1378, opened 2021-03-17
Get rid of Bintray leftovers

qwwdfsad:


gradle/publishing.gradle
```diff
@@ -29,8 +29,6 @@ if (!isBom) {
 task emptyJar(type: Jar) {
 }
 
-def bintrayUploadFlag = System.getenv("libs.bintray.upload") != null
```
shanshin, line 32: Perhaps it is worth updating the comment `Configures publishing of Maven artifacts to Bintray`

shanshin (approved):

build.gradle
```diff
@@ -139,21 +127,10 @@ allprojects {
     }
 
     repositories {
-        mavenLocal()
```
sandwwraith, line 142: mavenLocal is used when you build library with a locally built compiler, so I suggest to leave it at the end of the list

build.gradle
```diff
@@ -139,21 +127,10 @@ allprojects {
...
         mavenCentral()
-        maven { url "https://dl.bintray.com/kotlin/kotlinx" }
-        // Future replacement for kotlin-dev, with cache redirector
-        maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
```
sandwwraith, line 146: Where do we get dev versions then?

PR #1380, opened 2021-03-17
Rename JsonConf to JsonConfiguration

qwwdfsad:
* Previously it was named that way to prevent clash with deprecated separate JsonConfiguration
    * Also we may want to make it public in the future


sandwwraith:
Update api dump

qwwdfsad:
Oh wow, I double-checked before push that it's not published API and is not exposed in public.
Apparently, it's the problem in BCV: https://github.com/Kotlin/binary-compatibility-validator/issues/19

Anyway, updated dump

sandwwraith (approved):

PR #1397, opened 2021-04-02
Fix serializer lookup by KType for third party generics

mvdbos:
Fix serializer lookup by KType for third party generics

This lookup used to fail, because `SerializersModule.serializerByKTypeImpl()`
never tried `getContextual()` after constructSerializerForGivenTypeArgs() fails
to find a serializer (and that *will* fail, because `ThirdPartyBox` is not
annotated as @Serializable, since it is a third party class).

This does work when doing the look up by Java type with
`serializerByJavaTypeImpl.serializerByJavaTypeImpe()`.

The difference is that the former does this:

```kotlin
rootClass.constructSerializerForGivenTypeArgs(*serializers.toTypedArray())
```

And the latter does this:

```kotlin
rootClass.constructSerializerForGivenTypeArgs(*serializers.toTypedArray())
    ?: reflectiveOrContextual(rootClass)
```

This commit adds the call to `reflectiveOrContextual()` to
`SerializersModule.serializerByKTypeImpl()`, which fixes it for all platforms.

The accompanying test illustrates the scenario where this used to fail and is now succesful. Both for JVM and other platforms.

### DIFF
--- a/core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
@@ -1,0 +1,45 @@
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.modules.*
+import kotlin.test.*
+
+open class ThirdPartyGenericsTest {
+    // This is a 3rd party class that we can't annotate as @Serializable
+    data class ThirdPartyBox<T>(val contents: T)
+
+    // This is the item that we put in the ThirdPartyBox, we control it, so can annotate it
+    @Serializable
+    data class Item(val name: String)
+
+    // The serializer for the ThirdPartyBox<T>
+    class BoxSerializer<T>(dataSerializer: KSerializer<T>) : KSerializer<ThirdPartyBox<T>> {
+        @Serializable
+        data class BoxSurrogate<T>(val contents: T)
+
+        private val strategy = BoxSurrogate.serializer(dataSerializer)
+        override val descriptor: SerialDescriptor = strategy.descriptor
+
+        override fun deserialize(decoder: Decoder): ThirdPartyBox<T> {
+            return ThirdPartyBox(decoder.decodeSerializableValue(strategy).contents)
+        }
+
+        override fun serialize(encoder: Encoder, value: ThirdPartyBox<T>) {
+            encoder.encodeSerializableValue(strategy, BoxSurrogate(value.contents))
+        }
+    }
+
+    // Register contextual serializer for ThirdPartyBox<Item>
+    protected val boxWithItemSerializer = BoxSerializer(Item.serializer())
+    protected val serializersModule = SerializersModule {
+        contextual(boxWithItemSerializer)
+    }
+
+    @Test
+    fun testSurrogateSerializerFoundForGenericWithKotlinType() {
+        val serializer = serializersModule.serializer<ThirdPartyBox<Item>>()
+        assertEquals(boxWithItemSerializer.descriptor, serializer.descriptor)
+    }
+}
--- a/core/jvmTest/src/kotlinx/serialization/features/JvmThirdPartyGenericsTest.kt
+++ b/core/jvmTest/src/kotlinx/serialization/features/JvmThirdPartyGenericsTest.kt
@@ -1,0 +1,13 @@
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
+import kotlin.test.*
+
+class JvmThirdPartyGenericsTest : ThirdPartyGenericsTest() {
+    @Test
+    fun testSurrogateSerializerFoundForGenericWithJavaType() {
+        val filledBox = ThirdPartyBox(contents = Item("Foo"))
+        val serializer = serializersModule.serializer(filledBox::class.java)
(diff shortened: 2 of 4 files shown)


core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
```diff
@@ -0,0 +1,70 @@
+package kotlinx.serialization.features
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
```
sandwwraith, line 8: We usually prefer star imports in the code style

core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.serialization.modules.contextual
+import kotlinx.serialization.serializer
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+open class ThirdPartyGenericsTest {
```
sandwwraith, line 14: Why this class needs to be `open`?

core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    // Register contextual serializer for ThirdPartyBox<Item>
+    protected val boxWithItemSerializer = BoxSerializer(Item.serializer())
+    protected val serializersModule = SerializersModule {
+        contextual(boxWithItemSerializer)
+    }
+
+    @Test
+    fun testSurrogateSerializerFoundForGenericWithKotlinType() {
+        /*
+         * This used to fail, because SerializersModule.serializerByKTypeImpl() never tried `getContextual()` after
```
sandwwraith, line 48: This comment is kinda excessive, because a) you explained bug in the commit message b) 'Fixed by changing `SerializersModule.serializerByKTypeImpl()` line 117' is literally the change in the commit. So I suggest to remove it.

core/jvmTest/src/kotlinx/serialization/features/JvmThirdPartyGenericsTest.kt
```diff
@@ -0,0 +1,19 @@
+package kotlinx.serialization.features
+
+import kotlinx.serialization.serializer
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+open class JvmThirdPartyGenericsTest : ThirdPartyGenericsTest() {
```
sandwwraith, line 7: I think there's no need in inheritance for nested classes to be accessible. Also, `open` modifier here for `JvmThirdPartyGenericsTest ` is definitely redundant

sandwwraith (commented):
Great job, thanks for the PR! I do not have any objections besides mentioned code style issues

core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
mvdbos: Because `JvmThirdPartyGenericsTest` extends it.

core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
mvdbos: Yes, indeed. Left it in on purpose if any discussion would have followed. Will remove now.

core/jvmTest/src/kotlinx/serialization/features/JvmThirdPartyGenericsTest.kt
mvdbos: > I think there's no need in inheritance for nested classes to be accessible.

This is true, but it also accesses instance vals `boxWithItemSerializer` and `serializersModule`. Thought inheritance was better than repetition in this case, but also happy to copy-paste these vals to this test if you prefer.

Remove `open`.

core/commonTest/src/kotlinx/serialization/features/ThirdPartyGenericsTest.kt
mvdbos: You want this to be a `kotlinx.serialization.*` import? So if there is more than one import from a package, we use '*'? Or one level lower, e.g. `kotlinx.serialization.encoding.*`? Perhaps good to indicate the threshold in CONTRIBUTING.md? 

Out of interest, why this deviation from the ktlint ruleset (https://ktlint.github.io/#rule-import)?

sandwwraith: We follow official style guide (https://kotlinlang.org/docs/coding-conventions.html) which does not explicitly endorse or prohibit star imports. However, they're proven to be useful because one usually want to import extensions or dependent classes alongside the class itself; and it also looks very neat when library fits in 3-4 packages. For example, Ktor style guide explicitly recommends them for this reason: https://ktor.io/docs/code-style.html#with-star-imports

core/jvmTest/src/kotlinx/serialization/features/JvmThirdPartyGenericsTest.kt
sandwwraith: OK, good point

PR #1403, opened 2021-04-05
Fixed loss of precision in JS test for current kotlin master branch

shanshin:
Error in the current master may be corrected later.

"${Long.MIN_VALUE}" gives "-9223372036854776000" but not "-9223372036854775808"


sandwwraith:
This is a Kotlin compiler bug? Do you have a ticket for it?

shanshin:
> This is a Kotlin compiler bug? Do you have a ticket for it?

https://youtrack.jetbrains.com/issue/KT-39891

sandwwraith (approved):

PR #1406, opened 2021-04-06
Registering multiple contextual serializers for generic class fails

mvdbos:
This PR is meant for discussion, not merging.

The use case here is serializing a third party generic class with different contents: ThirdPartyBox<T>.
As is show in the manual, the BoxSerializer in this test is capable of serializing any ThirdPartyBox and takes as its constructor arg the serializer for T.

Different from the examples for generics in the manual, we have to use `contextual` here, because we don't control ThirdPartyBox<T>.
Unfortunately, adding another contextual serializer for the generic BoxSerializer fails, because SerializersModuleBuilders.registerSerializer() only looks at the toplevel type of generic types to determine if a serializer was already registered.

This is the assertion, see the code for the full setup:
```kotlin
        assertFailsWith(IllegalArgumentException::class) {
            SerializersModule {
                contextual(BoxSerializer(Item.serializer()))
                contextual(BoxSerializer(AnotherItem.serializer()))
            }

        }.also { it.message!!.contains("Serializer for ${ThirdPartyBox::class} already registered in this module") }
```

As solution could be to add type parameter info as a differentiator to the map, but this would not be possible for JavaScript,
because KClass.typeParameters is not available in JavaScript. How to solve this?

### DIFF
--- a/core/jvmTest/src/kotlinx/serialization/features/MultipleContextualGenericsTest.kt
+++ b/core/jvmTest/src/kotlinx/serialization/features/MultipleContextualGenericsTest.kt
@@ -1,0 +1,64 @@
+package kotlinx.serialization.features
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.contextual
+import kotlin.test.Test
+import kotlin.test.assertFailsWith
+
+open class MultipleContextualGenericsTest {
+    // This is a 3rd party class that we can't annotate as @Serializable
+    data class ThirdPartyBox<T>(val contents: T)
+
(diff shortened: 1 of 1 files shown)


sandwwraith:
Hi,
We have designed a solution for this, but it is not finished yet. If you take a look at the `ContextualSerializer` (https://github.com/Kotlin/kotlinx.serialization/blob/b275a2d013f502de4aaedbc5ea849d10fbca5255/core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt#L45), you find out that it does have additional `typeParametersSerializers` argument, which is dedicated for, surprise, type arguments of contextually serializable class. As for now plugin provides these arguments, but they're not used. Using this arguments, it is possible to add special function in module builder, that would allow to construct serializer, like that:

```kotlin
SerializersModule {
    contextual<Box> { typeArgs -> BoxSerializer(typeArgs[0]) }
}
```

mvdbos:
Great, that would help. Any idea about ETA?

sandwwraith:
@mvdbos No particular guarantees but I hope it would come in one next or two releases

sandwwraith:
See here: #1416

mvdbos:
@sandwwraith Great! Thanks

PR #1409, opened 2021-04-08
Make JsonConfiguration public and expose it for custom serializers

### LINKED ISSUE #1361: Make `kotlinx.serialization.json.Json.configuration` public
Hi,

I want to write a custom serializer, where I also need to access the configuration of the `Json` class . But it is marked a `internal` 

```kotlin
public sealed class Json(internal val configuration: JsonConf) : StringFormat {
```

So I have to use this hack to access the configuration

```kotlin
        val decoder: JsonDecoder = ...
        var classDiscriminatorConfig: String? = null
        Json(decoder.json) {
            classDiscriminatorConfig = classDiscriminator // the value I want to know
        }
```

So I would like to change the visibility of `kotlinx.serialization.json.Json.configuration` to `public`:

```kotlin
    public sealed class Json(val configuration: JsonConf) : StringFormat {
```

and also ` kotlinx.serialization.json.internal.JsonConf`  to be to `public` (and moving to package `kotlinx.serialization.json`)
```kotlin
package kotlinx.serialization.json
...
data class JsonConf
```

qwwdfsad:
Fixes #1361


formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -45,9 +45,12 @@ import kotlinx.serialization.modules.*
  *  // Deserialize from string to JSON tree, JSON-specific
  *  val deserializedToTree: JsonElement = json.parseJsonElement(stringOutput)
  * ```
+ *
+ * Json instance also exposes its [configuration] that can be used in custom serializers
+ * that rely on [JsonDecoder] and [JsonEncoder] for customizeable behaviour.
```
sandwwraith, line 50:
```suggestion
 * that rely on [JsonDecoder] and [JsonEncoder] for customizable behaviour.
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Configuration of the current [Json] instance available through [Json.configuration]
+ * and configured with [JsonBuilder] constructor.
+ *
+ * Can be used for debug purposes and for custom Json-specific serializers
+ * via [JsonEncoder] and [JsonDecoder].
+ *
+ * Standalone configuration object is meaningless and can nor be used outside of the
+ * [Json], neither new [Json] instance can be created from it.
```
sandwwraith, line 14: Maybe it makes sense to mention here that one can create Json instance from other Json instance

formats/json/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+@ExperimentalSerializationApi
+public class JsonConfiguration internal constructor(
+    public val encodeDefaults: Boolean = false,
+    public val ignoreUnknownKeys: Boolean = false,
+    public val isLenient: Boolean = false,
+    public val allowStructuredMapKeys: Boolean = false,
+    public val prettyPrint: Boolean = false,
+    @ExperimentalSerializationApi
+    public val prettyPrintIndent: String = "    ",
```
sandwwraith, line 24: Eventually we'll need to make a decision about this one

sandwwraith (approved):

formats/json/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Can be used for debug purposes and for custom Json-specific serializers
+ * via [JsonEncoder] and [JsonDecoder].
+ *
+ * Standalone configuration object is meaningless and can nor be used outside of the
+ * [Json], neither new [Json] instance can be created from it.
+ */
+@ExperimentalSerializationApi
+public class JsonConfiguration internal constructor(
+    public val encodeDefaults: Boolean = false,
```
sandwwraith, line 18: I think we need to express somehow that documentation for this properties is available in JsonBuilder

PR #1416, opened 2021-04-14
Support contextual serialization of generic classes

### LINKED ISSUE #1407: Functionality for registering and finding contextual serializers for third party generic classes is incomplete.
These might possibly be bugs, but determining how to solve them requires a design discussion.

### Summary

The use case here is serializing a third party generic class with different contents: ThirdPartyBox<T>.
As is show in the manual, the BoxSerializer in this test is capable of serializing any ThirdPartyBox and takes as its constructor arg the serializer for T.

Different from the examples for generics in the manual, we have to use contextual here, because we don't control ThirdPartyBox.

### Registering bug

Unfortunately, *registering* another contextual serializer for the generic BoxSerializer fails, because SerializersModuleBuilders.registerSerializer() only looks at the toplevel type of generic types to determine if a serializer was already registered.

See #1406 for details. That PR contains only a test showing the issue.

### Finding bug

Finding the serializer for a  (generic) third party type by KType fails, because `SerializersModule.serializerByKTypeImpl()`
never tries `getContextual()` after `constructSerializerForGivenTypeArgs()` fails to find a serializer (and that will fail, because ThirdPartyBox is not annotated as @serializable, since it is a third party class).

This does work when doing the look up by Java type with `serializerByJavaTypeImpl.serializerByJavaTypeImpl()`.

See #1397 for details. That PR contains a solution to be discussed, and tests to support it.

sandwwraith:
Via special lambda provider in SerializersModule

Fixes #1407


core/commonMain/src/kotlinx/serialization/modules/SerializersModuleCollector.kt
```diff
@@ -24,7 +24,14 @@ public interface SerializersModuleCollector {
     /**
      * Accept a serializer, associated with [kClass] for contextual serialization.
      */
-    public fun <T : Any> contextual(kClass: KClass<T>, serializer: KSerializer<T>)
+    public fun <T : Any> contextual(kClass: KClass<T>, serializer: KSerializer<T>): Unit =
+        contextual(kClass) { _ -> serializer }
```
whyoleg, line 28: why default implementation is needed here, as looks like it will be always overridden?

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -114,15 +114,16 @@ private fun SerializersModule.builtinSerializer(
...
+            rootClass.constructSerializerForGivenTypeArgs(*args)
+                ?: reflectiveOrContextual(rootClass, serializers)
         }
     }
 }
 
 @OptIn(ExperimentalSerializationApi::class)
-internal fun <T : Any> SerializersModule.reflectiveOrContextual(kClass: KClass<T>): KSerializer<T>? {
-    return kClass.serializerOrNull() ?: getContextual(kClass)
+internal fun <T : Any> SerializersModule.reflectiveOrContextual(kClass: KClass<T>, typeParameterSerializers: List<KSerializer<Any?>>): KSerializer<T>? {
```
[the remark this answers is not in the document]
sandwwraith: Indeed, thx

core/commonMain/src/kotlinx/serialization/modules/SerializersModuleCollector.kt
sandwwraith: Nope, it is not overriden, for example, in PolymorphismValidator (since it does not use contextual at all)

core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
```diff
@@ -42,11 +42,11 @@ import kotlin.reflect.*
 public class ContextualSerializer<T : Any>(
     private val serializableClass: KClass<T>,
     private val fallbackSerializer: KSerializer<T>?,
-    private val typeParametersSerializers: Array<KSerializer<*>>
+    private val typeArgumentsSerializers: Array<KSerializer<*>>
 ) : KSerializer<T> {
 
     private fun serializer(serializersModule: SerializersModule): KSerializer<T> =
-        serializersModule.getContextual(serializableClass) ?: fallbackSerializer ?: serializableClass.serializerNotRegistered()
+        serializersModule.getContextual(serializableClass, typeArgumentsSerializers.asList()) ?: fallbackSerializer ?: serializableClass.serializerNotRegistered()
```
qwwdfsad, line 49: [nit] probably worth transforming array to list in the ctor

core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
```diff
@@ -42,11 +42,11 @@ import kotlin.reflect.*
...
 
     // Used from auto-generated code
```
qwwdfsad, line 51: Could you please suppress this warning?

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
```diff
@@ -28,7 +29,21 @@ public sealed class SerializersModule {
...
     @ExperimentalSerializationApi
-    public abstract fun <T : Any> getContextual(kclass: KClass<T>): KSerializer<T>?
+    @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
+    @Deprecated(
+        "Deprecated in favor of overload with default parameter",
+        ReplaceWith("getContextual(kclass)"),
+        DeprecationLevel.ERROR
+    )
+    @LowPriorityInOverloadResolution
+    public fun <T : Any> getContextual(kclass: KClass<T>): KSerializer<T>? =
```
qwwdfsad, line 39: Why not making it HIDDEN straight away? Then `LowPriorityInOverloadResolution` will be redundant as well

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
```diff
@@ -28,7 +29,21 @@ public sealed class SerializersModule {
...
+        getContextual(kclass, emptyList())
+
+    @ExperimentalSerializationApi
+    public abstract fun <T : Any> getContextual(
```
qwwdfsad, line 43: Please add KDoc (also you can remove it from deprecated method)

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
```diff
@@ -74,11 +89,19 @@ public operator fun SerializersModule.plus(other: SerializersModule): Serializer
...
+@OptIn(ExperimentalSerializationApi::class)
 public infix fun SerializersModule.overwriteWith(other: SerializersModule): SerializersModule = SerializersModule {
     include(this@overwriteWith)
     other.dumpTo(object : SerializersModuleCollector {
         override fun <T : Any> contextual(kClass: KClass<T>, serializer: KSerializer<T>) {
-            registerSerializer(kClass, serializer, allowOverwrite = true)
+            registerSerializer(kClass, ContextualProvider.Argless(serializer), allowOverwrite = true)
+        }
+
+        override fun <T : Any> contextual(
```
qwwdfsad, line 100: missing KDoc

core/commonMain/src/kotlinx/serialization/modules/SerializersModuleCollector.kt
qwwdfsad, line 28:
```suggestion
        contextual(kClass) { serializer }
```

qwwdfsad (changes requested):
Great addition, looks great!

Let's fill in our KDocs and slightly update `serializers.md` (no need to write a full-blown section or example, just mentioning is enough) and it's good to go!

core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
sandwwraith: About unused constructor?

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
sandwwraith: I think it's possible

core/commonMain/src/kotlinx/serialization/modules/SerializersModuleBuilders.kt
```diff
@@ -37,18 +38,43 @@ public inline fun SerializersModule(builderAction: SerializersModuleBuilder.() -
...
+     * val module = SerializersModule {
+     *   // args[0] contains Int.serializer() or String.serializer(), depending on the property
+     *   contextual(Box::class) { args -> BoxSerializer(args[0]) }
+     * }
+     * ```
+     *
+     * Throws [SerializationException] if a module already has provider or serializer associated with a [kClass].
+     * To overwrite an already registered serializer, [SerializersModule.overwriteWith] can be used.
+     */
+    public override fun <T : Any> contextual(
```
sandwwraith, line 74: Shouldn't we mark this as experimental?

core/commonMain/src/kotlinx/serialization/ContextualSerializer.kt
qwwdfsad: Right

core/commonMain/src/kotlinx/serialization/modules/SerializersModuleBuilders.kt
qwwdfsad: It seems that it's okay to add it as stable in 1.2.0

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
```diff
@@ -23,12 +23,28 @@ import kotlin.reflect.*
  */
 public sealed class SerializersModule {
 
+    @ExperimentalSerializationApi
+    @Deprecated(
+        "Deprecated in favor of overload with default parameter",
+        ReplaceWith("getContextual(kclass)"),
+        DeprecationLevel.HIDDEN
+    )
```
qwwdfsad, line 31:
```suggestion
    ) // Was stable since 1.0.0, HIDDEN in 1.2.0 in a backwards-compatible manner
```

qwwdfsad (approved):
👍

core/commonMain/src/kotlinx/serialization/modules/SerializersModule.kt
sandwwraith: Btw, these functions (in SerializersModule, not the builder) are still marked as experimental. Should we lift this?

PR #1425, opened 2021-04-20
Class discriminator in subclasses

aikrikunov95:
Note that it's absent in subclasses instances.


sandwwraith:
There's a special section in this document called 'static types' that explains that problem. But I agree we can add a link to it: https://github.com/Kotlin/kotlinx.serialization/pull/1437

sandwwraith:
Closing in favor of #1492

PR #1430, opened 2021-04-22
Added support for nullable value, nested and empty collections in protobuf

shanshin:


formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufTaggedEncoder.kt
```diff
@@ -27,7 +27,11 @@ internal abstract class ProtobufTaggedEncoder : ProtobufTaggedBase(), Encoder, C
 
     protected open fun encodeTaggedInline(tag: ProtoDesc, inlineDescriptor: SerialDescriptor): Encoder = this.apply { pushTag(tag) }
 
-    public final override fun encodeNull(): Unit = encodeTaggedNull()
+    public final override fun encodeNull() {
+        if (!nullIsAcceptable) {
+            throw SerializationException("'null' is not supported in ProtoBuf")
```
sandwwraith, line 32: Probably we need to pass additional information here, tag or descriptor, to make an exception easier to understand e.g. 'null is not supported for property X because it has default value/nested collection/etc'

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufEncoding.kt
```diff
@@ -177,3 +181,18 @@ private class RepeatedEncoder(
...
 }
+
+private class NestedRepeatedEncoder(
+    proto: ProtoBuf,
+    @JvmField val writer: ProtobufWriter,
+    @JvmField val curTag: ProtoDesc,
+    descriptor: SerialDescriptor,
+    @JvmField val stream: ByteArrayOutput = ByteArrayOutput()
+) : ProtobufEncoder(proto, ProtobufWriter(stream), descriptor) {
+    // all elements are always has id = 1
```
sandwwraith, line 192:
```suggestion
    // all elements always have id = 1
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -97,7 +127,18 @@ internal open class ProtobufDecoder(
 
     override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
         return when (descriptor.kind) {
-            StructureKind.LIST -> RepeatedDecoder(proto, reader, currentTagOrDefault, descriptor)
+            StructureKind.LIST -> {
+                val tag = currentTagOrDefault
+                return if (tag != MISSING_TAG && this.descriptor != descriptor && this.descriptor.kind == StructureKind.LIST) {
```
sandwwraith, line 132: I think `this.descriptor.kind == StructureKind.LIST` the most important and rare clause so it should come first. Also, a comment e.g. '// decoding nested list' would explain things better, because it's hard to understand why this `if` needed at all

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -97,7 +127,18 @@ internal open class ProtobufDecoder(
...
+                    val reader = makeDelimited(reader, tag)
+                    // repeated decoder expects the first tag to be read already
+                    reader.readTag()
+                    // all elements are always has id = 1
```
sandwwraith, line 136:
```suggestion
                    // all elements always have id = 1
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -16,6 +16,10 @@ import kotlinx.serialization.modules.*
 import kotlinx.serialization.protobuf.*
 import kotlin.jvm.*
 
+private const val FILLED_OR_DEFAULT: Byte = 0
```
sandwwraith, line 19: I constantly think that this is a mistypo of 'FIELD'. Maybe we need a better name, e.g. PRESENT_OR DEFAULT

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -28,10 +32,36 @@ internal open class ProtobufDecoder(
...
+    private var currentAbsenceMode: Byte = FILLED_OR_DEFAULT
+
     init {
         populateCache(descriptor)
     }
 
+    private fun createAbsenceCache(descriptor: SerialDescriptor): ByteArray {
+        val elements = descriptor.elementsCount
+        val cache = ByteArray(elements)
+        if (elements < 32) {
```
sandwwraith, line 45: What if there are more than 32 elements?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -204,16 +245,28 @@ internal open class ProtobufDecoder(
         while (true) {
             val protoId = reader.readTag()
             if (protoId == -1) { // EOF
+                for (i in 0 until descriptor.elementsCount) {
```
sandwwraith, line 248: I'm afraid that this might fall into the 'accidentally quadratic' case. Have you tried to use a bit mask in Int as a replacement for FILLED? We still would need an array thought (because there are 3 possible states), but we wouldn't do a linear lookup every time, `Int.countLeadingZeroBits` or smth like that would be sufficient

shanshin: Another way to implement it are two arrays with bit values (final absence reading mark and null/empty collection mark)

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -28,10 +28,46 @@ internal open class ProtobufDecoder(
     private var indexCache: IntArray? = null
     private var sparseIndexCache: MutableMap<Int, Int>? = null
 
+    /*
+    Element decoding marks from given bytes.
+    The element number is the same as the bit position.
+    Marks for the lowest 64 elements always stores in single Long value, higher elements stores in long array.
```
sandwwraith, line 34:
```suggestion
    Marks for the lowest 64 elements are always stored in a single Long value, higher elements stores in long array.
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -200,20 +247,98 @@ internal open class ProtobufDecoder(
...
     override fun SerialDescriptor.getTag(index: Int) = extractParameters(index)
 
+    private fun findUnreadElementIndex(): Int {
+        val elementsCount = descriptor.elementsCount
+        while (lowerReadMark != -1L) {
+            val index = lowerReadMark.inv().countTrailingZeroBits()
+            lowerReadMark = lowerReadMark or (1L shl index)
+
+            if (!descriptor.isElementOptional(index)) {
+                val elementDescriptor = descriptor.getElementDescriptor(index)
```
sandwwraith, line 257: Is this possible to evaluate these conditions beforehand? So we won't iterate the whole bitmask every time

Probably we can mark every non-optional and nullable(or collection) field with one bits and others with zeroes in `prepareReadMarks` (read values would be zero too). Then it's possible to use countLeading/TrailingZeroBits right away without need to retrieve element descriptor

Although I'm not sure if such optimization would really make the difference in benchmarks. If you have strong objections, feel free to leave this as is

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -200,20 +247,98 @@ internal open class ProtobufDecoder(
...
+        if (elementsCount > Long.SIZE_BITS) {
+            val higherMarks = highReadMarks!!
+
+            for (slot in higherMarks.indices) {
+                // (slot + 1) because first element in high marks has index 64
+                val slotOffset = (slot + 1) * Long.SIZE_BITS
+                // store in a variable so as not to frequently use the array
+                var mark = higherMarks[slot]
+
+                while (mark != -1L) {
```
sandwwraith, line 277: This loop can be extracted in a function as the previous one

shanshin: This loop differs in that it contains `higherMarks[slot] = mark` statement and I have not found a way to write simple and fast function that returns an indication that an absence value was found and mark value after loop.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
shanshin: Evaluate these conditions beforehand will force to always read descriptor's elements properties - this will slow down the execution for cases with all required fields.

sandwwraith: `getElementDescriptor` is a simple O(1) array access. But ok, note taken

sandwwraith (approved):

Issue #1384, opened 2021-03-23
Pass additional arguments into custom serializer


KevinBassaDevelopment:
I want to be able to parse data that can have multiple types, and multiple names into 1 data model.

1. Example JSON:
`{ "doseNumberInt" = 5 }`
or
`{ "doseNumberMyDose" = { myDose : { value = "5" } } }`
or
`{ "doseNumberString" = { myDose = "5" }`

2. The model itself has just 1 property
`var doseNumber: Type?`

How I would imaging a possible solution:

```Type
object MultiTypeSerializer : JsonContentPolymorphicSerializer<Type>(Type::class) {

    override fun selectDeserializer(content: JsonElement): DeserializationStrategy<out Type> {
        // How to: access variable name and additional properties here (e.g. multiple SerialNames)
        // How to: define multiple serial descriptors here?
        val actualType= todoJsonKey.replace(todoVariableName,"")  //e.g. "doseNumberMyDose".repalce("doseNumber","")
        return if (actualType.equals("String")) {
            StringTypeSerializer
        } else if (actualType.equals("MyDose")
            MyDoseSerializer
        else 
          throw NotImplementedException("type  ${actualType} not implemented in MultiTypeSerializer ")
    }
}
```
To achieve this, it would be required to pass additional information from the data model into the serializer (currently only SerialName?)

Is this a usecase that makes sense? Is there a way to achieve this with the current possibilities?

shanshin:
Could you describe the problem being solved in more detail?
Polymorphic serializers generally do not have a stable common structure, therefore the polymorphic serializer descriptor does not describe fields.
The structure and set of attributes are controlled only by you in the `selectDeserializer` function. You can only "spy" the attribute names through the descriptors of the target serializers e.g. `MyDoseSerializer.descriptor`. However,  I recommend using global constants for serial names if they are needed in several places.

shanshin:
To show in more detail the capabilities of polymorphic deserialization, I will give the following example.
Model classes are
```
const val IntHolderAttributeName = "int32"
const val GeneratedOverriddenAttributeName = "overridden"

@Serializable(with = ParentSerializer::class)
open class Parent
data class IntHolder(val i: Int): Parent()
data class StringHolder(val text: String): Parent()
@Serializable
data class Generated(@SerialName(GeneratedOverriddenAttributeName) val a: Int, val b: String): Parent()
```

some of them have custom serializers
```
[...]

PR #1440, opened 2021-04-28
#1384 | Select deserializer with fieldname argument

KevinBassaDevelopment:
The feature `@JsonNames("name1";"name2")` introduced with 1.2.0 provides the new possibility to have different names for the same property.
To now also allow to distinguish the serialization strategy based on the element name, it must be passed on to custom `JsonContentPolymorphicSerializer`.

This resolved issue #1384

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonContentPolymorphicSerializer.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonContentPolymorphicSerializer.kt
@@ -34,7 +34,7 @@ import kotlin.reflect.*
  * For convenience, [serialize] method can lookup default serializer, but it is recommended to follow
  * standard procedure with [registering][SerializersModuleBuilder.polymorphic].
  *
- * Usage example:
+ * Usage example #1:
  * ```
  * interface Payment {
  *     val amount: String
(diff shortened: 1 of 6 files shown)


krottv:
@KevinBassaDevelopment 
Really nice feature. It would be more convenient (if even possible), if we could specify serializer right in the annotation. Do you think it is possible?

KevinBassaDevelopment:
> @KevinBassaDevelopment 
> Really nice feature. It would be more convenient (if even possible), if we could specify serializer right in the annotation. Do you think it is possible?

This request is not specific to this MR, but globally, so you could create a feature request if not already existing.

sandwwraith:
Hi! We've discussed your contribution and came to the thought that it maybe useful not only for `JsonContentPolymorphicSerializer`, but for `JsonTransformingSerializer` too.

Moreover, it may be useful to provide not only element name, but full JSON path. We do not have this functionality available in 1.2.1, but we have a prototype for it, so it may appear in the next release or two. So to avoid making double deprecations (we can't just change the signature of `selectDeserialzer`, we need to deprecate the old function and introduce new), it is reasonable to postpone this feature until JSON path is released.

However, if for some reason, JSON path won't be available after some time — do not hesitate to ping us again with this, we'll happily accept contribution only with `elementName`.

KevinBassaDevelopment:
Ok, thank you, is there already a ticket for this that I can watch?

sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/issues/1556

PR #1443, opened 2021-04-28
Added tests on calculated default value of properties depending on other properties

shanshin:
Tests for internal and external serializable classes. The default values for constructor properties, transient properties, body properties are checked

### DIFF
--- a/formats/json/build.gradle
+++ b/formats/json/build.gradle
@@ -24,3 +24,7 @@ kotlin {
         }
     }
 }
+
+compileTestKotlinJsLegacy {
+    exclude '**/PropertyInitializerTest.kt'
+}
--- a/formats/json/commonTest/src/kotlinx/serialization/features/PropertyInitializerTest.kt
+++ b/formats/json/commonTest/src/kotlinx/serialization/features/PropertyInitializerTest.kt
@@ -1,0 +1,105 @@
+@file:Suppress("MayBeConstant")
+
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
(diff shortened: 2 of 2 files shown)


formats/json/commonTest/src/kotlinx/serialization/features/PropertyExternalInitializerTest.kt
```diff
@@ -0,0 +1,49 @@
...
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.Json
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@Suppress("MemberVisibilityCanBePrivate", "unused", "ComplexRedundantLet")
+class PropertyExternalInitializerTest {
+    data class ExternalClass(
```
sandwwraith, line 12: I think both these classes can be in the same file, I don't see any objections

shanshin: Classes are separated by files to disable the compilation of external serializers in Legacy JS. Disable all initializer tests for the legacy JS?

sandwwraith: Ah, OK then, leave them as is

shanshin: Leave only one file - evaluated default value is unstable on legacy JS

sandwwraith (approved):

PR #1444, opened 2021-04-28
Added tests on caching serializer for `serializer()` function

shanshin:
Serializer cached for the object, abstract, and sealed classes.

Works since 1.5.20


sandwwraith (approved):
Let's merge when 1.5.20-M1 is available

PR #1452, opened 2021-05-05
Add Minecraft NBT to community formats

BenWoodworth:
Here's an implementation of Minecraft NBT format. It's a binary format, so I make it easy to serialize to/from ByteArrays. Also with okio sinks/sources, but only temporarily until kotlinx-io is back to being maintained, and is stable enough.

Requesting a new project for maven central/OSSRH has been slow, but I'll publish artifacts as soon as I get a chance :)


sandwwraith:
Great job, thanks!

PR #1470, opened 2021-05-11
Allow using value classes with primitive carriers as map keys without…

### LINKED ISSUE #1459: Value class cannot be used as JSON map key
**Describe the bug**
`kotlinx.serialization.json.internal.JsonEncodingException: Value of type 'MyId' can't be used in JSON as a key in the map. It should have either primitive or enum kind, but its kind is 'CLASS'.
    Use 'allowStructuredMapKeys = true' in 'Json {}' builder to convert such maps to [key1, value1, key2, value2,...] arrays.`

**To Reproduce**
```kotlin
@JvmInline
@Serializable
value class MyId(val id: String)

@Serializable
data class MyObject {
   val myMap: Map<MyId, Double>? = null
}

Json().decodeFromString(MyObject.serializer(), """{"myMap":{"hello", 1.0}}""")
```

**Expected behavior**
Since value classes are supposed to be (de)serialized by kotlinx.serialization as their underlying value, I would expect `MyId` to be usable anywhere that `String` is normally usable, such as the keys in a map.

Workaround is to use a serializer that mimics what kotlinx.serialization already does for value classes, but explicitly makes it a "primitive", such as:
```kotlin
object MyIdSerializer: KSerializer<MyId> {
    override val descriptor = PrimitiveSerialDescriptor("MyId", PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: MyId) {
        encoder.encodeString(value.id)
    }
    override fun deserialize(decoder: Decoder): MyId {
        return MyId(decoder.decodeString())
    }
}
```

**Environment**
 - Kotlin version: [e.g. 1.5.0]
 - Library version: [e.g. 1.2.0]
 - Kotlin platforms: JVM, JS, Native
 - Gradle version: 4.7

qwwdfsad:
… 'allowStructuredMapKeys'

Fixes #1459

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
@@ -32,7 +32,7 @@ internal inline fun <T, R1 : T, R2 : T> Json.selectMapMode(
     ifMap: () -> R1,
     ifList: () -> R2
 ): T {
-    val keyDescriptor = mapDescriptor.getElementDescriptor(0)
+    val keyDescriptor = mapDescriptor.getElementDescriptor(0).carrierDescriptor
     val keyKind = keyDescriptor.kind
     return if (keyKind is PrimitiveKind || keyKind == SerialKind.ENUM) {
         ifMap()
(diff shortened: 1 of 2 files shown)


qwwdfsad:
And here goes JS-specific compiler-related crash. PTAL :)

sandwwraith:
@qwwdfsad Just add `noLegacyJs` for your new tests. We do not live in IR-only bright world (yet!)

PR #1471, opened 2021-05-11
Fix regressions from the JSON encoder/decoder rewrite

qwwdfsad:


formats/json/commonTest/src/kotlinx/serialization/test/TestingFramework.kt
```diff
@@ -38,10 +38,10 @@ inline fun <reified T : Any> StringFormat.assertStringFormAndRestored(
     assertEquals(original, restored)
 }
 
-inline fun <reified T : Any> assertSerializedAndRestored(
+fun <T : Any> assertSerializedAndRestored(
```
sandwwraith, line 41: Why this change is needed here?

qwwdfsad: well, `inline reified` is not necessary for this function at all, but prevent IDEA debugger from working (AFAIR breakpoints didn't hit properly)

sandwwraith: Ok then

sandwwraith (approved):

sandwwraith:
Please preserve all commit messages with issue numbers

PR #1473, opened 2021-05-11
Support @JsonNames for enum values

### LINKED ISSUE #1458: @JsonNames does not work on enums
**Describe the bug**

The @JsonNames annotation to provide alternative names when decoding json does not work on enums.

**To Reproduce**

Create an enum with @JsonNames, for example:
```Kotlin
@Serializable
enum class SomeEnum {
    @JsonNames("someValue", "some_value", "test")
    SOME_VALUE
}
```

Trying to decode a json with one of the JsonNames will result in a SerializationException:
`kotlinx.serialization.SerializationException: com.zorabots.mqtt.model.filtering.SomeEnum does not contain element with name 'someValue'`

**Expected behavior**

"SOME_VALUE", "someValue", "some_value" or "test" should work, instead of only "SOME_VALUE".

**Environment**
 - Kotlin version: 1.5.0
 - Library version: 1.2.0
 - Kotlin platforms: JVM
 - Gradle version: 6.1.1

sandwwraith:
Fixes #1458


formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
```diff
@@ -79,7 +79,7 @@ private open class DynamicInput(
     }
 
     override fun decodeTaggedEnum(tag: String, enumDescriptor: SerialDescriptor): Int =
-        enumDescriptor.getElementIndexOrThrow(getByTag(tag) as String)
+        enumDescriptor.getJsonNameIndexOrThrow(json, getByTag(tag) as String)
```
qwwdfsad, line 82: Could you please also add a test for a dynamic decoder?

formats/json/commonTest/src/kotlinx/serialization/features/JsonAlternativeNamesTest.kt
```diff
@@ -26,10 +39,33 @@ class JsonAlternativeNamesTest : JsonTestBase() {
     private val inputString2 = """{"_foo":"foo"}"""
     private val json = Json { useAlternativeNames = true }
 
+    @Test
+    fun testEnumSupportsAlternativeNames() = noLegacyJs {
+        val input = """{"enumList":["VALUE_A", "someValue", "some_value", "VALUE_B"], "checkCoercion":"someValue"}"""
```
qwwdfsad, line 44: I see from the code that top-level enums are also properly supported (e.g. `Json.decodeFromString<AlternateEnumNames>("\"someValue\"")` works properly), but it would be nice to test it as well

qwwdfsad (approved):
Looks good modulo two small comments

formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
sandwwraith: It turns out DynamicDecoder does not support JsonNames at all, so I'll leave it for future PR

sandwwraith: And it also does not support coercing

qwwdfsad (approved):
Good to go

PR #1475, opened 2021-05-12
Handle EOF in skipElement correctly

### LINKED ISSUE #1474: JSON decodeFromString never terminates with some invalid inputs (JsonLexer.skipElement loops infinitely) => huge battery drain.
Duplicating from https://youtrack.jetbrains.com/issue/KT-46611

Using Kotlin 1.5.0 with org.jetbrains.kotlinx:kotlinx-serialization-json:1.2.0 (latest)

Given:
```
@Serializable
class Something

val json = Json { ignoreUnknownKeys = true }
```

.. this expression never terminates (with an exception) as it should, because the input is not valid JSON:
```
json.decodeFromString(Something.serializer(), """{"a":[{"b":[{"c":{}d",""e"":"}]}""")
```

Looks like the JsonLexer hits the end of input string, thus consumeNextToken() returns TC_EOF. After that, the while loop inside skipElement() keeps spinning forever.

sandwwraith:
Improve EOF exception in consumeStringLenient

Fixes #1474

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonLexer.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonLexer.kt
@@ -456,7 +456,7 @@ internal class JsonLexer(private val source: String) {
                     )
                     tokenStack.removeLast()
                 }
-                TC_EOF -> fail("Unexpected end of input, possibly due to malformed JSON")
+                TC_EOF -> fail("Unexpected end of input due to malformed JSON during ignoring unknown keys")
             }
             consumeNextToken()
             if (tokenStack.size == 0) return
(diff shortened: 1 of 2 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonLexer.kt
```diff
@@ -445,16 +446,17 @@ internal class JsonLexer(private val source: String) {
...
                 TC_END_OBJ -> {
                     if (tokenStack.last() != TC_BEGIN_OBJ) throw JsonDecodingException(
                         currentPosition,
                         "found } instead of ]",
                         source
                     )
-                    tokenStack.removeAt(tokenStack.size - 1)
+                    tokenStack.removeLast()
                 }
+                TC_EOF -> fail("Unexpected end of input, possibly due to malformed JSON")
```
qwwdfsad, line 459:
```suggestion
                TC_EOF -> fail("Unexpected end of input due to malformed JSON during ignoring unknown keys")
```

qwwdfsad (approved):
👍

PR #1479, opened 2021-05-13
Support for @JsonNames and coerceInputValues in Json.decodeFromDynamic

sandwwraith:


formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -63,3 +63,23 @@ internal fun SerialDescriptor.getJsonNameIndexOrThrow(json: Json, name: String):
         throw SerializationException("$serialName does not contain element with name '$name'")
     return index
 }
+
+@OptIn(ExperimentalSerializationApi::class)
+internal inline fun Json.shouldCoerceValue(
```
qwwdfsad, line 68: nit, but maybe it's better to name it `tryCoerceValue`?

All the call-sites of this method pass such values of lambdas that the part input is consumed. This is not something expected from the method with `shouldDoFoo(): Boolean` signature

formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
```diff
@@ -68,18 +69,45 @@ private open class DynamicInput(
...
     }
 
-    override fun decodeTaggedEnum(tag: String, enumDescriptor: SerialDescriptor): Int =
-        enumDescriptor.getJsonNameIndexOrThrow(json, getByTag(tag) as String)
+    private fun hasName(name: String) = value[name] !== undefined
+
+    override fun elementName(desc: SerialDescriptor, index: Int): String {
+        val mainName = desc.getElementName(index)
+        if (!json.configuration.useAlternativeNames) return mainName
+        // Fast path, do not go through ConcurrentHashMap.get
```
qwwdfsad, line 95: There is no concurrent map here

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
sandwwraith: DynamicDecoder and TreeDecoder do not consume anything, they support random access. But for Streaming parser this is true indeed.

qwwdfsad (approved):

PR #1492, opened 2021-05-17
Clarify docs section of static typing with polymorphism

elizarov:


sandwwraith (approved):
Please update the FQname in the failing test

PR #1494, opened 2021-05-17
Update JSON corpus test set

### LINKED ISSUE #1491: Perfomance drawback in big string encoding
**Describe the bug**
After update serializtion library from 1.0.0 to 1.2.1 I have noticed "infinite" encoding issue. When object contains long string with `"` char, encoding library stucks.
**To Reproduce**
https://github.com/MEJIOMAH17/kotlinx-serialization-perfomance-bug/blob/master/src/test/kotlin/PerformanceBug.kt 

**Expected behavior**
Works fine like 1.0.0

**Environment**
 - Kotlin version: 1.5.0
 - Library version: 1.2.1
 - Kotlin platforms: JVM

qwwdfsad:
* Test not only raw JSON-tree parser, but also streaming one and encoding
    * Update size in JsonStringBuilder slow-path to avoid excessive array-copies for large strings with escape symbols

Fixes #1491


qwwdfsad:
I've verified that this corpus also catches 1.2.0 issue and also re-run our performance tests and observed stable ~5% improvement

sandwwraith:
`SpecConformanceTest.kt` for some reason is parsed as a binary file, so I can't comment on it :(

I don't get the whole point of reusing `spec_cases` and putting them in Holder. Some of the original tests are supposed to fail, so why we do `assertSerializedAndRestored` on every one of them? Or is this just a string-escaping test that checks that everything works correctly if we put JSON in a string property?

Also, you can reuse `forEachSpecTest` in `testCompatibility`

sandwwraith:
Also `testSerializedGoFuzzCorpus` seems not to test actual parsing of things like arrays etc. It also puts whole JSON 'as a string', and then it is being escaped?

qwwdfsad:
>I don't get the whole point of reusing spec_cases and putting them in Holder.

`testCompatibility` stresses `parseToJsonElement` function, that is a path for `JsonTreeReader`. It doesn't touch `appendStringSlowPath`, for example. Wrapping it into a holder kicks in completely different code path via `StreamingJson*` classes

>Some of the original tests are supposed to fail, so why we do assertSerializedAndRestored on every one of them?

Not every string represents a valid `JsonElement`. But we must correctly encode (+escape) _any_ string as part of the `{"key":"$anyStringHere"}` expression

>Also testSerializedGoFuzzCorpus seems not to test actual parsing of things like arrays etc. It also puts the whole JSON 'as a string', and then it is being escaped?

Indeed, this one is much trickier to test, I don't know how to do it consistently and correctly

sandwwraith (approved):

Issue #1487, opened 2021-05-15
Exclude class discriminator


P1NG2WIN:
Right now in `Json { }` we cannot disable write `classDiscriminator`. In source code i found that class discriminator can be ignored if we change `writePolymorphic boolean` inside `encodePolymorphically` in `StreamingJsonEncoder`. Can i somehow change it without fully fork your library and change all layers? Maybe somehow with extension

Dominaezzz:
You can skip the writing of the `classDiscriminator` by using the serializer of the child class instead of the base class.

P1NG2WIN:
> You can skip the writing of the `classDiscriminator` by using the serializer of the child class instead of the base class.

Can you provide example?

sandwwraith:
```kotlin
val data = ChildClass("...")
Json.encodeToString<ParentClass>(data) // discriminator written
Json.encodeToString<ChildClass>(data) // discriminator omitted
```

P1NG2WIN:
In my case, this will not suit

sandwwraith:
What's your use case for excluding the discriminator? You won't be able to decode polymorphic data without it.

Whathecode:
Probably the other posted issue: https://github.com/Kotlin/kotlinx.serialization/issues/1486#issue-892136945

P1NG2WIN:
@sandwwraith I don't need to decode this json. He goes to the server. The server does not accept unknown fields. Now I forked the library and added a new field to the builder, so now I have two Json instances, one for the application that writes the class discriminator and the other for the api, which skips it when serializing.

sandwwraith:
I'll close this because #1247 already mentions this use-case — feel free to continue the discussion there

Issue #1247, opened 2020-12-10
Allow custom policy for adding a discriminator


joffrey-bion:
**What is your use-case and why do you need this feature?**

There has been several use cases mentioned in different issues that require to add the discriminator in more places, or avoid it in some places:

* for serialization-only use cases (such as sending data to an external API), we may not need discriminators at all. They bloat the payload at best, and can actually fail the requests at worst (https://github.com/Kotlin/kotlinx.serialization/issues/464)
* for interop with other deserialization libraries, possibly in more dynamic languages, we may want to have a discriminator for every type (https://github.com/Kotlin/kotlinx.serialization/issues/1128)
* even when using only Kotlinx Serialization itself, there are times when the deserializer has less information than the serializer about the type in the JSON, we may want to _always_ include the discriminator for subtypes of polymorphic types that are serializable (https://github.com/Kotlin/kotlinx.serialization/issues/1194). An example of that is sending events over a "pipe" (websocket or other). We know the concrete type at serialization time, but the deserializer at the other end doesn't know which one to use.

**Describe the solution you'd like**

It seems these use cases could be "easily" supported by having a configuration parameter for the inclusion of the discriminator, such as `discriminatorInclusionPolicy`, with the following values:

* `NEVER`: never include the discriminator
* `POLYMORPHIC_TYPES`: include the discriminator only when serializing a type that's polymorphic itself, the "parent" type (this is the current behaviour).
* `POLYMORPHIC_SUBTYPES`: include the discriminator for all instances of polymorphic types that are serializable, regardless of whether they are used as their parent or as themselves (`json.encodeToString<Parent>(SubType())` and `json.encodeToString<SubType>(SubType())` would behave the same way.
* `ALWAYS`, `ALL_TYPES`, or `ALL_OBJECT_TYPES`: always include the discriminator in types serialized as JSON objects (does not apply to primitive types)

Maybe the names could be refined, but I think you get the idea. Any thoughts on this?

Whathecode:
Thank you for this excellent summary and aggregating a set of diverse yet related issues!

I just wanted to note that the current two ways polymorphism is supported and implemented (normal and array polymorphism) might complicate this suggestion; I haven't given it much thought, though, and I expect it is certainly something the smart people behind this library could tackle. ;p

[...]

PR #1498, opened 2021-05-18
Allow disable class discriminator

P1NG2WIN:


P1NG2WIN:
https://github.com/Kotlin/kotlinx.serialization/issues/1487

sandwwraith:
Hi, thanks for your contribution. 
This feature depends on the other one — encoding a type discriminator for polymorphic subtypes. Depending on how that feature will look like, we may want to introduce an enum flag instead of a boolean flag, or do not add the flag and simply make classDiscriminator nullable. So let's postpone this now and resolve it alongside #1247

sandwwraith:
Can be replaced with #2450

PR #1500, opened 2021-05-18
Fix deprecations and warnings, enable -Werror

sandwwraith:


README.md
```diff
@@ -28,8 +28,6 @@ Kotlin serialization consists of a compiler plugin, that generates visitor code
   * [Android](#android)
   * [Multiplatform (Common, JS, Native)](#multiplatform-common-js-native)
   * [Maven](#maven)
-* [Kotlin Serialization Guide](docs/serialization-guide.md)
-* [Full API reference](https://kotlin.github.io/kotlinx.serialization/)
```
qwwdfsad, line 32: Maybe it's worth leaving it here even though it's a knit warning? I find it pretty useful

formats/hocon/build.gradle
```diff
@@ -1,26 +1,12 @@
...
 
 apply plugin: 'kotlin'
 apply plugin: 'kotlinx-serialization'
 
-if (rootProject.ext.jvm_ir_enabled) {
-    kotlin.target.compilations.all {
-        kotlinOptions.useIR = true
-    }
+compileKotlin {
+    kotlinOptions.allWarningsAsErrors = true
```
qwwdfsad, line 9: Does it affect all source-sets or only `main` one?

README.md
sandwwraith: Roman fixed this in master by moving these two bullets into 'Additional links' section outside of TOC

formats/hocon/build.gradle
sandwwraith: Since this is `compileKotlin`, not `compileTestKotlin`, I believe only the main one. I can re-check this if you'd like

qwwdfsad (approved):

PR #1505, opened 2021-05-21
Improved enum and onject serializer lookup tests

shanshin:
* added `.serializer()` invoke for enums with generated and external serializers
* added tests on object's serializer lookup and `.serializer()` invoke
* added `assertFails()` invoke for platforms on which the lookup is broken and needs to be fixed in the future

### DIFF
--- a/core/commonTest/src/kotlinx/serialization/SerializersLookupEnumTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/SerializersLookupEnumTest.kt
@@ -31,15 +31,15 @@ class SerializersLookupEnumTest {
     @Serializable(with = EnumExternalClassSerializer::class)
     enum class EnumExternalClass
 
-    @Serializer(forClass = EnumExternalObject::class)
+    @Serializer(forClass = EnumExternalClass::class)
     class EnumExternalClassSerializer {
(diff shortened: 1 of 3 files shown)


sandwwraith:
I'm not sure if the build is supposed to fail with ` e: /opt/buildAgent/work/b2fef8360e1bcf3d/core/commonTest/src/kotlinx/serialization/SerializersLookupEnumTest.kt: (60, 69): Unresolved reference: serializer` . I think not.

`// FIXME` comments are fixed by https://github.com/JetBrains/kotlin/pull/4401 , right?

shanshin:
> I'm not sure if the build is supposed to fail with ` e: /opt/buildAgent/work/b2fef8360e1bcf3d/core/commonTest/src/kotlinx/serialization/SerializersLookupEnumTest.kt: (60, 69): Unresolved reference: serializer` . I think not.
> 
> `// FIXME` comments are fixed by [JetBrains/kotlin#4401](https://github.com/JetBrains/kotlin/pull/4401) , right?

JetBrains/kotlin#4401 adds the `serializer` function to the companion - only after its merge build will be successful.
A `// FIXME`  comment indicates a different problem with a lookup on some platforms - it's better to solve it in another issue.

sandwwraith:
OK

PR #1509, opened 2021-05-21
Update jmh-gradle-plugin to latest version

MEJIOMAH17:
Motivation: I tried to run :benchmark:jmh task and recieve error. Version update fix this error.
Env: 
```
OS Name:                   Microsoft Windows 10 Home
OS Version:                10.0.19042 N/A Build 19042
jdk: Amazon corretto 11.0.9
```
```
Error: Could not find or load main class org.openjdk.jmh.runner.ForkedMain
```
jmh_error

Related topic
https://github.com/gradle/gradle/issues/9618

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,7 +74,7 @@ buildscript {
 
         // Various benchmarking stuff
         classpath "com.github.jengelman.gradle.plugins:shadow:4.0.2"
-        classpath "me.champeau.gradle:jmh-gradle-plugin:0.4.8"
+        classpath "me.champeau.gradle:jmh-gradle-plugin:0.5.3"
         classpath "net.ltgt.gradle:gradle-apt-plugin:0.21"
     }
 }


sandwwraith:
NB: we usually do `jmhJar -- no-daemon` and then invoke benchmarks manually with `java -jar ...`. This method is more reliable IMO because benchmarks launched outside of any Gradle processes

qwwdfsad:
Thanks! We indeed use `./gradlew --no-daemon cleanJmhJar jmhJar` and then run benchmarks from the resulting jar.
Running them directly from the daemon is a common pitfall that includes skews results

qwwdfsad (approved):
Thanks! We indeed use `./gradlew --no-daemon cleanJmhJar jmhJar` and then run benchmarks from the resulting jar.
Running them directly from the daemon is a common pitfall that includes skews results

PR #1515, opened 2021-05-25
@JsonClassDiscriminator and other serial info annotations in sealed, polymorphic and object serializables

sandwwraith:
As of now, requires a special branch of the Kotlin compiler to work (will likely be merged in 1.5.20)


formats/json/commonTest/src/kotlinx/serialization/features/JsonClassDiscriminatorTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+package kotlinx.serialization.features
+
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+import kotlinx.serialization.modules.*
+import kotlin.test.*
+
+class JsonClassDiscriminatorTest : JsonTestBase() {
```
shanshin, line 13: It would be nice to add a test with an example from the documentation 
```
@Serializable
@JsonTypeDiscriminator("class")
abstract class Base
 
@Serializable
@JsonTypeDiscriminator("error_class")
abstract class ErrorClass: Base()

@Serializable
class Message(val object: Base, val error: ErrorClass?)

val message = Json.decodeFromString<Message>("""{"object": {"class":"my.app.BaseMessage", "message": "not found"}, "error": {"error_class":"my.app.GenericError", "error_code": 404}}""")
```

It may also be necessary to check the case of inheritance of the discriminator by abstract classes, like
```
@Serializable
@JsonTypeDiscriminator("class")
abstract class Base
 
@Serializable
abstract class ErrorClass: Base()

@Serializable
class Message(val object: Base, val error: ErrorClass?)

...
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -79,3 +75,8 @@ private fun throwSerializerNotFound(type: String?, jsonTree: JsonObject): Nothin
         else "class discriminator '$type'"
     throw JsonDecodingException(-1, "Polymorphic serializer was not found for $suffix", jsonTree.toString())
 }
+
+internal fun SerialDescriptor.classDiscriminator(json: Json): String =
+    annotations.filterIsInstance<JsonClassDiscriminator>().singleOrNull()?.discriminator
```
shanshin, line 80: Iterating over annotations for each annotated class can slow down deserialization, it might be worth caching the discriminator.

core/commonMain/src/kotlinx/serialization/internal/ObjectSerializer.kt
```diff
@@ -17,6 +17,15 @@ import kotlinx.serialization.encoding.*
 @PublishedApi
 @OptIn(ExperimentalSerializationApi::class)
 internal class ObjectSerializer<T : Any>(serialName: String, private val objectInstance: T) : KSerializer<T> {
+    @PublishedApi
+    internal constructor(
+        serialName: String,
+        objectInstance: T,
+        classAnnotations: Array<Annotation>
+    ) : this(serialName, objectInstance) {
+        (this.descriptor as SerialDescriptorImpl).annotations = classAnnotations.asList()
```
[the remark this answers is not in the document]
sandwwraith: The problem is, the secondary constructor is called after all `init` blocks and properties initializers. I considered the following alternatives:
- Mutable (to some reasonable extent) serialDescriptorImpl — current solution. I can add `buildSerialDescriptorInternal` function to avoid downcast to be sure this won't break.
- Store the builder in property, call `.build` on every access to `public val descriptor` — I don't like copying every time
- Store descriptor in private mutable property, rebuild it from scratch (because we do not have a copying mechanism) every time. Seems reasonable, although a bit boilerplaite-ish. However, I do not see any strong objections to this solution, so I can rewrite everything to it

formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
sandwwraith: For most of the classes, this array will contain 0 or 1 element (because we do not have a lot of serial info class-level annotations). So I assume this linear lookup will be faster than two (SerialDescriptor -> DescriptorSchemaCache.Key) hash map accesses, or at least not slower

formats/json/commonTest/src/kotlinx/serialization/features/JsonClassDiscriminatorTest.kt
sandwwraith: We haven't designed a discriminator inheritance, but we can discuss it if necessary

sandwwraith:
Superceded by https://github.com/Kotlin/kotlinx.serialization/pull/1608 as we changed the design

PR #1516, opened 2021-05-25
Allow custom serializers to serialize CBOR byte strings

Fadenfire:
Currently attempting to use a custom serializer with `@ByteString` will throw an exception because `encodeSerializableValue` tries to cast the object to a `ByteArray`. This PR fixes that and allows custom serializers to be used with `@ByteString`.

I added `value is ByteArray` to the original check for `@ByteString` in `encodeSerializableValue` to make sure that the value is a `ByteArray`. If the object isn't a `ByteArray` but it's using the byte array descriptor then `beginCollection` will return a custom encoder that will build a byte array and then write the byte string in `endStructure`.

I changed the decoder similarly. If the field has `@ByteString` and it's using the byte array serializer then the byte array is decoded directly. If the field has `@ByteString` but is using a different serializer then `beginStructure` will return a custom decoder that will provide the elements in the byte array.

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
@@ -57,6 +57,40 @@ private open class CborListWriter(cbor: Cbor, encoder: CborEncoder) : CborWriter
     override fun encodeElement(descriptor: SerialDescriptor, index: Int): Boolean = true
 }
 
+private class CborByteStringWriter(private val cbor: Cbor, private val encoder: CborEncoder, size: Int) : CompositeEncoder {
+    override val serializersModule: SerializersModule
+        get() = cbor.serializersModule
+
+    private val buffer = ByteArray(size)
+    
+    override fun encodeByteElement(descriptor: SerialDescriptor, index: Int, value: Byte) {
+        buffer[index] = value
+    }
+
+    override fun endStructure(descriptor: SerialDescriptor) {
+        encoder.encodeByteString(buffer)
+    }
+    
+    private fun throwWrongTypeError(): Nothing = throw SerializationException("Only bytes can be encoded to a byte string")
+
+    override fun encodeBooleanElement(descriptor: SerialDescriptor, index: Int, value: Boolean) = throwWrongTypeError()
+    override fun encodeShortElement(descriptor: SerialDescriptor, index: Int, value: Short) = throwWrongTypeError()
+    override fun encodeCharElement(descriptor: SerialDescriptor, index: Int, value: Char) = throwWrongTypeError()
+    override fun encodeIntElement(descriptor: SerialDescriptor, index: Int, value: Int) = throwWrongTypeError()
+    override fun encodeLongElement(descriptor: SerialDescriptor, index: Int, value: Long) = throwWrongTypeError()
+    override fun encodeFloatElement(descriptor: SerialDescriptor, index: Int, value: Float) = throwWrongTypeError()
+    override fun encodeDoubleElement(descriptor: SerialDescriptor, index: Int, value: Double) = throwWrongTypeError()
+    override fun encodeStringElement(descriptor: SerialDescriptor, index: Int, value: String) = throwWrongTypeError()
+
+    @ExperimentalSerializationApi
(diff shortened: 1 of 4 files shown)


sandwwraith:
This problem is actually deeper than it seems. Check for `ByteArraySerializer` is a recommended way to provide a short-cut (see custom formats guide: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#format-specific-types). Although CBOR uses `== ByteArraySerializer().descriptor` instead of `=== ByteArraySerializer`, they should be generally interchangeable (we may reconsider this and classify this as a cbor implementation bug — this is the first possible solution).

Another problem is, you can't just use `ByteArraySerializer().descriptor` in your serializer, you need to specify your custom serial name. For primitive types, we have a special `PrimitiveSerialDescriptor` function which is used as shown here: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#primitive-serializer . However, there's no analog to wrap a non-primitive descriptor with a new one with a different name. Once we had `SerialDescriptor.withName` extension for that, but it was removed. Adding it back is the second possible solution.

The third solution is to accept your proposal: add a mandatory isInstance check. This way means we also need to change Protobuf implementation and corresponding documentation, and probably notify all custom format authors that this is a preferable approach now.

Fadenfire:
I think the second solution is probably the best. I didn't know that descriptors had to the be unique for the type, I though that they had to be unique for the serialized representation.

Why was `withName` removed? Is `PrimitiveSerialDescriptor` meant to be its replacement (at least for primitives)? I'll close this PR and make another to add `withName` back unless you think that there is a better alternative.

sandwwraith:
withName was removed because there weren't any use cases for it back then, and we planned to eventually replace it with `buildClassSerialDescriptor`. It's also an open discussion whether it should be an extension or a constructor-like function (e.g. `WrappedDescriptor`)

Fadenfire:
Using a constructor function seems to fit the style of the other descriptor builders better than an extension function.

Just to clarify, if something like `WrappedDescriptor(original, name)` where to be implemented then I would define my custom serializer like this:
```kotlin
class CustomByteStringSerializer : KSerializer<CustomByteString> {
    override val descriptor: SerialDescriptor = WrappedDescriptor(ByteArraySerializer().descriptor, "CustomByteString")

    override fun serialize(encoder: Encoder, value: CustomByteString) {
        encoder.encodeSerializableValue(ByteArraySerializer(), byteArrayOf(value.a, value.b, value.c))
    }

    override fun deserialize(decoder: Decoder): CustomByteString {
        val array = decoder.decodeSerializableValue(ByteArraySerializer())

        return CustomByteString(array[0], array[1], array[2])
    }
}
```
Instead of like this:
```kotlin
class CustomByteStringSerializer : KSerializer<CustomByteString> {
    override val descriptor: SerialDescriptor = ByteArraySerializer().descriptor

    override fun serialize(encoder: Encoder, value: CustomByteString) {
        ByteArraySerializer().serialize(encoder, byteArrayOf(value.a, value.b, value.c))
    }

    override fun deserialize(decoder: Decoder): CustomByteString {
        val array = ByteArraySerializer().deserialize(decoder)

        return CustomByteString(array[0], array[1], array[2])
    }
}
```

pdvrieze:
Custom serializers can just delegate to a ByteArraySerializer to do the actual serializing. Given the implementation and the fact that it fundamentally involves bytearrays there isn't really any efficiency issue either.

Fadenfire:
@pdvrieze I can’t use `ByteArraySerializer` directly because then CBOR will serialize it as a list of bytes instead of a byte string. That’s what this PR was supposed to solve but then I found out that according to [this section of the documentation](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#format-specific-types) encoders are allowed to override `encodeSerializableValue` to serialize certain types differently than others. So unless I misunderstand something, custom serializers that wrap another type should always use `encodeSerializableValue` instead of calling the serializer directly.

sandwwraith:
So, after a design discussion with the team, we decided to go with the second option — a descriptor wrapper with a constructor-like function. Most likely it will be named `SerialDescrptor(serialName: String, original: SerialDescriptor)`, as a `PrimitiveSerialDescriptor` analog.

You can contribute this function if you'd like to, or we can take it from here.

Fadenfire:
Sounds good. I have created #1547 to add `SerialDescrptor`.

PR #1520, opened 2021-05-29
Update README.md

diegowholikesorange:
Updating example in Readme to include the actual signature of encodeToString(..) which also requires SerializationStrategy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ data class Project(val name: String, val language: String)
 fun main() {
     // Serializing objects
     val data = Project("kotlinx.serialization", "Kotlin")
-    val string = Json.encodeToString(data)  
+    val string = Json.encodeToString(Project.serializer(), data)  
     println(string) // {"name":"kotlinx.serialization","language":"Kotlin"} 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ data class Project(val name: String, val language: String)
 fun main() {
     // Serializing objects
     val data = Project("kotlinx.serialization", "Kotlin")
-    val string = Json.encodeToString(data)  
+    val string = Json.encodeToString(Project.serializer(), data)
```
Dominaezzz, line 52:
```suggestion
    val string = Json.encodeToString<Project>(data)  
```
Would be more consistent I think.

diegowholikesorange: Yeah sorry, I had problems using it and needed to provide the serialiser. But that's probably only because I customised it with @Serializable(with = MyProjectSerializer::class). Feel free to ignore/reject.

sandwwraith:
Custom serializer should also be recognized when using a reified one-argument extension. Probably you just haven't imported it — try `import kotlinx.serialization.json.*`

PR #1521, opened 2021-05-29
Fix typo in documentation

Dominaezzz:
I'm fairly certain these two have been swapped around.


sandwwraith:
You're right, thanks!

PR #1528, opened 2021-05-31
Introduce @EncodeDefault annotation with two modes: ALWAYS and NEVER

### LINKED ISSUE #1091: Provide @AlwaysEncode / @EncodeDefault annotation

sandwwraith:
Fixes #1091

As usual, corresponding plugin changes required

### DIFF
--- a/formats/json/commonTest/src/kotlinx/serialization/features/SkipDefaults.kt
+++ b/formats/json/commonTest/src/kotlinx/serialization/features/SkipDefaults.kt
@@ -5,11 +5,14 @@
 package kotlinx.serialization.features
 
 import kotlinx.serialization.*
+import kotlinx.serialization.EncodeDefault.Mode.*
 import kotlinx.serialization.json.*
+import kotlinx.serialization.test.noLegacyJs
 import kotlin.test.*
 
 class SkipDefaultsTest {
-    private val json = Json { encodeDefaults = false }
+    private val jsonDropDefaults = Json { encodeDefaults = false }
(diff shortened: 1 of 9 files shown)


formats/json/commonTest/src/kotlinx/serialization/features/SkipDefaults.kt
```diff
@@ -5,18 +5,28 @@
...
+    data class DifferentModes(
+        val a: String = "a",
+        @EncodeDefault val b: String = "b",
+        @EncodeDefault(ALWAYS) val c: String = "c",
+        @EncodeDefault(NEVER) val d: String = "d"
+    )
+
     @Test
     fun serializeCorrectlyDefaults() {
         val jsonWithDefaults = Json { encodeDefaults = true }
```
shanshin, line 32: reuse `jsonEncodeDefaults`

shanshin (approved):

qwwdfsad (approved):
👍 

PTAL at my documentation update

ArcticLampyrid (approved):

PR #1538, opened 2021-06-08
Introduce Animal Sniffer check

qwwdfsad:
Currently, Android API level 14 and Java API for 1.8 are enforced


qwwdfsad:
Enforced Android API level is yet TBD, but it's still worth reviewing the change

build.gradle
```diff
@@ -154,5 +155,21 @@ subprojects {
 
 }
 
+subprojects {
+    // Can't be applied to BOM
+    if (project.name == "kotlinx-serialization-bom") return
```
sandwwraith, line 160: benchmarks could also be excluded here, i think

qwwdfsad: sure, good idea

sandwwraith (approved):

PR #1547, opened 2021-06-11
Add factory function to wrap a serial descriptor with a custom name

Fadenfire:
See #1516 for context.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -115,7 +115,8 @@ public fun PrimitiveSerialDescriptor(serialName: String, kind: PrimitiveKind): S
 @ExperimentalSerializationApi
 public fun SerialDescriptor(serialName: String, original: SerialDescriptor): SerialDescriptor {
     require(serialName.isNotBlank()) { "Blank serial names are prohibited" }
-    require(serialName != original.serialName) { "The name of the wrapped descriptor cannot be the same as the name of the original descriptor" }
+    require(original.kind !is PrimitiveKind) { "For primitive descriptors please use 'PrimitiveSerialDescriptor' instead" }
+    require(serialName != original.serialName) { "The name of the wrapped descriptor ($serialName) cannot be the same as the name of the original descriptor (${original.serialName})" }
     
     return WrappedSerialDescriptor(serialName, original)
 }
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -89,6 +89,41 @@ public fun PrimitiveSerialDescriptor(serialName: String, kind: PrimitiveKind): S
 }
 
 /**
+ * Factory to create a new descriptor that is identical to [original] except that the name is equal to [serialName].
+ * Should be used when you want to serialize a type as another non-primitive type.
+ * Don't use this if you want to serialize a type as a primitive value, use [PrimitiveSerialDescriptor] instead.
+ * 
+ * Example:
+ * ```
+ * @Serializable(CustomSerializer::class)
+ * class CustomType(val a: Int, val b: Int, val c: Int)
+ *
+ * class CustomSerializer: KSerializer<CustomType> {
+ *     override val descriptor = SerialDescriptor("CustomType", IntArraySerializer().descriptor)
+ *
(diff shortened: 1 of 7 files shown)


core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -88,6 +88,41 @@ public fun PrimitiveSerialDescriptor(serialName: String, kind: PrimitiveKind): S
...
+ *         val array = decoder.decodeSerializableValue(IntArraySerializer())
+ *         return CustomType(array[0], array[1], array[2])
+ *     }
+ * }
+ * ```
+ */
+@ExperimentalSerializationApi
+public fun SerialDescriptor(serialName: String, original: SerialDescriptor): SerialDescriptor {
+    require(serialName.isNotBlank()) { "Blank serial names are prohibited" }
+    require(serialName != original.serialName) { "The name of the wrapped descriptor cannot be the same as the name of the original descriptor" }
```
EdwarDDay, line 118: I think, it would be a good idea to just return the `original` descriptor instead of throwing an exception.

sandwwraith: Can't agree here — if one wants keep original name and descriptor, they simply use it without creating wrapper. Name that matches original is clearly a programming mistake here, so we must indicate it

core/commonTest/src/kotlinx/serialization/WrappedSerialDescriptorTest.kt
```diff
@@ -0,0 +1,70 @@
+package kotlinx.serialization
+
+import kotlinx.serialization.builtins.ListSerializer
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.builtins.serializer
```
sandwwraith, line 5: We usually prefer star imports in code style

sandwwraith (commented):
Great job! Can you please also add to cbor tests your use-case with a custom serializer over byte array?

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
EdwarDDay: I thought more of a dynamic serial name. But yeah, I'm not sure if this is really a case here.

core/commonTest/src/kotlinx/serialization/WrappedSerialDescriptorTest.kt
Fadenfire: Sorry, IntelliJ automatically added those. I'll fix it.

Fadenfire:
Unfortunately, I ran into a problem when I tried to add tests for custom CBOR byte strings. When the CBOR encoder/decoder checks for `@ByteString` on a property it also checks that the descriptor equals `ByteArraySerializer().descriptor`. Because a wrapped descriptor doesn't have the same name as the original descriptor it doesn't equal the original so the CBOR encoder/decoder will ignore `@ByteString` on a type with a wrapped descriptor.

Here is the code for the function that CBOR uses to check for `@ByteString`:
```kotlin
private fun SerialDescriptor.isByteString(index: Int): Boolean {
    val elementDescriptor = getElementDescriptor(index)
    val byteArrayDescriptor = ByteArraySerializer().descriptor

    return (elementDescriptor == byteArrayDescriptor || elementDescriptor == byteArrayDescriptor.nullable) &&
}
```

The only solution I see is to remove the check for `ByteArraySerializer().descriptor` in `SerialDescriptor.isByteString`. There is still a check for `ByteArraySerializer().descriptor` in `encodeSerializableValue` so removing the check in `SerialDescriptor.isByteString` shouldn't cause any problems. I'm not really sure why the check in `SerialDescriptor.isByteString` is there.

Do you think removing the check for `ByteArraySerializer().descriptor` in `SerialDescriptor.isByteString` is safe? Is there a better alternative?

sandwwraith:
I think it's would be safe to do so, if it is sufficient to do that so custom serializer works. Probably custom serializers just didn't come into mind when this piece of code was written

Fadenfire:
I have fixed `SerialDescriptor.isByteString` and added tests for using a custom serializer with CBOR byte strings.

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
qwwdfsad, line 118: Please also add `serialName` to error message for better diagnostics

sandwwraith (approved):

sandwwraith:
Many thanks! It was a pleasure working with you

Issue #1551, opened 2021-06-13
Contextual serializer fails for map keys


Sxtanna:
**Describe the bug**
Contextual serialization does not work on map keys, even when the serializer descriptor is of STRING primitive kind.

**To Reproduce**

#### Code
```kotlin
@Serializable
data class Pass(val id: Long, @Contextual val uuid: UUID)

@Serializable
data class Fail(val id: Long, val data: Map<@Contextual UUID, String>)

fun main() {
    val uuid = object : KSerializer<UUID> {
        override val descriptor = PrimitiveSerialDescriptor("java.util.UUID", PrimitiveKind.STRING)
    
        override fun serialize(encoder: Encoder, value: UUID) {
            encoder.encodeString(value.toString())
        }
        
        override fun deserialize(decoder: Decoder): UUID {
            return UUID.fromString(decoder.decodeString())
        }
    }
    
    val json = Json {
        serializersModule = SerializersModule {
            contextual(uuid)
        }
    }
    
    println(json.encodeToString(Pass(0, UUID.randomUUID())))
    println(json.encodeToString(Fail(0, mapOf(UUID.randomUUID() to "Hello"))))
}
```

#### Output:
```
{"id":0,"uuid":"e8ccb154-3062-4f29-857e-62ddb2c0d786"}

Exception in thread "main" kotlinx.serialization.json.internal.JsonEncodingException: Value of type 'kotlinx.serialization.ContextualSerializer<UUID>' can't be used in JSON as a key in the map. It should have either primitive or enum kind, but its kind is 'CONTEXTUAL'.
Use 'allowStructuredMapKeys = true' in 'Json {}' builder to convert such maps to [key1, value1, key2, value2,...] arrays.
	at kotlinx.serialization.json.internal.JsonExceptionsKt.InvalidKeyKindException(JsonExceptions.kt:69)
	at kotlinx.serialization.json.internal.WriteModeKt.switchMode(WriteMode.kt:56)
	at kotlinx.serialization.json.internal.StreamingJsonEncoder.beginStructure(StreamingJsonEncoder.kt:80)
	at kotlinx.serialization.encoding.Encoder$DefaultImpls.beginCollection(Encoding.kt:274)
	at kotlinx.serialization.encoding.AbstractEncoder.beginCollection(AbstractEncoder.kt:18)
	at kotlinx.serialization.internal.MapLikeSerializer.serialize(CollectionSerializers.kt:118)
[...]

PR #1552, opened 2021-06-14
Allow for Contextual serializers as  map keys in Json

pdvrieze:
This addresses #1551 and allows contextual serializers to be used as map keys (it supports both value and non-value variants). It adds tests for both.

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
@@ -8,6 +8,7 @@ package kotlinx.serialization.json.internal
 import kotlinx.serialization.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.SerializersModule
 import kotlin.jvm.JvmField
 
 internal enum class WriteMode(@JvmField val begin: Char, @JvmField val end: Char) {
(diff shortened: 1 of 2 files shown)


sandwwraith:
Can you reopen this against `dev`, please?

pdvrieze:
Oops, sorry, it was even developed against dev.

formats/json/commonTest/src/kotlinx/serialization/json/JsonMapKeysTest.kt
```diff
@@ -94,4 +124,29 @@ class JsonMapKeysTest : JsonTestBase() {
             Json
         )
     }
+
+    @Test
+    fun testContextualValuePrimitivesAreAllowedAsValueMapKeys() =  noLegacyJs {
```
sandwwraith, line 129: These tests do not work with legacy JS compiler?

formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt
```diff
@@ -32,8 +32,19 @@ internal inline fun <T, R1 : T, R2 : T> Json.selectMapMode(
...
 ): T {
-    val keyDescriptor = mapDescriptor.getElementDescriptor(0).carrierDescriptor
-    val keyKind = keyDescriptor.kind
+    var keyDescriptor = mapDescriptor.getElementDescriptor(0).carrierDescriptor
+    var keyKind = keyDescriptor.kind
+
+    if (keyKind == SerialKind.CONTEXTUAL) {
+        keyDescriptor = serializersModule.getContextualDescriptor(keyDescriptor)
+            ?: throw InvalidKeyKindException(keyDescriptor)
+        keyKind = if (keyDescriptor.isInline) {
```
sandwwraith, line 41: That's the same unwrapping that the `.carrierDescriptor` does. I think it all can be merged in the same function like

```kotlin
internal fun SerialDescriptor.carrierDescriptor(module: SerializersModule): SerialDescriptor =
    when {
        kind == SerialKind.CONTEXTUAL -> module.getContextualDescriptor(this)?.carrierDescriptor(module) ?: this
        isInline -> getElementDescriptor(0).carrierDescriptor(module)
        else -> this
    }
```

pdvrieze:
I've addressed both. The test indeed works without IR (only the inline one doesn't).

sandwwraith (approved):
Great job, thanks!

PR #1558, opened 2021-06-16
Added test on serialization of sealed classes from different files

shanshin:


formats/json/commonTest/src/kotlinx/serialization/features/SealedClassesSerializationTest.kt
```diff
@@ -257,4 +259,15 @@ class SealedClassesSerializationTest : JsonTestBase() {
...
     }
+
+    @Test
+    fun testClassesFromDifferentFiles() {
+        val p: SealedParent = SealedChild(5)
+        if (isJvm() || isJsLegacy()) {
+            assertEquals("""{"type":"first child","i":1,"j":5}""", Json.encodeToString(p))
+        } else {
+            // This would be fixed for native and JS/IR in plugin in 1.5.20 - remove condition and else branch in this case
+            assertFails { Json.encodeToString(p) }
```
sandwwraith, line 270: Is this failing because we can't retrieve serializer? Then we can use `encodeToString(SealedParent.serializer())`

Also it makes sense to test deserialization with `assertJsonFormAndRestored` or `asssertSerializedAndRestored`

sandwwraith (approved):

PR #1569, opened 2021-06-24
JVM integration with InputStream and OutputStream

sandwwraith:


formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -0,0 +1,76 @@
...
+@ExperimentalSerializationApi
+public fun <T> Json.decodeFromStream(
+    deserializer: DeserializationStrategy<T>,
+    stream: InputStream,
+    charset: Charset = Charsets.UTF_8
+): T {
+    val lexer = JsonReaderLexer(stream, charset)
+    val input = StreamingJsonDecoder(this, WriteMode.OBJ, lexer)
+    val result = input.decodeSerializableValue(deserializer)
+    lexer.expectEof()
```
sandwwraith, line 63: Probably we shouldn't expect eof here since there are use-cases for deserializing more than one object from stream (i.e. produce `Sequence<T>`)

formats/json/commonTest/src/kotlinx/serialization/json/BasicTypesSerializationTest.kt
```diff
@@ -36,8 +36,8 @@ class BasicTypesSerializationTest : JsonTestBase() {
         testPrimitive("string", "\"string\"", useStreaming)
     }
 
-    private inline fun <reified T : Any> testPrimitive(primitive: T, expectedJson: String, useStreaming: Boolean) {
-        val json = default.encodeToString(primitive, false)
+    private inline fun <reified T : Any> testPrimitive(primitive: T, expectedJson: String, useStreaming: JsonTestingMode) {
+        val json = default.encodeToString(primitive, JsonTestingMode.TREE)
```
qwwdfsad, line 40: It seems to be a typo from the previous version of the code, `useStreaming` should be used.
Also, it's worth renaming it to `jsonTestingMode` here

qwwdfsad: And in our test-specific `decode*` and `encode*` methods as well

formats/json/commonTest/src/kotlinx/serialization/json/JsonMapKeysTest.kt
```diff
@@ -80,7 +80,7 @@ class JsonMapKeysTest : JsonTestBase() {
         }
     }
 
-    private inline fun <reified T: Any> verifyProhibition(value: T, streaming: Boolean) {
+    private inline fun <reified T: Any> verifyProhibition(value: T, streaming: JsonTestingMode) {
```
qwwdfsad, line 83: parameter naming here as well

formats/json/commonTest/src/kotlinx/serialization/json/JsonTestBase.kt
```diff
@@ -6,54 +6,87 @@ package kotlinx.serialization.json
...
+        fun value(i: Int) = values()[i]
+    }
+}
 
 abstract class JsonTestBase {
     protected val default = Json { encodeDefaults = true }
     protected val lenient = Json { isLenient = true; ignoreUnknownKeys = true; allowSpecialFloatingPointValues = true }
 
-    internal inline fun <reified T : Any> Json.encodeToString(value: T, useStreaming: Boolean): String {
+    internal inline fun <reified T : Any> Json.encodeToString(value: T, useStreaming: JsonTestingMode): String {
```
qwwdfsad, line 28: naming here and below

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -0,0 +1,76 @@
...
+ * Serializes given [value] to [stream] using [charset] and serializer retrieved from the reified type parameter.
+ *
+ * @throws [SerializationException] if the given value cannot be serialized to JSON.
+ * @throws [IOException] If an I/O error occurs and stream can't be written to.
+ */
+@ExperimentalSerializationApi
+public inline fun <reified T> Json.encodeToStream(
+    value: T,
+    stream: OutputStream,
+    charset: Charset = Charsets.UTF_8
```
qwwdfsad, line 44: A charset is quite a serious commitment in our public API.
Let's stick with UTF-8 only and document it as well?

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
qwwdfsad: Right, this seems to be beyond the scope of our parsing.
In our string-related code, this assertion verified our own parsing logic is correct, but it's not our responsibility for streams.

I've checked out Jackson and it ignores content right after finished object as well

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TwitterFeedStreamBenchmark.kt
```diff
@@ -0,0 +1,84 @@
...
+    private val jsonIgnoreUnknwn = Json { ignoreUnknownKeys = true }
+    private val objectMapper: ObjectMapper =
+        jacksonObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+
+
+    private var file: Path? = null
+
+    @Setup
+    fun init() {
+        file = Files.createTempFile("json_benchmark", "tmp")
```
qwwdfsad, line 35: [nit] typically, for such benchmarks I prefer measurement from the stream that is already presented in the memory (i.e. BAIS/BAOS).

While it definitely not represent the "real"-ish workload, it does not skew our measurements and we can be sure that e.g. the benchmark score is not affected by the quality and/or load of the disk or file system,
In the end, the goal of this benchmark is to get precise (for this particular data) comparison with Jackson and we do not want our differences to be obscured or hidden by 3-rd party factors

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TwitterFeedStreamBenchmark.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlin.io.path.deleteIfExists
+import kotlin.io.path.outputStream
+
+@Warmup(iterations = 7, time = 1)
+@Measurement(iterations = 7, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@Fork(2)
+open class TwitterFeedStreamBenchmark {
```
qwwdfsad, line 22: I see that a lot of `decode*` benchmarks dropped their score by ~15%.

It seems to be a serious hit and it's definitely worth pinpointing it before merging this PR, so we are sure we cannot avoid it (or can and then will do)

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonStringBuilder.kt
```diff
@@ -75,7 +76,7 @@ internal actual class JsonStringBuilder {
              * We ar already on slow path and haven't guessed the capacity properly.
              * Reserve +2 for backslash-escaped symbols on each iteration
              */
-            ensureTotalCapacity(sz + 2)
+            sz = ensureTotalCapacity(sz , 2)
```
qwwdfsad, line 79: excessive indent

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonStringBuilder.kt
```diff
@@ -25,9 +25,10 @@ package kotlinx.serialization.json.internal
  * 3) We pool char arrays in order to save excess resizes, allocations
  *    and nulls-out of arrays.
  */
-internal actual class JsonStringBuilder {
-    private var array = CharArrayPool.take()
-    private var size = 0
+internal actual open class JsonStringBuilder(protected var array: CharArray) {
```
qwwdfsad, line 28:
```suggestion
internal actual open class JsonStringBuilder(@JvmField protected var array: CharArray) {
```

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonToWriterStringBuilder.kt
```diff
@@ -0,0 +1,41 @@
+package kotlinx.serialization.json.internal
+
+import java.io.OutputStream
+import java.io.Writer
+import java.nio.charset.Charset
+
+
+internal class JsonToWriterStringBuilder(private val writer: Writer) : JsonStringBuilder(
+    // maybe this can also be taken from the pool, but currently initial char array size there is 128, which is too low.
+    CharArray(BATCH_SIZE)
```
qwwdfsad, line 10: It seems to be pretty huge allocation (e.g. it always misses TLAB). Could you please ensure it doesn't dominate small objects serialization?

If so, it's worth either reducing its size or pool a few instances of this

qwwdfsad (changes requested):
I haven't got a deep look into decoding yet, but it seems that there is already enough actionable points here

chris-hatton:
Question regarding the behaviour of this change.  Will `decodeFromStream` behave such that I can call it serially against the same stream, each time consuming only as much of the stream as is necessary to form the complete object?
e.g. given the stream `{"someKey":"someValue1"}{"someKey":"someValue2"}`
I can call `decodeFromStream()` twice to get both objects?
This is the characteristic I am looking for, to be able to read a `Flow<T>` from a long-lived HTTP response stream.
Thanks for your efforts @sandwwraith 🙏 This is a hotly awaited improvement.

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
JakeWharton: Moshi will leave your stream intact allowing you to do things like JSON-RPC which has newline-delimited objects. Granted, we also require that you pass in a buffered stream so that we can read only what's necessary and leave everything else in the buffer you then have access to.

BenWoodworth:
I'm curious, was using [`Reader`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Reader.html)/[`Writer`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Writer.html)s considered (instead of `Stream`s)? That would've been my first thought for JSON/String formats, and avoids [dealing with character encodings](https://github.com/Kotlin/kotlinx.serialization/pull/1569/files/8678e127d653b334ff5288fd6b617cbbd02c633c#r659932844).

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
chris-hatton: I have a use case for implementing a `Flow<T>` from a long-lived HTTP response stream.  To have the lexer expect the end of the stream here seems wrong - why impose this opinion on the stream?  The lexers interest in the stream should begin and end with the fully-formed JSON content IMO

pdvrieze: having thought about the streaming case I would say that a flow of objects would be a valid case to handle. Probably flow is the right type - both for encoding and decoding. There are some interesting interactions with coroutines there though and flows should probably have their own entrypoint.

pdvrieze: As a separate matter, these are provided as extension functions. It would be interesting to have an interface like `StringFormat` or `BinaryFormat` (or a combination interface) that provides the consistency between formats. There is also a case to support the `Reader` and `Writer` interfaces next to/instead of InputStream/OutputStream (that would also avoid bringing in charset explicitly while still allowing non-utf8 to be used).

sandwwraith:
@BenWoodworth `Reader/Writer` is used internally. API provides methods with `Input/OutputStream`s because it's more versatile and allows to implement charset-specific parsers in the future

@chris-hatton Yes, I think we can do it — I'll add a test for it

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
sandwwraith: I don't see many problems in it — Charset isn't going anywhere from `java.nio`, AFAIK. We still can implement charset-specific decoders if we need to.

Of course, we also can add overload with charset later, if there would be demand for it, but we'll end up with two almost identical overloads instead of a function with default parameter

qwwdfsad: It will obstruct us from properly optimizing UTF-8 decoding because we will be forced to maintain the rest of the charsets (think about suffix tree), apart from that charsets are quite slow on their own.

Moreover, most of the modern applications are using UTF-8 anyway and it's unlikely we'll stumble across such demand. So I suggest avoiding introducing charsets into our API and rely on UTF-8 only, waiting for a proper demand.

qwwdfsad (changes requested):
I've dug through the profile and it seems that JVM cannot optimize string access across `CharSequence` interface properly.

I've tried to tweak it here and there, but it's quite hard to ensure all the invariants with the existing limitations.

I'd suggest you to do the following:

Get the base JsonLexer with the only state -- `currentPosition` and utility functions for slow-paths: `skipElement`, various `fail` functions, maybe boolean/numbers consumption (char sequence is an input parameter of a function then). Everything else is copy-pasted between streaming and string implementations.

At this moment, the performance model is quite clear and expected degradation should be insignificant (educated guess -- 2-4%).
Then you can start commonizing (handling via `CharSequence` interface in the base json lexer) the parts of parsing where the compiler is smart enough to optimize everything away.

I expect that the biggest offenders (things you cannot commonize) will be just a few functions that were written in a compact and polished manner -- `skipWhitespaces`, `tryConsumeComma`, `canConsumeValue` and `peekNextToken`. Everything else will probably be working as expected

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
pdvrieze: @qwwdfsad I agree on the character set thing. Especially in the context of taking a "writer". The api to handle the charsets is already there. The only counterpoint is the speed of parsing streams vs readers.

qwwdfsad: Good to know, thanks!

Leonid is on the vacation right now, so the final decision is not yet made, but we've discussed this and other cases on our design meeting and seems that we have a consensus regarding encodings -- UTF-8 dominates all along, and everyone is mostly fine to be UTF-8 first. Even Java changes towards that: [JEP-400](https://openjdk.java.net/jeps/400). Moshi is also a good example of that.

So it's _likely_ that all our new API will be using UTF-8 as well and other encoding will be put aside and be supported on a best-effort basis

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package kotlinx.serialization.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.internal.*
+import java.io.*
+
+/**
+ * Serializes the [value] with [serializer] into a [stream] using JSON format and UTF-8 encoding..
```
qwwdfsad, line 12: Redundant dot

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonToWriterStringBuilder.kt
```diff
@@ -0,0 +1,41 @@
...
+                // to per-symbol path in appendQuoted on large strings,
+                // but this approach is adequate for current stage, too.
+                array = CharArray(requiredSize.coerceAtLeast(currentSize * 2))
+            }
+            return 0
+        }
+        return oldSize
+    }
+
+    private fun dumpAndReset(sz: Int = size) {
```
qwwdfsad, line 31: [nit] it's just `flush` :)

qwwdfsad (approved):
Good to go 🚀 

Please don't forget to file issues for future improvements -- UTF-8 parsing and multishot streams

sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/issues/1662

slavonnet:
String. Serializer work ok on big JSon. Stream   get  random.eof exception  I sinppe replace Strung Encoder t to  Stream Encoder  in converter factory . Small size json is ok. Buffer rewrite by gzip wtite all

PR #1570, opened 2021-06-24
CBOR Integer Encoding Length Fix

davertay:
This PR fixes an issue where the packed length is sub-optimal for all positive integers that have the MSB bit set.

CBOR integers are packed with a set number of bytes, bucketed by length (8bit, 16bit, 32bit etc) determined by the bit length of the source. Negative integers are denoted with a different Major Type (1 for -ve, 0 for +ve) but are otherwise treated as an unsigned value when packing/unpacking.

The original code bucketing check uses the *signed* `MAX_VALUE` constants which means any values with the MSB set will fall into the next highest bucket, and thus pack into double the expected number of bytes.

e.g. 65535 (uint16, all bits high) is encoded with two extra zero bytes as `0x19 0x00 0x00 0xFF 0xFF` whereas it should encode to `0x19 0xFF 0xFF`.

The RFC does not explicitly state that integers must always pack to the fewest bytes possible, and the format certainly allows it so it is still valid CBOR. e.g. you could pack the value 1 into the space large enough for a 64bit integer and it can still be decoded successfully. However this is certainly at odds with the "C" in "CBOR".

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
@@ -166,30 +166,29 @@ internal class CborEncoder(private val output: ByteArrayOutput) {
     }
 
     private fun composeNumber(value: Long): ByteArray =
-        if (value >= 0) composePositive(value) else composeNegative(value)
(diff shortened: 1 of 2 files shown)


sandwwraith (approved):
Good catch, thanks for your help!

PR #1591, opened 2021-07-08
Add guide for delegating serializers and wrapping serial descriptor

### LINKED ISSUE #1588: Document new 'SerialDescriptor(original, name)' function in guide on custom serializers
See https://github.com/Kotlin/kotlinx.serialization/pull/1547 and https://github.com/Kotlin/kotlinx.serialization/pull/1516

sandwwraith:
Fixes #1588

### DIFF
--- a/docs/serializers.md
+++ b/docs/serializers.md
@@ -393,19 +393,20 @@ using [encodeSerializableValue][Encoder.encodeSerializableValue] and
 import kotlinx.serialization.builtins.IntArraySerializer
 
 class ColorIntArraySerializer : KSerializer<Color> {
-    override val descriptor = SerialDescriptor("Color", IntArraySerializer().descriptor)
+    private val delegateSerializer = IntArraySerializer()
+    override val descriptor = SerialDescriptor("Color", delegateSerializer.descriptor)
 
     override fun serialize(encoder: Encoder, value: Color) {
-      val data = intArrayOf(
-        (value.rgb shr 16) and 0xFF,
-        (value.rgb shr 8) and 0xFF,
-        value.rgb and 0xFF
-      )
-      encoder.encodeSerializableValue(IntArraySerializer(), data)
+        val data = intArrayOf(
+            (value.rgb shr 16) and 0xFF,
+            (value.rgb shr 8) and 0xFF,
+            value.rgb and 0xFF
+        )
+        encoder.encodeSerializableValue(delegateSerializer, data)
     }
 
     override fun deserialize(decoder: Decoder): Color {
-        val array = decoder.decodeSerializableValue(IntArraySerializer())
+        val array = decoder.decodeSerializableValue(delegateSerializer)
         return Color((array[0] shl 16) or (array[1] shl 8) or array[2])
     }
 }
--- a/docs/serializers.md
+++ b/docs/serializers.md
@@ -412,13 +412,13 @@ class ColorIntArraySerializer : KSerializer<Color> {
 }
 ```
 
-Note that we can't use default `Color.serializer().descriptor` here, because formats that rely
+Note that we can't use default `Color.serializer().descriptor` here because formats that rely
 on the schema may think that we would call `encodeInt` instead of `encodeSerializableValue`.
 Neither we can use `IntArraySerializer().descriptor` directly — otherwise, formats that handle int arrays specially
 can't tell if `value` is really a `IntArray` or a `Color`. Don't worry, this optimization would still kick in
 when serializing actual underlying int array.
 
-> Example of how format can treat byte arrays specially is shown in the [formats guide](formats.md#format-specific-types).
+> Example of how format can treat arrays specially is shown in the [formats guide](formats.md#format-specific-types).
 
 Now we can use the serializer:
 
--- a/docs/serializers.md
+++ b/docs/serializers.md
@@ -17,6 +17,7 @@ In this chapter we'll take a look at serializers in more detail, and we'll see h
   * [Using top-level serializer function](#using-top-level-serializer-function)
 * [Custom serializers](#custom-serializers)
   * [Primitive serializer](#primitive-serializer)
+  * [Delegating serializers](#delegating-serializers)
   * [Composite serializer via surrogate](#composite-serializer-via-surrogate)
   * [Hand-written composite serializer](#hand-written-composite-serializer)
   * [Sequential decoding protocol (experimental)](#sequential-decoding-protocol-experimental)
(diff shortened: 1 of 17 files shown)


docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+An implementation of [KSerializer] for our original `Color` class is going to perform a conversion between
+`Color` and `IntArray`, but delegate the actual serialization logic to the `IntArraySerializer`
+using [encodeSerializableValue][Encoder.encodeSerializableValue] and
+[decodeSerializableValue][Decoder.decodeSerializableValue].
+
+```kotlin
+import kotlinx.serialization.builtins.IntArraySerializer
+
+class ColorIntArraySerializer : KSerializer<Color> {
+    override val descriptor = SerialDescriptor("Color", IntArraySerializer().descriptor)
```
pdvrieze, line 396: Would it perhaps be a better example to have a `val delegateSerializer = IntArraySerializer()` property instead of repeated constructor calls?

pdvrieze:
This guide is useful and hopefully encourages more use of delegates/surrogates. It may be useful to highlight the use cases solved by surrogates and delegates (there are plenty in the issue tracker). As an idea for the future, Api/plugin support might be worthwhile to make the threshold to either even lower.

docs/serializers.md
Dominaezzz: repeated how? `descriptor` is only initialised once.

pdvrieze: @Dominaezzz, the serializer is used in the encode/decode functions below. While using an attribute to store the delegate serializer can use more memory, this is normally not particularly significant, and limited to the lifetime of the serializer itself (`ColorIntArraySerializer` in this case).

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
 
 <!--- TEST -->
 
+### Delegating serializers
+
+In the previous example, we represented the `Color` class as string.
```
qwwdfsad, line 383:
```suggestion
In the previous example, we represented the `Color` class as a string.
```

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+String is considered to be a primitive type, therefore we used `PrimitiveClassDescriptor` and specialized `encodeString` method. 
+Now let's see what our actions would be if we have to serialize `Color` as other non-primitive type, let's say `IntArray`.
```
qwwdfsad, line 385:
```suggestion
Now let's see what our actions would be if we have to serialize `Color` as another non-primitive type, let's say `IntArray`.
```

docs/serializers.md
qwwdfsad: I second that, let's not encourage people to use our constructor-like factories all over the place, they can be quite slow (though it's not the case here)

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+
+```kotlin
+import kotlinx.serialization.builtins.IntArraySerializer
+
+class ColorIntArraySerializer : KSerializer<Color> {
+    override val descriptor = SerialDescriptor("Color", IntArraySerializer().descriptor)
+
+    override fun serialize(encoder: Encoder, value: Color) {
+      val data = intArrayOf(
+        (value.rgb shr 16) and 0xFF,
```
qwwdfsad, line 400: indents are incorrect, should be 4 spaces (to be aligned with Kotlin and our own code style)

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+
+    override fun deserialize(decoder: Decoder): Color {
+        val array = decoder.decodeSerializableValue(IntArraySerializer())
+        return Color((array[0] shl 16) or (array[1] shl 8) or array[2])
+    }
+}
+```
+
+Note that we can't use default `Color.serializer().descriptor` here, because formats that rely
+on schema may think that we would call `encodeInt` instead of `encodeSerializableValue`.
```
qwwdfsad, line 415:
```suggestion
on the schema may think that we would call `encodeInt` instead of `encodeSerializableValue`.
```

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+    }
+
+    override fun deserialize(decoder: Decoder): Color {
+        val array = decoder.decodeSerializableValue(IntArraySerializer())
+        return Color((array[0] shl 16) or (array[1] shl 8) or array[2])
+    }
+}
+```
+
+Note that we can't use default `Color.serializer().descriptor` here, because formats that rely
```
qwwdfsad, line 414:
```suggestion
Note that we can't use default `Color.serializer().descriptor` here because formats that rely on
```

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+class Color(val rgb: Int)
+
+fun main() {
+    val green = Color(0x00ff00)
+    println(Json.encodeToString(green))
+}  
+```    
+
+As you can see, such array representation is not very useful in JSON,
+but may save your space when used with a `ByteArray` and a binary format.
```
qwwdfsad, line 435:
```suggestion
but may save some space when used with a `ByteArray` and a binary format.
```

qwwdfsad (approved):
Everything else looks great, thanks for taking care of it!

docs/serializers.md
```diff
@@ -377,6 +378,71 @@ Both `Color` properties are serialized as strings.
...
+        value.rgb and 0xFF
+      )
+      encoder.encodeSerializableValue(IntArraySerializer(), data)
+    }
+
+    override fun deserialize(decoder: Decoder): Color {
+        val array = decoder.decodeSerializableValue(IntArraySerializer())
+        return Color((array[0] shl 16) or (array[1] shl 8) or array[2])
+    }
+}
```
pdvrieze, line 411: The fix for the delegate property would be (also with fixed indentation):
```suggestion
class ColorIntArraySerializer : KSerializer<Color> {
    private val delegateSerializer = IntArraySerializer()
    override val descriptor = SerialDescriptor("Color", delegateSerializer.descriptor)

    override fun serialize(encoder: Encoder, value: Color) {
        val data = intArrayOf(
            (value.rgb shr 16) and 0xFF,
            (value.rgb shr 8) and 0xFF,
            value.rgb and 0xFF
        )
        encoder.encodeSerializableValue(delegateSerializer, data)
    }

    override fun deserialize(decoder: Decoder): Color {
        val array = decoder.decodeSerializableValue(delegateSerializer)
        return Color((array[0] shl 16) or (array[1] shl 8) or array[2])
    }
}
```

PR #1596, opened 2021-07-12
Switch on deep recursive function when nested level of JS…

### LINKED ISSUE #1594: StackOverflowError when parsing deep JSON
**Describe the bug**
JsonTreeReader class of kotlinx-serialization-json library uses recursion to parse JSON which causes StackOverflowError when JSON is deep enough. 

**To Reproduce**
```Kotlin
var json = """{ "a": %}"""
for (i in 0..12) {
    json = json.replace("%", json)
}
json = json.replace("%", "0")
Json.parseToJsonElement(json)
```

**Expected behavior**
Json parsed without crashing the app.

**Environment**
 - Kotlin version: 1.5.10
 - Library version: 1.2.1
 - Kotlin platforms: JVM
 - Gradle version: 7.0.2

qwwdfsad:
…ON is too deep

Fixes #1594

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
@@ -13,8 +13,16 @@ internal class JsonTreeReader(
     private val lexer: JsonLexer
 ) {
     private val isLenient = configuration.isLenient
+    private var stackDepth = 0
 
-    private fun readObject(): JsonElement {
+    private fun readObject(): JsonElement = readObjectImpl {
+        read()
+    }
+
+    private suspend fun DeepRecursiveScope<Unit, JsonElement>.readObject(): JsonElement =
+        readObjectImpl { callRecursive(Unit) }
+
+    private inline fun readObjectImpl(reader: () -> JsonElement): JsonObject {
         var lastToken = lexer.consumeNextToken(TC_BEGIN_OBJ)
         if (lexer.peekNextToken() == TC_COMMA) lexer.fail("Unexpected leading comma")
         val result = linkedMapOf<String, JsonElement>()
@@ -22,7 +30,7 @@ internal class JsonTreeReader(
             // Read key and value
             val key = if (isLenient) lexer.consumeStringLenient() else lexer.consumeString()
             lexer.consumeNextToken(TC_COLON)
-            val element = read()
+            val element = reader()
             result[key] = element
             // Verify the next token
             lastToken = lexer.consumeNextToken()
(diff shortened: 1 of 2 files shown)


qwwdfsad:
This is a prototype and I do not expect it to be merged as is, this is rather a PR to evaluate the approach in general

qwwdfsad:
Huh, it seems that the recursive upper-bound should be lower for K/N targets

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
```diff
@@ -75,9 +83,28 @@ internal class JsonTreeReader(
         return when (lexer.peekNextToken()) {
             TC_STRING -> readValue(isString = true)
             TC_OTHER -> readValue(isString = false)
-            TC_BEGIN_OBJ -> readObject()
+            TC_BEGIN_OBJ -> {
+                // 500 is an arbitrary constant
+                val result = if (++stackDepth > 500) {
```
elizarov, line 88: It seems that it should be checking level for equality, e.g. (`++stackDepth == RECUSIVE_LIMIT`). In the current implemented a new `DeepRecursiveFunction` invocation context will be created on each recursion level beyond the limit.

formats/json/commonTest/src/kotlinx/serialization/json/JsonElementDecodingTest.kt
```diff
@@ -40,4 +40,15 @@ class JsonElementDecodingTest : JsonTestBase() {
         assertEquals(expectedString, element.toString())
         assertEquals(value, Json.decodeFromJsonElement(element))
     }
+
+    @Test
+    fun testDeepRecursion() {
+        // Reported as https://github.com/Kotlin/kotlinx.serialization/issues/1594
+        var json = """{ "a": %}"""
+        for (i in 0..12) {
+            json = json.replace("%", json)
```
elizarov, line 49: This does not seem "deep enough" to reliably trigger stack overflow on JVM. Likely it passes by chance. I'd suggest going from 2^12 to 2^16 (or maybe even up to 2^20) to be absolutely sure it'll give StackOverflowException because of the regular recursion.

elizarov (changes requested):

formats/json/commonTest/src/kotlinx/serialization/json/JsonElementDecodingTest.kt
qwwdfsad: It seems more than enough. `2^13 already` takes up to 2 seconds on JVM and everything above is just way too slow.
I've checked that this test reliably crashes all backends prior to fix

qwwdfsad:
@sandwwraith ready to merge, PTAL

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
```diff
@@ -75,9 +83,33 @@ internal class JsonTreeReader(
...
+                /*
+                 * If the object has the depth of 200 (an arbitrary "good enough" constant), it means
+                 * that it's time to switch to stackless recursion to avoid StackOverflowError.
+                 * This case is quite rare and specific, so more complex nestings (e.g. through
+                 * the chain of JsonArray and JsonElement) are not supported.
+                 */
+                val result = if (++stackDepth == 200) {
+                    readDeepRecursive()
+                } else {
+                    readObject()
```
sandwwraith, line 96: I may be missing something here, but how do we actually make a recursive call when we're on e.g. stack depth 202 (object inside an object that triggered callRecursive)

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
qwwdfsad: Not really because once we're in `readDeepRecursive`, we no longer can re-enter this function (moreover, when nested objects are separated via `JsonArray`, equality check will prevent us from maintaining multiple deep-recursive scopes)

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
qwwdfsad: On the depth of 200 we switch to `readDeepRecursive`.
`readDeepRecursive` is a recursive function that invokes `DeepRecursiveScope<Unit, JsonElement>.readObject()` which, when parsing, invokes `callRecursive` (aka `readDeepRecursive`). The stack depth is no longer maintained as soon as we're in `readDeepRecursive`

elizarov (approved):

sandwwraith (approved):

PR #1608, opened 2021-07-20
@JsonClassDiscriminator and @InheritableSerialInfo

sandwwraith:


core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
```diff
@@ -67,14 +67,28 @@ import kotlin.reflect.*
...
 @OptIn(ExperimentalSerializationApi::class)
 public class PolymorphicSerializer<T : Any>(override val baseClass: KClass<T>) : AbstractPolymorphicSerializer<T>() {
-    public override val descriptor: SerialDescriptor =
+
+    @PublishedApi // should we allow user access to this constructor?
+    internal constructor(
+        baseClass: KClass<T>,
+        classAnnotations: Array<Annotation>
+    ) : this(baseClass) {
+        _descriptor.annotations = classAnnotations.asList()
```
pdvrieze, line 76: Wouldn't it be better to make this the primary constructor and have the current primary become secondary? That also means that the descriptor can remain immutable.

sandwwraith: Unfortunately, old (e.g. 1.5.10) plugins are written in the way that they always pick the primary constructor. Changing its signature would break them

shanshin (approved):

core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
qwwdfsad: It's worth dropping a comment here, especially when we do not have such compatibility tests.

Also, it forces us to use this pretty crafty hack. Do we have a chance to change that in the future?

core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
```diff
@@ -67,14 +67,28 @@ import kotlin.reflect.*
...
+    }
+
+    private val _descriptor: SerialDescriptorImpl =
```
qwwdfsad, line 79: Suggest change here to avoid mutable descriptor:

```
public class PolymorphicSerializer<T : Any>(override val baseClass: KClass<T>) : AbstractPolymorphicSerializer<T>() {

    @PublishedApi // should we allow user access to this constructor?
    internal constructor(
        baseClass: KClass<T>,
        classAnnotations: Array<Annotation>
    ) : this(baseClass) {
        annotations = classAnnotations
    }


    private var annotations = emptyArray<Annotation>()

    public override val descriptor: SerialDescriptor =  buildSerialDescriptor("kotlinx.serialization.Polymorphic", PolymorphicKind.OPEN) {
        element("type", String.serializer().descriptor)
        element(
            "value",
            buildSerialDescriptor(
                "kotlinx.serialization.Polymorphic<${baseClass.simpleName}>",
                SerialKind.CONTEXTUAL
            )
        )
        annotations = this@PolymorphicSerializer.annotations.asList()
    }.withContext(baseClass)
```

core/commonMain/src/kotlinx/serialization/SealedSerializer.kt
```diff
@@ -77,6 +77,17 @@ public class SealedClassSerializer<T : Any>(
...
 
+    @PublishedApi
+    internal constructor(
+        serialName: String,
+        baseClass: KClass<T>,
+        subclasses: Array<KClass<out T>>,
+        subclassSerializers: Array<KSerializer<out T>>,
+        classAnnotations: Array<Annotation>
+    ) : this(serialName, baseClass, subclasses, subclassSerializers) {
+        (this.descriptor as SerialDescriptorImpl).annotations = classAnnotations.asList()
```
qwwdfsad, line 88: same here, would be nice to keep our hacks local and comment them

core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
sandwwraith: I think this hack can be fixed next time we raise minimal Kotlin version

core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
sandwwraith: the problem is, `buildSerialDescriptor`, as initializing expression, is executed before secondary constructor

sandwwraith: See also https://github.com/Kotlin/kotlinx.serialization/pull/1515#discussion_r642580350

qwwdfsad: Let's make it lazy then?

formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
```diff
@@ -79,3 +75,8 @@ private fun throwSerializerNotFound(type: String?, jsonTree: JsonObject): Nothin
         else "class discriminator '$type'"
     throw JsonDecodingException(-1, "Polymorphic serializer was not found for $suffix", jsonTree.toString())
 }
+
+internal fun SerialDescriptor.classDiscriminator(json: Json): String =
+    annotations.filterIsInstance<JsonClassDiscriminator>().singleOrNull()?.discriminator
```
qwwdfsad, line 80: It will create a separate `ArrayList` instance, better to avoid that and use plain loop instead

qwwdfsad (changes requested):

core/commonMain/src/kotlinx/serialization/SealedSerializer.kt
qwwdfsad: Don't forget to leave an extensive comment with a migration strategy

core/commonMain/src/kotlinx/serialization/PolymorphicSerializer.kt
qwwdfsad: Won't expect-actual help us here?

sandwwraith: I can try

formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt
sandwwraith: Won't `asSequence` solve this problem?

PR #1611, opened 2021-07-23
Add missing experimental status to HOCON in list of supported formats

bluefireoly:
The API for HOCON is annotated with `@ExperimentalSerializationApi`, therefore it should be noted in the list of supported formats as well, because it has been done for the other experimental formats in the same way.


sandwwraith (approved):
Indeed, thanks!

PR #1614, opened 2021-07-27
CBOR: ignore tags when reading

DavidJRobertson:
Currently, if the CBOR-encoded input contains any [tags](https://datatracker.ietf.org/doc/html/rfc7049#section-2.4) an exception will be thrown. The default behaviour should be to ignore any tags.


sandwwraith (approved):
Thank you very much!

Issue #940, opened 2020-07-28
Explicit module-descriptor for Java 9 and later


TheMrMilchmann:
**What is your use-case and why do you need this feature?**
As of Kotlin 1.4-M3, Java 9 module descriptors have been added to the standard library artifacts ([KT-21266](https://youtrack.jetbrains.com/issue/KT-21266)). It would be nice to see this also happen to libraries in the Kotlin ecosystem going forward as this would enable Kotlin JVM applications to benefit from modern JDK tools (such as jlink). As kotlinx.serialization has no remaining unmodularized dependencies, it seems like an ideal candidate to receive module support.

**Describe the solution you'd like**
Ideally, an explicit descriptor would be added using the Multi-Release JAR mechanism ([JEP 238](https://openjdk.java.net/jeps/238)) added in JDK 9 (just as it was done for the standard library).

TheMrMilchmann:
I'm currently facing a related issue with a fairly simple intermediate solution:

I would love to publish a library with an explicit module descriptor and a dependency on `kotlinx.serialization` but this is not possible as the latter does not have a stable module name. This could be solved by providing a stable [`Automatic-Module-Name`](https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers) entry in the JVM artifact's manifest.

TheMrMilchmann:
Explicit module descriptors were added by #1624 and are available in [1.3.0-RC](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.3.0-RC). Thanks, @lion7!

PR #1624, opened 2021-08-03
Add explicit module-info's for JPMS compatability

lion7:
In this pull request I added a few `module-info.java` files so the `kotlinx.serialization` library can be used in Java JPMS (Jigsaw) projects. I also extended the root `build.gradle` script so that the appropriate subprojects do a separate compilation of the `module-info.java` file. 

To maintain backwards compatibility with Java 8, I marked the resulting JAR as a multi-release JAR with the `module-info.class` moved to `META-INF/versions/9/module-info.class` which is similar as how it's done in the Kotlin stdlib.  See https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jdk8/build.gradle#L80, https://github.com/JetBrains/kotlin/blob/master/buildSrc/src/main/kotlin/LibrariesCommon.kt#L22 and https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jdk8/java9/module-info.java for more details.

Note that moving the `module-info.class` was / is not really necessary for JPMS to work, but I believe it's a bit cleaner than just adding the compiled `module-info.class` to the root of the JAR.

This change allows bundling projects using `jlink`, such as requested in https://github.com/Kotlin/kotlinx.serialization/issues/940. `jlink` requires that all dependencies are explicit JPMS (Jigsaw) modules. That means all JAR's must contain a `module-info.class`, automatic modules don't work with `jlink`. 
In 1 of my own projects I am trying to bundle an OpenJFX / TornadoFX application, which currently works using a SNAPSHOT version of this branch.

Note that at least JDK 9 is needed to build this branch (for compiling the `module-info.java` file). Existing code can of course be compiled using a Java 8 target.

### DIFF
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -97,7 +97,7 @@ object Java9Modularity {
                 val actualOutput = outputCaptureStream.toString(Charsets.UTF_8).trim()
 
                 if (actualOutput != expectedOutput) {
-                    throw IllegalStateException("Module-info does not match!\n$actualOutput")
+                    throw IllegalStateException("Module-info requirements section does not match!\n$actualOutput")
                 }
             }
         }
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -1,0 +1,108 @@
+import org.gradle.api.*
+import org.gradle.api.file.*
+import org.gradle.api.tasks.bundling.*
+import org.gradle.api.tasks.compile.*
+import org.gradle.kotlin.dsl.*
+import org.gradle.util.GUtil.*
+import org.jetbrains.kotlin.gradle.dsl.*
+import org.jetbrains.kotlin.gradle.plugin.*
+import org.jetbrains.kotlin.gradle.plugin.mpp.*
+import org.jetbrains.kotlin.gradle.plugin.mpp.pm20.*
+import org.jetbrains.kotlin.gradle.targets.jvm.*
+import java.io.*
+import java.lang.module.*
+import java.util.spi.*
+
+object Java9Modularity {
+
+    @JvmStatic
+    @JvmOverloads
+    fun Project.configureJava9ModuleInfo(multiRelease: Boolean = true) {
+        val kotlin = extensions.findByType<KotlinProjectExtension>() ?: return
+        val jvmTargets = kotlin.targets.filter { it is KotlinJvmTarget || it is KotlinWithJavaTarget<*> }
+        if (jvmTargets.isEmpty()) {
+            logger.warn("No Kotlin JVM targets found, can't configure compilation of module-info!")
+        }
+        jvmTargets.forEach { target ->
+            target.compilations.forEach { compilation ->
+                val defaultSourceSet = compilation.defaultSourceSet.kotlin
+                val moduleInfoSourceFile = defaultSourceSet.find { it.name == "module-info.java" }
+
+                if (moduleInfoSourceFile == null) {
+                    logger.info("No module-info.java file found in ${defaultSourceSet.srcDirs}, can't configure compilation of module-info!")
+                } else {
+                    val targetName = toCamelCase(target.targetName)
+                    val compilationName = if (compilation.name != KotlinCompilation.MAIN_COMPILATION_NAME) toCamelCase(compilation.name) else ""
+                    val compileModuleInfoTaskName = "compile${compilationName}ModuleInfo$targetName"
+                    val checkModuleInfoTaskName = "check${compilationName}ModuleInfo$targetName"
+
+                    val compileKotlinTask = compilation.compileKotlinTask as AbstractCompile
+                    val modulePath = compileKotlinTask.classpath
+                    val moduleInfoClassFile = compileKotlinTask.destinationDirectory.file("module-info.class").get().asFile
+
+                    val compileModuleInfoTask = registerCompileModuleInfoTask(compileModuleInfoTaskName, modulePath, compileKotlinTask.destinationDirectory, moduleInfoSourceFile)
+                    tasks.getByName(compilation.compileAllTaskName).dependsOn(compileModuleInfoTask)
+
+                    val checkModuleInfoTask = registerCheckModuleInfoTask(checkModuleInfoTaskName, modulePath, moduleInfoClassFile)
+                    checkModuleInfoTask.configure { dependsOn(compilation.compileAllTaskName) }
+                    tasks.getByName("check").dependsOn(checkModuleInfoTask)
+                }
+            }
+
+            if (multiRelease) {
+                tasks.getByName<Jar>(target.artifactsTaskName) {
+                    rename("module-info.class", "META-INF/versions/9/module-info.class")
+                    manifest {
+                        attributes("Multi-Release" to true)
+                    }
+                }
+            }
+        }
+    }
+
+    private fun Project.registerCompileModuleInfoTask(taskName: String, modulePath: FileCollection, destinationDir: DirectoryProperty, moduleInfoSourceFile: File) =
+        tasks.register(taskName, JavaCompile::class) {
+            dependsOn(modulePath)
+            source(moduleInfoSourceFile)
+            classpath = files()
+            destinationDirectory.set(destinationDir)
+            sourceCompatibility = JavaVersion.VERSION_1_9.toString()
+            targetCompatibility = JavaVersion.VERSION_1_9.toString()
+            doFirst {
+                options.compilerArgs = listOf(
+                    "--release", "9",
+                    "--module-path", modulePath.asPath,
+                    "-Xlint:-requires-transitive-automatic"
+                )
+            }
+        }
+
+    private fun Project.registerCheckModuleInfoTask(taskName: String, modulePath: FileCollection, moduleInfoClassFile: File) =
+        tasks.register(taskName) {
+            dependsOn(modulePath)
+            doLast {
+                val jdeps = ToolProvider.findFirst("jdeps").orElseThrow { IllegalStateException("Tool 'jdeps' is not available") }
+                val moduleDescriptor = moduleInfoClassFile.inputStream().use { ModuleDescriptor.read(it) }
+                val moduleName = moduleDescriptor.name()
+                val expectedOutput = moduleDescriptor.toJdepsOutput(moduleInfoClassFile)
+
+                val outputCaptureStream = ByteArrayOutputStream()
+                val printStream = PrintStream(outputCaptureStream, true, Charsets.UTF_8)
+                jdeps.run(
+                    printStream, printStream,
+                    "--multi-release", "9",
+                    "--module-path", (modulePath + files(moduleInfoClassFile.parentFile)).asPath,
+                    "--check", moduleName
+                )
+                val actualOutput = outputCaptureStream.toString(Charsets.UTF_8).trim()
+
+                if (actualOutput != expectedOutput) {
+                    throw IllegalStateException("Module-info does not match!\n$actualOutput")
+                }
+            }
+        }
+
+    private fun ModuleDescriptor.toJdepsOutput(file: File, separator: String = System.lineSeparator()) =
+        "${name()} (${file.parentFile.toURI().toString().replace("file:", "file://")})$separator  [Module descriptor]$separator" +
+                requires().sortedBy { it.name() }.joinToString(separator) { requirement -> "    requires $requirement;" }
+}
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -9,6 +9,10 @@ repositories {
     mavenCentral()
 }
 
+dependencies {
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.30-RC")
+}
+
 kotlinDslPluginOptions {
     experimentalWarning.set(false)
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -42,3 +42,5 @@ tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
         )
     }
 }
+
+Java9Modularity.configureJava9ModuleInfo(project)
--- a/core/jvmMain/src/module-info.java
+++ b/core/jvmMain/src/module-info.java
@@ -1,0 +1,10 @@
+module kotlinx.serialization.core {
+    requires transitive kotlin.stdlib;
+
+    exports kotlinx.serialization;
+    exports kotlinx.serialization.builtins;
+    exports kotlinx.serialization.descriptors;
+    exports kotlinx.serialization.encoding;
+    exports kotlinx.serialization.internal;
+    exports kotlinx.serialization.modules;
+}
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,5 +1,10 @@
 # Building Kotlin Serialization from the source
 
+## JDK version
+
+To build Kotlin Serialization JDK version 9 or higher is required.
+This is needed to compile the `module-info` file included for JPMS support.
+
 ## Runtime library
 
 Kotlin Serialization runtime library itself is a [multiplatform](http://kotlinlang.org/docs/reference/multiplatform.html) project.
--- a/formats/cbor/build.gradle
+++ b/formats/cbor/build.gradle
@@ -29,3 +29,5 @@ kotlin {
         }
     }
 }
+
+Java9Modularity.configureJava9ModuleInfo(project)
--- a/formats/cbor/jvmMain/src/module-info.java
+++ b/formats/cbor/jvmMain/src/module-info.java
@@ -1,0 +1,6 @@
+module kotlinx.serialization.cbor {
+    requires transitive kotlin.stdlib;
+    requires transitive kotlinx.serialization.core;
+
+    exports kotlinx.serialization.cbor;
+}
--- a/formats/hocon/build.gradle
+++ b/formats/hocon/build.gradle
@@ -13,7 +13,7 @@ dependencies {
     compile project(':kotlinx-serialization-core')
     api 'org.jetbrains.kotlin:kotlin-stdlib'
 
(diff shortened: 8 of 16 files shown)


lion7:
After some more testing I observed that the Kotlin compiler does not check if the correct `requires <package>` is present when importing a class. The Java compiler does check this when it's running in 'module mode', which is automatically activated when a `module-info.java` is present. As a result of this behaviour I got `ClassNotFoundErrors` at runtime (instead of compile errors) because some of the needed `requires <package>` lines were missing in my `module-info.java` files.

To circumvent this problem altogether I took an alternative route where the `module-info.java` is generated by the `jdeps` tool. The `jdeps` tool can generate a `module-info.java` file by inspecting the imports of all class files in a (non-modular) JAR. For this to work I only had to change the Gradle build script. It basically comes down to the following steps:

- Create a `Jar` task that generates a non-modular JAR with the same contents / classes as the main `jvmJar` task.
- Use `jdeps` to generate a `module-info.java` based on this non-modular JAR.
- Create a `JavaCompile` task that compiles the generated file into a `module-info.class` file. Here I explicitly specify the module-path that should be used during compilation. I also had to patch the main classes dir so the compiler 'knows' which packages can be exported using `exports <package>` (I actually figured this out after finding [this post](https://discuss.kotlinlang.org/t/kotlin-support-for-java-9-module-system/2499/27) of @ilya-g).
- Add the compiled file as `META-INF/versions/9/module-info.class` to the main `jvmJar` task and mark it as a multi-release JAR.

build.gradle
```diff
@@ -175,3 +175,7 @@ subprojects {
 apply from: rootProject.file('gradle/compiler-version.gradle')
 apply from: rootProject.file("gradle/dokka.gradle")
 apply from: rootProject.file("gradle/benchmark-parsing.gradle")
+
+subprojects {
```
sandwwraith, line 179: Such clause will also affect utility projects like `benchmarks` and `guide`. Although it's not a big deal (they probably don't even have jar task), logically module configuration can belong here, where publishing itself is configured: https://github.com/Kotlin/kotlinx.serialization/blob/f7a7141fd8c6080484a1b37b5ada2525cf28369e/build.gradle#L154

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.gradle.api.tasks.compile.AbstractCompile
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.kotlin.dsl.register
+import java.util.spi.ToolProvider
+
+object Java9Modularity {
+    @JvmStatic
+    fun configureJava9ModuleInfo(project: Project) {
+        project.afterEvaluate {
+            val jdeps = ToolProvider.findFirst("jdeps").orElseThrow { IllegalStateException("Tool 'jdeps' is not available") }
```
sandwwraith, line 12: I'd throw this exception as late as possible (when the actual task `doLast` configured), to make sure that it won't disrupt normal local development without publishing

lion7: Good point, I moved the `orElseThrow` to the location where `jdeps` is actually invoked.

build.gradle
lion7: Yeah, those 2 projects don't have a `Jar` task so nothing happened there.
I moved the clause anyway, so it is called at a more logical place.

ilya-g (changes requested):
I believe it would be ok to use jdeps once to find the dependencies of a module, but after that the resulting module-info file should be committed in the repository and maintained manually in order to have control on what is being exported.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,63 @@
...
+                    destinationDirectory.set(classesDir)
+                    source(generatedDir)
+                    classpath = files()
+                    doFirst {
+                        val moduleName = generatedDir.listFiles().first().name
+                        options.compilerArgs = listOf(
+                            "--release", "9",
+                            "--module-path", compileKotlinJvm.classpath.asPath,
+                            "--patch-module", "$moduleName=${compileKotlinJvm.destinationDir}",
+                            "-Xlint:-requires-transitive-automatic"
```
ilya-g, line 47: This will fail (as expected) when running build on JDK 8. Do we have a manual about setting up the environment for building kotlinx.serialization? It may need a note about required JDKs and their location.

lion7: Ah, good point. Is that something I should add to the README?

lion7:
I could add the resulting `module-info` to the repo and configure only the compilation task, but that does have the downside that the imported packages are not checked by the Kotlin compiler. In other words, if you use a new package from a library and forget to add the needed `requires`, it would still all compile and package (but it will break at runtime).

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: I figured `docs/building.md` would be the correct place.
An alternative would be to only configure the compilation task of the `module-info` if the JDK is version 9 or higher.
Maybe with a warning that the resulting JAR is not JPMS compatible due to the used JDK?

ilya-g:
AFAIK, `requires` are specified not for packages, but for modules. So, it's not the case when you started using new package from a library, but rather when you add a dependency on a new library.

sandwwraith:
Is it possible to go with a similar to binary compatibility validator and apiCheck approach? I.e. add two tasks:

- `moduleInfoDump` will generate human-readable module-info.java to commit to repository
- `moduleInfoCheck` will check that no new dependencies were accidentally introduced

sandwwraith:
But it seems that we anyway have to use Java 9 to compile module-info.java even if it is committed to the repository? Or it can be simply ignored by JDK 8

lion7:
My mistake, `requires` is per module and not per package. But still, if you would add a dependency it would not break the compilation of the module-info, even if it is at that point incomplete.

As for the required JDK: to compile the module-info file you would need at least JDK 9. But since the module-info file is compiled separately you could choose to skip the compilation completely instead of throwing an exception. Of course the resulting JAR would not work with the Java 9 module system if that direction is chosen. 

My personal preference would be to make JDK 9+ mandatory for building (but not for running) Kotlin serialization. To support older JVM runtimes the compiled `module-info.class` file should be stored  in the multi-release way, so for example a JVM 8 runtime won't break when loading all class files.

lion7:
I do like the idea of having a separate moduleInfoCheck task that visually shows the repository version of the module-info vs a generated one. I think this is actually the purpose of jdeps 'check' command, I'll do a bit of research there.

lion7:
Per your suggestions I reworked my branch so it now has the (generated) `module-info` files stored in the repository.
I also reworked the `configureJava9ModuleInfo` method so it works correctly with Kotlin multiplatform JVM targets.

For checking the contents of the `module-info` I added a `jvmCheckModuleInfo` task. This uses `jdeps` under the hood to diff the compiled `module-info` with the file that `jdeps` would generate. 

This is the output when the contents of the `module-info` file is identical to the generated output:
```
> Task :kotlinx-serialization-core:jvmCheckModuleInfo
kotlinx.serialization.core (file:///C:/IdeaProjects/lion7/kotlinx.serialization/core/build/libs/kotlinx-serialization-core-jvm-1.2.2-SNAPSHOT.jar)
  [Module descriptor]
    requires mandated java.base;
    requires transitive kotlin.stdlib;
```


And this is the output when the contents of the `module-info` file is different (an extra, unused requires) to the generated output:
```
> Task :kotlinx-serialization-core:jvmCheckModuleInfo
kotlinx.serialization.core (file:///C:/IdeaProjects/lion7/kotlinx.serialization/core/build/libs/kotlinx-serialization-core-jvm-1.2.2-SNAPSHOT.jar)
  [Module descriptor]
    requires mandated java.base;
    requires transitive kotlin.stdlib;
    requires some.library;
  [Suggested module descriptor for kotlinx.serialization.core]
    requires mandated java.base;
    requires transitive kotlin.stdlib;
  [Transitive reduced graph for kotlinx.serialization.core]
    requires mandated java.base;
    requires transitive kotlin.stdlib;
```

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.kotlin.dsl.findByType
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
+import java.util.spi.ToolProvider
+
+object Java9Modularity {
+    @JvmStatic
+    fun Project.configureJava9ModuleInfo(moduleName: String) {
+        val multiplatform = extensions.findByType<KotlinMultiplatformExtension>() ?: return
```
sandwwraith, line 12: Will it work with `Hocon` format that is JVM-only?

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,65 @@
...
+            logger.warn("No Kotlin JVM targets found with Java enabled, can't configure Java 9 source set!")
+        }
+        jvmTargets.forEach { target ->
+            val mainCompilation = target.compilations.getByName("main")
+            val inputClasspath = mainCompilation.compileKotlinTask.classpath
+            val outputClassesDir = mainCompilation.compileKotlinTask.destinationDir
+            val outputJarTask = tasks.getByName(target.artifactsTaskName) as Jar
+
+            val java9SourceSet = multiplatform.sourceSets.create("${target.targetName}Java9")
+            val java9Compilation = target.compilations.create("java9") {
```
sandwwraith, line 24: Am I correct that this compilation is always created and compiled on every build?

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,65 @@
...
+                manifest {
+                    attributes(mapOf("Multi-Release" to true))
+                }
+            }
+
+            tasks.register("${target.targetName}CheckModuleInfo") {
+                dependsOn(outputJarTask)
+                doLast {
+                    val jdeps = ToolProvider.findFirst("jdeps").orElseThrow { IllegalStateException("Tool 'jdeps' is not available") }
+                    jdeps.run(
```
sandwwraith, line 55: It would be nice if we could detect if `jdeps` reports different dependencies from the current module-info and fail the task. This is needed to check automatically on CI that no new dependencies are accidentally introduced.

However, this may block us from removing some dependencies manually

formats/json/jvmJava9/src/module-info.java
```diff
@@ -0,0 +1,7 @@
+module kotlinx.serialization.json {
+    requires transitive kotlin.stdlib;
+    requires transitive kotlinx.serialization.core;
```
sandwwraith, line 3: Although json definitely uses some things from `kotlinx.serialization.internal` package, it isn't mentioned in 'requires' here. Can this cause issues?

If not, I'd prefer removing `exports kotlinx.serialization.*.internal` compiletely.

buildSrc/src/main/kotlin/Java9Modularity.kt
sandwwraith: I also don't see task to 'write' module-info. But from your comment I assume the same `CheckModuleInfo` task can be invoked and results copied from console, right?

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: No, I deliberately left modularity support out for the `Hocon` format because the structure of that (sub)project was quite different from the other formats. But it does make sense to add JPMS support for `Hocon` too.

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: Yes, if the method `configureJava9ModuleInfo` is called on a project It will always setup the sourceset and it's associated compilation.

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: To automatically check differences I could capture the (console) output that comes from `jdeps` and throw an exception if something along the lines of `[Suggested module descriptor for xxx]` is in the output. If you remove dependencies you would need to remove the appropriate `requires` line from the module-info too, which I believe to be the desired effect.

For 'writing' the module-info you would indeed need to copy the suggested module-descriptor manually from the console output into the repository version of the module-info.

formats/json/jvmJava9/src/module-info.java
lion7: This line requires the `kotlinx.serialization.core` module, not the package. Since the `kotlinx.serialization.core` module exports the `kotlinx.serialization.internal` package it can be used by `kotlinx.serialization.json`. 

As an alternative, we could export `kotlinx.serialization.internal` to only the format modules.

lion7:
I reworked the whole modularity configuration a bit. It now also supports 'regular' Kotlin projects such as the `Hocon` format. Another thing that has changed, is that the `module-info.java` file now resides in the main sourceset instead of in a separate sourceset. And instead of creating a full-blown compilation, now only a single `JavaCompile` task (`compileModuleInfo[Jvm]`) is registered to compile the relevant `module-info` file. This also prevents unwanted effects if someone uses `kotlin.sourceSets.forEach {}` or the equivalent for compilations.  

A nice side-effect of the dedicated `JavaCompile` task is that I don't have to know the Java module name up front. The `checkModuleInfo` task does need to know the module name, but that task can simply parse the (by that time) compiled module descriptor and retrieve the name from there. The same `checkModuleInfo` task now also throws an exception if the module descriptor is different compared to what `jdeps` reports. This check has become quite strict, I was a bit unsure if I should keep the exact matching of the output or that pattern matching would be better suitable there..

Lastly, I couldn't remove the export of `kotlinx.serialization.internal` in the `core` module since it's used by the formats, but I could remove the exports of internal packages in all the formats without breaking stuff.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,115 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.File
+import java.io.PrintStream
+import java.lang.module.ModuleDescriptor
+import java.util.spi.ToolProvider
+
+object Java9Modularity {
+
+    @JvmStatic
+    fun Project.configureJava9ModuleInfo(multiRelease: Boolean = false) {
```
sandwwraith, line 27: Shouldn't it be `true` by default? You don't pass `multiRelease` explicitly anywhere

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.gradle.api.file.FileCollection
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.api.tasks.compile.AbstractCompile
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.kotlin.dsl.attributes
+import org.gradle.kotlin.dsl.findByType
+import org.gradle.kotlin.dsl.getByName
+import org.gradle.kotlin.dsl.register
+import org.gradle.util.GUtil.toCamelCase
+import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
```
sandwwraith, line 13: We prefer star imports in the code style

sandwwraith (approved):
Looks good to me (modulo comments). I'm still not sure whether need to put module-info in a separate compilation (as in previous version) or just as a separate task (current), but it seems it doesn't matter much

sandwwraith:
NB: I've changed base branch to `dev`, but it seems that now your branch contains unrelated commits. Can you please rebase your PR on latest `dev`?

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: Correct, changed that.

buildSrc/src/main/kotlin/Java9Modularity.kt
lion7: Done

lion7:
I rebased my branch on `dev ` as suggested and force-pushed it. I also updated the Hocon config library to a version that better supports JPMS, see https://github.com/lightbend/config/issues/546. Could you double check?

sandwwraith:
It looks like module-info.java in the main source set causes issues in binary compatibility validator: https://github.com/Kotlin/binary-compatibility-validator/issues/68

So I think I'll merge this after that issue would be resolved

sandwwraith:
I've checked resulting jars — it has `META-INF/versions/9/module-info.class` with major version 53 and all other classes have major version 50, as intended

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,108 @@
...
+                jdeps.run(
+                    printStream, printStream,
+                    "--multi-release", "9",
+                    "--module-path", (modulePath + files(moduleInfoClassFile.parentFile)).asPath,
+                    "--check", moduleName
+                )
+                val actualOutput = outputCaptureStream.toString(Charsets.UTF_8).trim()
+
+                if (actualOutput != expectedOutput) {
+                    throw IllegalStateException("Module-info does not match!\n$actualOutput")
```
sandwwraith, line 100:
```suggestion
                    throw IllegalStateException("Module-info requierements section does not match!\n$actualOutput")
```

Since we check only requirements, not exports, it's better to highlight this

sandwwraith:
Many thanks for your efforts!

lion7:
You're welcome. I'm planning to create a PR for `kotlinx.datetime` too, as I'm using both in my projects.

sandwwraith:
@lion7 Have you opened the project jvmMain souceset in the IDE? I'm now experiencing problems with it: https://youtrack.jetbrains.com/issue/KTIJ-19614
Maybe a separate sourceset for module-info was a better idea.

lion7:
Follow up is in https://github.com/Kotlin/kotlinx.serialization/pull/1665

PR #1629, opened 2021-08-09
Update descriptors doc

martinbonnin:


qwwdfsad (commented):
Could you please elaborate on this change? 
Why do you think it's worth replacing concise API designed especially for descriptor builders by a more verbose one?

martinbonnin:
I didn't find the symbols 🤔 . Let me check again...

martinbonnin:
Ah! Looks like `element<>` is not released yet? I'm using 1.2.2 and the changes seem relatively recent. Feel free to close this issue.

martinbonnin:
For the record, I'd say that I would maybe favor consistency over shorter syntax here. It's weird having to pass a descriptor for classes and not for primitive types. But also I'm like super new to this API so not the best judge here.

PR #1638, opened 2021-08-17
Duct-taped prototype of JSON path

qwwdfsad:
* Probably has use-after-free
    * Does not work with polymorphism
    * Does not work with JsonTransformingSerializer
    * Inline classes support status -- unknown
    * Is not tested

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
@@ -1,0 +1,81 @@
+package kotlinx.serialization.json.internal
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+
+internal class JsonPath {
+    /*
+     * Serial descriptor OR map key.
+     * Map key can also implement SD, so it also has '-2' as current index
+     */
+    private var currentObjectPath = arrayOfNulls<Any?>(8)
+    private var currentIndices = IntArray(8)
+    private var currentDepth = -1
+
+    fun pushObject(sd: SerialDescriptor) {
+        if (++currentDepth == currentObjectPath.size) {
+            resize()
+        }
+        currentObjectPath[currentDepth] = sd
+        currentIndices[currentDepth] = -1
+    }
+
+    fun updateCurrentMapKey(key: Any?) {
+        if (currentIndices[currentDepth] != -2 && ++currentDepth == currentObjectPath.size) {
+            resize()
+        }
+        currentObjectPath[currentDepth] = key
+        currentIndices[currentDepth] = -2
+    }
+
+    fun updateCurrentIndex(index: Int) {
+        currentIndices[currentDepth] = index
+    }
+
+    fun pop() {
+        // When we are ending map, we pop the last key and the outer field as well
+        if (currentIndices[currentDepth] == -2) {
+            currentIndices[currentDepth] = -1
+            currentDepth--
+        }
+        // Against top-level maps
+        if (currentDepth != -1) currentDepth--
+    }
+
+    @OptIn(ExperimentalSerializationApi::class)
+    fun currentPath(): String {
+        if (currentDepth == -1) return ""
+        return buildString {
+            append("Root type: ")
+            append(currentObjectPath[0])
+            append(", leaf type: ")
+            append(currentObjectPath[currentDepth])
+            append(", \npath: ")
+            repeat(currentDepth + 1) {
+                val element = currentObjectPath[it]
+                if (element is SerialDescriptor) {
+                    if (element.kind == StructureKind.LIST) {
+                        append("[")
+                        append(currentIndices[it])
+                        append("]")
+                    } else {
+                        val idx = currentIndices[it]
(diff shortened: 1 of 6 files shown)


qwwdfsad:
This is a no-production ready prototype, it would be nice if you could take a look at the approach to see if it makes sense to continue pushing it further.

The current performance hit of the decoding is around 5%

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -74,12 +76,30 @@ internal open class StreamingJsonDecoder(
...
 
+    override fun <T> decodeSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        deserializer: DeserializationStrategy<T>,
+        previousValue: T?
+    ): T {
+        // TODO kludge for Map
+        val value = super.decodeSerializableElement(descriptor, index, deserializer, previousValue)
+        if (mode == WriteMode.MAP && index == 0) {
```
sandwwraith, line 87: Shouldn't it be `if (index % 2 == 0)`?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
```diff
@@ -0,0 +1,81 @@
+package kotlinx.serialization.json.internal
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+
+internal class JsonPath {
+    /*
+     * Serial descriptor OR map key.
+     * Map key can also implement SD, so it also has '-2' as current index
```
sandwwraith, line 9: `Map key can also implement SD` - how? Serial descriptor is not `@Serializable`

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    fun updateCurrentIndex(index: Int) {
+        currentIndices[currentDepth] = index
+    }
+
+    fun pop() {
+        // When we are ending map, we pop the last key and the outer field as well
+        if (currentIndices[currentDepth] == -2) {
+            currentIndices[currentDepth] = -1
+            currentDepth--
```
sandwwraith, line 39: Also not sure why we should reset `currentIndices[currentDepth]` if on next `pushObject/updateCurrentMapKey` we overwrite this value

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -74,12 +76,30 @@ internal open class StreamingJsonDecoder(
...
+
     override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
-        return when (mode) {
+        val index = when (mode) {
             WriteMode.OBJ -> decodeObjectIndex(descriptor)
             WriteMode.MAP -> decodeMapIndex()
             else -> decodeListIndex() // Both for LIST and default polymorphic
         }
+        if (mode != WriteMode.MAP) {
+            lexer.path.updateCurrentIndex(index)
```
sandwwraith, line 100: You can move this to `.also {}` in `WriteMode.OBJ` and `else ->`. Not sure if this would save any time though.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
sandwwraith: Also, I don't get why it's after decodeSerializableElement, not before

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
```diff
@@ -0,0 +1,81 @@
...
+        // Against top-level maps
+        if (currentDepth != -1) currentDepth--
+    }
+
+    @OptIn(ExperimentalSerializationApi::class)
+    fun currentPath(): String {
+        if (currentDepth == -1) return ""
+        return buildString {
+            append("Root type: ")
+            append(currentObjectPath[0])
```
sandwwraith, line 50: `SerialDescriptor.toString` also prints all properties, which may cause output to be bloated. I'd think we better stick with serial name only here.

sandwwraith (commented):
5% seems too much — changes are relatively simple. Is this from storing objects (SerialDescriptors) in an array or what?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonPath.kt
qwwdfsad: But serializable class that implements it is :)

This is a prototype, so there is no tests for that, but it may be the case anyway

neppug-Megvu2-wizqys:
> This is a not production ready prototype, it would be nice if you could take a look at the approach to see if it makes sense to continue pushing it further.
> 
> 
> 
> The current performance hit of the decoding is around 5%

PR #1651, opened 2021-08-29
Fix annotation name

gmulders:
The annotation mentioned in the docs was `@Serializer`, whereas from the context it is clear that is should be `@Serializable`.


sandwwraith (approved):
Indeed, you're right

PR #1656, opened 2021-09-03
Fix SerialFormat doc: Change `BinaryFormat` to `SerialFormat`

BenWoodworth:
Fixed this line in the SerialFormat interface's documentation:
```
It is safe to operate with instances of `BinaryFormat` and call its methods.
                                         ^^^^^^^^^^^^
```
and changed it to SerialFormat.


sandwwraith (approved):

sandwwraith:
Thanks!

PR #1665, opened 2021-09-07
Use a separate source set for compiling the module descriptor to prevent IDE issues

lion7:
IDE issues described in https://youtrack.jetbrains.com/issue/KTIJ-19614


lion7:
@sandwwraith this should resolve your highlighting issues, could you double check the JAR's themselves are not changed?

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -24,85 +19,84 @@ object Java9Modularity {
...
         jvmTargets.forEach { target ->
+            val artifactTask = tasks.getByName<Jar>(target.artifactsTaskName) {
+                if (multiRelease) {
+                    manifest {
+                        attributes("Multi-Release" to true)
+                    }
+                }
+            }
+
             target.compilations.forEach { compilation ->
```
sandwwraith, line 30: Won't this also affect `test` compilation?

lion7: Yes, it always did affect the `test` compilation (in case you want a black-box test where the `test` compilation is a module of its own. But it only actually does something if a `module-info.java` is present in the test sources..

sandwwraith:
Yeah JARs still contain `META-INF/versions/9/module-info.class`. Thanks again!

PR #1667, opened 2021-09-10
doc: proofread json.md

p7nov:


docs/json.md
```diff
@@ -804,27 +811,27 @@ manually, using only the [KSerializer] class.
...
-There are several tidbits on custom serializers with [Json].
+Here are some useful things about custom serializers with [Json]:
 
 * [Encoder] can be cast to [JsonEncoder], and [Decoder] to [JsonDecoder], if the current format is [Json].
 * `JsonDecoder` has the [decodeJsonElement][JsonDecoder.decodeJsonElement] method and `JsonEncoder` 
-  has the [encodeJsonElement][JsonEncoder.encodeJsonElement] method.
-  which basically retrieve/insert an element from/to a current position in the stream.
-* Both [`JsonDecoder`][JsonDecoder.json] and [`JsonEncoder`][JsonEncoder.json] have the `json` property 
+  has the [encodeJsonElement][JsonEncoder.encodeJsonElement] method,
+  which basically retrieve and insert an element from/to a current position in the stream.
```
sandwwraith, line 819: `retrieve/insert` was correct. 'retrieve' corresponds to 'decodeJsonElement' and 'from', 'insert' to 'encodeJsonElement' and 'to'

sandwwraith (commented):
Most of the guide was specifically written in the way it uses 'we' instead of 'you'. Do you find the latter more preferable style?

p7nov:
> Most of the guide was specifically written in the way it uses 'we' instead of 'you'. Do you find the latter more preferable style?

Yes, it's our Kotlin-wide writing guideline (see [here](https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit#heading=h.n0wijcuj1poc). When we show what a user should or can do, we always use second person.
"We" stands only for the Kotlin team as a whole, i.e. in wordings like "we're presenting a feature".

We (tech writing team) know that "we" is still being used, mostly in the libraries' docs. We're going to rewrite all such texts eventually to make everything consistent.

docs/json.md
p7nov: Yeah, I see. But the use of `/` in place of conjunctions is generally a bad style because it's hard to parse for non-native speakers. 
Anyway, rewrote this in the clearest way possible :)

sandwwraith:
@p7nov Style guide also mentions active voice. Out of curiosity, why this was chosen? Most of the documentation and technical writing in the world is written using passive voice

p7nov:
> @p7nov Style guide also mentions active voice. Out of curiosity, why this was chosen? Most of the documentation and technical writing in the world is written using passive voice

Short answer: 
- unlike passive, active is understandable for international audiences of different backgrounds. Most languages in the world don't even have this category.
- Active writing is better in terms of style - you always know who does what.

Most big software companies who have their established writing styles avoid passive where possible: [Apple](https://books.apple.com/us/book/apple-style-guide/id1161855204), [Google](https://developers.google.com/tech-writing/one/active-voice), [MS](https://docs.microsoft.com/en-us/style-guide/grammar/verbs), [JetBrains](https://confluence.jetbrains.com/pages/viewpage.action?title=JetBrains+Writing+Style+Guide&spaceKey=MRKT#JetBrainsWritingStyleGuide-Activevoice) :) 

Passive is widely used in scientific and highly technical content for professionals with a specific background, which is not exactly our case.

Issue #1633, opened 2021-08-12
Provide ByteArrayAsBase64StringSerializer


sandwwraith:
See the https://github.com/Kotlin/kotlinx.serialization/issues/1533 for context

cy6erGn0m:
@sandwwraith Where do you recommend to mention it in the docs?

sandwwraith:
`LongAsStringSerializer` is mentioned in builtin classes (https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/builtin-classes.md#long-numbers-as-strings), although I think better place for base64 serializer is somewhere at the end of JSON settings

GeorgePap-719:
Hey @sandwwraith, is this still up for grabs?

sandwwraith:
@GeorgePap-719 I believe so, as the previous attempt (#1676) wasn't finished.

PR #1676, opened 2021-09-15
ByteArrayAsBase64StringSerializer implementation (#1633)

cy6erGn0m:
Open questions:
- [ ] How do I mention including the module in the docs?
- [ ] Where should I put the module: keep in formats or what?

### DIFF
--- a/formats/base64/commonMain/src/kotlinx.serialization.base64/ByteArrayAsBase64StringSerializer.kt
+++ b/formats/base64/commonMain/src/kotlinx.serialization.base64/ByteArrayAsBase64StringSerializer.kt
@@ -1,0 +1,25 @@
+package kotlinx.serialization.base64
+
+import kotlinx.serialization.*
+import kotlinx.serialization.base64.impl.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+
+/**
+ * Serializer that encodes and decodes [ByteArray] using [Base64](https://en.wikipedia.org/wiki/Base64) encodings.
+ * This is usually makes sense with text formats like JSON.
+ */
+public object ByteArrayAsBase64StringSerializer : KSerializer<ByteArray> {
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor(
+            "kotlinx.serialization.ByteArrayAsBase64StringSerializer",
+            PrimitiveKind.STRING
+    )
+
+    override fun deserialize(decoder: Decoder): ByteArray {
+        return decode(decoder.decodeString())
+    }
+
+    override fun serialize(encoder: Encoder, value: ByteArray) {
+        encoder.encodeString(encode(value))
+    }
+}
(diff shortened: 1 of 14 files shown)


formats/base64/commonMain/src/kotlinx.serialization.base64/ByteArrayAsBase64StringSerializer.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.serialization.base64.impl.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+
+/**
+ * Serializer that encodes and decodes [ByteArray] using [Base64](https://en.wikipedia.org/wiki/Base64) encodings.
+ * This is usually makes sense with text formats like JSON.
+ */
+public object ByteArrayAsBase64StringSerializer : KSerializer<ByteArray> {
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor(
```
guness, line 13: I come up with the same idea today before I see this PR. So this is a good feature. On the other hand, can we find a shorter name?
i.e:
`ByteArrayBase64Serializer`
or `Base64BytesSerializer`
or `Base64Serializer`
or `ByteArraySerializer` (replacing existing solution.)

cy6erGn0m: Well, there are already similar serializers in the library with same naming schema so I see no reason to break it

guness: the only example I can find was `LongAsStringSerializer` among 30+ other serializers. Since `Base64` already implies `String` we can at least reduce it to `ByteArrayAsBase64Serializer`. Anyway, it is up to you. It is your contribution(which I like), mine is just a mere suggestion.

ryanholden8:
We are facing the same need for this since our server end only works with Base64 binary data in JSON.

Thank you for your guys work on this. Look forward to using it soon!

sandwwraith:
It seems that the development of this has stalled. @cy6erGn0m Do you have plans to pick it up, or maybe you can give away your unfinished work to someone else?

cy6erGn0m:
> It seems that the development of this has stalled. @cy6erGn0m Do you have plans to pick it up, or maybe you can give away your unfinished work to someone else?

Don't know what happened, let me update this and see what state it's at. 

As far as I understand, this is still makes sense, right ?

sandwwraith:
Yes, I think it still makes sense to have such serializer in the core library

ryanholden8:
Any update on this PR? We've copied the implementation to a few projects and it works great. Base64 with JSON is a very common situation.

sandwwraith:
cc @cy6erGn0m

sandwwraith:
By the way, there would be built-in MPP support for Base64 in Kotlin 1.8.20: https://kotlinlang.org/docs/whatsnew-eap.html#experimental-support-for-base64-encoding-in-standard-library

PR #1688, opened 2021-09-21
Remove @ExperimentalSerializationApi from base serial format interfaces

### LINKED ISSUE #1668: Kotlin 1.5.30: Leaking @ExperimentalSerializationApi in stable APIs
**Describe the bug**
Some stable APIs are being reported as experimental.

**To Reproduce**
```kotlin
import kotlinx.serialization.builtins.serializer

Json.encodeToString(Int.serializer(), 0) // No warnings

Json.encodeToString(0)
//   ^^^^^^^^^^^^^^ 
//   This declaration is experimental and its usage should be marked with
//   '@kotlinx.serialization.ExperimentalSerializationApi' or
//   '@OptIn(kotlinx.serialization.ExperimentalSerializationApi::class)'
```

**Expected behavior**
No opt-in warnings on stable APIs

**Environment**
 - Kotlin version: 1.5.30
 - Library version: 1.2.2, 1.3.0-RC
 - Kotlin platforms: JVM (probably others)
 - Gradle version: 6.8

sandwwraith:
because they are unstable for inheritance only, and experimentality is now leaking in function signatures

Fixes #1668

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/SerialFormat.kt
+++ b/core/commonMain/src/kotlinx/serialization/SerialFormat.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization
@@ -26,7 +26,6 @@ import kotlinx.serialization.modules.*
  *
  * It is safe to operate with instances of `SerialFormat` and call its methods.
  */
-@ExperimentalSerializationApi
 public interface SerialFormat {
     /**
      * Contains all serializers registered by format user for [Contextual] and [Polymorphic] serialization.
@@ -46,7 +45,6 @@ public interface SerialFormat {
  *
  * It is safe to operate with instances of `BinaryFormat` and call its methods.
  */
-@ExperimentalSerializationApi
 public interface BinaryFormat : SerialFormat {
 
     /**
@@ -70,7 +68,6 @@ public interface BinaryFormat : SerialFormat {
  *
  * It is safe to operate with instances of `StringFormat` and call its methods.
  */
-@ExperimentalSerializationApi
 public interface StringFormat : SerialFormat {
 
     /**


sandwwraith:
We also may need to revisit experimental annotation on `SerialDescriptor` members (AFAIR, it was for inheritance too).

Also, `JsonConfiguration` looks like it may be stable already.

qwwdfsad (changes requested):
Please add a comment that these interfaces are experimental for implementation

sandwwraith:
It's been there all along: https://github.com/Kotlin/kotlinx.serialization/blob/82134bd7cbca9da53ce2f73bd0047b96c76e47db/core/commonMain/src/kotlinx/serialization/SerialFormat.kt#L22

BenWoodworth:
I don't think it's necessary (documenting that 3rd-party formats aren't stable seems like enough), just something I've been pondering the past week. But conceivably something like this could be used if you still want the implementing classes to explicitly opt in, bringing the instability front-and-center to the format dev:

```kotlin
interface SerialFormat {
    @ExperimentalSerializationApi
    @Suppress("PropertyName")
    @Deprecated(
        "SerialFormat interface is not stable for inheritance in 3rd party libraries, " +
                "as new methods might be added to this interface or contracts of the existing methods can be changed.",
        level = DeprecationLevel.HIDDEN,
    )
    val SerialFormat_interface_is_not_stable_for_inheritance: Unit // or maybe Nothing

    // ...
}

class ThirdPartySerialFormat : SerialFormat {
    @ExperimentalSerializationApi
    override val SerialFormat_interface_is_not_stable_for_inheritance: Unit = Unit
}
```

I don't think this would be a breaking change with existing 3rd party formats at runtime (e.g. no `NoSuchMethodError` since the property can never be accessed), but you would of course need to add the override in when compiling against kxs 1.3.0.

qwwdfsad (approved):

sandwwraith:
@BenWoodworth Idea is good, and I think it was used in other parts somewhere (or in kotlinx.coroutines), but It looks like an overkill for this task. Interfaces are small, and usually inherited by format authors who read the documentation and know what's what

BenWoodworth:
Hah I figured. Cool to know it's been used elsewhere, and glad to see this merged! :)

PR #1690, opened 2021-09-21
Promote JsonConfiguration and its usages to stable

qwwdfsad:


sandwwraith:
It seems we need `@OptIn(ExperimentalSerializationApi::class)` on `JsonConfiguration.toString()` because it uses still experimental `prettyPrintIndent` and `explicitNulls`

sandwwraith (approved):

Issue #1662, opened 2021-09-06
Allow to decode more than one object from InputStream


sandwwraith:
This is a meta issue for collecting various use-cases for designing an API that will allow to actually 'stream' objects from `InputStream`, instead of decoding a single one using `Json.decodeFromStream`. 

Please describe your use case in detail, if this is an HTTP long-polling stream or WebSocket, first-party or third-party API, or just a very large file.

chris-hatton:
Really happy to see this question being asked.  I'm happy to give a concrete example:

I'm implementing an App where a Mobile Device or Desktop client scans food product barcodes and we want to look up a name, description and photos of the product from upstream, 3rd party services.
The general state of food product barcodes, is that no single 'source of truth' API exists; it's necessary to integrate with multiple 3rd party services to achieve good coverage across a typical mix of groceries.
So, my back-end takes the barcode request from the client and will broadcast this request to multiple upstream API's concurrently.
Now, these services each have their own varied response times, and I wish to update my client as each response comes in, as soon as possible, to achieve a well-performing user interface.  It therefore makes sense to me, that the response from my back-end server to the client, be a relatively long-lived HTTP stream upon which I will 'stream' several query responses, one after the other, until all upstream backends have provided their answers, then I'll close the response to the client.

Sorry if that was longer-winded than you were looking for; but maybe something about it helps.
In more concise technical language; I want to be able to issue a request to my back end, and have several asynchronous processes provide answers to the client within the same HTTP response, in the most timely manner possible; and that means streaming several complete JSON objects on the same stream, closing the stream after the last one is transmitted.

rgmz:
Our use case is ingesting >1GB JSON files provided by a vendor, read from the file system. For example:
```json
{
    "items": [
        { ... }
    ]
}
```

Presently, we're using [Vert.x's](https://vertx.io) [JsonParser](https://vertx.io/docs/apidocs/io/vertx/core/parsetools/JsonParser.html) utility; however, being able to do this with Kotlin Serialization would be great. 

<details>
<summary>Vert.x JsonParser example</summary>

```kotlin
package foo;
​
import io.vertx.core.json.JsonObject
import io.vertx.core.parsetools.JsonEventType
import io.vertx.core.parsetools.JsonParser
import io.vertx.kotlin.core.file.openOptionsOf
import io.vertx.kotlin.coroutines.CoroutineVerticle
import io.vertx.kotlin.coroutines.await
​
/**
[...]

PR #1691, opened 2021-09-22
Prototype decodeToSequence to read multiple objects from stream lazily

sandwwraith:
See https://github.com/Kotlin/kotlinx.serialization/issues/1662

### DIFF
--- a/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
+++ b/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
@@ -94,30 +94,42 @@ public enum class LazyStreamingFormat {
     /**
      * Declares that objects in the input stream are separated by whitespace characters.
      *
-     * Stream must start with a first object and there are some (maybe none) whitespace chars between objects.
+     * The stream is read as multiple JSON objects separated by any number of whitespace characters between objects.
+     * Each individual object is parsed lazily, when it is requested from the resulting sequence.
+     *
      * Whitespace character is either ' ', '\n', '\r' or '\t'.
+     *
+     * Example of `WHITESPACE_SEPARATED` stream content:
+     * ```
+     * """{"key": "value"}{"key": "value2"}   {"key2": "value2"}"""
+     * ```
      */
     WHITESPACE_SEPARATED,
 
     /**
-     * Declares that objects in the input stream are wrapped in the JSON array. Elements of the array are still parsed lazily,
-     * so there shouldn't be problems if the array total size is larger than available memory.
+     * Declares that objects in the input stream are wrapped in the JSON array.
+     * Each individual object in the array is parsed lazily when it is requested from the resulting sequence.
      *
-     * Stream must start with json array start character and objects in it must be separated with commas and optional whitespaces.
-     * Stream must end with array end character, otherwise, [JsonDecodingException] would be thrown.
-     * Dangling chars after array end are not permitted.
+     * The stream is read as multiple JSON objects wrapped into a JSON array.
+     * The stream must start with an array start character `[` and end with an   array end character `]`,
+     * otherwise, [JsonDecodingException] is thrown.
+     *
+     * Example of `ARRAY_WRAPPED` stream content:
+     * ```
+     * """[{"key": "value"}, {"key": "value2"},{"key2": "value2"}]"""
+     * ```
      */
     ARRAY_WRAPPED,
 
     /**
      * Declares that parser itself should select between [WHITESPACE_SEPARATED] and [ARRAY_WRAPPED] modes.
-     * Selection is performed by looking on the first meaningful character of the stream.
+     * The selection is performed by looking on the first meaningful character of the stream.
      *
      * In most cases, auto-detection is sufficient to correctly parse an input.
-     * However, in the input is _whitespace-separated stream of the arrays_, parser would select incorrect mode.
-     * For such cases, [LazyStreamingFormat] must be specified explicitly.
+     * If the input is _whitespace-separated stream of the arrays_, parser could select an incorrect mode,
+     * for that [LazyStreamingFormat] must be specified explicitly.
      *
-     * Example of exceptional case:
+     * Example of an exceptional case:
      * `[1, 2, 3]   [4, 5, 6]\n[7, 8, 9]`
      */
     AUTO_DETECT;
@@ -129,11 +141,11 @@ public enum class LazyStreamingFormat {
  *
  * Elements must all be of type [T].
  * Elements are parsed lazily when resulting [Sequence] is evaluated.
- * Resulting sequence is tied to the stream and constrained to be evaluated only once.
+ * Resulting sequence is tied to the stream and can be evaluated only once.
  *
- * **Resource caution:** this method does not close [stream] when the parsing is finished neither provides method to close it manually.
+ * **Resource caution:** this method neither closes the [stream] when the parsing is finished nor provides a method to close it manually.
  * It is a caller responsibility to hold a reference to a stream and close it. Moreover, because stream is parsed lazily,
- * closing it before returned sequence is evaluated fully would result in [IOException] from decoder.
+ * closing it before returned sequence is evaluated completely will result in [IOException] from decoder.
  *
  * @throws [SerializationException] if the given JSON input cannot be deserialized to the value of type [T].
  * @throws [IOException] If an I/O error occurs and stream can't be read from.
--- a/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
+++ b/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
@@ -94,7 +94,7 @@ public enum class DecodeSequenceMode {
     /**
      * Declares that objects in the input stream are separated by whitespace characters.
      *
-     * The stream is read as multiple JSON objects separated by any number of whitespace characters between objects.
+     * The stream is read as multiple JSON objects separated by any number of whitespace characters between objects. Starting and trailing whitespace characters are also permitted.
      * Each individual object is parsed lazily, when it is requested from the resulting sequence.
      *
      * Whitespace character is either ' ', '\n', '\r' or '\t'.
--- a/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
+++ b/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
@@ -111,7 +111,7 @@ public enum class DecodeSequenceMode {
      * Each individual object in the array is parsed lazily when it is requested from the resulting sequence.
      *
      * The stream is read as multiple JSON objects wrapped into a JSON array.
-     * The stream must start with an array start character `[` and end with an   array end character `]`,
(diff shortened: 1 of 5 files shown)


formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
```diff
@@ -0,0 +1,100 @@
...
+private class JsonIteratorWsSeparated<T>(
+    private val json: Json,
+    private val lexer: ReaderJsonLexer,
+    private val deserializer: DeserializationStrategy<T>
+) : Iterator<T> {
+    override fun next(): T =
+        StreamingJsonDecoder(json, WriteMode.OBJ, lexer, deserializer.descriptor)
+            .decodeSerializableValue(deserializer)
+
+    override fun hasNext(): Boolean = lexer.isNotEof()
```
qwwdfsad, line 62: This implementation has inconsistent restrictions with both itself and `decodeFromString` -- the stream can start with an arbitrary number of whitespace, but cannot end with anything but EOF.
Leading and trailing WSes should be allowed in every mode

formats/json/jvmTest/src/kotlinx/serialization/features/JsonStreamFlowTest.kt
```diff
@@ -4,16 +4,17 @@
...
+import kotlinx.serialization.json.internal.JsonDecodingException
 import kotlinx.serialization.test.assertFailsWithMessage
-import org.junit.Ignore
 import org.junit.Test
 import java.io.*
-import kotlin.test.assertEquals
+import kotlin.test.*
 
 class JsonStreamFlowTest {
     val json = Json {}
```
qwwdfsad, line 20: Should be just `Json`, IDEA highlights that

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
```diff
@@ -0,0 +1,100 @@
...
+            lexer.consumeNextToken(TC_END_LIST)
+            if (lexer.isNotEof()) {
+                if (lexer.peekNextToken() == TC_BEGIN_LIST) lexer.fail("There is a start of the new array after the one parsed to sequence. " +
+                        "${LazyStreamingFormat.ARRAY_WRAPPED.name} mode doesn't merge consecutive arrays.\n" +
+                        "If you need to parse a stream of arrays, please use ${LazyStreamingFormat.WHITESPACE_SEPARATED.name} mode instead.")
+                lexer.expectEof()
+            }
+            return false
+        }
+        // We may allow unclosed bracket on .isLenient, but it's usually not what this flag do
```
qwwdfsad, line 96: This comment is redundant

formats/json/jvmTest/src/kotlinx/serialization/features/JsonStreamFlowTest.kt
```diff
@@ -24,7 +25,16 @@ class JsonStreamFlowTest {
...
+    private suspend inline fun <reified T> Json.readFromStream(iss: InputStream): Flow<T> = flow {
+        val serial = serializer<T>()
+        val iter = iterateOverStream(iss, serial)
+        while (iter.hasNext()) {
+            emit(iter.next())
+        }
+    }.flowOn(Dispatchers.IO)
+
+    val inputStringWsSeparated = """{"data":"a"}{"data":"b"}{"data":"c"}"""
+    val inputStringWrapped = """[{"data":"a"},{"data":"b"},{"data":"c"}]"""
```
qwwdfsad, line 37: All these can be private

qwwdfsad (changes requested):

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -80,3 +80,93 @@ public fun <T> Json.decodeFromStream(
...
+     * Declares that objects in the input stream are not separated by meaningful characters.
+     *
+     * Stream must start with a first object and there are some (maybe none) whitespace chars between objects.
+     * Whitespace character is either ' ', '\n', '\r' or '\t'.
+     */
+    WHITESPACE_SEPARATED,
+
+    /**
+     * Declares that objects in the input stream are wrapped in the array. Elements of the array are still parsed lazily,
+     * so there shouldn't be problems if array total size is larger than application memory.
```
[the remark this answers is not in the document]
pdvrieze: typo in "avilable" should be "available"

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -80,3 +80,93 @@ public fun <T> Json.decodeFromStream(
...
+/**
+ * Description of [decodeToSequence]'s JSON input shape.
+ *
+ * The sequence represents a stream of objects parsed one by one;
+ * [LazyStreamingFormat] defines a separator between these objects.
+ * Normally, these objects are not separated by meaningful characters ([WHITESPACE_SEPARATED])
+ * or the whole stream is a large array of objects separated with commas ([ARRAY_WRAPPED]).
+ */
+@ExperimentalSerializationApi
+public enum class LazyStreamingFormat {
```
pdvrieze, line 93: I'm wondering whether the name should include Json as an indicator that this is specific to Json. Alternatively a generic class (not in the json package) may make sense to be used by multiple formats (but should then be defined a bit more generically). The issue I'm considering is multiple formats in the same code base where both formats support stream decoding (eg. ktor content negotiation).

formats/json/jvmTest/src/kotlinx/serialization/features/JsonStreamFlowTest.kt
```diff
@@ -36,16 +46,116 @@ class JsonStreamFlowTest {
...
     @Test
     fun testDecodeSeveralItems() {
-        val ins = ByteArrayInputStream(inputString.encodeToByteArray())
+        val ins = ByteArrayInputStream(inputStringWsSeparated.encodeToByteArray())
         assertFailsWithMessage<SerializationException>("EOF") {
             json.decodeFromStream<StringData>(ins)
         }
     }
 
+    private inline fun <reified T> Iterator<T>.assertNext(expected: T) {
```
qwwdfsad, line 60: Seems like this functionality might have slightly more tests. WDYT about adding `JsonTestingMode` for streams?
AUTO and WS can be tested as is, and for `ARRAY` we can just wrap the strings into `[]`

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
sandwwraith: The name is not final for now, so we can come up with a better one. Although I don't think that making it generic would be a good idea because whitespace vs array wrapping looks like the choice only for json.

formats/json/jvmTest/src/kotlinx/serialization/features/JsonStreamFlowTest.kt
sandwwraith: I don't think it will give benefits. Underlying parsing logic is the same as in `JsonTestingMode.JAVA_STREAMS`, the only difference we need to test is object separators. Since all of the tests with `JsonTestingMode` parse one object at a time, separators will likely be untested.

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
qwwdfsad: @pdvrieze it could be done, but alas, it still will be JSON-centric as our main target here.
Making it more generalized will only confuse `Json` format users, but will bring insignificant benefits for other formats.
Unlike, for example, `SerialDescriptor`, there shouldn't be much value in sharing this mode between e.g. XML and JSON formats

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -80,3 +80,93 @@ public fun <T> Json.decodeFromStream(
 @ExperimentalSerializationApi
 public inline fun <reified T> Json.decodeFromStream(stream: InputStream): T =
     decodeFromStream(serializersModule.serializer(), stream)
+
+/**
+ * Description of [decodeToSequence]'s JSON input shape.
+ *
+ * The sequence represents a stream of objects parsed one by one;
+ * [LazyStreamingFormat] defines a separator between these objects.
+ * Normally, these objects are not separated by meaningful characters ([WHITESPACE_SEPARATED])
```
qwwdfsad, line 89:
```suggestion
 * Typically, these objects are not separated by meaningful characters ([WHITESPACE_SEPARATED]),
```

qwwdfsad (approved):
Good to go! 🔥 

Please proofread my documentation changes prior to merge

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
pdvrieze: @qwwdfsad My main intuition is that the format would probably be format specific. But the concept would work for XML (either separate xml messages - I know this is technically nonconformant - that never stops people from wanting to parse/generate almostXML, or a wrapper "list"). The current name (in a different package) would also make sense in the XML case (and other formats), which means that systems that support both have naming conflicts (renaming imports would fix this).

As a different aside, perhaps **`...Format`** has challenges given what normally format means in the context of the library.

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -80,3 +80,105 @@ public fun <T> Json.decodeFromStream(
...
+     * ```
+     */
+    WHITESPACE_SEPARATED,
+
+    /**
+     * Declares that objects in the input stream are wrapped in the JSON array.
+     * Each individual object in the array is parsed lazily when it is requested from the resulting sequence.
+     *
+     * The stream is read as multiple JSON objects wrapped into a JSON array.
+     * The stream must start with an array start character `[` and end with an   array end character `]`,
```
sandwwraith, line 114:
```suggestion
     * The stream must start with an array start character `[` and end with an array end character `]`,
```

formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt
```diff
@@ -80,3 +80,105 @@ public fun <T> Json.decodeFromStream(
...
+ * [LazyStreamingFormat] defines a separator between these objects.
+ * Typically, these objects are not separated by meaningful characters ([WHITESPACE_SEPARATED]),
+ * or the whole stream is a large array of objects separated with commas ([ARRAY_WRAPPED]).
+ */
+@ExperimentalSerializationApi
+public enum class LazyStreamingFormat {
+    /**
+     * Declares that objects in the input stream are separated by whitespace characters.
+     *
+     * The stream is read as multiple JSON objects separated by any number of whitespace characters between objects.
```
sandwwraith, line 97:
```suggestion
     * The stream is read as multiple JSON objects separated by any number of whitespace characters between objects. Starting and trailing whitespace characters are also permitted.
```

qwwdfsad (approved):

PR #1740, opened 2021-10-22
Hocon encoder

### LINKED ISSUE #1609: Encode an object into a Hocon config
It is possible to decode a Hocon config file using `Hocon.decodeFromConfig(config: Config): T`.

However, there seems to be no way to encode `T` back into a Hocon `Config` file.

osipxd:
Just started to work on Hocon encoder implementation and want to make my work "transparent" to kotlinx.serialization team.
If this feature is not needed I'll stop working on it, otherwise, any comments are welcome!

### Use cases

- **Generate default config.** For now, it is possible only to provide default config files from resources.
- **Edit config from an app.** This feature might be useful for apps having both text config and UI for configuration.

### Essentials

- [x] Primitives encoding
- [x] Lists encoding
- [x] Maps encoding
- [x] Nested objects encoding
- [x] Polymorphic types encoding
- [x] `useConfigNamingConvention` support
- [x] Reasonable exception messages
- [x] Add KDoc comments

### Nice to have

- [x] `encodeDefaults` support
- [ ] Annotation like `JsonClassDiscriminator`
- [ ] Add description (origin) to config values

Closes #1609

### DIFF
--- a/formats/hocon/build.gradle
+++ b/formats/hocon/build.gradle
@@ -12,17 +12,9 @@ compileKotlin {
     }
 }
 
-configurations {
-    apiElements {
-        attributes {
-            attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 8)
-        }
-    }
-    runtimeElements {
-        attributes {
-            attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 8)
-        }
-    }
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 }
 
 
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -35,6 +35,13 @@ public sealed class Hocon(
     public fun <T> decodeFromConfig(deserializer: DeserializationStrategy<T>, config: Config): T =
         ConfigReader(config).decodeSerializableValue(deserializer)
 
+    @ExperimentalSerializationApi
+    public fun <T> encodeToConfig(serializer: SerializationStrategy<T>, value: T): Config {
+        val encoder = HoconConfigEncoder(this)
+        encoder.encodeSerializableValue(serializer, value)
+        return encoder.conf
+    }
+
     /**
      * The default instance of Hocon parser.
      */
@@ -251,6 +258,10 @@ public sealed class Hocon(
 @ExperimentalSerializationApi
 public inline fun <reified T> Hocon.decodeFromConfig(config: Config): T =
     decodeFromConfig(serializersModule.serializer(), config)
+
+@ExperimentalSerializationApi
+public inline fun <reified T> Hocon.encodeToConfig(value: T): Config =
+    encodeToConfig(serializersModule.serializer(), value)
 
 /**
  * Creates an instance of [Hocon] configured from the optionally given [Hocon instance][from]
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -37,9 +37,12 @@ public sealed class Hocon(
 
     @ExperimentalSerializationApi
     public fun <T> encodeToConfig(serializer: SerializationStrategy<T>, value: T): Config {
-        val encoder = HoconConfigEncoder(this)
+        lateinit var configValue: ConfigValue
+        val encoder = HoconConfigEncoder(this) { configValue = it }
         encoder.encodeSerializableValue(serializer, value)
-        return encoder.conf
+
+        check(configValue is ConfigObject) { TODO("Write reasonable error here") }
+        return (configValue as ConfigObject).toConfig()
     }
 
     /**
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -41,6 +41,7 @@ public sealed class Hocon(
 
     /**
      * Encodes the given [value] into a [Config] using the given [serializer].
+     * @throws SerializationException If list or primitive type passed as a [value].
      */
     @ExperimentalSerializationApi
     public fun <T> encodeToConfig(serializer: SerializationStrategy<T>, value: T): Config {
@@ -48,9 +49,11 @@ public sealed class Hocon(
         val encoder = HoconConfigEncoder(this) { configValue = it }
         encoder.encodeSerializableValue(serializer, value)
 
-        check(configValue is ConfigObject) {
-            "Value of type '${configValue.valueType()}' can't be used at the root of HOCON Config." +
-                    "It should be either object or map."
+        if (configValue !is ConfigObject) {
+            throw SerializationException(
+                "Value of type '${configValue.valueType()}' can't be used at the root of HOCON Config." +
+                        "It should be either object or map."
+            )
         }
         return (configValue as ConfigObject).toConfig()
     }
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -25,10 +25,11 @@ import kotlinx.serialization.modules.*
  */
 @ExperimentalSerializationApi
 public sealed class Hocon(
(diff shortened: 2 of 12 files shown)


formats/hocon/build.gradle
```diff
@@ -12,17 +12,9 @@ compileKotlin {
...
-        }
-    }
-    runtimeElements {
-        attributes {
-            attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 8)
-        }
-    }
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
```
osipxd, line 17: With these options `TARGET_JVM_VERSION_ATTRIBUTE ` is assigned to value `8` in module.json and `configurations` block is not needed.

Also without this, I was unable to update `apiDump` because of errors like this:
```
Symbol is declared in unnamed module which is not read by current module
```
also, there was the warning:
```
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```

osipxd, line 17: With these options `TARGET_JVM_VERSION_ATTRIBUTE ` is assigned to value `8` in module.json and `configurations` block is not needed.

Also without this, I was unable to update `apiDump` because of errors like this:
```
Symbol is declared in unnamed module which is not read by current module
```
also, there was the warning:
```
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```

osipxd:
> We're currently busy with an upcoming release of Kotlin and Kover, so will likely return to your PR shortly after the 9th of November.
@qwwdfsad sorry for bothering you but is there anybody who is able to review this PR?

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconEncoderTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    @Serializable
+    class ConfigWithIterables(
+        val array: BooleanArray,
+        val set: Set<Int>,
+        val list: List<String>,
+    )
+
+    @Test
+    fun `encode config with iterables`() {
```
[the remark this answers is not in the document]
qwwdfsad: Yep, that's okay

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconEncoderTest.kt
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    fun encodeConfigWithEnum() {
+        val obj = ConfigWithEnum(RegularEnum.VALUE)
+        val config = Hocon.encodeToConfig(obj)
+
+        config.assertContains("e = VALUE")
+    }
+
+    @Serializable
+    class ConfigWithIterables(
```
qwwdfsad, line 44: Could you please also add a test for the class that contains a set or a list of other classes with nullability?

E,g,
```
class Foo(
val set: Set<Bar?>?
)
```

with both nulls and non-nulls as container and as its elements when its not null

qwwdfsad (changes requested):
The overall changes look good, thanks a lot for the effort! I have a few pretty minor comments and it will be good to go.

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
```diff
@@ -78,4 +77,14 @@ class HoconRootMapTest {
         // com.typesafe:config lib parse empty config as empty object
         ConfigFactory.parseString(emptyConfig)
     }
+
+    @Test
+    fun testEncodeList() {
+
```
qwwdfsad, line 83: These ones probably have to do something :)

formats/hocon/build.gradle
qwwdfsad: @sandwwraith could you please take a look? I remember you've been tweaking this part not so long ago

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconRootObjectsTest.kt
osipxd: Ooops! Forgot to remove these declarations

formats/hocon/build.gradle
sandwwraith: I found a discussion about this and it seems like that we simply haven't checked whether `java.sourceCompatibility` works or not.

I've compared module files in both versions — they have identical attributes, so this change should be OK

qwwdfsad (approved):
Great work, thanks for the contribution and your patience!

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -25,24 +25,45 @@ import kotlinx.serialization.modules.*
...
+     * Decodes the given [config] into a value of type [T] using the given serializer.
+     */
     @ExperimentalSerializationApi
     public fun <T> decodeFromConfig(deserializer: DeserializationStrategy<T>, config: Config): T =
         ConfigReader(config).decodeSerializableValue(deserializer)
 
     /**
-     * The default instance of Hocon parser.
+     * Encodes the given [value] into a [Config] using the given [serializer].
+     * @throws SerializationException If list or primitive type passed as a [value].
```
sandwwraith, line 44:
```suggestion
     * @throws SerializationException If a list or a primitive type is passed as a [value].
```

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -25,24 +25,45 @@ import kotlinx.serialization.modules.*
...
-    public companion object Default : Hocon(false, false, "type", EmptySerializersModule) {
-        private val NAMING_CONVENTION_REGEX by lazy { "[A-Z]".toRegex() }
+    public fun <T> encodeToConfig(serializer: SerializationStrategy<T>, value: T): Config {
+        lateinit var configValue: ConfigValue
+        val encoder = HoconConfigEncoder(this) { configValue = it }
+        encoder.encodeSerializableValue(serializer, value)
+
+        if (configValue !is ConfigObject) {
+            throw SerializationException(
+                "Value of type '${configValue.valueType()}' can't be used at the root of HOCON Config." +
```
sandwwraith, line 54:
```suggestion
                "Value of type '${configValue.valueType()}' can't be used as the root of HOCON Config." +
```

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -25,24 +25,45 @@ import kotlinx.serialization.modules.*
...
+                        "It should be either object or map."
```
sandwwraith, line 55:
```suggestion
                        "It should be either an object or a map."
```

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconEncoderTest.kt
```diff
@@ -0,0 +1,130 @@
+package kotlinx.serialization.hocon
+
+import kotlinx.serialization.*
+import org.junit.*
+
+class HoconEncoderTest {
```
sandwwraith, line 6: Sorry if I missed that, but I don't see the test that checks that serializing an Int, List, etc. leads to a `SerializationException("Value of type '${configValue.valueType()}' can't be used at the root of HOCON Config.")`. Can you add it, please? And for other exception cases too, if necessary.

osipxd: Ok, I'll add it

sandwwraith: It seems that work is almost finished; can you add this test (and resolve conflicts with https://github.com/Kotlin/kotlinx.serialization/pull/1795) so we can include your work in the release planned for this week?

osipxd: Ok. I'l try to finish it ASAP

osipxd: Added tests for unsupported root objects and map keys in [two](https://github.com/Kotlin/kotlinx.serialization/pull/1740/commits/fe4c06760b7c0abfd8c2a714e7cd17ae3b30e41b) [last](https://github.com/Kotlin/kotlinx.serialization/pull/1740/commits/c4edfe91551375e3cef38e0a329ed99dcde72ce8) commits.

sandwwraith (approved):
Great job, thanks for all your efforts!

PR #1749, opened 2021-11-01
Add an encodeCollection extension

### LINKED ISSUE #1748: Add extension for encoding collections

There is an extension for encoding structures (`encodeStructure`) but no extension for encoding collection which would be useful for custom serializers.

I'd like for an extension to be added. It's signature would be:
```
public inline fun Encoder.encodeCollection(
    descriptor: SerialDescriptor,
    collectionSize: Int,
    block: CompositeEncoder.() -> Unit
)
```

ansman:
This is akin to encodeStructure but for collections.

This fixes #1748

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
+++ b/core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
@@ -486,3 +486,24 @@ public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, block: C
         if (ex == null) composite.endStructure(descriptor)
     }
 }
+
+/**
+ * Begins a collection, encodes it using the given [block] and ends it.
+ */
+public inline fun Encoder.encodeCollection(
+    descriptor: SerialDescriptor,
+    collectionSize: Int,
+    block: CompositeEncoder.() -> Unit
+) {
+    val composite = beginCollection(descriptor, collectionSize)
+    var ex: Throwable? = null
+    try {
+        composite.block()
+    } catch (e: Throwable) {
+        ex = e
+        throw e
+    } finally {
+        // End structure only if there is no exception, otherwise it can be swallowed
+        if (ex == null) composite.endStructure(descriptor)
+    }
+}
--- a/core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
+++ b/core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
@@ -473,13 +473,17 @@ public interface CompositeEncoder {
 /**
  * Begins a structure, encodes it using the given [block] and ends it.
  */
-public inline fun Encoder.encodeStructure(
-    descriptor: SerialDescriptor,
-    crossinline block: CompositeEncoder.() -> Unit
-) {
-    with(beginStructure(descriptor)) {
-        block()
-        endStructure(descriptor)
+public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, block: CompositeEncoder.() -> Unit) {
+    val composite = beginStructure(descriptor)
+    var ex: Throwable? = null
+    try {
+        composite.block()
+    } catch (e: Throwable) {
+        ex = e
+        throw e
+    } finally {
+        // End structure only if there is no exception, otherwise it can be swallowed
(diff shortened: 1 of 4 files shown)


core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -486,3 +486,24 @@ public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, block: C
...
+    try {
+        composite.block()
+    } catch (e: Throwable) {
+        ex = e
+        throw e
+    } finally {
+        // End structure only if there is no exception, otherwise it can be swallowed
+        if (ex == null) composite.endStructure(descriptor)
+    }
+}
```
pdvrieze, line 509: Given the way that collections work, it would make sense to put the element iteration in the helper, not just the structure part. Why not use something like:
```kotlin
fun encodeCollection(
    descriptor: SerialDescriptor,
    collection: Collection<T>,
    encodeElement: CompositeEncoder.(T) -> Unit
)
```

ansman: You can encode things using a collection structure that aren't collections themselves. The simplest example are arrays.

I don't understand the first part, are you talking about which file the extension lives in or something else?

pdvrieze: @ansman The first part refers to the idea that you would possibly want `encodeCollection` to actually enforce/provide the expected semantics of a collection rather than leaving this to the caller. As to arrays, you can always overload the helper for array types.

ansman: I like that idea but I don't want to reduce the flexibility of not enforcing a collection. Instead two extensions could be added (one that accepts a collection size and one that accepts a collection).

core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -486,3 +486,39 @@ public inline fun Encoder.encodeStructure(descriptor: SerialDescriptor, block: C
...
+
+/**
+ * Begins a collection, calls [block] with each item and ends the collections.
+ */
+public inline fun <E> Encoder.encodeCollection(
```
qwwdfsad, line 514: Please add a test or a usage of this signature

qwwdfsad (changes requested):
Thanks for taking care of it!

Could you please also make the new signatures [right](https://github.com/Kotlin/kotlinx.serialization/issues/1750)?

It's our mistake that we haven't marked the original signatures as `crossinline`, so the ugly workaround with `catch` is necessary. We aim to break it (on the source level) in 1.4.0, but these can be done in the desired way from the very beginning.

With `crossinline`, it can be as straightforward as `start() block() end()` without catch and finally blocks

ansman:
Sure can!

core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
ansman: Done, I added a test in the json module which isn't ideal but I couldn't really figure out how to write an effective test in the core module.

core/commonMain/src/kotlinx/serialization/encoding/Encoding.kt
```diff
@@ -473,16 +473,41 @@ public interface CompositeEncoder {
...
-    var ex: Throwable? = null
-    try {
-        composite.block()
-    } catch (e: Throwable) {
-        ex = e
-        throw e
-    } finally {
-        // End structure only if there is no exception, otherwise it can be swallowed
-        if (ex == null) composite.endStructure(descriptor)
+public inline fun Encoder.encodeStructure(
```
qwwdfsad, line 476: Sorry if I wasn't clear in my message, I've meant only new signatures (`encodeCollection` overloads).

With the change in the existing signature, I cannot merge it as is, as it will be breaking change and will prevent us from releasing 1.3.2.

ansman: Ah, fixed.

qwwdfsad (approved):
Thanks!

PR #1777, opened 2021-11-24
Properly handle top-level value classes in encodeToJsonElement

### LINKED ISSUE #1774: encodeToJsonElement on an inline class throws "No tag in stack for requested element"
**Describe the bug**

```
@Serializable @JvmInline
value class Name(val v: String)
```
Encoding a value of that class to a JSON string works as expected, but `encodeToJsonElement` throws the exception in the title.

This only occurs when serializing the value directly. It works as expected when embedded in a data class.

It can be worked around by creating a custom serializer that effectively does the same thing as the generated one would.

**To Reproduce**
For a unit test that reproduces it and shows various workarounds that do work, see https://github.com/bartvanheukelom/jbali/blob/java17/src/jvmTest/java/org/jbali/kotser/TaggedEncoderBugTest.kt

**Excepted Behaviour**

1. For it to, well, work :D
2. Generally speaking, I've seen various situations over the years where encoding something to a string worked but to an element didn't. This surprises me, since I would expect these 2 encoders to have like 80% of their implementation in common, where the differences in the 20% wouldn't cause bugs such as these. But that is just an outsider's perspective, so please correct me if I'm wrong! I'm not complaining, this is after all a great and free library, but I am curious what's up.

**Environment**
 - Kotlin version: 1.6.0
 - Library version: 1.3.0 & 1.3.1
 - Kotlin platforms: JVM
 - Gradle version: 7.2

sandwwraith:
Fixes #1774

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt
@@ -5,15 +5,14 @@
 
 package kotlinx.serialization.json.internal
 
-import kotlinx.serialization.ExperimentalSerializationApi
-import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
-import kotlinx.serialization.internal.NamedValueEncoder
+import kotlinx.serialization.internal.*
 import kotlinx.serialization.json.*
-import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.*
 import kotlin.collections.set
-import kotlin.jvm.JvmField
+import kotlin.jvm.*
 
 internal fun <T> Json.writeJson(value: T, serializer: SerializationStrategy<T>): JsonElement {
     lateinit var result: JsonElement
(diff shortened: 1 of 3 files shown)


bartvanheukelom:
Thanks, and sorry, I think I've broken it again :]
Please see my reply in #1774

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt
```diff
@@ -69,7 +70,7 @@ private sealed class AbstractJsonTreeEncoder(
 
     override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) {
         // Writing non-structured data (i.e. primitives) on top-level (e.g. without any tag) requires special output
-        if (currentTagOrNull != null || serializer.descriptor.kind !is PrimitiveKind && serializer.descriptor.kind !== SerialKind.ENUM) {
+        if (currentTagOrNull != null || !serializer.descriptor.needTopLevelTag) {
```
qwwdfsad, line 73: It seems like it also fixes "improper" CONTEXTUAL usages reported https://github.com/Kotlin/kotlinx.serialization/issues/1774#issuecomment-977976962, right?

sandwwraith: Nope, it's better to reuse here `carrierDescriptor` then

qwwdfsad (approved):

PR #1782, opened 2021-11-29
Fix incorrect handling of object end when JsonTreeReader (JsonElement) is used with decodeToSequence

### LINKED ISSUE #1779: Json.decodeToSequence is unable to parse a stream of whitespace-separated json objects
**Describe the bug**
A call to `next()` function throws the following exception:
`kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 96: Expected beginning of the string, but got {`

I am trying to decode a string containing two objects separated by a whitespace:
`{...} {...}`

After stepping through the library code in android studio I got to the point where `JsonTreeReader` read the first object. But I did not get to the point where `stackDepth` is decremented back to `0`.

I am sure that the type `com.example.package.MyType` is parseable because I tried it with `Json.decodeFromString()` which works fine.

It seems that the inputstream is not split by whitespace before it is parsed and the parser is also not taking care of this.

**To Reproduce**
Given the following string in `byteArray`: 
`{"type":"com.example.package.MyType","id":"user2117487202","id2":"user2117487202","someInt":2} {"type":"com.example.package.MyType","id":"user2117487202","id2":"user2117487202","someInt":2}`
It has 1 whitespace as separator between the two objects.
```
val sequence = Json.decodeToSequence<Message>(ByteArrayInputStream(byteArray), 
                                              DecodeSequenceMode.WHITESPACE_SEPARATED)
val iter = sequence.iterator()
val m1 = iter.next() // <-- here is throws 
```

**Expected behavior**
First call to `iter.next()` gives the first value of type `MyType` and second call gives the second value.

**Environment**
I am not sure where to find the versions but I found this:
`implementation 'org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.1'`

 - Kotlin version: [e.g. 1.3.30]
 - Library version: [e.g. 0.11.0]
 - Kotlin platforms: [e.g. JVM, JS, Native or their combinations]
 - Gradle version: [e.g. 4.10]
 - IDE version (if bug is related to the IDE) [e.g. IntellijIDEA 2019.1, Android Studio 3.4]
 - Other relevant context [e.g. OS version, JRE version, ... ]

sandwwraith:
Fixes #1779

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization.json.internal
@@ -34,8 +34,10 @@ internal class JsonTreeReader(
             result[key] = element
             // Verify the next token
             lastToken = lexer.consumeNextToken()
-            if (lastToken != TC_COMMA && lastToken != TC_END_OBJ) {
-                lexer.fail("Expected end of the object or comma")
+            when(lastToken) {
+                TC_COMMA -> Unit // no-op, can continue with `canConsumeValue` that verifies the token after comma
+                TC_END_OBJ -> break // `canConsumeValue` can return incorrect result, since it checks token _after_ TC_END_OBJ
+                else -> lexer.fail("Expected end of the object or comma")
             }
         }
         // Check for the correct ending
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
@@ -4,7 +4,7 @@
 
 package kotlinx.serialization.json.internal
 
-import kotlinx.serialization.*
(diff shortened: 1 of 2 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
```diff
@@ -34,8 +34,10 @@ internal class JsonTreeReader(
             result[key] = element
             // Verify the next token
             lastToken = lexer.consumeNextToken()
-            if (lastToken != TC_COMMA && lastToken != TC_END_OBJ) {
-                lexer.fail("Expected end of the object or comma")
+            when(lastToken) {
```
qwwdfsad, line 37:
```suggestion
            when (lastToken) {
```

formats/json/jvmTest/src/kotlinx/serialization/features/JsonStreamFlowTest.kt
```diff
@@ -108,6 +111,29 @@ class JsonStreamFlowTest {
...
+    @Test
+    fun testJsonElement() {
+        val list = listOf<JsonElement>(
+            buildJsonObject { put("foo", "bar") },
+            buildJsonObject { put("foo", "baz") },
+            JsonPrimitive(10),
+            JsonPrimitive("abacaba"),
+            buildJsonObject { put("foo", "qux") }
+        )
+        val input = """${list[0]} ${list[1]} ${list[2]}    ${list[3]}    ${list[4]}"""
```
qwwdfsad, line 123: Please also add a test for array-wrapped

qwwdfsad (approved):

PR #1788, opened 2021-12-07
Bump Kotlin to 1.6.0 in README.md

martinbonnin:


sandwwraith:
Thanks for the PR! We usually specify a version of Kotlin that was used to build the library, although 1.6.0 is of course compatible with 1.3.1

We plan to release the next patch version soon with 1.6.10, so I'll close this.

PR #1804, opened 2021-12-22
Update Dokka to 1.6.0 and migrate settings

IgnatBeresnev:
It's pretty much the diff we use for our [ integration tests](https://github.com/Kotlin/dokka/blob/master/integration-tests/gradle/projects/serialization/serialization.diff) in Dokka. 

I've generated html docs via simple `./gradlew dokkaHtmlMultiModule` and uploaded it to GH pages for you to click around and maybe find some bugs :) If your usual setup is more sophisticated, I can re-generate. https://ignatberesnev.github.io/dokka-serialization-1-6/

Also, coroutines have some hacks for dokka in [buildSrc/build.gradle.kts](https://github.com/Kotlin/kotlinx.coroutines/blob/67caffe88f5ad461713c1114a2f2c700a3eefe18/buildSrc/build.gradle.kts#L51), I'm not sure if this needs to be added here as well. Perhaps, @qwwdfsad can shed some light on it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,6 +99,11 @@ knit {
 // Build API docs for all modules with dokka before running Knit
 knitPrepare.dependsOn "dokka"
 
+apply plugin: 'org.jetbrains.dokka'
+dependencies {
+    dokkaPlugin("org.jetbrains.kotlinx:dokka-pathsaver-plugin:$knit_version")
+}
+
 allprojects {
     group 'org.jetbrains.kotlinx'
 
--- a/docs/basic-serialization.md
+++ b/docs/basic-serialization.md
@@ -646,12 +646,14 @@ The next chapter covers [Builtin classes](builtin-classes.md).
 
 <!--- MODULE /kotlinx-serialization-core -->
 <!--- INDEX kotlinx-serialization-core/kotlinx.serialization -->
-[kotlinx.serialization.encodeToString]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/encode-to-string.html
-[Serializable]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/-serializable/index.html
-[kotlinx.serialization.decodeFromString]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/decode-from-string.html
-[Required]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/-required/index.html
-[Transient]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/-transient/index.html
-[SerialName]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization/-serial-name/index.html
+
+[kotlinx.serialization.encodeToString]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-string.html
+[Serializable]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/index.html
+[kotlinx.serialization.decodeFromString]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-string.html
+[Required]: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-required/index.html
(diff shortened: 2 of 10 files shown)


sandwwraith:
It looks like the new dokka has conflicts with `knit`: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxSerialization_Ko/3740335?showRootCauses=false&expandBuildChangesSection=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&showLog=3740335_6250_518.6245.6482.6488&logFilter=debug

sandwwraith:
Not sure what happened here, maybe coroutines had the same problem. You could've asked @qwwdfsad , but he's on vacation now

qwwdfsad:
I remember patching Knit for that, so I'd suggest updating to 0.3.0 and seeing if the problem is gone.
Alas I don't recall whether any changes in configuration are required, but it's easy to check with `git blame` on coroutines around the line with Knit update

IgnatBeresnev:
Yeah, I actually thought about that earlier today :) Will update the PR soon

IgnatBeresnev:
@sandwwraith Hi! I can't trigger teamcity builds myself, can you re-run it?

* Updated knit to 0.3.0, it complained about indexes, so I had to add `dokka-pathsaver-plugin`
* Ran knit locally and it updated a bunch of files. To make it easier to review, pushed it as different commits.
* Updated github pages that you can have a look at: https://ignatberesnev.github.io/dokka-serialization-1-6/

Now `knitCheck` passes locally. There are still differences in configuration [compared to coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/gradle/dokka.gradle.kts) (like custom MPP configs and external links), but as I don't fully understand the motivation behind it, I didn't copy anything over. Let me know if anything specific needs to be added or if something doesn't work as expected, I'll dig into it.

sandwwraith:
Thank you! 

The only concern I have for now is that the documentation links have been changed: from e.g. `https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx-serialization-json/kotlinx.serialization.json/-json/index.html` to `https://ignatberesnev.github.io/dokka-serialization-1-6/kotlinx-serialization-json/kotlinx.serialization.json/-json-element/index.html`

I'm afraid this may break things for people that used these links to write tutorials, blog posts, etc. This is not the end of the world but may be quite inconvenient. Are there any options to leave links as in the previous version?

I also have several questions regarding website design

1. At the bottom, it says `© 2021 Copyright`. I assume we need to change this text to smth like 'Copyright JetBrains s.r.o' or something. How to do this?
2. I assume the version in the header (1.3.3-SNAPSHOT) is taken from `gradle.properties`. Is there any configuration key or other property to override it? We usually deploy a new version of the docs pages manually, and on local machines, this version is always -SNAPSHOT, which is of course not what we want.

qwwdfsad:
FTR: we've discussed it separately and decided that changing the format of the links is acceptable for us

IgnatBeresnev:
Both header version and footer text are customizable via gradle configuration. For `gradle.build.kts`, it'll look something like this
```kotlin
import org.jetbrains.dokka.base.DokkaBaseConfiguration
import org.jetbrains.dokka.gradle.DokkaTaskPartial
import org.jetbrains.dokka.base.DokkaBase

tasks.withType<DokkaTaskPartial>().configureEach {
    moduleVersion.set("1.234.567")
    pluginConfiguration<DokkaBase, DokkaBaseConfiguration> {
        footerMessage = ("Copyright JetBrains s.r.o")
    }
}
```

As for the links, the difference seems to come from `all-modules-plugin` (used for multimodule builds), it used to pack each module's docs differently. AFAIK, there's no migration guide for this and the workarounds that I can think of would take some time to implement. You could either move files around to represent the same file structure as before or create missing pages with redirects or "Page has been moved" signs

sandwwraith:
Ok, thanks for the explanations!

sandwwraith (approved):

sandwwraith:
@IgnatBeresnev Looks like the favicon is missing from GH pages: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/-mode/index.html 

Any ideas where it might gone?

IgnatBeresnev:
Good question... I'll have a look

IgnatBeresnev:
This seems to be a problem specific to GH Pages + Chrome. Everything works locally and everything works when using Firefox.

I tried various workarounds, but none of them worked, not sure why. As there's no easy fix, we'll probably get to it sometime later.

Was the favicon present in 1.4?

sandwwraith:
Orange favicon was present before this PR; also, it can be seen at the root: https://kotlin.github.io/kotlinx.serialization/
but not on any subpages somehow (eg here: https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/add.html)

Favicon is not crucial of course, take your time. Just create a proper ticket in Dokka so you won't forget it.

IgnatBeresnev:
Yep, I described the problem and what I tried here (it was mentioned above): https://github.com/Kotlin/dokka/issues/2317

I'll have a look why it worked before then... should be easy enough to just copy it over, right? ... right? :cold_sweat:

Issue #1657, opened 2021-09-03
Document new annotations in serialization guide


sandwwraith:

sandwwraith:
And `@JsonClassDiscriminator` too

PR #1816, opened 2022-01-06
guide: add note about @JsonClassDiscriminator

rocketraman:
Document the `@JsonClassDiscriminator ` attribute in the guide.


rocketraman:
> I think there also should be an explanation about inheriting this attribute (and InheritableSerialInfo in general), so it looks this task is a bit bigger than it looks at the first sight. There's a ticket for that: #1657

Thanks. Feel free to close this drive-by contribution so the job can be done right.

PR #1818, opened 2022-01-07
Change Ktoml to TOML

Peanuuutz:
Didn't expect existing support for TOML but that one still doesn't implement some important features after months and is in slow progress. :(

### DIFF
--- a/formats/README.md
+++ b/formats/README.md
@@ -93,6 +93,16 @@ Allows serialization and deserialization of objects to and from [MsgPack](https:
 This library allows serialization and deserialization of objects into and from Android
 [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences).
 
+### Toml (Multiplatform)
+
+* GitHub repo: [Peanuuutz/tomlkt](https://github.com/Peanuuutz/tomlkt)
+* Artifact ID: `net.peanuuutz:tomlkt`
+* Platform: all supported platforms
+
+This library allows serialization and deserialization of map-like entries to and from [TOML](https://toml.io/).
+Due to no support for Date Time in Kotlin Multiplatform without [additional library](https://github.com/Kotlin/kotlinx-datetime),
+currently these types are not available.
+
 ### XML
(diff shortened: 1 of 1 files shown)


qwwdfsad (commented):
Hi, it seems like the library just hit its very first release -- congrats!

I would like to slightly postpone the merge to see that the format will be developed/released/used first.

qwwdfsad:
>Didn't expect existing support for TOML but that one still doesn't implement some important features after months and is in slow progress. :(

By the way, It may be the case that it will be much easier and more impactful to cooperate with the author of `ktoml` and open a few PRs. Maintaining a library is time-consuming work and maybe here it will be better to collaborate to cut the costs instead of having two identical libraries

Peanuuutz:
> By the way, It may be the case that it will be much easier and more...

Thanks for review! I'll try to build up collaboration. :blush:

Peanuuutz:
Changed my mind. Just fix the issue that `Ktoml` is not the actual format, `TOML` is.

formats/README.md
```diff
@@ -59,13 +59,12 @@ This library allows serialization and deserialization of objects to and from [Av
...
 
-### Ktoml 
+### TOML (Multiplatform) 
 * GitHub repo: [akuleshov7/ktoml](https://github.com/akuleshov7/ktoml)
 * Artifact ID: `com.akuleshov7:ktoml-core`
-* Platforms: multiplatform, all Kotlin supported platforms
+* Platforms: all supported platforms
 
-Fully Native and Multiplatform Kotlin serialization library for serialization/deserialization of TOML format.
-This library contains no Java code and no Java dependencies and it implements multiplatform parser, decoder and encoder of TOML.
```
sandwwraith, line 68: Isn't this line also relevant?

Peanuuutz: Yes, just want to make the title be like others.

sandwwraith: Not all format descriptions fit in one line or sentence (see Avro or XML), so IMO it's better to leave it as is

PR #1821, opened 2022-01-09
Support serialization of collections that are not lists

### LINKED ISSUE #1421: Serialization of Set as Collection does not work
**Describe the bug**

I'm using Retrofit with kotlinx.serialization and the following code:

```
@Serializable
data class PackagesWrapper(
    val purls: Collection<String>
)

@POST("api/packages/bulk_search")
suspend fun getPackageVulnerabilities(@Body packageUrls: PackagesWrapper): List<PackageVulnerabilities>
```

When calling `service.getPackageVulnerabilities(PackagesWrapper(packageMap.keys))` where `keys` is a `Set`, I get

```
java.lang.ClassCastException: class java.util.LinkedHashMap$LinkedKeySet cannot be cast to class java.util.List (java.util.LinkedHashMap$LinkedKeySet and java.util.List are in module java.base of loader 'bootstrap')
```

**To Reproduce**

Checkout https://github.com/oss-review-toolkit/ort/tree/vc2kxs-bug and run `./gradlew :advisor:test --tests org.ossreviewtoolkit.advisor.advisors.VulnerableCodeTest`.

**Expected behavior**

As a `Set` is a `Collection` that can be serialized to a JSON array, I'd expected this to just work. However, `ArrayListSerializer` seems to hard-code in `ListLikeSerializer<E, List<E>, ArrayList<E>>(element)` that the "list like" collection *needs* to be a `List`.

**Environment**
 - Kotlin version: 1.4.32
 - Library version: 1.1.0
 - Kotlin platforms: JVM
 - Gradle version: 7.0
 - IDE version (if bug is related to the IDE): IntellijIDEA 2021.1
 - Other relevant context: Windows 10 64-bit

sschuberth:
So far there was an implicit assumption hard-coded that collections are
always lists. However, sets are also collections, and can be serialized
to JSON arrays just like lists.

This change allows to serialize generic collections independently of the
concrete implementation.

Fixes #1421.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
@@ -54,7 +54,7 @@ public sealed class AbstractCollectionSerializer<Element, Collection, Builder> :
 }
 
 @PublishedApi
-internal sealed class ListLikeSerializer<Element, Collection, Builder>(
+internal sealed class CollectionLikeSerializer<Element, Collection, Builder>(
     private val elementSerializer: KSerializer<Element>
 ) : AbstractCollectionSerializer<Element, Collection, Builder>() {
 
@@ -143,7 +143,7 @@ internal abstract class PrimitiveArrayBuilder<Array> internal constructor() {
 internal abstract class PrimitiveArraySerializer<Element, Array, Builder
 : PrimitiveArrayBuilder<Array>> internal constructor(
     primitiveSerializer: KSerializer<Element>
-) : ListLikeSerializer<Element, Array, Builder>(primitiveSerializer) {
+) : CollectionLikeSerializer<Element, Array, Builder>(primitiveSerializer) {
     final override val descriptor: SerialDescriptor = PrimitiveArrayDescriptor(primitiveSerializer.descriptor)
 
     final override fun Builder.builderSize(): Int = position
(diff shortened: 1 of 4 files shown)


sschuberth:
Reviews anyone? @sandwwraith maybe?

sandwwraith:
@sschuberth Sorry, I was on vacation. I'll return to your PR in a couple of days.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -70,13 +70,13 @@ internal sealed class ListLikeSerializer<Element, Collection, Builder>(
         }
     }
 
-    protected final override fun readAll(decoder: CompositeDecoder, builder: Builder, startIndex: Int, size: Int) {
+    final override fun readAll(decoder: CompositeDecoder, builder: Builder, startIndex: Int, size: Int) {
```
pdvrieze, line 73: Any reason you remove the protected visibility from these functions? They are internal and none of your changes seem to reference them in a way incompatible with protected. This is the same for most other removals of protected.

sschuberth: Please see the commit message title for this commit. This is just a cleanup of redundant visibility modifiers reported by IDE inspection.

sandwwraith:
Please also execute `./gradlew apiDump` so check task would not fail

sschuberth:
> Please also execute `./gradlew apiDump` so check task would not fail

I've added a new commit that updates the API dump for all introduced changes at once, I hope that's fine. Let me know if you want API dump changes squashed into the commit that do the actual changes instead.

sandwwraith (approved):
Thank you, great job!

sandwwraith:
Oops. I've just realized that this PR is opened against `master`, while it should be merged into `dev`. I think I will leave it in master and cherry-pick to `dev` manually.

PR #1825, opened 2022-01-11
Update README with plugin setup using Gradle version catalog

lukaszkalnik:
Adds information to README how to setup plugin using Gradle version catalog.
This is helpful mainly because for Gradle version catalog the full plugin ID is needed (which is currently not explicitly mentioned).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Make sure you have the corresponding Kotlin plugin installed in the IDE, no addi
 You can set up the serialization plugin with the Kotlin plugin using 
 [Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block):
 
-Kotlin DSL:
+##### Kotlin DSL:
 
 ```kotlin
 plugins {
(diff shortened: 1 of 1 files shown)


sandwwraith:
Thanks for the PR! It looks like this feature is still incubating:

> Central declaration of dependencies is an incubating feature. It requires the activation of the VERSION_CATALOGS feature preview. 

I do not think that advising incubating Gradle features is a wise idea. Moreover, this feature is just yet another way for managing dependency declaration — one can use `gradle.properties`, `settings.gradle.kts`, constants in `buildSrc`, or even custom Gradle plugins. The goal of README is not to provide all possible ways to declare a dependency on kotlinx.serialization, but to provide a quick start reference for creating a small project or for copy-pasting coordinates and versioning into existing scripts. I think developers that use advanced dependency management can figure themselves what to do with coordinates and versions.

lukaszkalnik:
Yes, you are right. My PR was probably too broad.
My intention was to make more explicit what the actual plugin ID is, because in the "Kotlin DSL" part the ID is only mentioned using the `kotlin()` shorthand function.
But the full plugin ID is mentioned a couple of lines below in the "Groovy DSL" part, so I guess that should suffice.

PR #1827, opened 2022-01-12
Introduce java.lang.Class.serializerOrNull() that is equivalent of KClass.serializerOrNull(), but attempts not to instantiate KClass when possible

qwwdfsad:


core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
```diff
@@ -74,6 +74,19 @@ public fun SerializersModule.serializer(type: Type): KSerializer<Any> =
...
+ * Retrieves a [KSerializer] for the given [java.lang.Class] instance of the Kotlin type or returns `null` if none is found.
+ * The given class must be annotated with [Serializable] or be one of the built-in types.
+ *
+ * The API avoids instantiation of the correspnding [KClass][Class.kotlin] on the best-effort basis.
+ *
+ * This is a [Class] counterpart of [KClass.serializerOrNull] with all the restrictions the original function implies,
+ * including the general recommendation to avoid uses of this API.
+ */
+@InternalSerializationApi
+public fun <T : Any> Class<T>.serializerOrNull(): KSerializer<T>? =
```
qwwdfsad, line 87: Keep in mind that we already have `serializerOrNull` on `Type` that does exactly this and a bit of extra (e.g. checks for generics), I've just found out that making a separate overload is much easier from the maintenance standpoint than avoiding `KClass` instantiation on each path of `serializerOrNull(Type)`

core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
```diff
@@ -74,6 +74,19 @@ public fun SerializersModule.serializer(type: Type): KSerializer<Any> =
 public fun SerializersModule.serializerOrNull(type: Type): KSerializer<Any>? =
     serializerByJavaTypeImpl(type, failOnMissingTypeArgSerializer = false)
 
+/**
+ * Retrieves a [KSerializer] for the given [java.lang.Class] instance of the Kotlin type or returns `null` if none is found.
+ * The given class must be annotated with [Serializable] or be one of the built-in types.
+ *
+ * The API avoids instantiation of the correspnding [KClass][Class.kotlin] on the best-effort basis.
```
ihostage, line 81: typo
```suggestion
 * The API avoids instantiation of the corresponding [KClass][Class.kotlin] on the best-effort basis.
```

core/jvmMain/src/kotlinx/serialization/SerializersJvm.kt
sandwwraith: I'm still not sure we should have a separate overload, especially `@Internal` one. Maybe just rewrite `is Class<*>` branch in the implementation?

qwwdfsad: That's a great point, I'll give it a try

qwwdfsad:
@sandwwraith  PTAL

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -137,7 +136,7 @@ internal fun <T : Any> SerializersModule.reflectiveOrContextual(kClass: KClass<T
  *
  * The recommended way to retrieve the serializer is inline [serializer] function and [`serializer(KType)`][serializer]
  *
- * This API is not guaranteed to work consistent across different platforms or
+ * This API is not guaranteed to work consistently across different platforms or
```
sandwwraith, line 139: pls fix it in the docs for `KClass.serializer()` too, then

sandwwraith (approved):

core/commonMain/src/kotlinx/serialization/Serializers.kt
qwwdfsad: Already done :)

PR #1832, opened 2022-01-19
Provide documentation for @EncodeDefault and @JsonClassDiscriminator

### LINKED ISSUE #1657: Document new annotations in serialization guide

sandwwraith:
Fixes #1657

Also, most of the unnecessary trailing spaces were removed


docs/basic-serialization.md
```diff
@@ -413,36 +413,76 @@ value is equal to the default one, produces the following exception.
...
+
+```kotlin
+@Serializable
+data class Project(
+    val name: String,
+    @EncodeDefault val language: String = "Kotlin"
+)
+```
+
+This annotation instructs framework to always serialize property, regardless of its value or format settings.
```
qwwdfsad, line 459:
```suggestion
This annotation instructs the framework to always serialize property, regardless of its value or format settings.
```

docs/json.md
```diff
@@ -410,6 +410,62 @@ control over the resulting JSON object:
 
 <!--- TEST -->
 
+It is also possible to specify different class discriminators for different hierarchies. Instead of Json instance property, use [JsonClassDiscriminator] annotation directly on base serializable class:
```
qwwdfsad, line 413:
```suggestion
It is also possible to specify different class discriminators for different hierarchies. Instead of Json instance property, it is possible to use [JsonClassDiscriminator] annotation directly on base serializable class:
```

docs/json.md
```diff
@@ -410,6 +410,62 @@ control over the resulting JSON object:
...
+This annotation is _inheritable_, so all subclasses of `Base` will have the same discriminator:
+
+```kotlin
+@Serializable // Class discriminator is inherited from Base
+sealed class ErrorClass: Base()
+```
+
+> To learn more about inheritable serial annotations, see documentation for [InheritableSerialInfo].
+
+Note that it is not possible to explicitly specify different class discriminators in subclasses of `Base`. Only hierarchies with empty intersection can have different discriminators.
```
qwwdfsad, line 430:
```suggestion
Note that it is not possible to explicitly specify different class discriminators in subclasses of `Base`. Only hierarchies with empty intersections can have different discriminators.
```

docs/json.md
```diff
@@ -410,6 +410,62 @@ control over the resulting JSON object:
...
+
+Discriminator specified in the annotation have priority over discriminator in Json configuration:
```
qwwdfsad, line 432:
```suggestion
Discriminator specified in the annotation has priority over discriminator in Json configuration:
```

qwwdfsad (approved):

docs/json.md
sandwwraith: 'it is possible to' phrase repeating twice, it may be inconvenient to read

Issue #1839, opened 2022-01-24
Do not use tree-based decoding for polymorphism


qwwdfsad:
Currently, even for regular usages, `Json` operates in two modes -- streaming-based decoding and, in the case of polymorphism, `JsonElement`-based decoding.

The latter complicates maintainability, predictability of code execution and debuggability (e.g. printing proper error reporting, `toString`, code bloating and a separate-code path used for polymorphism-unrelated functionality).

My proposal is to completely get rid of tree-based deciding on the path of `decodeFromString` and `decodeFromSteam` and maintain it only for `JsonElement` serialization

Issue #1266, opened 2020-12-25
Unable to catch MissingFieldException 


carbotaniuman:
**What is your use-case and why do you need this feature?**

I'd like to programmatically catch MissingFieldExceptions without having to resort to reading the stacktrace in order to check. 

**Describe the solution you'd like**

Make MissingFieldException public.

qwwdfsad:
Could you please elaborate on what exactly are you trying to achieve? E.g. provide a snippet of code and/or a use-case.
Then we could understand why it is important to differentiate MFE from other serialization exceptions

carbotaniuman:
For me it's just to fulfill an API contract and return a better error, so if you don't want to commit to this in the public API my hack works fine for me.

qwwdfsad:
> if you don't want to commit to this in the public API

Well, it's not really about us wanting it or not, but about the rationale behind the decision. In general, we're trying to keep public API as small as possible, but still concise and pragmatic. So before making it public, we should know why it can be useful.

>For me it's just to fulfill an API contract and return a better error

Could you please elaborate on why you can't catch plain `SerializationException` there? Or how `MissingFieldException` handling is different from `SerializationException`>

bukajsytlos:
I second this and also not having the field name as an attribute of exception.
I my case, when I am parsing json request and there is a missing field and I want to provide custom error response message I would have to parse message to get info I want, instead of catching specific exception which would have this field attribute and I can do whatever I want with it.

bukajsytlos:
@qwwdfsad ping

qwwdfsad:
>I second this and also not having the field name as an attribute of exception.

For example in 1.1.0 it will be not the name, but name**s**: https://github.com/Kotlin/kotlinx.serialization/blob/dev/core/commonMain/src/kotlinx/serialization/SerializationException.kt#L70 And it's nice for us to have an ability to freely change this API without preserving backwards compatibility.

Could you please elaborate on what's wrong with the default error message and how can we improve it?
Or, additionally, why you have to differentiate e.g. invalid data (e.g. expected string, but have int) from missing field exception?

bukajsytlos:
for example internationalisation

bukajsytlos:
well as in my example, when using serialization for http requests, I want to tell client why is his input bad, what's wrong with it, and which field is responsible

jeffk-kiva:
[...]

PR #1841, opened 2022-01-24
Incorporate JsonPath into exception messages

qwwdfsad:
While figuring out what to do with polymorphism I've come to the conclusion that #1839 have to be done, otherwise the change is way too widespread and intrusive while being really hard to maintain, so I've reverted this work halfway through and stopped on this

### DIFF
--- a/formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
+++ b/formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
@@ -28,6 +28,11 @@ class JsonPathTest : JsonTestBase() {
     }
 
     @Test
+    fun testMissingKey() {
+        expectPath("$.i.d[1]") { Json.decodeFromString<Outer>("""{"a":42, "i":{"d":{1:{}}""") }
+    }
+
+    @Test
     fun testMalformedRootObject() {
         expectPath("$") { Json.decodeFromString<Outer>("""{{""") }
     }
--- a/formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
+++ b/formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
@@ -33,6 +33,13 @@ class JsonPathTest : JsonTestBase() {
     }
 
     @Test
+    fun testUnknownKeyIsProperlyReported() {
+        expectPath("$.i") { Json.decodeFromString<Outer>("""{"a":42, "i":{"foo":42}""") }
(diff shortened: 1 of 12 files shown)


qwwdfsad:
The latest commit is kinda controversial, it seems like we have to make #1266 finally public for both usage and construction

formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
```diff
@@ -32,6 +32,13 @@ class JsonPathTest : JsonTestBase() {
         expectPath("$.i.d[1]") { Json.decodeFromString<Outer>("""{"a":42, "i":{"d":{1:{}}""") }
     }
 
+    @Test
+    fun testUnknownKeyIsProperlyReported() {
+        expectPath("$.i") { Json.decodeFromString<Outer>("""{"a":42, "i":{"foo":42}""") }
+        expectPath("$") { Json.decodeFromString<Outer>("""{"x":{}, "a": 42}""") }
+//        Json.decodeFromString<Outer>("""{"a":42, "x":{}}""")
```
qwwdfsad, line 39: This is the only place where misattribution happens, but I'm not sure how to fix it without a huge performance impact, so maybe we can live with that

formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    @Test
+    fun testMalformedMap() {
+        expectPath("$.i.d\n") { Json.decodeFromString<Outer>("""{"a":42, "i":{ "d": []""") }
+        expectPath("$\n") { Json.decodeFromString<Map<Int, String>>("""[]""") }
+    }
+
+    @Test
+    fun testMapValue() {
+        expectPath("$.i.d[42]\n") { Json.decodeFromString<Outer>("""{"a":42, "i":{ "d": {42: {"xx":"bar"}}""") }
```
JakeWharton, line 79: Shouldn't these be quoted such as `$.i.d['42']` to indicate object keys and not array indexing? As of now it is indistinguishable (and thus ambiguous) from the 43rd element of an array.

formats/json/commonTest/src/kotlinx/serialization/JsonPathTest.kt
sandwwraith: Please add this comment directly to the code

sandwwraith (approved):

Issue #683, opened 2020-01-30
Use EnumSerializer for explicitly serializable enum instead of auto-generated one


qwwdfsad:
And inject a customized descriptor (or use generated by `EnumSerializer` if enum doesn't have customized annotations).

Multiple reasons to do it:

    * Less bytecode (not a big deal, in fact)
    * Completely **identical** behaviour of `@Serializable enum class Foo { ... }` and `enum class Foo`
    * Ability to do fixes, changes and performance optimization on the library level without changes in the plugin

Issue #1372, opened 2021-03-15
Do not call Enum.values() on each call to decodeEnumValue for explicitly-serializable enums


qwwdfsad:
Steps to reproduce:

```
@Serializable
enum class AnnotatedEnum {
    E1,
    E2
}
```

Its `serializer.deserialize` code contains the following sequence: `AnnotatedEnum.values()[decoder.decodeEnum($$serialDesc)];` It triggers an allocation of values array on each call.

The desired behaviour is to pre-cache this array in serializer's state

shanshin:
Fixed via #1851

PR #1851, opened 2022-01-31
Use EnumSerializer for explicitly serializable enum instead of auto-generated

shanshin:
Resolves Kotlin/kotlinx.serialization#683 Kotlin/kotlinx.serialization#1372

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/Enums.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Enums.kt
@@ -7,6 +7,7 @@ package kotlinx.serialization.internal
 import kotlinx.serialization.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
+import kotlin.jvm.Volatile
 
 /*
  * Descriptor used for explicitly serializable enums by the plugin.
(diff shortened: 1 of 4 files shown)


qwwdfsad:
Is it going to 1.7.0 or 1.7.20?

shanshin:
> Is it going to 1.7.0 or 1.7.20?

`1.7.20`

shanshin:
Merge after switching kotlinx.serialization to the kotlin  version `1.7.20`

core/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -49,20 +50,54 @@ internal class EnumDescriptor(
...
+}
+
+// TODO we can create another class for factories, it may be a copy-paste of this class or a subclass for some `AbstractEnumSerializer` with common `serialize`, `deserialize` and `toString` functions
 @PublishedApi
 @OptIn(ExperimentalSerializationApi::class)
 internal class EnumSerializer<T : Enum<T>>(
     serialName: String,
     private val values: Array<T>
 ) : KSerializer<T> {
+    @Volatile
```
sandwwraith, line 86: Is `@Volatile` really needed here? I don't see why

shanshin: Indeed, there is no need for it in the current implementation

sandwwraith (approved):

Issue #1348, opened 2021-02-20
Intrinsics for serializer() function


qwwdfsad:
Currently, our format API can be used in two ways: with explicit serializer, and with inferred from the type parameters, for exaple:
```
// Explicit
val jsonString = Json.decodeToString(MyClass.serializer(), MyClass(...))
val data = Json.encodeFromString(MyClass.serializer(), jsonString)

// Implicit
val jsonString = Json.decodeToString(MyClass(...)) // Inferred type-parameter T
val data = Json.encodeFromString<MyClass>(jsonString) // Or data: MyClass = Json.encodeFromString(sonString)
```

We always promote usages of the latter, because it has the same meaning, but in the meantime is more expressive and concise. It is based on the experimental `typeOf` function and we are able to retrieve the serializer using `KType`.

Unfortunately, it has two downsides:

- To retrieve the serializer, we have to use reflection. It prevents usage of this API with Graal (https://github.com/Kotlin/kotlinx.serialization/issues/1125) and also complicates Android usages
- Serializer retrieval is slow and often [slower than the serialization itself](https://www.reddit.com/r/Kotlin/comments/kd3wce/kotlin_json_performance_libraries/). The recommended and concise API shouldn't behave this way.

### Proposal

Introduce the notion of intrinsic to the serialization plugin. `serializer()` and `serializerOfNull` are going to be evaluated in the compile-time. Additionally, this evaluation will happen transitively through all `reified inline` functions that leverage `serializer<T>`.

To properly resolve context-specific serializers, more compact `contextualOrDefault(defaultSerializer)` function will be used instead.

sandwwraith:
@NorbertSandor We can't give any particular timeframe for now, but we aim to provide this before Kotlin 1.6

sdeleuze:
If possible we would live to get it in Kotlin 1.7 for Spring needs, especially related to GraalVM native images.

snowe2010:
seems like this has been bumped several times. What work needs to be done here? Is there anything the community can do to help out?

sandwwraith:
@snowe2010 The work has to be done in the compiler plugin — currently, they can't provide intrinsics, and we have to change that. I'm working at this task at the moment, and I hope I will be able to present results soon

snowe2010:
@sandwwraith ah, I see. thank you for your work! good luck!

PedroAlvarado:
It's been some time since the last update on this issue. Is it fair to expect this fixed some time next year?

[...]

PR #1853, opened 2022-02-03
Fix: retrieving external serializer defined on interface

Whathecode:
For now, I only added a failing unit test.

By looking at [`SerializersLookupObjectTest`](https://github.com/Kotlin/kotlinx.serialization/blob/master/core/commonTest/src/kotlinx/serialization/SerializersLookupObjectTest.kt#L63), I got the impression that `serializer(KType)` look up is not supported on JS legacy, which is why I also ignored the test on this platform. Is this correct?


Whathecode:
Once the compiler is fixed, this will fix the JVM runtime.

And yes, it _also_ adds unit tests to verify that when that time comes.

I don't know how/where to contribute to the compiler plugin. Is it open source?

sandwwraith:
Yeah, the plugin is in the same repository as Kotlin: https://github.com/JetBrains/kotlin/tree/master/plugins/kotlin-serialization/kotlin-serialization-compiler

Although it might be tedious to catch up with what's happening in there, so please don't feel obliged to do this.

sandwwraith:
I think I'll close this for now, as this tests serve illustrating purposes to highlight the problem and won't be merged until the fix.

By the way, intrinsics (#1348) kinda fix the problem (although not in all cases)

PR #1878, opened 2022-03-04
`buildClassSerialDescriptor` KDoc typo, `typeParamSerializer` -> `tSerializer`

aSemy:
I think `typeParamSerializer` is a mistake, and should be `tSerializer`


sandwwraith:
You're right, thanks!

sandwwraith (approved):

PR #1885, opened 2022-03-21
Wrong link to class discriminator key customisation docs

iruizmar:
Hey 👋🏻 

Only changing docs, so I'm basing it on master.

Thanks,


sandwwraith (approved):

sandwwraith:
Good catch, thanks!

PR #1897, opened 2022-04-07
Make DeserializationStrategy covariant at declaration-site

Lukellmann:
The type parameter `T` of `DeserializationStrategy` only appears in an `out` position, namely as the return type of `deserialize()`.
It can therefore be annotated with `out` to increase API flexibility compared to `DeserializationStrategy` being invariant.
E.g. a function taking `DeserializationStrategy<Number>` would now also accept an object of type `DeserializationStrategy<Int>`.

This would also result in `DeserializationStrategy` being more similar to the symmetric `SerializationStrategy` interface which is already contravariant at declaration-site (`interface SerializationStrategy<in T>`), `SerializationStrategy` is a consumer, `DeserializationStrategy` a producer.

This change is
* **binary backwards compatible**: generics are erased at runtime
* **source backwards compatible**: previously specified `out` projections (use-site covariance) will only report a warning about a redundant projection

Please tell me when there is a reason why this shouldn't be changed or wasn't changed before, maybe I'm missing something.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/KSerializer.kt
+++ b/core/commonMain/src/kotlinx/serialization/KSerializer.kt
@@ -5,7 +5,6 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.descriptors.*
-import kotlinx.serialization.descriptors.elementNames
 import kotlinx.serialization.encoding.*
 
 /**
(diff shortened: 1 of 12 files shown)


sandwwraith:
Hi! I'm terribly sorry this PR was stale for this long; it completely fell out of my sight :( 

I've dug through our initial design notes and haven't found any mentions regarding `DeserializationStrategy` variance — it seems that it was indeed simply just forgotten about, so there shouldn't be any problems adding it.

Given that the next release is likely going to be 1.5.0, now seems to be an excellent time to introduce this change. Can you please rebase your PR on the latest `dev` and check that it's still running?

Lukellmann:
> Hi! I'm terribly sorry this PR was stale for this long; it completely fell out of my sight :(

No problem :)

> Given that the next release is likely going to be 1.5.0, now seems to be an excellent time to introduce this change. Can you please rebase your PR on the latest `dev` and check that it's still running?

Rebase is done 👍

Issue #1789, opened 2021-12-07
Support decoding partial streams


martinbonnin:
The following test:

```kotlin
  @Serializable
  class SimpleObject(val name: String)

  @Test
  fun testDecode() {
    val inputStream  = """
    {
      "name": "Object1"
    }
    {
      "name": "Object2"
    }
    """.trimIndent().byteInputStream()
    println("1")
    println(Json.decodeFromStream<SimpleObject>(inputStream).name)
    println("2")
    println(Json.decodeFromStream<SimpleObject>(inputStream).name)
  }
```

fails with:

```
Unexpected JSON token at offset 25: Expected EOF after parsing, but had { instead
JSON input: kotlinx.serialization.json.internal.ArrayAsSequence@2c4ca0f9
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 25: Expected EOF after parsing, but had { instead
JSON input: kotlinx.serialization.json.internal.ArrayAsSequence@2c4ca0f9
	at app//kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
	at app//kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
	at app//kotlinx.serialization.json.internal.AbstractJsonLexer.fail(AbstractJsonLexer.kt:524)
	at app//kotlinx.serialization.json.internal.AbstractJsonLexer.fail$default(AbstractJsonLexer.kt:523)
	at app//kotlinx.serialization.json.internal.AbstractJsonLexer.expectEof(AbstractJsonLexer.kt:163)
	at app//kotlinx.serialization.json.JvmStreamsKt.decodeFromStream(JvmStreams.kt:66)
```

Is there a specific reason to mandate a trailing EOF? Being able to parse partial streams would be nice. For an example to extract a Json from a markdown file or to read an array of objects without comas like the example above.

PS: there could be a `decodeFromStreamFully()` that also closes the inputStream

martinbonnin:
> Behaviour of decodeFromStream reflects decodeFromString, which also requires fully-formed string with no dangling characters.

Any reason `decodeFromStream` doesn't close the `inputStream` in that case? That'd save potential leaks if the caller forgets to wrap in `.use {}`?

> You can try to use decodeToSequence if you need consume stream lazily or partially, I think it would suite your needs

[...]

Issue #535, opened 2019-08-16
Support for linuxArm32Hfp target


atomgomba:
**What is your use-case and why do you need this feature?**

For programs running on Raspberry Pi written in Kotlin be able to use the serialization library. Protobuf is very useful for machine-to-machine communication.

**Describe the solution you'd like**

I'd like `org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:*` to resolve for this target. Currently Gradle fails and outputs this:

<details>
  <summary><code>Could not resolve all files for configuration ':linuxArm32HfpCompileKlibraries'.</code></summary>
<pre>
Execution failed for task ':compileKotlinLinuxArm32Hfp'.
> Could not resolve all files for configuration ':linuxArm32HfpCompileKlibraries'.
   > Could not resolve org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:0.11.1.
     Required by:
         project :
      > Unable to find a matching variant of org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:0.11.1:
          - Variant 'iosArm32-api' capability org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:0.11.1:
              - Incompatible attribute:
                  - Required org.jetbrains.kotlin.native.target 'linux_arm32_hfp' and found incompatible value 'ios_arm32'.
              - Other attributes:
                  - Found org.gradle.status 'release' but wasn't required.
                  - Required org.gradle.usage 'kotlin-api' and found compatible value 'kotlin-api'.
                  - Required org.jetbrains.kotlin.platform.type 'native' and found compatible value 'native'.
          - Variant 'iosArm64-api' capability org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:0.11.1:
              - Incompatible attribute:
                  - Required org.jetbrains.kotlin.native.target 'linux_arm32_hfp' and found incompatible value 'ios_arm64'.
              - Other attributes:
                  - Found org.gradle.status 'release' but wasn't required.
                  - Required org.gradle.usage 'kotlin-api' and found compatible value 'kotlin-api'.
                  - Required org.jetbrains.kotlin.platform.type 'native' and found compatible value 'native'.
          - Variant 'iosX64-api' capability org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:0.11.1:
              - Incompatible attribute:
[...]

PR #1901, opened 2022-04-13
IO Streams

shanshin:
Preview. Don't merge!


formats/json-okio/commonMain/src/kotlinx/serialization/json/OkioStreams.kt
```diff
@@ -0,0 +1,120 @@
+package kotlinx.serialization.json
```
pdvrieze, line 1: This duplicates packages with the regular json library. Isn't this invalid when using Java 9+ modules?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonToStringWriter.kt
```diff
@@ -0,0 +1,10 @@
+package kotlinx.serialization.json.internal
+
+internal expect class JsonToStringWriter constructor() : JsonWriter {
+    override fun writeChar(char: Char)
+    override fun writeLong(value: Long)
+    override fun write(text: String)
+    override fun writeQuoted(text: String)
```
pdvrieze, line 7: It seems a little double up to put the semantics of writing the different parts of json here, where every implementation has to re-implement the same translation semantics (I know that it generally is the same as the platform `.toString()`)

shanshin: In this case, only `writeLong` looks redundant. `Long.toString` is used here, but this is not the optimal solution in JVM, because such an implementation internally creates a temporary buffer for characters and creates a new `String` instance. In the case of small classes (for example, `data class Foo(var i: Int)`), array allocation can greatly slow down performance.
This can be optimized, but within a different task.

The implementations of other functions may differ significantly for different platforms and writers, so it is not optimal to put them in the general code.

formats/json-okio/commonMain/src/kotlinx/serialization/json/OkioStreams.kt
shanshin: Thanks! My bad.

shanshin:
This implementation is not suitable for release: since `json-tests` contains a dependency on `okio`, this reduces the number of native platforms on which tests can be run.

sandwwraith:
I see that currently there's an issue with targets that we do support but okio doesn't: `linuxArm32Hfp, linuxArm64, iosArm32`.

I think we should simply ignore them in our tests too. Here are my thoughts:

- We can't delete these targets, as there's certain demand (https://github.com/Kotlin/kotlinx.serialization/issues/535). Maybe we can further research demand, but grep.app doesn't allow to 'dump' all repositories by the query/
- These targets are actually not tested even now. We don't have linuxArm agents. Regular linux agents test only `linuxX64` target. 
- We can always ask Square to add these targets to okio and add them later

The only alternative way I see is to move tests to a separate source set instead of a module. To avoid confusion with IDEA, such source set would be included only in -json module for local development, but into both -json and -json-okio modules during the build.

WDYT?

shanshin:
> I see that currently there's an issue with targets that we do support but okio doesn't: linuxArm32Hfp, linuxArm64, iosArm32.
> 
> I think we should simply ignore them in our tests too. Here are my thoughts:
> 
> We can't delete these targets, as there's certain demand (https://github.com/Kotlin/kotlinx.serialization/issues/535). Maybe we can further research demand, but grep.app doesn't allow to 'dump' all repositories by the query
> These targets are actually not tested even now. We don't have linuxArm agents. Regular linux agents test only linuxX64 target.
> We can always ask Square to add these targets to okio and add them later
> The only alternative way I see is to move tests to a separate source set instead of a module. To avoid confusion with IDEA, such source set would be included only in -json module for local development, but into both -json and -json-okio modules during the build.
> 
> WDYT?

Targets `linux ARM 32 HFP`, `linux ARM 64`, `iOS ARM 32`, `minGW x86` was disabled for project `kotlinx-serialization-json-tests`

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JvmJsonStreams.kt
```diff
@@ -0,0 +1,266 @@
+package kotlinx.serialization.json.internal
+
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.charset.Charset
+
+internal class JsonToJavaStreamWriter(private val stream: OutputStream) : JsonWriter {
+    private val buffer = ByteArrayPool.take()
+    private var array = CharArrayPool.take()
+    private var index: Int = 0
```
sandwwraith, line 10: I'd suggest renaming to 'byteIndex` or smth to clearly indicate which array this index for.

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JvmJsonStreams.kt
```diff
@@ -0,0 +1,266 @@
...
+                ensure(4)
+                /* ktlint-disable no-multi-spaces */
+                write(codePoint shr 18 or 0xf0) // 11110xxx
+                write(codePoint shr 12 and 0x3f or 0x80) // 10xxxxxx
+                write(codePoint shr 6 and 0x3f or 0x80) // 10xxyyyy
+                write(codePoint and 0x3f or 0x80) // 10yyyyyy
+                /* ktlint-enable no-multi-spaces */
+            }
+            else -> {
+                throw IllegalArgumentException("Unexpected code point: $codePoint")
```
sandwwraith, line 251: This should be `JsonEncodingException`

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JvmJsonStreams.kt
```diff
@@ -0,0 +1,266 @@
...
+                        if (c >= 0x80) break
+                        write(c) // 0xxxxxxx
+                        i++
+                    }
+                }
+
+                c < 0x800 -> {
+                    // Emit a 11-bit character with 2 bytes.
+                    ensure(2)
+                    /* ktlint-disable no-multi-spaces */
```
sandwwraith, line 163: We don't need these ktlint comments since we don't use it

sandwwraith: Also, it may be a gesture of good will to mention the source of the code (it's from okio afair?) in the comments, even though we are not obliged to do this by license

sandwwraith (changes requested):

formats/json-okio/api/kotlinx-serialization-json-okio.api
```diff
@@ -0,0 +1,7 @@
+public final class kotlinx/serialization/json/okio/OkioStreamsKt {
+	public static final fun decodeFromOkio (Lkotlinx/serialization/json/Json;Lkotlinx/serialization/DeserializationStrategy;Lokio/BufferedSource;)Ljava/lang/Object;
+	public static final fun decodeOkioToSequence (Lkotlinx/serialization/json/Json;Lokio/BufferedSource;Lkotlinx/serialization/DeserializationStrategy;Lkotlinx/serialization/json/DecodeSequenceMode;)Lkotlin/sequences/Sequence;
+	public static synthetic fun decodeOkioToSequence$default (Lkotlinx/serialization/json/Json;Lokio/BufferedSource;Lkotlinx/serialization/DeserializationStrategy;Lkotlinx/serialization/json/DecodeSequenceMode;ILjava/lang/Object;)Lkotlin/sequences/Sequence;
+	public static final fun encodeToOkio (Lkotlinx/serialization/json/Json;Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Lokio/BufferedSink;)V
```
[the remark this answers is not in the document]
shanshin: `Sink.buffer()` is not a virtual function, its extension, therefore, for each encoding will be created new instance of `RealBufferedSink` and `Buffer`. The main impact turns out if someone passes `BufferedSink` (I think this will happen most often) - in this case, a second `Buffer` will be created in the new class, which uses an additional `Segment` - thereby we get memory overruns (two Segments when), extra allocations and a second useless copying of data in memory.

Alternatively, we can create several functions, one will take `BufferedSink`, the other a pure `Sink` and call first function with `.buffer()`.

formats/json-okio/commonTest/src/kotlinx/serialization/json/okio/TestTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import okio.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+
+class TestTest {
```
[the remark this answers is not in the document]
shanshin: Removed

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -77,7 +77,8 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
 public fun JsonPrimitive(value: Nothing?): JsonNull = JsonNull
 
 // JsonLiteral is deprecated for public use and no longer available. Please use JsonPrimitive instead
-internal class JsonLiteral internal constructor(
+@InternalSerializationApi
+public class JsonLiteral internal constructor(
```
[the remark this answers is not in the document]
shanshin: If change it to 
```
@PublishedApi
internal class JsonLiteral internal constructor(
...
```
I got 
```
> Task :kotlinx-serialization-json-tests:compileTestKotlinNative FAILED
e: /.../kotlinx.serialization/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonParserTest.kt: (19, 50): Cannot access 'JsonLiteral': it is internal in 'kotlinx.serialization.json'
...
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -18,7 +18,8 @@ import kotlin.jvm.*
  * [JsonDecoder] which reads given JSON from [AbstractJsonLexer] field by field.
  */
 @OptIn(ExperimentalSerializationApi::class)
-internal open class StreamingJsonDecoder(
+@InternalSerializationApi
+public open class StreamingJsonDecoder(
```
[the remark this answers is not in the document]
shanshin: Thanks)

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
```diff
@@ -13,7 +13,7 @@ private fun toHexChar(i: Int) : Char {
 }
 
 @SharedImmutable
-internal val ESCAPE_STRINGS: Array<String?> = arrayOfNulls<String>(93).apply {
+public val ESCAPE_STRINGS: Array<String?> = arrayOfNulls<String>(93).apply {
```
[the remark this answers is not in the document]
shanshin: Yep

formats/json-tests/jvmTest/src/kotlinx/serialization/features/JsonJvmStreamsTest.kt
```diff
@@ -72,16 +72,11 @@ class JsonJvmStreamsTest {
...
             val s = generateRandomUnicodeString(strLen)
-            try {
                 val serializer = String.serializer()
                 val b = ByteArrayOutputStream()
                 Json.encodeToStream(serializer, s, b)
                 val restored = Json.decodeFromStream(serializer, ByteArrayInputStream(b.toByteArray()))
                 assertEquals(s, restored)
-            } catch (e: Throwable) {
-                // Not assertion error to preserve cause
-                throw IllegalStateException("Unexpectedly failed test, cause string: $s", e)
```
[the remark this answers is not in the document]
shanshin: Removed it for debugging, forgot to return it (

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -129,6 +124,60 @@ public sealed class Json(
     }
 }
 
+/**
+ * Description of [decodeToSequence]'s JSON input shape.
```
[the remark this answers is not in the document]
shanshin: I think mention `decodeToSequence` and `decodeFromOkioToSequence` is not a good idea. `decodeFromOkioToSequence` located in another module, it turns out the `kotlinx-serialization-json` module will know about the `kotlinx-serialization-json-okio` module.
If  specify only `decodeToSequence`, then there will be an erroneous idea that the утгь is used only in this function.

formats/json-okio/api/kotlinx-serialization-json-okio.api
```diff
@@ -0,0 +1,7 @@
+public final class kotlinx/serialization/json/okio/OkioStreamsKt {
+	public static final fun decodeFromOkio (Lkotlinx/serialization/json/Json;Lkotlinx/serialization/DeserializationStrategy;Lokio/BufferedSource;)Ljava/lang/Object;
```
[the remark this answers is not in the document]
sandwwraith: Yep, this would probably be better but we haven't thought about that atm

formats/json-okio/api/kotlinx-serialization-json-okio.api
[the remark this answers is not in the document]
sandwwraith: Can't we simply do `if (sink is BufferedSink) /*use it*/ else sink.buffer()`?

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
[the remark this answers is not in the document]
sandwwraith: Try `@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")`

formats/json-okio/api/kotlinx-serialization-json-okio.api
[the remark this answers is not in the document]
shanshin: I think we can add such a check, but in this case I think it should be explicitly specified in KDoc

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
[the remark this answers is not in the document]
shanshin: Same error

formats/json-okio/api/kotlinx-serialization-json-okio.api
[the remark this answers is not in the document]
sandwwraith: Yeah, why not

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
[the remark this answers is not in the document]
sandwwraith: Weird, that works for me

Anyway, those tests (JsonParserTest and JsonArraySerializerTest) looks to me like we just forgot to change JsonLiteral to JsonPrimitive, so we can rewrite them and leave JsonLiteral internal

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
```diff
@@ -31,7 +33,7 @@ internal val ESCAPE_STRINGS: Array<String?> = arrayOfNulls<String>(93).apply {
 }
 
 @SharedImmutable
-internal val ESCAPE_MARKERS: ByteArray = ByteArray(93).apply {
+public val ESCAPE_MARKERS: ByteArray = ByteArray(93).apply {
```
sandwwraith, line 36: I've looked into it and it seems that this val isn't used anywhere in public, so it can be reverted back to internal

formats/json/api/kotlinx-serialization-json.api
```diff
@@ -358,3 +358,43 @@ public final class kotlinx/serialization/json/JvmStreamsKt {
...
+public abstract interface class kotlinx/serialization/json/internal/SerialReader {
+	public abstract fun read ([CII)I
+}
+
+public final class kotlinx/serialization/json/internal/StreamingJsonDecoderKt {
+	public static final fun decodeStringToJsonTree (Lkotlinx/serialization/json/Json;Lkotlinx/serialization/DeserializationStrategy;Ljava/lang/String;)Lkotlinx/serialization/json/JsonElement;
+}
+
+public final class kotlinx/serialization/json/internal/StringOpsKt {
+	public static final fun getESCAPE_MARKERS ()[B
```
sandwwraith, line 389: I think this definitely should be converted to `@PublishedApi internal`. While other internal API can be potentially implemented by very curious users willing to implement their own stream adaptation (and we're ok with that) and can't mess our implementation much, these vals are giving access to internal data structures (and given that `Array` is mutable, may do very bad things)

I've tried to use `@file:Suppress` in various places and it worked normally, so I can push this change myself after you finish with other fixes

sandwwraith (changes requested):

shanshin:
Benchmark comparison for implementation from PR and dev


Test |   | PR (ops/ms) | dev (ops/ms) | change
-- | -- | -- | -- | --
JacksonComparisonBenchmark.jacksonFromString |   | 363.431 | 368.535 | -1.38%
JacksonComparisonBenchmark.jacksonSmallToString |   | 5882.674 | 5923.969 | -0.70%
JacksonComparisonBenchmark.jacksonToString |   | 1142.508 | 1125.109 | +1.55%
JacksonComparisonBenchmark.jacksonToStringWithEscapes |   | 885.894 | 887.015 | -0.13%
JacksonComparisonBenchmark.kotlinFromString |   | 929.297 | 885.595 | +4.93%
JacksonComparisonBenchmark.kotlinSmallToOkio |   | 5363.857 |   |  
JacksonComparisonBenchmark.kotlinSmallToStream |   | 6527.822 | 51.192 | +12 651.64%
JacksonComparisonBenchmark.kotlinSmallToString |   | 11594.878 | 11588.499 | +0.06%
JacksonComparisonBenchmark.kotlinToOkio |   | 692.897 |  |  
JacksonComparisonBenchmark.kotlinToStream |   | 753.526 | 49.273 | +1 429.29%
JacksonComparisonBenchmark.kotlinToString |   | 1633.087 | 1622.768 | +0.64%
JacksonComparisonBenchmark.kotlinToStringWithEscapes |   |  1103.148 | 1138.291 | -3.09%

sandwwraith (approved):
Great work!

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
```diff
@@ -0,0 +1,139 @@
...
+import kotlinx.serialization.json.internal.*
+import kotlinx.serialization.json.okio.internal.JsonToOkioStreamWriter
+import kotlinx.serialization.json.internal.decodeToSequenceByReader
+import kotlinx.serialization.json.okio.internal.OkioSerialReader
+import okio.*
+
+/**
+ * Serializes the [value] with [serializer] into a [target] using JSON format and UTF-8 encoding.
+ *
+ * If [target] is not a [BufferedSink] then called [Sink.buffer] for it to create buffered wrapper.
```
martinbonnin, line 19: Sorry I'm late to the party! Why not making the "buffer" requirement in the API itself? 

The current API fails if someone expects to be able to resume reading the stream after decoding an element.

See https://github.com/Kotlin/kotlinx.serialization/issues/1789#issuecomment-988943550 for a related discussion.

JakeWharton: For encoding this isn't a problem. You are always free to call `buffer()` and then `emit()` as implementation details.

The problem you are referring to, however, is for decoding. Decoding absolutely should take a `BufferedSource` and not a `Source`. This is what Moshi and Wire do as well.

https://github.com/square/moshi/blob/moshi-parent-1.13.0/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java#L57

https://github.com/square/wire/blob/4.4.0/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt#L145

martinbonnin: Right! Sorry for the confusion, I never realised how asymmetric this was 👍 . Maybe it's worth removing the "BufferSink" mention from the kdoc then? Since it's an implementation detail, the caller doesn't really need to know about it?

And in all cases, make the `BufferedSource` mandatory for `decodeFromSource`:

```kotlin
public fun <T> Json.decodeFromSource(
    deserializer: DeserializationStrategy<T>,
    source: BufferedSource
): T
```

JakeWharton: Agree on both points!

sandwwraith: Thanks for your suggestions! Looks like we probably need to change existing functions working with Java streams applying the same logic.

However, this may lead to some asymmetry in the API. What do you think would be better — change both `encodeToSink/decodeFromSource` so they accept buffered versions or leave `encodeToSink` intact?

martinbonnin: >  change both encodeToSink/decodeFromSource so they accept buffered versions or leave encodeToSink intact?

No strong opinion from me. Both [Moshi](https://github.com/square/moshi/blob/moshi-parent-1.13.0/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java#L80) and [Wire](https://github.com/square/moshi/blob/moshi-parent-1.13.0/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java#L80) use `BufferedSink` so I'd lean towards `BufferedSink` here as well for consistency/aesthetics reasons. 

Since the sink is ending up being buffered in all cases, it's pretty easy to let the caller call `.buffer()` if required. Plus it might even save an `emit()` when writing multiple Json to the same `BufferedSink` (So tiny performance gain maybe...). But if there's a use case for passing a raw `Sink`, that works with me too.

ahulyk:
@JakeWharton hi, do we need any extra integration from  okhttp/retrofit side?

PR #1913, opened 2022-04-26
Fix issue with registering default polymorphic serializer

piotrtomiak:
It looks like the code was copied and the update `deserializer` -> `serializer` was not made in that single place.

The following would throw an exception (but shouldn't):
```
SerializersModule {
    polymorphicDefaultDeserializer(AutomationActivityV1::class) { AutomationActivityV1Serializer }
    polymorphicDefaultSerializer(AutomationActivityV1::class) { AutomationActivityV1Serializer }
}
```
whereas the following would not (but should):
```
SerializersModule {
    polymorphicDefaultSerializer(AutomationActivityV1::class) { AutomationActivityV1Serializer }
    polymorphicDefaultSerializer(AutomationActivityV1::class) { AutomationActivityV1Serializer }
}
```


sandwwraith:
It's already fixed in `dev`: https://github.com/Kotlin/kotlinx.serialization/pull/1849

piotrtomiak:
Sorry, wrongly assumed that `master` is the development branch.

PR #1917, opened 2022-04-27
Mark block in corresponding encodeStructure/decodeStructure extension…

### LINKED ISSUE #1750: Mark block of encodeStructure, decodeStructure and encodeCollection as crossinline
And reduce the bytecode size of inlined try-catch-finally block that is totally redundant.

It may be a source incompatible change, so it can only be done in the scope of the next major release

qwwdfsad:
…s as crossinline to reduce amount of bytecode

Fixes #1750

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
+++ b/core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
@@ -562,19 +562,12 @@ public interface CompositeDecoder {
  */
 public inline fun <T> Decoder.decodeStructure(
     descriptor: SerialDescriptor,
-    block: CompositeDecoder.() -> T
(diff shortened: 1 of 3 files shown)


sandwwraith:
Such a change would prohibit early returns from these lambdas, thus this change is non source-compatible. Are we OK with that?

qwwdfsad:
Yep, the ticket explicitly mentions that.
The incompatibility requires to add a FQN to return, you can see an example in [AbstractPolymorphicSerializer.kt](https://github.com/Kotlin/kotlinx.serialization/pull/1917/files#diff-f4aaa3a70d42617e0c5daeb7f40cedde4dabcb26ff9eff5b6f13af8e6fe1e091)

sandwwraith (approved):

PR #1919, opened 2022-04-27
Do not use tree-based decoding for fast-path polymorphism

### LINKED ISSUE #1839: Do not use tree-based decoding for polymorphism

qwwdfsad:
Fixes #1839

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
@@ -44,9 +44,10 @@ open class PolymorphismOverheadBenchmark {
     @Benchmark
     fun base() = json.decodeFromString(Impl.serializer(), implString)
 
+    // As of 1.3.x
     // Baseline -- 1500
     // v1, no skip -- 2000
-    // v2, with skip -- 3000
+    // v2, with skip -- 3000 [withdrawn]
     @Benchmark
     fun poly() = json.decodeFromString(serializer, polyString)
 
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
@@ -1,0 +1,53 @@
+package kotlinx.benchmarks.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import kotlinx.serialization.modules.*
+import org.openjdk.jmh.annotations.*
+import java.util.concurrent.*
+
+@Warmup(iterations = 7, time = 1)
+@Measurement(iterations = 5, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@Fork(1)
+open class PolymorphismOverheadBenchmark {
+
+    @Serializable
+    @JsonClassDiscriminator("poly")
+    data class PolymorphicWrapper(val i: @Polymorphic Poly, val i2: Impl) // amortize the cost a bit
+
+    @Serializable
+    data class BaseWrapper(val i: Impl, val i2: Impl)
+
+    @JsonClassDiscriminator("poly")
+    interface Poly
+
+    @Serializable
+    @JsonClassDiscriminator("poly")
+    class Impl(val a: Int, val b: String) : Poly
+
+    private val impl = Impl(239, "average_size_string")
+    private val module = SerializersModule {
+        polymorphic(Poly::class) {
+            subclass(Impl.serializer())
+        }
+    }
+
+    private val json = Json { serializersModule = module }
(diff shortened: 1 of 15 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -33,7 +34,37 @@ internal open class StreamingJsonDecoder(
...
+         * 1) Read the very next key of JSON structure
+         * 2) If it matches*  the descriminator key, read the value, remember current position
+         * 3) Return the value, recover an initial position
+         * 4) Right after starting the structure beginning, immediately skip over discriminator
+         *    using position from '2'.
+         * In such scenario we neither process the same input twice nor create aux data structures.
+         *
+         * (*) -- if it doesn't match, fallback to dSVP method.
+         */
+        if (deserializer !is AbstractPolymorphicSerializer<*> || json.configuration.useArrayPolymorphism) {
```
qwwdfsad, line 55: @sandwwraith PTAL.

This is a working prototype with actual measurements, please take a look and see if the overall approach is safe and sound. It still requires polishing (e.g. I'm pretty sure the current version is incorrect for IO steaming, we just don't have enough test coverage yet) which I'm eager to do as soon as we'll agree on the overall approach

qwwdfsad:
I believe this is the approach we are going to stick with.

The PR only missing test that will fail on streaming (because such lookahead is only available on String or completely read buffer), but it shouldn't prevent you from reviewing it properly

qwwdfsad:
Added reproducer for streaming and fixed it as well

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -193,7 +243,7 @@ internal open class StreamingJsonDecoder(
     }
 
     private fun handleUnknown(key: String): Boolean {
-        if (configuration.ignoreUnknownKeys) {
+        if (configuration.ignoreUnknownKeys || (discriminatorHolder?.trySkip(key) ?: false)) {
```
sandwwraith, line 246: Wouldn't it be better to make `trySkip` an extension function on `DiscriminatorHolder?` instead of `?: false`?

formats/json/jvmMain/src/kotlinx/serialization/json/internal/JsonLexerJvm.kt
```diff
@@ -174,4 +174,7 @@ internal class ReaderJsonLexer(
     override fun appendRange(fromIndex: Int, toIndex: Int) {
         escapedString.append(_source, fromIndex, toIndex - fromIndex)
     }
+
+    // Can be carefully implemented but postponed for now
+    override fun consumeLeadingMatchingValue(keyToMatch: String, isLenient: Boolean): String? = null
```
sandwwraith, line 179: Have you created a ticket/todo so we won't forget about it?

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val polyString = json.encodeToString<Poly>(impl)
+    private val serializer = serializer<Poly>()
+
+    // 5000
+    @Benchmark
+    fun base() = json.decodeFromString(Impl.serializer(), implString)
+
+    // Baseline -- 1500
+    // v1, no skip -- 2000
+    // v2, with skip -- 3000
```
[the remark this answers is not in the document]
sandwwraith: I think these comments can be safely deleted now?

sandwwraith (approved):
lgtm

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/PolymorphismOverheadBenchmark.kt
```diff
@@ -0,0 +1,53 @@
...
+@Warmup(iterations = 7, time = 1)
+@Measurement(iterations = 5, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@Fork(1)
+open class PolymorphismOverheadBenchmark {
+
+    @Serializable
+    @JsonClassDiscriminator("poly")
```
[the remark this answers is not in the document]
qwwdfsad: Just to keep string shorter

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
qwwdfsad: Good point

PR #1924, opened 2022-05-04
Docs: Add decodeSequentially possible skipping decodeElementIndex

hfhbd:
Already in the KDoc at `decodeSequentially`, but missing at `decodeElementIndex`.


sandwwraith:
Indeed, this is a good idea, thanks!

sandwwraith (approved):

PR #1938, opened 2022-05-16
Update docs: use kotlinlang.org for API reference

zoobestik:


sandwwraith:
It seems most of these links return 404 now

zoobestik:
@sandwwraith It should work after merge. Cause is important to up dokka deps into 1.6.0 -> 1.6.10. Without this changes we don't have index.html. If you will push it into main repo separated branch we can test it.

sandwwraith (approved):

PR #1941, opened 2022-05-18
Fix misleading token description

TheMrMilchmann:
The token is actually a colon (`:`) and not a semicolon (`;`) as the description suggests. Since the actual token is included in the message, it is not terrible. However, in my opinion, it still makes sense to fix this misleading description.


qwwdfsad (approved):
Thanks!

qwwdfsad:
Cherry-picked manually

PR #1962, opened 2022-06-03
Get rid of @ExperimentalUnsignedTypes on our declarations that levera…

qwwdfsad:
…ge non-experimental unsigned types


sandwwraith:
I think we can remove experimentality, I don't remember any problems with them

qwwdfsad:
I'll do it in a separate PR then

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.jvm.*
 /**
  * [JsonDecoder] which reads given JSON from [AbstractJsonLexer] field by field.
  */
-@OptIn(ExperimentalSerializationApi::class, ExperimentalUnsignedTypes::class)
+@OptIn(ExperimentalSerializationApi::class)
```
sandwwraith, line 19: You've missed `StreamingJsonEncoder`

sandwwraith (changes requested):
We also need to change a bit of documentation here: https://github.com/Kotlin/kotlinx.serialization/blob/8be6845927414a844ffebedc871dd8bc0d4b8aee/docs/builtin-classes.md#L71

sandwwraith (approved):

PR #1969, opened 2022-06-16
Support quoting unsigned integers when used as map keys

### LINKED ISSUE #1966: Number literals are not encoded as quotes when used as keys
**Describe the bug**
If you have a map with a custom key (encoding to a number literal) this will result in invalid json
**To Reproduce**
```kotlin
@Serializable
data class Test(val map: Map<Snowflake, Int>)

fun main() {
    val string = Json.encodeToString(Test(mapOf(Snowflake(1UL) to 2)))
    println(string) // {"map":{1:2}}
    println(Json.decodeFromString<Test>(string)) // actually works, despite JSON being invalid
}

@Serializable(with = Snowflake.Serializer::class)
class Snowflake(val value: ULong) {

    @OptIn(ExperimentalSerializationApi::class, ExperimentalUnsignedTypes::class)
    internal object Serializer : KSerializer<Snowflake> {
        override val descriptor: SerialDescriptor =
            @OptIn(ExperimentalUnsignedTypes::class) ULong.serializer().descriptor

        override fun deserialize(decoder: Decoder): Snowflake =
            Snowflake(decoder.decodeInline(descriptor).decodeSerializableValue(ULong.serializer()))

        override fun serialize(encoder: Encoder, value: Snowflake) {
            encoder.encodeInline(descriptor).encodeSerializableValue(ULong.serializer(), value.value)
        }
    }
}
```

**Expected behavior**
The key will be quoted

**Environment**
 - Kotlin version: 1.6.21
 - Library version: 1.3.3
 - Kotlin platforms: tested on JVM but I assume it is the same for other platforms
 - Gradle version: 6.4.2

sandwwraith:
Fixes #1966

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2017-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.serialization.json.internal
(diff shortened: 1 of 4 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
```diff
@@ -160,10 +160,18 @@ internal class StreamingJsonEncoder(
...
-            ComposerForUnsignedNumbers(composer.sb), json, mode, null
+            composerForUnsignedNumbers(), json, mode, null
         )
         else super.encodeInline(inlineDescriptor)
 
+    private fun composerForUnsignedNumbers(): Composer {
+        // If we're inside encodeInline().encodeSerializableValue, we should preserve the forceQuoting state
+        // inside the composer, but not in the encoder (otherwise we'll get into `if (forceQuoting) encodeString(value.toString())` part
+        // and unsigned numbers would be encoded incorrectly)
+        return if (composer is ComposerForUnsignedNumbers) composer
```
sandwwraith, line 171: I can lift `forceQuoting` to the base `Composer` class to avoid is-instance check. Although I'm not sure this micro-optimization would make any difference.

qwwdfsad (approved):

PR #1973, opened 2022-06-27
Update value classes guide

qwwdfsad:


docs/value-classes.md
```diff
@@ -1,33 +1,31 @@
-# Serialization and inline classes (experimental, IR-specific)
+# Serialization and value classes (IR-specific)
```
sandwwraith, line 1: While new file name better reflects its content, it may broke someone else's guides and articles that used hyperlinks to the old file. I suggest we put `inline-classes.md` back with some kind of notice about renaming or try more complicated [technique](https://stackoverflow.com/a/65485810):

> However, if your README is a symlink to a file in another directory and otherwise named appropriately, GitHub will follow the symlink when it renders the page, provided the path is relative to your repository.

sandwwraith (changes requested):

sandwwraith (approved):

PR #1975, opened 2022-06-29
Actualize note about inline classes.

mcpiroman:


sandwwraith:
Thanks for your contribution! It seems #1973 already covers this topic, so I have to close this.

Issue #1168, opened 2020-10-28
Serialization fails with UTF-8 with BOM


FlorianObermayer:
**Describe the bug**

`kotlinx.serialization` cannot deserialize `UTF-8 with BOM` strings.

**To Reproduce**

```kotlin
package com.example.myapplication

import kotlinx.serialization.Serializable
import kotlinx.serialization.SerializationException
import kotlinx.serialization.json.Json
import org.junit.Test

class ExampleUnitTest {

    /**
    Throws:
    kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 0: Expected '{, kind: CLASS'
    JSON input: ﻿{}

    at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
    at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
    at kotlinx.serialization.json.internal.JsonReader.fail(JsonReader.kt:333)
    at kotlinx.serialization.json.internal.StreamingJsonDecoder.beginStructure(StreamingJsonDecoder.kt:39)
    at com.example.myapplication.ExampleUnitTest$test deserialization of UTF-8 with BOM$Data$$serializer.deserialize(ExampleUnitTest.kt)
    at com.example.myapplication.ExampleUnitTest$test deserialization of UTF-8 with BOM$Data$$serializer.deserialize(ExampleUnitTest.kt:12)
    at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:63)
    at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:33)
    at kotlinx.serialization.json.Json.decodeFromString(Json.kt:85)
    at com.example.myapplication.ExampleUnitTest.test deserialization of UTF-8 with BOM(ExampleUnitTest.kt:16)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
[...]

Issue #167, opened 2018-07-06
Convert Int to Boolean with custom serializer


xdma:
Hi, I've already spent a lot of time trying to write custom serializer, to replace Int (TINYINT from mysql) to Boolean.

Using **Gson** I do it without problems, something like this (java):
` public class BooleanSerializer implements JsonSerializer<Boolean>, JsonDeserializer<Boolean> {

        @Override
        public JsonElement serialize(Boolean arg0, Type arg1, JsonSerializationContext arg2) {
            return new JsonPrimitive(arg0 ? 1 : 0);
        }

        @Override
        public Boolean deserialize(JsonElement arg0, Type arg1, JsonDeserializationContext arg2) throws JsonParseException {
            return arg0.getAsInt() == 1;
        }
    }

GsonBuilder().registerTypeAdapter(Boolean.class, serializer)
`

Is it possible to achieve the same result here and how?
Thank you.

Skeptick:
As I understand it, this is not possible at the moment. I use for this:
```kotlin
@Serializable
data class BooleanInt internal constructor(val value: Boolean) {

    @Serializer(forClass = BooleanInt::class)
    companion object : KSerializer<BooleanInt> {
        override fun save(output: KOutput, obj: BooleanInt) =
            output.writeIntValue(if (obj.value) 1 else 0)

        override fun load(input: KInput): BooleanInt =
            BooleanInt(input.readNullable(IntSerializer) == 1)
    }
}
```
And:
```kotlin
@Serializable
data class Example(
    @SerialName("my_bool") val myBool: BooleanInt
)
```

xdma:
Yes, I did the same, but it's not very convenient

sandwwraith:
Or you can use `@Serializable(with=...)` on boolean properties which you need to convert with custom serializer.

Currently, indeed, there is no option to override default serializers globally.

pdvrieze:
Given the way the serialization code is generated by the plugin overriding the only straightforward way is to have a delegating writer and reader where the boolean option is redirected.

qwwdfsad:
@xdma 
Even though globally one can't override `Boolean` serializer, it's possible to achieve locally.
For your custom serialization format, it's possible to write custom `KSerializer`, which can hook into `writeBooleanElementValue` and `readBooleanElementValue`.

If you need such feature in JSON (do you?), we can provide different implementation/builder which allows to hookup into specific serialization points by `::class` or `KSerialClassDesc`

[...]

PR #1981, opened 2022-07-04
UTF8 Boom and integer boolean

umerov1999:
Add support utf8 Bom
Add support integer as boolean

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
@@ -68,5 +68,7 @@ internal fun StringBuilder.printQuoted(value: String) {
 internal fun String.toBooleanStrictOrNull(): Boolean? = when {
     this.equals("true", ignoreCase = true) -> true
     this.equals("false", ignoreCase = true) -> false
+    this.equals("1", ignoreCase = true) -> true
+    this.equals("0", ignoreCase = true) -> false
     else -> null
 }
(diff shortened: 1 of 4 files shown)


sandwwraith:
Hello and thank you for your contribution! Please familiarise yourself with the guidelines here: https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md#submitting-prs 
Note that `All development (both new features and bug fixes) is performed in the dev branch` — you should start your work from `dev` branch too, since it may have significant differences from master (for example, we have recently merged to `dev` okio support)

Regarding the contents: as I can see, there are two separate features

1. Coercing integer to boolean. This is a very non-standard extension and not a part of JSON. Most of the existing libraries don't do that either. So I doubt this should be baked in here. I suggest leaving such functionality to custom serializers. See https://github.com/Kotlin/kotlinx.serialization/issues/167 for examples.

2. Parsing BOM. First of all, BOM is a property of a document/file, not a byte stream per se. Moreover, if your `String` does contain BOM, it is really out of the place (see related discussion here: https://github.com/Kotlin/kotlinx.serialization/issues/1168#issuecomment-717925264). BOM handling in string parsing is inappropriate. Regarding `InputStream`, the BOM is usually handled by charset (in our codebase, this is `java.nio.Charset` in the `InputStreamReader`). Since UTF-8 spec [does not encourage](https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-without-bom/2223926#2223926) BOM usage, Java implementation does not recognize it, and neither do we. My vision is that BOM should be handled on the application level (e.g. after opening the file) to correctly determine the encoding, not at the input-stream or deserialization level (see also https://github.com/square/okio/issues/364). If you want to discuss that separately, please open a new issue.

PR #1982, opened 2022-07-06
Improved okio support

shanshin:
- when encoding and decoding, only BufferedSink/BufferedSource are used to eliminate ambiguities
- at the end of encoding, emitCompleteSegments is called instead of flash for BufferedSink
- internal extension functions changed to @PublishedApi + internal

### DIFF
--- a/formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
+++ b/formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
@@ -53,6 +53,9 @@ public inline fun <reified T> Json.encodeToBufferedSink(
 /**
  * Deserializes JSON from [source] using UTF-8 encoding to a value of type [T] using [deserializer].
  *
+ * Note that this functions expects that exactly one object would be present in the source
+ * and throws an exception if there are any dangling bytes after an object.
+ *
  * @throws [SerializationException] if the given JSON input cannot be deserialized to the value of type [T].
  * @throws [okio.IOException] If an I/O error occurs and source can't be read from.
  */
@@ -67,6 +70,9 @@ public fun <T> Json.decodeFromBufferedSource(
 /**
  * Deserializes the contents of given [source] to the value of type [T] using UTF-8 encoding and
  * deserializer retrieved from the reified type parameter.
+ *
+ * Note that this functions expects that exactly one object would be present in the stream
+ * and throws an exception if there are any dangling bytes after an object.
  *
  * @throws [SerializationException] if the given JSON input cannot be deserialized to the value of type [T].
  * @throws [okio.IOException] If an I/O error occurs and source can't be read from.
--- a/formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
+++ b/formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
@@ -2,6 +2,8 @@
  * Copyright 2017-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
+@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
+
 package kotlinx.serialization.json.okio
 
 import kotlinx.serialization.*
@@ -16,19 +18,16 @@ import okio.*
 /**
  * Serializes the [value] with [serializer] into a [target] using JSON format and UTF-8 encoding.
  *
- * If [target] is not a [BufferedSink] then called [Sink.buffer] for it to create buffered wrapper.
- *
  * @throws [SerializationException] if the given value cannot be serialized to JSON.
  * @throws [okio.IOException] If an I/O error occurs and sink can't be written to.
  */
 @ExperimentalSerializationApi
-public fun <T> Json.encodeToSink(
+public fun <T> Json.encodeToBufferedSink(
     serializer: SerializationStrategy<T>,
     value: T,
-    target: Sink
+    target: BufferedSink
 ) {
-    val buffered = if (target is BufferedSink) target else target.buffer()
-    val writer = JsonToOkioStreamWriter(buffered)
+    val writer = JsonToOkioStreamWriter(target)
     try {
         encodeByWriter(writer, serializer, value)
     } finally {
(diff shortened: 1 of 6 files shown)


formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/internal/OkioJsonStreams.kt
```diff
@@ -42,7 +42,7 @@ internal class JsonToOkioStreamWriter(private val target: BufferedSink) : JsonWr
     }
 
     override fun release() {
-        target.flush()
+        target.emitCompleteSegments()
```
JakeWharton, line 45: Since you are not wrapping the sink and instead are accepting an already-buffered copy, there is no need to call any flush/emit/emitCompleteSegment function at the end. Bytes will always be fully available in the caller's control.

So this call (or the whole release concept) can be removed.

JakeWharton: A `BufferedSink` will automatically call `emitCompleteSegments` after each granular call to a `write*` function. So this call will always be a no-op as the previous `write*` will have already invoked it and sent any complete segments to the underlying `Sink`.

The only time you would get runaway memory usage is if you wrote directly to the `BufferedSink`'s `Buffer` rather than to the `BufferedSink` itself. You would only do this if for some reason you wanted to enqueue more bytes than a single segment before emitting them a single unit.

So a user doing a loop of encodes will see segment-sized writes automatically and then they can choose whether they want to add a call to `emit()` inside their loop or if no new elements show up (or whatever policy they want).

Generally I recommend that the person who creates the `BufferedSink` should be the one responsible for any calls to `flush`/`emit`/`emitCompleteSegments` because they have the application-specific knowledge to the behavior they want (in addition to the automatic internal calls to `emitCompleteSegments`).

shanshin: The serializer uses the functions `writeUtf8` and `writeUtf8CodePoint` - they do not call either `emit` or `emitCompleteSegments` inside

JakeWharton: The implementation of those functions on `BufferedSink` are as follows.

* `writeUtf8`: https://github.com/square/okio/blob/ad6f80d954bdd727f9757f85d4109d7b42665604/okio/c/commonMain/kotlin/okio/internal/-RealBufferedSink.kt#L52-L56
* `writeUtf8CodePoint`: https://github.com/square/okio/blob/ad6f80d954bdd727f9757f85d4109d7b42665604/okio/src/commonMain/kotlin/okio/internal/-RealBufferedSink.kt#L68-L72

Both write to the backing `Buffer` and then transfer any complete segments from that buffer to the underlying `Sink`.

shanshin: Won't this cause problems with custom implementations of `BufferedSink`? Because we can be sure that `emit...` is called only for `RealBufferedSink`, however, we accept any inheritors of `BufferedSink` in the function.

shanshin: Perhaps it should be indicated in the documentation that the users themselves should monitor the buffer and explicitly emit it to the `Sink` if they use a custom implementation?
Then there really won't be any sense in call `emit...` in the serialization code.

JakeWharton: Realistically there are no third-party subtypes of the `Buffered*` interfaces. The interfaces should be sealed now that such a thing is possible in Kotlin. `Sink` is the type you create and then you use the library to buffer it for you. 

`BufferedSink` would be a class with a private ctor but `Buffer` needs to be both a `BufferedSink` and `BufferedSource` which necessitated the interface usage.

Leaving this emit call certainly won't harm anything, but in practice it will 100% be a no-op. Sticking with the comparison to Moshi and Wire that got us to the point about accepting a `BufferedSink`, you'll notice that they do not contain such calls either.

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
```diff
@@ -45,54 +44,47 @@ public fun <T> Json.encodeToSink(
...
+    return decodeByReader(deserializer, OkioSerialReader(source))
 }
 
 /**
  * Deserializes the contents of given [source] to the value of type [T] using UTF-8 encoding and
  * deserializer retrieved from the reified type parameter.
  *
- * If [source] is not a [BufferedSource] then called [Source.buffer] for it to create buffered wrapper.
- *
  * Note that this functions expects that exactly one object would be present in the stream
```
sandwwraith, line 74: This limitation arised from original `decodeFromStream` function due to the overbuffering. Since we already accepting here `BufferedSource`, and anyone can freely continue to use this source after us, maybe we should reconsider this limitation for this particular function

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
```diff
@@ -21,8 +21,8 @@ public interface SerialReader {
 }
 
 /** @suppress */
-@InternalSerializationApi
-public fun <T> Json.encodeByWriter(writer: JsonWriter, serializer: SerializationStrategy<T>, value: T) {
+@PublishedApi
+internal fun <T> Json.encodeByWriter(writer: JsonWriter, serializer: SerializationStrategy<T>, value: T) {
```
sandwwraith, line 25: Should we also internalize `JsonWriter/SerialReader`?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
```diff
@@ -58,9 +58,9 @@ public fun <T> Json.decodeToSequenceByReader(
 }
 
 /** @suppress */
```
sandwwraith, line 60: Suppress isn't needed here?

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
```diff
@@ -45,95 +44,79 @@ public fun <T> Json.encodeToSink(
...
+    target: BufferedSink
+): Unit = encodeToBufferedSink(serializersModule.serializer(), value, target)
 
 
 /**
  * Deserializes JSON from [source] using UTF-8 encoding to a value of type [T] using [deserializer].
  *
- * If [source] is not a [BufferedSource] then called [Source.buffer] for it to create buffered wrapper.
- *
- * Note that this functions expects that exactly one object would be present in the source
```
sandwwraith, line 59: But this function still does `lexer.expectEof()`, why remove this part of documentation?

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/OkioStreams.kt
shanshin: Now it's hard to rewrite it, because the character buffer in `ReaderJsonLexer` is used. As part of another task, it will be possible to work out this optimization. 

Also, the `ReaderJsonLexer` itself can be optimized if moved `CharArrayPool` to the multiplatform code.

Issue #1930, opened 2022-05-06
Expose exception details so I can handle specific errors differently, and customise the error message


aSemy:
**What is your use-case and why do you need this feature?**

I have a Spring Boot REST API and I want to control what error codes are returned.

Jackson exceptions provides specific data about the JSON mapping failure.

```kotlin
@ControllerAdvice
class ControllerExceptionHandler {

    @ExceptionHandler(JsonProcessingException::class)
    fun handleJsonError(exception: JsonProcessingException): ResponseEntity<Any> {

        val error = when (exception) {
            // ...
            is UnrecognizedPropertyException -> "${cause.propertyName} unrecognised")
            // ...
        }

        return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error)
    }
```

I can't do this with Kotilnx Serialization because the exceptions are internal, and data about the exceptions gets concatenated into a string.

```kotlin
@ExceptionHandler(kotlinx.serialization.SerializationException::class)
fun handleKotlinxSerializationException(exception: SerializationException): ResponseEntity<Any> {
    when (exception) {
        // ERROR Cannot access 'MissingFieldException': it is internal in 'kotlinx.serialization'
        is kotlinx.serialization.MissingFieldException -> ...
    }
}
```

**Describe the solution you'd like**

I'd like to do the same with the Kotlinx Serialization exceptions. Here are some suggestions:

1. make the existing exceptions public, and add fields so I can access the exact problem
   ```kotlin
   @PublishedApi
   public class MissingFieldException(  // make public
      val fieldNames: List<String>, // expose the specific problem
      val cause: Exception?,
    ) : SerializationException("Field ... was missing...", cause)
   ```
2. Keep the exceptions internal, but provide some public data class and add it to `SerializationException`
   ```kotlin
   sealed interface SerializationExceptionDetail {
       data class MissingFields(
           val missingFields: List<String>
       ): SerializationExceptionDetail
       
       data class UnknownField(
           val unknownFields: List<String>
       ): SerializationExceptionDetail
       
       // ...
   }
   ```
   ```kotlin
   public open class SerializationException(val detail: SerializationExceptionDetail) : IllegalArgumentException {
   ```

PR #1983, opened 2022-07-06
Make missing field exception public

### LINKED ISSUE #1266: Unable to catch MissingFieldException

qwwdfsad:
Fixes #1266

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
+++ b/core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
@@ -69,8 +69,7 @@ public open class SerializationException : IllegalArgumentException {
  * @see KSerializer
  */
 @ExperimentalSerializationApi
-public class MissingFieldException
-private constructor(
+public class MissingFieldException(
     missingFields: List<String>, message: String?, cause: Throwable?
 ) : SerializationException(message, cause) {
 
@@ -106,19 +105,6 @@ private constructor(
         "Field '$missingField' is required for type with serial name '$serialName', but it was missing",
         null
     )
-
-    @PublishedApi // Constructor for stacktrace recovery
-    internal constructor(message: String?, cause: Throwable?) : this(
-        // Kludge for stacktrace recovery
-        kotlin.run {
-            val msg = message ?: return@run emptyList()
-            val prefix = "Field '"
-            val suffix = "' is required"
-            val pIdx = msg.indexOf(prefix)
-            val sIdx = msg.indexOf(suffix)
-            if (pIdx != -1 && sIdx != -1) listOf(msg.substring(pIdx + prefix.length, sIdx))
-            else emptyList()
-        }, message, cause)
 
     @PublishedApi // Constructor used by the generated serializers
     internal constructor(missingField: String) : this(
--- a/core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
+++ b/core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
@@ -4,6 +4,8 @@
 
 package kotlinx.serialization
 
+import kotlinx.serialization.encoding.*
+
 /**
  * A generic exception indicating the problem in serialization or deserialization process.
  *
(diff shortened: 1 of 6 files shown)


qwwdfsad:
I still cannot wrap my head around the proper solution for #1930, so it's unlikely to fit into 1.4.0 timeline

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -57,15 +59,73 @@ public open class SerializationException : IllegalArgumentException {
...
-    internal constructor(fieldNames: List<String>, serialName: String) : this(if (fieldNames.size == 1) "Field '${fieldNames[0]}' is required for type with serial name '$serialName', but it was missing" else "Fields $fieldNames are required for type with serial name '$serialName', but they were missing", null)
+@ExperimentalSerializationApi
+public class MissingFieldException
+private constructor(
+    missingFields: List<String>, message: String?, cause: Throwable?
+) : SerializationException(message, cause) {
+
+    /**
+     * List of fields that were required but not found during deserialization.
+     * Contains at least one element.
```
shanshin, line 79: It is indicated that it contains at least one element, but there is an exceptional case 
`else emptyList()` - we assume that this will never happen at all?

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -57,15 +59,73 @@ public open class SerializationException : IllegalArgumentException {
...
  */
-@PublishedApi
-internal class MissingFieldException
-// This constructor is used by coroutines exception recovery
-internal constructor(message: String?, cause: Throwable?) : SerializationException(message, cause) {
-    // This constructor is used by the generated serializers
-    constructor(fieldName: String) : this("Field '$fieldName' is required, but it was missing", null)
-    internal constructor(fieldNames: List<String>, serialName: String) : this(if (fieldNames.size == 1) "Field '${fieldNames[0]}' is required for type with serial name '$serialName', but it was missing" else "Fields $fieldNames are required for type with serial name '$serialName', but they were missing", null)
+@ExperimentalSerializationApi
+public class MissingFieldException
```
shanshin, line 72: If we made the `JvmMissingFieldsExceptionTest` public, then we can refine some tests (for example) so that they use `assertFailsWith(JvmMissingFieldsExceptionTest::class, null, block)` and check the missing fields explicitly with `missingFields` property.

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
qwwdfsad: Exactly. We control the plugin and parse out the string for at least one element.
For the sake of consistency I've replaced such cases with `listOf("[unknown field name]")`

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
shanshin: The golden mask check should work for all IR backends. Probably, this test was written when JS/IR was still unstable.

formats/json-tests/jvmTest/src/kotlinx/serialization/JvmMissingFieldsExceptionTest.kt
```diff
@@ -132,10 +129,10 @@ class JvmMissingFieldsExceptionTest {
     }
 
 
-    private inline fun assertFailsWithMessages(messages: List<String>, block: () -> Unit) {
-        val exception = assertFailsWith(SerializationException::class, null, block)
-        assertEquals("kotlinx.serialization.MissingFieldException", exception::class.qualifiedName)
-        val missedMessages = messages.filter { !exception.message!!.contains(it) }
+    private inline fun assertFailsWithMessages(fields: List<String>, block: () -> Unit) {
+        val exception = assertFailsWith(MissingFieldException::class, null, block)
+        val missedMessages = fields.filter { !exception.message!!.contains(it) }
```
shanshin, line 134: M.b. `missedFields`?
Such a check is not very reliable `!exception.message!!.contains(it)`, maybe always put names in single quotes and check `!exception.message!!.contains("'$it'")`

formats/json-tests/jvmTest/src/kotlinx/serialization/JvmMissingFieldsExceptionTest.kt
```diff
@@ -132,10 +129,10 @@ class JvmMissingFieldsExceptionTest {
...
+        assertEquals(exception.missingFields.sorted(), fields.sorted())
         assertTrue(missedMessages.isEmpty(), "Expected message '${exception.message}' to contain substrings $missedMessages")
```
sandwwraith, line 136:
```suggestion
        assertTrue(missedMessages.isEmpty(), "Expected message '${exception.message}' to contain substrings $fields")
```

sandwwraith (approved):

sandwwraith:
What about `UnknownFieldException`? It should be easier to expose. We can handle JSON exceptions later.

qwwdfsad:
@sandwwraith  I found `UnknownFieldException` particularly useless -- `index` itself has no meaningful information and everything else is missing right now. So it should be reworked along with a plugin, prior to that there is not much sense to make this exception public

PR #1992, opened 2022-07-14
Added support for the unsigned primitives and arrays as built-in

### LINKED ISSUE #1480: Unable to use UInt in custom serializer
**Describe the bug**

I tried to use `UInt` in a custom serializer. Unfortunately, it throws an exception at runtime.

**To Reproduce**

```kotlin
@Serializable(with = Foo.Serializer::class)
class Foo(val bar: UInt) {
	internal class Serializer : KSerializer<Foo> {
		override val descriptor: SerialDescriptor =
			buildClassSerialDescriptor("org.example.Foo") {
				element<UInt>("bar")
			}
		// ...
	}
}
```

This produces the following exception at runtime:

```
kotlinx.serialization.SerializationException: Serializer for class 'UInt' is not found.
	at kotlinx.serialization.internal.Platform_commonKt.serializerNotRegistered(Platform.common.kt:91)
```

The first line in the stacktrace pointing at my code is at the `element<UInt>("bar")` line.

**Environment**
 - Kotlin version: 1.5.0
 - Library version: both 1.1.0 and 1.2.1 produce this exception
 - Kotlin platforms: JVM
 - Gradle version: 7.0
 - Java version: 11.0.11 (OpenJDK)

### LINKED ISSUE #1745: Unsigned Integers behave differently than signed integers
**Describe the bug**

I would have expected UInt's to work as Int's now that they are a stable feature in Kotlin.

**To Reproduce**
```Json.encodeToString(10)``` produces "10" as expected
However
```Json.encodeToString(10U)``` produces an exception:
```Serializer for class 'UInt' is not found.
kotlinx.serialization.SerializationException: Serializer for class 'UInt' is not found.
```
If i explicitly pass the serializer it works as expected: `Json.encodeToString(UInt.serializer(), 10U)`

**Expected behavior**

The extension function `json.encodeToString(10U)` would not crash.

Is this a bug or the intended behaviour?

**Environment**
 - Kotlin version: 1.5.31
 - Library version: 1.3.0
 - Kotlin platforms: JVM

### LINKED ISSUE #1989: Add support for unsigned numbers arrays

Use-cases for `U[X]Array`, are similar as for `[X]Array`.

Add serializers for `UByteArray`, `UShortArray`, `UIntArray` and `ULongArray`.

```kotlin
val bytes = Json.decodeFromString<UByteArray>("[0, 127, 255]")
prtintln(bytes.contentToString()) // [0, 127, 255]
```

shanshin:
Fixes #1745
Fixes #1480
Resolves #1989

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
@@ -15,6 +15,7 @@ import kotlin.native.concurrent.*
 import kotlin.reflect.*
 import kotlin.time.Duration
 
+@Suppress("OPT_IN_USAGE")
 @SharedImmutable
 private val BUILTIN_SERIALIZERS = mapOf(
     String::class to String.serializer(),
@@ -26,12 +27,20 @@ private val BUILTIN_SERIALIZERS = mapOf(
     FloatArray::class to FloatArraySerializer(),
     Long::class to Long.serializer(),
     LongArray::class to LongArraySerializer(),
+    ULong::class to ULong.serializer(),
+    ULongArray::class to ULongArraySerializer(),
     Int::class to Int.serializer(),
     IntArray::class to IntArraySerializer(),
+    UInt::class to UInt.serializer(),
+    UIntArray::class to UIntArraySerializer(),
     Short::class to Short.serializer(),
     ShortArray::class to ShortArraySerializer(),
+    UShort::class to UShort.serializer(),
+    UShortArray::class to UShortArraySerializer(),
     Byte::class to Byte.serializer(),
     ByteArray::class to ByteArraySerializer(),
+    UByte::class to UByte.serializer(),
+    UByteArray::class to UByteArraySerializer(),
     Boolean::class to Boolean.serializer(),
     BooleanArray::class to BooleanArraySerializer(),
     Unit::class to Unit.serializer(),
--- a/core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
@@ -15,7 +15,7 @@ import kotlin.native.concurrent.*
 import kotlin.reflect.*
 import kotlin.time.Duration
 
-@Suppress("OPT_IN_USAGE")
+@OptIn(ExperimentalUnsignedTypes::class)
 @SharedImmutable
 private val BUILTIN_SERIALIZERS = mapOf(
     String::class to String.serializer(),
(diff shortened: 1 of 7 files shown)


shanshin:
No docs yet

core/commonMain/src/kotlinx/serialization/internal/Primitives.kt
```diff
@@ -15,6 +15,7 @@ import kotlin.native.concurrent.*
 import kotlin.reflect.*
 import kotlin.time.Duration
 
+@Suppress("OPT_IN_USAGE")
```
sandwwraith, line 18: Why `@Suppress` instead of regular `@OptIn`?

formats/json-tests/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
```diff
@@ -52,6 +52,21 @@ class SerializersLookupTest : JsonTestBase() {
...
+    @Test
+    fun testUnsigned() = noLegacyJs {
+        assertSame(UByte.serializer(), serializer<UByte>())
+        assertSame(UShort.serializer(), serializer<UShort>())
+        assertSame(UInt.serializer(), serializer<UInt>())
+        assertSame(ULong.serializer(), serializer<ULong>())
+    }
+    @Test
+    @OptIn(ExperimentalUnsignedTypes::class)
+    fun testUnsignedArrays() {
```
sandwwraith, line 64: Is this test working on legacy JS? If so, why previous doesn't?

formats/json-tests/commonTest/src/kotlinx/serialization/features/inline/UnsignedIntegersTest.kt
```diff
@@ -54,4 +86,55 @@ class UnsignedIntegersTest : JsonTestBase() {
...
+
+    @Test
+    fun testRoot() {
+        assertJsonFormAndRestored(UByte.serializer(), 220U, "220")
+        assertJsonFormAndRestored(UShort.serializer(), 65000U, "65000")
+        assertJsonFormAndRestored(UInt.serializer(), 2147483657U, "2147483657")
+        assertJsonFormAndRestored(ULong.serializer(), 9223372036854775817U, "9223372036854775817")
+    }
+
+    @Suppress("OPT_IN_USAGE")
```
sandwwraith, line 98: Again, why use suppress instead of opt-in?

formats/json-tests/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
shanshin: Same problem as in `testCustomGeneric`:
```
TypeError: tmp$.serializer is not a function
  ...
```

sandwwraith (approved):

PR #1999, opened 2022-07-20
Minor Kdoc update

martinbonnin:
Follow up from https://github.com/Kotlin/kotlinx.serialization/pull/1982. Since a `BufferedSink` is now required, no need to add this in the documentation.


sandwwraith (approved):

sandwwraith:
Great catch, thank you!

PR #2007, opened 2022-08-08
Fixed decoding of huge JSON data for okio streams

### LINKED ISSUE #2006: decodeFromBufferedSource throws JsonDecodingException  on big JSON
**Describe the bug**
Trying to use new Okio streams, introduced recently and encountering `JsonDecodingException `when trying to decode big JSON via `decodeFromBufferedSource()`.

` decodeFromString()` works fine.

**To Reproduce**
```
@Serializable
data class Node(
    val children: List<Node>?
)

fun test() {
        fun getNodes(count: Int, depth: Int) : List<Node> {
            val ret = mutableListOf<Node>()
            if (depth == 0) return ret
            for (i in 0 until count) {
                ret.add(Node(getNodes(1, depth - 1)))
            }
            return ret
        }

        val rootNode = Node(getNodes(1000, 10))

        val json = Serializer.json.encodeToString(rootNode)

        val buffer = okio.Buffer().apply {
            write(json.toByteArray())
        }

//        val decodedNode = Serializer.json.decodeFromString<Node>(json) // works fine
        val decodedNode = Serializer.json.decodeFromBufferedSource<Node>(buffer) // throws JsonDecodingException 
    }
```

Exception error: 

```
kotlinx.serialization.json.internal.JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.children[107].children[0].children[108]
    JSON input: .....hildren":[{"children":[{"children":[{"children":[{"children"
```

I've created github repository with reproducible example: https://github.com/Artman12/OkioTest

**Expected behavior**
`decodeFromBufferedSource` decodes JSON correctly.

**Environment**
 - Kotlin version: 1.7.10
 - Library version: 1.4.0-RC
 - Kotlin platforms: JVM, Native

shanshin:
Fixes #2006


formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonHugeDataSerializationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        val ret = mutableListOf<Node>()
+        if (depth == 0) return ret
+        for (i in 0 until count) {
+            ret.add(Node(createNodes(1, depth - 1)))
+        }
+        return ret
+    }
+
+    @Test
+    fun test() = noLegacyJs {
```
sandwwraith, line 30: Why it's `noLegacyJs`?

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonHugeDataSerializationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        // create some huge instance
+        val rootNode = Node(createNodes(1000, 10))
+
+        //  Encoding will always be true for a standard `encodeToString` - we leave this assumption so as not to insert a huge string into the sources
```
sandwwraith, line 34: Did you mean 'successful' instead of 'true'?

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -184,7 +184,6 @@ internal abstract class AbstractJsonLexer {
 
     open fun consumeNextToken(expected: Char) {
         ensureHaveChars()
-        val source = source
```
sandwwraith, line 187: It was inserted here because access to a local variable is faster than to a field getter, but we need to test with benchmarks that JVM can optimize this for us

shanshin: We cannot use a local variable here, because in the implementation in `ReaderJsonLexer` this field is mutable.
If during the call `prefetchOrEof(cpos)`the `source` field will be changed, then in this case we will read the old cached array

shanshin: You can leave it so, or make the `source` field immutable.

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonHugeDataSerializationTest.kt
shanshin: I described the situation incorrectly.
The `assertJsonFormAndRestored` function, when checking the encoding, will call `Json.encodeToString(...)` - since the string `expectedJson` was received by the same function, the test will always consider the encoding to the string is correct, even if there was actually an error there

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
sandwwraith: Ok then

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonHugeDataSerializationTest.kt
sandwwraith: I think that's fine since we're testing streams

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
shanshin: I have added a commit with corrections, please watch it (`ReaderJsonLexer#source` now immutable)

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonHugeDataSerializationTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun test() {
+        // create some huge instance
+        val rootNode = Node(createNodes(1000, 10))
+
+        val expectedJson = Json.encodeToString(Node.serializer(), rootNode)
+
+        /*
+          The assertJsonFormAndRestored function, when checking the encoding, will call Json.encodeToString(...) for `JsonTestingMode.STREAMING`
+          since the string `expectedJson` was generated by the same function, the test will always consider
+          the encoding to the `STREAMING` mode is correct, even if there was actually an error there.
```
sandwwraith, line 36:
```suggestion
          the encoding to the `STREAMING` mode is correct, even if there was actually an error there. So only TREE, JAVA_STREAMS and OKIO are actually being tested here
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/JsonLexer.kt
```diff
@@ -14,32 +14,41 @@ private const val DEFAULT_THRESHOLD = 128
...
 internal class ReaderJsonLexer(
     private val reader: SerialReader,
-    private var _source: CharArray = CharArray(BATCH_SIZE)
+    charsBuffer: CharArray = CharArray(BATCH_SIZE)
 ) : AbstractJsonLexer() {
     private var threshold: Int = DEFAULT_THRESHOLD // chars
+    private val sourceIntern: ArrayAsSequence = ArrayAsSequence(charsBuffer)
 
-    override var source: CharSequence = ArrayAsSequence(_source)
+    override val source: CharSequence = sourceIntern
```
sandwwraith, line 42: I think it's also possible to do override with a narrow type (`override val source: ArrayAsSequence`) if that's what you're trying to do here

shanshin: Ok, changed visibility for `ArrayAsSequence`

sandwwraith (approved):

PR #2008, opened 2022-08-09
Provide a way for customization Dokka's HTML output

nikpachoo:
We want to customize the header and the footer in coroutines API reference on kotlinlang.org to provide a better user experience and consistency between all parts of kotlinlang.org.

Starting from the 1.6.20 version, Dokka supports customization of HTML output. You can read about it [here](https://kotlin.github.io/dokka/1.6.20/user_guide/base-specific/frontend/#custom-html-pages).
To use this opportunity, we need a small preparation on your side.
These changes allow HTML customization.
The rest we will do on our side.

### DIFF
--- a/dokka-templates/README.md
+++ b/dokka-templates/README.md
@@ -1,0 +1,12 @@
+# Dokka's template customization
+To provide unified navigation for all parts of [kotlinlang.org](https://kotlinlang.org/),
+the Kotlin Website Team uses this directory to place custom templates in this folder
+during the website build time on TeamCity.
+
+It is not practical to place these templates in the kotlinx.serialization repository because they change from time to time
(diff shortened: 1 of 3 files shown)


dokka-templates/README.md
```diff
@@ -0,0 +1,4 @@
+# Customize Dokka's HTML. 
+To customize Dokka's HTML output, place a file in this folder.
```
sandwwraith, line 2: Am I correct that the actual insertion of template and customization happens during the build of kotlin's website on TC? If so, let's describe this process here instead of a generic README to avoid confusion.

nikpachoo: Thank you, I've updated the readme.

sandwwraith (approved):

PR #2011, opened 2022-08-12
buildClassSerialDescriptor: KDoc typo, listDescriptor -> listSerialDescriptor

sya-ri:
`listDescriptor` doesn't exist now. I think it should be `listSerialDescriptor`.


sandwwraith (approved):
You're right, thank you!

PR #2013, opened 2022-08-16
Add info about importing encodeToString()/decodeToString() to docs

lukasz-kalnik-gcx:
I stumbled upon this issue that I haven't used autocompletion, but just typed in the IDE:
```
val data: String // some JSON string
Json.decodeFromString<MyType>(data)
```
so the IDE showed an error:
image

This happened because in this case the member function was used `Json.decodeFromString(deserializer: DeserializationStrategy<T>, string: String)`, because the extension function `StringFormat.decodeFromString(String)` has not been imported automatically.

All in all, I think it would be helpful for future users who also get stuck on this problem that the documentation at least mentions the correct import.

### DIFF
--- a/docs/basic-serialization.md
+++ b/docs/basic-serialization.md
@@ -59,7 +59,7 @@ import kotlinx.serialization.json.*
 ### JSON encoding
 
 The whole process of converting data into a specific format is called _encoding_. For JSON we encode data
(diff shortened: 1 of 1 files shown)


lukasz-kalnik-gcx:
On the other hand I just noticed in my screenshot that the IDE actually suggested me to import the extension function... 🤦 So I just overlooked it.
So I leave it up to you to decide if this extended documentation would be helpful or not.

Issue #2003, opened 2022-08-02
Serializing `object` classes is 10x slower than data classes


chrisbanes:
Copying over my [Android Microbenchmarks](https://developer.android.com/topic/performance/benchmarking/microbenchmark-overview) results (ran on Google Pixel 3): 

| Benchmark                       | Time (ns) | Allocations |
|---------------------------------|-----------|-------------|
| object_reflection_serializer    | 56,077    | 27          |
| object_reflection_json          | 69,069    | 40          |
| object_intrinsic_serializer     | 56,211    | 30          |
| object_intrinsic_json           | 118,768   | 73          |
| dataclass_reflection_serializer | 2,450     | 8           |
| dataclass_reflection_json       | 6,462     | 28          |
| dataclass_intrinsic_serializer  | 2,687     | 11          |
| dataclass_intrinsic_json        | 9,927     | 42          |

Also interesting is that the intrinsic is usually (marginally) slower than using reflection.

**To Reproduce**
All benchmarks are here: https://github.com/chrisbanes/kotlin-serialization-object-perf

**Expected behavior**
I'd expect objects to be faster than data classes (or at least similar).

**Environment**
 - Kotlin version: 1.7.0
 - Library version: 1.3.3 (1.4-RC is very similar numbers though)
 - Kotlin platforms: JVM/Android

qwwdfsad:
The problem has multiple layers in it.

First of all, intrinsics are not yet implemented (https://github.com/Kotlin/kotlinx.serialization/issues/1348), which is our biggest issue right now. I suggest reading through the issue to understand why it's important.

The second problem is that we have seemingly the same code having a significant difference in its performance characteristics. 

All lookups in all benchmarks use reflection under the hood and use basically the same code path (more or less).
Optimistically, we assume that data classes are serialized much more frequently, thus it is prioritized on the `serializer()` code path and will always be faster. 

Though, it shouldn't be **ten** times faster. I've ran my own benchmarks under the profiler and pin-pointed the root cause:

https://github.com/Kotlin/kotlinx.serialization/blob/16a85df254f4f1e317554eb61ee1fbe914800aa4/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt#L104-L115

The problem here is simple -- we either have to use `getDeclaredMethod` and pay for it throwing an exception if the companion doesn't have `serializer` method (which is the case for `object`!), thus x10 performance loss on "uncommon path".

[...]

PR #2015, opened 2022-08-18
Implemented serializers caching for lookup

shanshin:
Relates #2003

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2017-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 package kotlinx.benchmarks.json
 
 import kotlinx.serialization.*
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
@@ -1,0 +1,45 @@
+package kotlinx.benchmarks.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+import org.openjdk.jmh.annotations.*
+import java.util.concurrent.*
+
+@Warmup(iterations = 7, time = 1)
+@Measurement(iterations = 7, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+@Fork(2)
+open class LookupOverheadBenchmark {
+
+    @Serializable
+    class Holder(val a: String)
+
+    @Serializable
+    class Generic<T>(val a: T)
+
+    private val data = """{"a":""}"""
+
+    @Serializable
+    object Object
+
+    @Benchmark
+    fun dataReified() = Json.decodeFromString<Holder>(data)
+
+    @Benchmark
+    fun dataPlain() = Json.decodeFromString(Holder.serializer(), data)
+
+    @Benchmark
+    fun genericReified() = Json.decodeFromString<Generic<String>>(data)
+
+    @Benchmark
+    fun genericPlain() = Json.decodeFromString(Generic.serializer(String.serializer()), data)
+
+    @Benchmark
+    fun objectReified() = Json.decodeFromString<Object>("{}")
+
+    @Benchmark
+    fun objectPlain() = Json.decodeFromString(Object.serializer(), "{}")
+}
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
@@ -24,7 +24,15 @@ open class LookupOverheadBenchmark {
     @Serializable
     class Generic<T>(val a: T)
 
+    @Serializable
+    class DoubleGeneric<T1, T2>(val a: T1, val b: T2)
+
+    @Serializable
+    class PentaGeneric<T1, T2, T3, T4, T5>(val a: T1, val b: T2, val c: T3, val d: T4, val e: T5)
+
     private val data = """{"a":""}"""
+    private val doubleData = """{"a":"","b":0}"""
+    private val pentaData = """{"a":"","b":0,"c":1,"d":true,"e":" "}"""
 
     @Serializable
     object Object
@@ -40,6 +48,18 @@ open class LookupOverheadBenchmark {
 
     @Benchmark
     fun genericPlain() = Json.decodeFromString(Generic.serializer(String.serializer()), data)
+
+    @Benchmark
+    fun doubleGenericReified() = Json.decodeFromString<DoubleGeneric<String, Int>>(doubleData)
+
+    @Benchmark
+    fun doubleGenericPlain() = Json.decodeFromString(DoubleGeneric.serializer(String.serializer(), Int.serializer()), doubleData)
+
+    @Benchmark
+    fun pentaGenericReified() = Json.decodeFromString<PentaGeneric<String, Int, Long, Boolean, Char>>(pentaData)
+
+    @Benchmark
+    fun pentaGenericPlain() = Json.decodeFromString(PentaGeneric.serializer(String.serializer(), Int.serializer(), Long.serializer(), Boolean.serializer(), Char.serializer()), pentaData)
 
     @Benchmark
     fun objectReified() = Json.decodeFromString<Object>("{}")
--- a/build.gradle
+++ b/build.gradle
@@ -173,7 +173,7 @@ subprojects {
     afterEvaluate { // Can be applied only when the project is evaluated
         animalsniffer {
             sourceSets = [sourceSets.main]
-            annotation = "kotlinx.serialization.json.internal.SuppressAnimalSniffer"
+            annotation = (name == "kotlinx-serialization-core")? "kotlinx.serialization.internal.SuppressAnimalSniffer" : "kotlinx.serialization.json.internal.SuppressAnimalSniffer"
         }
         dependencies {
             signature 'net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature'
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -18,10 +18,19 @@ import kotlin.native.concurrent.SharedImmutable
 import kotlin.reflect.*
 
 /**
- * A wrapper for a factory of serializers of non-parameterized types that can cache them.
+ * Cache class for non-parametrized and non-contextual serializers.
  */
 @SharedImmutable
-private val CACHED_SERIALIZER_FACTORY = createCachedFactoryWrapper { it.serializerOrNull()?.cast() }
+private val SERIALIZERS_CACHE = createCache { it.serializerOrNull()?.cast() }
+
+/**
+ * Cache class for parametrized and non-contextual serializers.
+ */
+@SharedImmutable
+private val PARAMETRIZED_SERIALIZERS_CACHE = createParametrizedCache { clazz, types ->
+    val serializers = null.serializersForParameters(types, true)!!
+    clazz.parametrizedSerializerOrNull(types, serializers)?.cast()
+}
 
 /**
  * Retrieves a serializer for the given type [T].
@@ -62,9 +71,7 @@ public fun serializerOrNull(type: KType): KSerializer<Any?>? = EmptySerializersM
  * @throws SerializationException if serializer cannot be created (provided [type] or its type argument is not serializable and is not registered in [this] module).
  */
 @OptIn(ExperimentalSerializationApi::class)
-public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
-    serializerByKTypeImpl(type, failOnMissingTypeArgSerializer = true) ?: type.kclass()
-        .platformSpecificSerializerNotRegistered()
+public fun SerializersModule.serializer(type: KType): KSerializer<Any?> = serializerCommon(type)
 
 /**
  * Attempts to create a serializer for the given [type] and fallbacks to [contextual][SerializersModule.getContextual]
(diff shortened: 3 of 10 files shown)


qwwdfsad:
Results:

```
Baseline
Benchmark                                Mode  Cnt   Score   Error   Units
LookupOverheadBenchmark.dataPlain       thrpt   14  20.943 ± 0.077  ops/us
LookupOverheadBenchmark.dataReified     thrpt   14   4.777 ± 0.028  ops/us
LookupOverheadBenchmark.genericPlain    thrpt   14   8.740 ± 0.022  ops/us
LookupOverheadBenchmark.genericReified  thrpt   14   0.323 ± 0.014  ops/us
LookupOverheadBenchmark.objectPlain     thrpt   14  48.704 ± 0.073  ops/us
LookupOverheadBenchmark.objectReified   thrpt   14   0.927 ± 0.004  ops/us

This patch
Benchmark                                Mode  Cnt   Score   Error   Units
LookupOverheadBenchmark.dataPlain       thrpt   14  21.232 ± 0.057  ops/us
LookupOverheadBenchmark.dataReified     thrpt   14   6.002 ± 0.126  ops/us
LookupOverheadBenchmark.genericPlain    thrpt   14   8.466 ± 0.180  ops/us
LookupOverheadBenchmark.genericReified  thrpt   14   0.589 ± 0.005  ops/us
LookupOverheadBenchmark.objectPlain     thrpt   14  48.713 ± 0.059  ops/us
LookupOverheadBenchmark.objectReified   thrpt   14   7.837 ± 0.229  ops/us

This patch + new kotlin.reflect
Benchmark                                Mode  Cnt   Score   Error   Units
LookupOverheadBenchmark.dataPlain       thrpt   14  21.241 ± 0.245  ops/us
LookupOverheadBenchmark.dataReified     thrpt   14  15.708 ± 0.210  ops/us
LookupOverheadBenchmark.genericPlain    thrpt   14   8.597 ± 0.145  ops/us
LookupOverheadBenchmark.genericReified  thrpt   14   2.143 ± 0.002  ops/us
LookupOverheadBenchmark.objectPlain     thrpt   14  48.793 ± 0.242  ops/us
LookupOverheadBenchmark.objectReified   thrpt   14  30.176 ± 0.219  ops/us
```

build.gradle
```diff
@@ -173,7 +173,7 @@ subprojects {
     afterEvaluate { // Can be applied only when the project is evaluated
         animalsniffer {
             sourceSets = [sourceSets.main]
-            annotation = "kotlinx.serialization.json.internal.SuppressAnimalSniffer"
+            annotation = (name == "kotlinx-serialization-core")? "kotlinx.serialization.internal.SuppressAnimalSniffer" : "kotlinx.serialization.json.internal.SuppressAnimalSniffer"
```
qwwdfsad, line 176: Works okay for me, nothing to worry about

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
```diff
@@ -0,0 +1,45 @@
+package kotlinx.benchmarks.json
```
qwwdfsad, line 1: Copyright is missing

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -14,8 +14,15 @@ import kotlinx.serialization.builtins.TripleSerializer
...
 import kotlinx.serialization.modules.*
 import kotlin.jvm.*
+import kotlin.native.concurrent.SharedImmutable
 import kotlin.reflect.*
 
+/**
+ * A wrapper for a factory of serializers of non-parameterized types that can cache them.
+ */
+@SharedImmutable
+private val CACHED_SERIALIZER_FACTORY = createCachedFactoryWrapper { it.serializerOrNull()?.cast() }
```
qwwdfsad, line 24: There are multiple interpretations of this name, Id suggest `SERIALIZERS_CACHE` and that's

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -79,26 +86,44 @@ private fun SerializersModule.serializerByKTypeImpl(
...
+private fun KClass<Any>.compiledParametrizedSerializerImpl(serializers: List<KSerializer<Any?>>): KSerializer<out Any>? {
+    // TODO remove after review!
+    // useless because same constructSerializerForGivenTypeArgs actually wil be called below?
+    // rootClass.compiledSerializerImpl() ?:
+    // useless because BUILTIN_SERIALIZERS types are non-paramtrized
+    // rootClass.builtinSerializerOrNull()
+    return constructSerializerForGivenTypeArgs(*serializers.toTypedArray())
 }
 
 @OptIn(ExperimentalSerializationApi::class)
```
qwwdfsad, line 120: This opt-in is no longer needed

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    override fun get(key: KClass<*>, isNullable: Boolean): KSerializer<Any?>? {
+        return cache.getOrPut(key.java) {
+            val pair = compute(key)?.let { SerializerPair(it, it.nullable) }
+            CacheEntry(pair)
+        }.serializer(isNullable)
+    }
+}
+
+private class SerializerPair(val nonNull: KSerializer<out Any>, val nullable: KSerializer<out Any?>)
```
qwwdfsad, line 61: `val`s either should be `private` or `@JvmField`

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -79,26 +86,44 @@ private fun SerializersModule.serializerByKTypeImpl(
...
     val typeArguments = type.arguments
         .map { requireNotNull(it.type) { "Star projections in type arguments are not allowed, but had $type" } }
-    val result: KSerializer<Any>? = when {
-        typeArguments.isEmpty() -> rootClass.serializerOrNull() ?: getContextual(rootClass)
-        else -> builtinSerializer(typeArguments, rootClass, failOnMissingTypeArgSerializer)
-    }?.cast()
-    return result?.nullable(isNullable)
+
+    val serializer: KSerializer<out Any>? = if (typeArguments.isEmpty()) {
+        // if serializer cached - return it immediately because nullable already processed
```
qwwdfsad, line 91:
```suggestion
        // if serializer cached - return it immediately because nullability is already processed
```

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -79,26 +86,44 @@ private fun SerializersModule.serializerByKTypeImpl(
...
+        CACHED_SERIALIZER_FACTORY.get(rootClass, isNullable)?.let { return it }
+        getContextual(rootClass)
+    } else {
+        val serializers = if (failOnMissingTypeArgSerializer)
+            typeArguments.map(::serializer)
+        else {
+            typeArguments.map { serializerOrNull(it) ?: return null }
+        }
+        rootClass.parametrizedSerializerOrNull(typeArguments, serializers) ?: getContextual(rootClass, serializers)
```
qwwdfsad, line 100: Also should be cached on type-argument basis

qwwdfsad (commented):
Approach in general looks okay, but parameterized serializer also should be cached

shanshin:
Results with latest changes 
```
Benchmark                                Mode  Cnt   Score   Error   Units
LookupOverheadBenchmark.dataPlain       thrpt   14  21.152 ± 0.495  ops/us
LookupOverheadBenchmark.dataReified     thrpt   14  16.309 ± 0.044  ops/us
LookupOverheadBenchmark.genericPlain    thrpt   14   8.232 ± 0.079  ops/us
LookupOverheadBenchmark.genericReified  thrpt   14   9.537 ± 0.079  ops/us
LookupOverheadBenchmark.objectPlain     thrpt   14  48.672 ± 0.086  ops/us
LookupOverheadBenchmark.objectReified   thrpt   14  28.955 ± 1.601  ops/us
```

qwwdfsad (approved):
I'm completely okay with ClassValue/caching part.

I would like @sandwwraith to check general code correctness, code style and consistency with his changes prior to merge, I haven't dived deep into the semantics of `builtinSerializer` changes

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -66,69 +80,72 @@ public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
...
-@OptIn(ExperimentalSerializationApi::class)
-internal fun <T : Any> SerializersModule.reflectiveOrContextual(
-    kClass: KClass<T>,
-    typeArgumentsSerializers: List<KSerializer<Any?>>
-): KSerializer<T>? {
-    return kClass.serializerOrNull() ?: getContextual(kClass, typeArgumentsSerializers)
+/**
+ * Returns null only if `failOnMissingTypeArgSerializer == false` and at least one parameter serializer not found.
+ */
+private fun SerializersModule?.serializersForParameters(
```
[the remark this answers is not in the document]
shanshin: This was an optimization for the previous implementation, but now it is more confusing.

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -66,10 +82,10 @@ public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
  * Returns `null` if serializer cannot be created (provided [type] or its type argument is not serializable and is not registered in [this] module).
  */
 @OptIn(ExperimentalSerializationApi::class)
-public fun SerializersModule.serializerOrNull(type: KType): KSerializer<Any?>? {
-    return serializerByKTypeImpl(type, failOnMissingTypeArgSerializer = false)
-}
+public fun SerializersModule.serializerOrNull(type: KType): KSerializer<Any?>? =
+    serializerByKTypeImpl(type, failOnMissingTypeArgSerializer = false)
 
+// the `this` is empty if there is no need to search for a contextual serializer
```
sandwwraith, line 88: I'm not sure this comment explains anything, because empty module already means there's no context serializers to search (and you don't check anywhere if it's empty)

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -14,8 +14,24 @@ import kotlinx.serialization.builtins.TripleSerializer
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 import kotlin.jvm.*
+import kotlin.native.concurrent.SharedImmutable
 import kotlin.reflect.*
 
+/**
+ * Cache class for non-parametrized and non-contextual serializers.
+ */
+@SharedImmutable
```
sandwwraith, line 23: `@SharedImmutable` for cache looks very counter-intuitive — until one finds that cache implementation on Native doesn't cache anything. Probably if we change this to `@ThreadLocal`, it wouldn't be a big waste, and we may want to make a simple cache for Native later.

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -179,7 +187,48 @@ public fun <T : Any> KClass<T>.serializer(): KSerializer<T> = serializerOrNull()
...
+        Map::class, MutableMap::class, LinkedHashMap::class -> LinkedHashMapSerializer(
+            serializers[0],
+            serializers[1]
+        )
+        Map.Entry::class -> MapEntrySerializer(serializers[0], serializers[1])
+        Pair::class -> PairSerializer(serializers[0], serializers[1])
+        Triple::class -> TripleSerializer(serializers[0], serializers[1], serializers[2])
+        else -> {
+            if (isReferenceArray(this)) {
+                return ArraySerializer(typeArguments[0].classifier as KClass<Any>, serializers[0])
```
sandwwraith, line 223: You don't need `return` here because you already have `return when(this)`

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
```diff
@@ -0,0 +1,124 @@
...
+
+
+private class SerializerPair(private val nonNull: KSerializer<out Any>, private val nullable: KSerializer<out Any?>) {
+    fun get(isNullable: Boolean): KSerializer<Any?> {
+        return (if (isNullable) nullable else nonNull).cast()
+    }
+
+    companion object {
+        fun from(serializer: KSerializer<out Any>?): SerializerPair? {
+            return serializer?.let { SerializerPair(it, it.nullable) }
```
sandwwraith, line 109: I'm not sure creating a pair of serializers right away is a great idea. Since this is a cache only for top-level serializers, I strongly feel that many nullable serializers will never be used, since the non-nullable case is much, much more often. In the end, they'll be just hard-referenced garbage in memory

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+
+
+private class ConcurrentHashMapParametrizedCache(private val compute: (KClass<Any>, List<KType>) -> KSerializer<Any>?) : ParametrizedSerializerCache {
+    private val cache = ConcurrentHashMap<Class<*>, ParametrizedCacheEntry>()
+
+    override fun get(key: KClass<Any>, isNullable: Boolean, types: List<KType>): Result<KSerializer<Any?>?> {
+        return cache.getOrPut(key.java) { ParametrizedCacheEntry() }
```
sandwwraith, line 94: Two `.getOrPut` in a row doesn't look very fast. Maybe we need additional benchmarks (with 2,3,4 or more type params) to determine if we need the parameterized cache at all.

Again, I think most of the cases for top-level generic serializers would be `List<Something>()` or `MyGeneric<X>()` where there are at most several (up to 10) different X so probably specialized solution may be more efficient.

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
```diff
@@ -0,0 +1,124 @@
...
+ * Serializers are computed using provided [factory] function.
+ *
+ * `null` values are not supported, though there aren't any technical limitations.
+ */
+internal actual fun createParametrizedCache(factory: (KClass<Any>, List<KType>) -> KSerializer<Any>?): ParametrizedSerializerCache {
+    return if (useClassValue) ClassValueParametrizedCache(factory) else ConcurrentHashMapParametrizedCache(factory)
+}
+
+@SuppressAnimalSniffer
+private class ClassValueCache(private val compute: (KClass<*>) -> KSerializer<Any>?) : SerializerCache {
```
sandwwraith, line 43: Have we also considered a strategy for limiting max size/cleaning these caches?

While serialization is generally used often throughout the application, it is not a rare case when it is also used on the startup for e.g. configuration reading. In this case, `decodeFromString<MyConfiguration>()` creates a cache entry that is never used again and is never removed.

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
sandwwraith: Regardless on decision on this particular feature, we also probably need some kind of benchmarks for memory footprint/throughput

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
qwwdfsad: That was actually my call to avoid limits for now.

In our initial mental model, these caches are nothing but static constant overhead.

E.g. if one uses `decodeFromString<MyConfiguration>()`, they opt-in into **some** footprint -- `companion object`, a few more loaded classes, few kilobytes of descriptors and other serialization-related stuff reachable from this companion and so on. This cache is an epsilon for this footprint, but instead of being reachable from a companion, it is reachable from here.

Nothing prevents us from introducing public `clearCaches` if there is need to, but for now it doesn't seem to be the case

sandwwraith: Given that configuration is likely one-two classes, I don't mind that static overhead. Nullability problem (see other comment) is worse IMO

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
qwwdfsad: I second that, that's a good observation.
A separate cache for nullable serializer might be good

core/commonMain/src/kotlinx/serialization/Serializers.kt
shanshin: A vestige from the old implementation

core/jvmMain/src/kotlinx/serialization/internal/Caching.kt
shanshin: I added tests for generic class `class PentaGeneric<T1, T2, T3, T4, T5>(val a: T1, val b: T2, val c: T3, val d: T4, val e: T5)` and got the following results:
with cache
```
LookupOverheadBenchmark.pentaGenericPlain     thrpt   14  3.285 ± 0.077  ops/us
LookupOverheadBenchmark.pentaGenericReified   thrpt   14  2.763 ± 0.042  ops/us
```

without cache
```
LookupOverheadBenchmark.pentaGenericPlain     thrpt   14  3.318 ± 0.017  ops/us
LookupOverheadBenchmark.pentaGenericReified   thrpt   14  0.894 ± 0.007  ops/us
```

So I think it makes sense to use a cache even with a large number of types

sandwwraith (approved):

PR #2020, opened 2022-08-27
Fix core module configuration in hocon module

### LINKED ISSUE #2019: kotlinx-serialization-hocon can not be used without kotlinx-serialization-core
**Describe the bug**
It is impossible to use `kotlinx-serialization-hocon` without `kotlinx-serialization-core` added explicitly because `core` module is added with `implementation` configuration in `hocon` module:
https://github.com/Kotlin/kotlinx.serialization/blob/d5bc32dc11d2a8a08c36663b93839d4fe91f4507/formats/hocon/build.gradle#L22

**To Reproduce**
1. Add `kotlinx-serialization-hocon` to dependencies.
2. Try to use annotation `@Serializable`.
3. You will get an errors like:
```
Unresolved reference: Serializable
```

**Expected behavior**
`core` module should be added to dependencies transitively

**Environment**
 - Kotlin version: 1.7.10
 - Library version: 1.3.3 and 1.4.0

osipxd:
Fixes #2019

### DIFF
--- a/formats/hocon/build.gradle
+++ b/formats/hocon/build.gradle
@@ -19,7 +19,7 @@ java {
 
 
 dependencies {
-    implementation project(':kotlinx-serialization-core')
+    api project(':kotlinx-serialization-core')
     api 'org.jetbrains.kotlin:kotlin-stdlib'
 
     api 'com.typesafe:config:1.4.1'


sandwwraith (approved):

sandwwraith:
Indeed, it was a misconfiguration. Thanks!

Issue #1207, opened 2020-11-16
Can't get interface serializer


InsanusMokrassar:
Bug is related to the fact that I am unable to get serializer of interfaces.

**To Reproduce**
You can use [this template](https://github.com/InsanusMokrassar/TelegramBotAPI-bot_template/tree/32a7f08a13f0465f4ed81d5bfaba9b94a40cff53) and try to put in the `main` function next code:

```kotlin
import dev.inmo.tgbotapi.types.ChatMember.abstracts.AdministratorChatMember

fun main {
    AdministratorChatMember.serializer()
}
```

And you will see that currently it is impossible to get serializer.

**Expected behavior**

This call must return [serializer](https://github.com/InsanusMokrassar/TelegramBotAPI/blob/488158d8fbc05b1952dd3ae4ef3b37a3bd27f14d/tgbotapi.core/src/commonMain/kotlin/dev/inmo/tgbotapi/types/ChatMember/abstracts/AdministratorChatMember.kt#L20-L26) because of I have [set it manually](https://github.com/InsanusMokrassar/TelegramBotAPI/blob/488158d8fbc05b1952dd3ae4ef3b37a3bd27f14d/tgbotapi.core/src/commonMain/kotlin/dev/inmo/tgbotapi/types/ChatMember/abstracts/AdministratorChatMember.kt#L8)

**Environment**
 - Kotlin version: 1.4.10
 - Library version: 1.0.1
 - Kotlin platforms: JVM (possibly JS too)
 - Gradle version: 6.7

qwwdfsad:
Thanks for the report, it is indeed a bug.

Unfortunately, `serializer< AdministratorChatMember>` doesn't work either

InsanusMokrassar:
> Thanks for the report, it is indeed a bug.
> 
> Unfortunately, `serializer< AdministratorChatMember>` doesn't work either

I have created template and put it in issue. I hope, it will help

InsanusMokrassar:
Hello, @sandwwraith , is there some assumption about when this issue will be solved?

BenWoodworth:
I'm having this issue with Kotlin 1.5.0 and kotlinx.serialization 1.2.0

```kt
fun main() {
    // Runtime exception:
    //   Serializer for class 'MyInterface' is not found.
    //   Mark the class as @Serializable or provide the serializer explicitly.
    serializer<MyInterface>()
    
    // Compilation error:
    //   Unresolved reference: serializer
[...]

PR #2026, opened 2022-09-02
Prefer Contextual serializer of default Polymorphic serializer.

### LINKED ISSUE #1645: Using contextual serialization and encoding functions with reified type parameter doesn't work for generic interfaces
**Describe the bug**
Using contextual serialization and encoding functions with reified type parameters doesn't work for generic interfaces. It works if I pass a serializer manually or change the interface to an abstract class.

**To Reproduce**

Code:
```kotlin
@Serializable(with = OptionSerializer::class)
/*sealed*/ interface Option<out T>
class Some<out T>(val value: T) : Option<T>
object None : Option<Nothing>

@Serializable
class OptionSurrogate<out T>(val isPresent: Boolean, val value: T?)

fun <T> Option<T>.toSurrogate() =
    when (this) {
        is Some<T> -> OptionSurrogate(true, value)
        is None -> OptionSurrogate(false, null)
        else -> throw IllegalArgumentException()
    }

fun <T> OptionSurrogate<T>.toOption() =
    if (isPresent) Some(value!!)
    else None

class OptionSerializer<T>(valueSerializer: KSerializer<T>) : KSerializer<Option<T>> {
    val surrogateSerializer = OptionSurrogate.serializer(valueSerializer)
    override val descriptor: SerialDescriptor
        get() = surrogateSerializer.descriptor

    override fun serialize(encoder: Encoder, value: Option<T>) =
        encoder.encodeSerializableValue(surrogateSerializer, value.toSurrogate())

    override fun deserialize(decoder: Decoder): Option<T> =
        decoder.decodeSerializableValue(surrogateSerializer).toOption()
}

class Dummy

@Serializer(forClass = Dummy::class)
object DummySerializer

fun main() {
    val format0 = Json {
        serializersModule = SerializersModule {
        }
    }
    try {
        println(format0.encodeToString<Option<Dummy>>(Some(Dummy())))
    } catch (e: Exception) {
        e.printStackTrace()
    }

    val format1 = Json {
        serializersModule = SerializersModule {
            contextual(DummySerializer)
        }
    }
    try {
        println(format1.encodeToString<Option<Dummy>>(Some(Dummy())))
    } catch (e: Exception) {
        e.printStackTrace()
    }

    println(Json.encodeToString(OptionSerializer(DummySerializer), Some(Dummy())))
}

```

Output:
```
kotlinx.serialization.SerializationException: Serializer for class 'Dummy' is not found.
	at kotlinx.serialization.internal.Platform_commonKt.serializerNotRegistered(Platform.common.kt:91)
	at kotlinx.serialization.internal.PlatformKt.platformSpecificSerializerNotRegistered(Platform.kt:29)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:60)
	at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
	at kotlinx.serialization.SerializersKt__SerializersKt.builtinSerializer$SerializersKt__SerializersKt(Serializers.kt:96)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializerByKTypeImpl$SerializersKt__SerializersKt(Serializers.kt:84)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:59)
	at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
	at OptionKotlinxSerializationTestKt.main(OptionKotlinxSerializationTest.kt:118)
	at OptionKotlinxSerializationTestKt.main(OptionKotlinxSerializationTest.kt)
kotlinx.serialization.SerializationException: Serializer for class 'Option' is not found.
	at kotlinx.serialization.internal.Platform_commonKt.serializerNotRegistered(Platform.common.kt:91)
	at kotlinx.serialization.internal.PlatformKt.platformSpecificSerializerNotRegistered(Platform.kt:29)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:60)
	at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
	at OptionKotlinxSerializationTestKt.main(OptionKotlinxSerializationTest.kt:121)
	at OptionKotlinxSerializationTestKt.main(OptionKotlinxSerializationTest.kt)
{"isPresent":true,"value":{}}
```

As one can see, if I specify an empty `SerializersModule`, I get "Serializer for class 'Dummy' is not found."; if I add `DummySerializer` to it, I get "Serializer for class 'Option' is not found."; It works if I pass `OptionSerializer(DummySerializer)` manually.

It works with no problem if I change the `Option` interface to an open, abstract, or sealed class.

**Expected behavior**
`{"isPresent":true,"value":{}}` should be printed for the second `println` statement.

**Environment**
 - Kotlin version: [1.5.21]
 - Library version: [1.2.2]
 - Kotlin platforms: [JVM]
 - Gradle version: [7.1.1]
 - IDE version (if bug is related to the IDE) [IntelliJ IDEA 2021.2 (Community Edition), Build #IC-212.4746.92, built on July 27, 2021]
 - Other relevant context [OS version Windows 10, JRE version 11.0.11+9-b1504.13 amd64]

AdlerFleurant:
According to documentation, contextual are preferred over other serializers. However, this was not the case for interfaces.

Polymorphic Serializer was always returned even if a contextual serializer was defined. This change ensures that contextual serializers actually win over other serializers.

Fixes #1645

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -80,7 +80,7 @@ private fun SerializersModule.serializerByKTypeImpl(
     val typeArguments = type.arguments
         .map { requireNotNull(it.type) { "Star projections in type arguments are not allowed, but had $type" } }
     val result: KSerializer<Any>? = when {
-        typeArguments.isEmpty() -> rootClass.serializerOrNull() ?: getContextual(rootClass)
+        typeArguments.isEmpty() -> getContextual(rootClass) ?: rootClass.serializerOrNull()
         else -> builtinSerializer(typeArguments, rootClass, failOnMissingTypeArgSerializer)
     }?.cast()
     return result?.nullable(isNullable)
--- a/core/commonTest/src/kotlinx/serialization/modules/ContextualInterfaceTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/modules/ContextualInterfaceTest.kt
@@ -1,0 +1,31 @@
+package kotlinx.serialization.modules
+
+import kotlinx.serialization.*
+import kotlinx.serialization.encoding.*
+import kotlin.test.*
+
+open class ContextualInterfaceTest {
+    sealed interface Dummy {
+        val data: String
+    }
+
+    @Serializable
+    data class DummyImpl(override val data: String) : Dummy
+
+    object DummySerializer : KSerializer<Dummy> {
+        override val descriptor = DummyImpl.serializer().descriptor
+
+        override fun serialize(encoder: Encoder, value: Dummy) =
+            DummyImpl.serializer().serialize(encoder, DummyImpl(value.data))
+
+        override fun deserialize(decoder: Decoder): Dummy = DummyImpl.serializer().deserialize(decoder)
+    }
+
+    private val module = SerializersModule { contextual(Dummy::class, DummySerializer) }
+
+    @Test
+    fun testContextualWinsOverCompiled() {
+        val serializer = module.serializer<Dummy>()
+        assertEquals(DummySerializer, serializer, "Contextual serializer should win over polymorphicSerializer.")
+    }
+}
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/SerializersLookupTest.kt
@@ -197,24 +197,6 @@ class SerializersLookupTest : JsonTestBase() {
         assertEquals("[[null]]", Json.encodeToString(serializer, listOf(listOf<IntBox?>(null))))
     }
 
-    @Test
-    fun testCompiledWinsOverContextual() {
-        val contextual = object : KSerializer<Int> {
-            override val descriptor: SerialDescriptor = Int.serializer().descriptor
-
-            override fun serialize(encoder: Encoder, value: Int) {
-                fail()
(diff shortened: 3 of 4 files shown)


sandwwraith:
I'm not sure this is a good idea to change the behavior of a stable function in such a significant way.
Also, documentation for `SerializersModule.serializer(type: KType)` is correct: `Attempts to create a serializer for the given [type] and fallbacks to [contextual][SerializersModule.getContextual] lookup for non-serializable types.`. Interfaces are always polymorphically serializable, so there's no contradiction. 

However, probably specific exceptions for interfaces should be made, if we take into consideration #1207

pdvrieze:
Another problem with preferring contextual (in the absense of a `@Contextual` annotation is that it would be inconsistent with how serializers are statically resolved (by the plugin) as there will not be a context per definition. As to #1207, the main challenge there is runtime lookup of a serializer that is statically attached to an interface.

AdlerFleurant:
@sandwwraith and @pdvrieze, I think it makes sense not to prefer contextual serializers. It is still however preferable to use contextual serializers over the default PolymorphicSerializer when a contextual serializer is defined. I will be updating this so that contextual serializers are preferred over the default polymorphicSerializer instead of all serializers.

Would that work?

sandwwraith:
This is still a big semantic change that can break someone's code, unfortunately. I think you can always specify the serializer explicitly in `encodeToString` if you want.

PR #2031, opened 2022-09-06
Add functions that can be used by compiler plugin for intrinsics

sandwwraith:
Intrinsics for `serializer<T>()` function should be available in the compiler plugin in Kotlin 1.8.0

### DIFF
--- a/core/commonTest/src/kotlinx/serialization/CachedSerializersTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/CachedSerializersTest.kt
@@ -4,8 +4,10 @@
 
 package kotlinx.serialization
 
+import kotlinx.serialization.modules.*
 import kotlinx.serialization.test.noJsLegacy
 import kotlin.test.*
+import kotlin.time.*
 
 class CachedSerializersTest {
     @Serializable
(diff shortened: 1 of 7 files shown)


core/commonTest/src/kotlinx/serialization/SerializersLookupEnumTest.kt
```diff
@@ -74,6 +74,7 @@ class SerializersLookupEnumTest {
     }
 
     @Test
+    @Ignore // @Polymorphic enum doesn't make sense and is not supported in intrinsics
```
sandwwraith, line 77: We need to chose what to do with `@Polymorphic enum`. My suggestion is to remove this test completely.
Optionally — add diagnostic that prohibits `@Polymorphic` on enums and objects

qwwdfsad: I suggest removing it. The diagnostic would be a good next step -- could you please file a ticket?

core/commonTest/src/kotlinx/serialization/CachedSerializersTest.kt
```diff
@@ -34,4 +36,22 @@ class CachedSerializersTest {
...
+        val m = SerializersModule {  }
+        val direct = measureTime {
+            Object.serializer()
+        }
+        val directMs = direct.inWholeMicroseconds
+        val indirect = measureTime {
+            m.serializer<Object>()
+        }
+        val indirectMs = indirect.inWholeMicroseconds
+        if (indirectMs > directMs + (directMs / 4)) error("Direct ($directMs) and indirect ($indirectMs) times are too far apart")
```
qwwdfsad, line 54: Interesting idea, not sure if it won't be flaky, but let's check it out

core/commonTest/src/kotlinx/serialization/SerializersLookupObjectTest.kt
```diff
@@ -75,7 +75,8 @@ class SerializersLookupObjectTest {
     }
 
     @Test
-    fun testEnumPolymorphic() {
+    @Ignore // @Polymorphic object doesn't make sense and is not supported in intrinsics
```
qwwdfsad, line 78: Same here for diagnostic

qwwdfsad (approved):

PR #2040, opened 2022-09-22
Deprecate .isNullable in SerialDescriptor builder

### LINKED ISSUE #1929: `buildClassSerialDescriptor` does not create nullable descriptor
Originally reported as https://youtrack.jetbrains.com/issue/KT-52300

**To Reproduce**
```kotlin
val descriptor = buildClassSerialDescriptor("FooBar") {
    isNullable = true
}
assertTrue(descriptor.isNullable) // throws
```
**Expected behavior**

Either we should allow to create nullable descriptors via builder or remove `nullable` property from it, leaving only `descriptor.nullable` extension. The latter would always append `?` to serial name.

**Environment**
 - Library version: 1.3.2

sandwwraith:
because it wasn't working properly and better alternative exists.

Fixes #1929


qwwdfsad (commented):
Could you please elaborate on the part "wasn't working properly"?

qwwdfsad:
Ohhhh 🤦

qwwdfsad (approved):

Issue #1405, opened 2021-04-05
Provide way to encode unquoted String in JSON format


sandwwraith:
Mainly to write big numbers. See reasoning here: https://github.com/Kotlin/kotlinx.serialization/issues/1051#issuecomment-812130020

JakeWharton:
Another use case is embedding known JSON documents into others without having to go through an intermediate parsing step. As long as it's named sufficiently scary this can be a powerful escape hatch.

nikolaykasyanov:
Should this issue also include decoding JSON primitives as raw strings (to be able to convert them to smth mentioned `BigDecimal`)?

Kantis:
Do you see any downsides to this approach? I suppose it might be prone to breaking, but it seems to work at least.

```
object BigDecimalSerializer : KSerializer<BigDecimal> {

    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("BigDecimal", PrimitiveKind.STRING)

    override fun deserialize(decoder: Decoder): BigDecimal =
        (decoder as JsonDecoder).decodeJsonElement().jsonPrimitive.content.toBigDecimal()


    override fun serialize(encoder: Encoder, value: BigDecimal) {
        (encoder as JsonEncoder).encodeJsonElement(JsonPrimitive(JsonBigDecimal(value)))
    }
}

internal class JsonBigDecimal(value: BigDecimal) : Number() {

    private val serialized: String = value.toString()

    override fun toString() = serialized

    override fun toByte() = notSupported()
    override fun toChar() = notSupported()
    override fun toDouble() = notSupported()
    override fun toFloat() = notSupported()
    override fun toInt() = notSupported()
    override fun toLong() = notSupported()
    override fun toShort() = notSupported()

    private fun notSupported(): Nothing = throw IllegalStateException("This method should not be invoked, ever.")
}
```

DrMetallius:
This solution unfortunately only works only for decoding. When encoding, the value will get converted to `double` by `JsonLiteralSerializer`.

Kantis:
> 
> 
> This solution unfortunately only works only for decoding. When encoding, the value will get converted to `double` by `JsonLiteralSerializer`.

Darn. I accidentally asserted my expected _too precise for double_ number through a JSON parsing lib which ended up doing the same doble-conversion. 🙄 

Added a warning to my original comment and code snippet.

nikolaykasyanov:
[...]

Issue #1051, opened 2020-09-08
How to serialize BigInteger and BigDouble?


MMairinger:
Currently I'm in the need of serializing numbers with many decimal digits (30+ decimal digits to be more precise). To deserialzie a JSON file with this many decimal points I used Java's BigIntegers and BigDecimal types which works fine. The problem that arises is when I want to serialize that value. It will be serialized as an Integer or Double respectively which cuts and rounds the actual value.

One example value is this, received in a JSON file.
`0.083532162010669708251953125000`
After deserializing I will have exactly a value of that above as a JsonLiteral. But when I serialize this value I will get a result of:
`0.08353216201066971`
Which is not my desired result.
A snippet from my unit tests showing the differences between expected and actual:
![image](https://user-images.githubusercontent.com/29223706/92460085-d1de1580-f1c7-11ea-8e5d-ba029abdb1ba.png)

My questions are now how should I serialized BigIntegers and BigDoubles and will there be support for this kind of data types in the future?

elizarov:
Can you, please, provide the example code you use to serialize and deserialize your data between JSON and BigIntegers?

MMairinger:
`var deserializedJson= json.decodeFromString(JsonClassSerializer, jsonToParse)
        var convertedKotlin= convertJsonToKotlin(deserializedJson)
        var convertedJson= convertKotlinToJson(convertedKotlin)
        var serializedJson= generateJsonOutputFromKotlin(convertedJson, json)`

The above sequence is what I went through. I took a JSON file as String input, configured my Json object and then decoded that input into an object instance of some JsonClass containing properties like val x: JsonArray and so on.
Then I convert that object into a different a different type that contains useable data types like for val x: List<Map<String, Any?>> and stuff like that.
Then that instance is converted back into a JsonClass instance, because we can't serialize Any types.
Finally the json output is generated from that JsonClass and here the problem arises. Because the standard serializer for JsonArray writes Numbers as either Long or Double.

The initial deserialization is correct, the Json -> Kotlin conversion is correct, transforming it back from Kotlin -> Json is still correct but the standard serializer when you annotate your class with @Serializable writes BigDecimal and BigIntegers as Double and Long respectively.

To clarify I did not use a custom serializer, for anything here. I simply let kotlinx handle the serialization of the intermediate type and then converted that to the actual types.

This is how I converted a JsonLiteral to BigInteger and so on:
[...]

Issue #1557, opened 2021-06-15
Expose an API to read primitive values of JSON keys as-is as a string


nikolaykasyanov:
**What is your use-case and why do you need this feature?**

The primary use case for this I have in mind is to be able to serialize numbers into types like `BigInteger` without loosing precision caused by currently unavoidable intermediate floating point number conversions.

This issue is the "read" counterpart of #1405.

At the moment, the only way I could find to achieve that is to set `isLenient` to `true` and use `decodeString` in `KSerializer.deserialize` implementations. This is unfortunately not acceptable because it requires enabling lenient parsing for the whole payload.

Another way is to have numbers serialized as strings but that's not always an option because 1) it might require changing already existing web APIs and 2) people don't always control APIs they have to deal with.

**Describe the solution you'd like**

From the top of my head, adding a mehod like to `Decoder` would do the trick:

```
fun decodeRawPrimitive(): String
```

This method would read a value as long as its primitive (number, bool, string) and return it as a string without any processing.

For example, reading the value of `"name": "Joe"` would return `"\"Joe\""`, reading `"value": 0.43` would yield `"0.43"`, reading `"isExcited": true` would yield `"true"`.

If it encounters an object or array, it would throw, just like other `decode<PrimitiveType>` methods.

pdvrieze:
@nikolaykasyanov Looking at the library design I'm not sure how this would actually work. The problem with "raw" decoding is that it is inherently format specific. As an example, how would this work for protocol buffers or some other binary format? For some formats (like Android Parcelables) there are no delimiters/sizes so it is even impossible to read a "raw" value as you don't know its size.

In the case that you can control the on-the-wire format the solution is easy, use a custom serializer (for `BigDecimal`) to encode/decode to string rather than another value. It is also possible to special case specific formats in the serializer/deserializer although you should do so judiciously (with a generic fallback). What you really want is a format-specific way to handle the serialization of particular values in a "special" way. I would say that the solution would have to be some approach to register special casing to formats that get invoked based upon the registered serializer.

In summary:

- **Problem**: Handle custom special serialization for particular types eg.: BigDecimal, UByteArray,...
- **Restriction**: It should be flexible and not rely on hardcoding
- **Proposed solution direction**: Formats to provide override hooks based upon serializer/serialname(of the type). These overrides aren't quite serializers as they would be having the specific format/encoder/decoder as parameter (and possibly ways to do format specific things such as reading raw values/bytes).

[...]

Issue #1600, opened 2021-07-13
Error trying to decode literal String "null" in 1.2.2


bsimon36:
**Describe the bug**
Attempting to decode json with String value of "null" throws:

Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 13: Unexpected 'null' value instead of string literal

**To Reproduce**

@Serializable
data class Test(val str: String)

JSON.decodeFromString<Test>("{\"str\":\"null\"}")

**Expected behavior**
No error, should decode to Test("null")

**Environment**
 - Kotlin version: 1.5.20
 - Library version: 1.2.2
 - Kotlin platforms: [e.g. JVM, JS, Native or their combinations]
 - Gradle version: [e.g. 4.10]
 - IDE version (if bug is related to the IDE) [e.g. IntellijIDEA 2019.1, Android Studio 3.4]
 - Other relevant context [e.g. OS version, JRE version, ... ]

qwwdfsad:
Affects only `isLenient = true` json instances

ajhuntsman:
I just ran into this bug when trying to deserialize a JSON file that has translated content between German and English.

As it turns out, "null" is the German word for the English word "zero".

Attention Germany: I'm a big fan of Hefeweizen and schnitzel, but you may want to look into this in all your software.

Saxintosh:
Please... fix this bug ASAP! 🙏

PR #2041, opened 2022-09-22
allow encoding literal JSON values

aSemy:
This PR provides a new function for encoding raw JSON content, without quoting it as a string. This allows for encoding JSON numbers of any size or precision, so BigDecimal and BigInteger can be supported.

[Link to view the updated documentation](https://github.com/aSemy/kotlinx.serialization/blob/feat/1051-big_decimals/docs/json.md#encoding-raw-literal-json-content-experimental)

fix: #1051 
fix: #1405

related: 

* #1557
* https://github.com/ionspin/kotlin-multiplatform-bignum/issues/219

The implementation is similar to how unsigned numbers are handled.

* `JsonUnquotedLiteral()` is a new function that allows creating literal JSON content.
* I added `val coerceToInlineType` to `JsonLiteral`, so that `JsonUnquotedLiteral` could use `encodeInline()` 
* I defined `val jsonUnquotedLiteralDescriptor` as a 'marker', for use with `encodeInline()`
* `ComposerForUnquotedLiterals` (based on `ComposerForUnsignedNumbers`) will 'override' the encoder when a `JsonLiteral` has the `jsonUnquotedLiteralDescriptor` marker, and will encode the content as a string *without* surrounding quotes.

~I'm not sure how to proceed with decoding literal numbers.~ I just realised, there's no need for special implementation - primitives are all decoded the same.


aSemy:
I think more tests and specialist knowledge is needed, but the approach is ready for review. I'm really keen to provide a way of supporting BigDecimal, as it's been a blocker in a couple of my projects for a long time.

YanishR (commented):
When do you guys foresee there will be progress on this? I have a use case for it in my project

sandwwraith:
Hi, thanks for your PR! I think I have time for it on this week, stay tuned

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -61,6 +63,17 @@ public fun JsonPrimitive(value: Number?): JsonPrimitive {
     return JsonLiteral(value, isString = false)
 }
 
+/**
+ * Creates [JsonPrimitive] from the given number.
+ */
+public fun JsonLiteralNumber(value: String?): JsonPrimitive {
+    if (value == null) return JsonNull
+    if (!jsonLiteralNumberRegex.matches(value)) throw JsonEncodingException("not a number: $value")
```
sandwwraith, line 71: Regex validating would block various other use-cases, like the one mentioned here: https://github.com/Kotlin/kotlinx.serialization/issues/1405#issuecomment-813381394

Now I'm a bit hesitant about whether we should support this (because we can't parse it back as the same raw value, but it is not needed), but it looks like a good use case. Supporting it would also require renaming this function into something like `JsonUnquotedLiteral` to avoid any connection to numbers, so it's better to think about this from the beginning.

sandwwraith (changes requested):

sandwwraith:
I think we should have additional design discussion on a general approach (using `encodeInline` and `JsonPrimitive(isString = true` instead of a separate `JsonEncoder.encodeUnquoted` function); I'll keep you posted

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
aSemy: In order to keep this PR focused I just wanted to completely prevent encoding anything except JSON numbers. I thought that allowing encoding raw strings would be a much larger discussion that would prevent an incremental improvement. Of course, changing this PR to allow for encoding raw strings is also possible!

formats/json/commonMain/src/kotlinx/serialization/json/internal/Composers.kt
```diff
@@ -60,6 +60,13 @@ internal class ComposerForUnsignedNumbers(writer: JsonWriter, private val forceQ
     }
 }
 
+@SuppressAnimalSniffer
+internal class ComposerForLiteralNumbers(writer: JsonWriter, private val forceQuoting: Boolean) : Composer(writer) {
+    override fun printQuoted(value: String) {
+        if (forceQuoting) printQuoted(value) else super.print(value)
```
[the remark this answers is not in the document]
aSemy: thanks, fixed by changing `printQuoted(value)` to `super.printQuoted(value)`

formats/json/commonMain/src/kotlinx/serialization/json/internal/Composers.kt
```diff
@@ -60,6 +60,13 @@ internal class ComposerForUnsignedNumbers(writer: JsonWriter, private val forceQ
     }
 }
 
+@SuppressAnimalSniffer
```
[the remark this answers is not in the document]
sandwwraith: AnimalSniffer is a plugin that detects and reports usages of Java 8 API. It is useful for checking if your library can run on old Android. Now, given that the default Kotlin target is 1.8 and Android has desugaring, everything still works, but the plugin still complains.

You should move annotation to `TreeJsonEncoder.inlineUnsignedNumberEncoder` from `TreeJsonEncoder.encodeTaggedInline`

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
```diff
@@ -169,6 +173,14 @@ internal class StreamingJsonEncoder(
         else ComposerForUnsignedNumbers(composer.writer, forceQuoting)
     }
 
+    private fun composerForLiteralNumbers(): Composer {
```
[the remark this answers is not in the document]
aSemy: I think that works well 👍 . I've implemented it.

formats/json-tests/jvmTest/src/kotlinx/serialization/BigDecimalTest.kt
```diff
@@ -0,0 +1,220 @@
...
+}
+
+@Serializable
+private data class BigDecimalHolder(
+    @Contextual
+    val bd: BigDecimal,
+    @Contextual
+    val bdList: List<@Contextual BigDecimal>,
+    val bdMap: Map<@Contextual BigDecimal, @Contextual BigDecimal>,
+)
```
aSemy, line 197: I think the repetition of `@Contextual` here would be reduced by https://github.com/Kotlin/kotlinx.serialization/issues/507. (Of course here this can be set with a file-annotation, but in other codebases it's likely that BigDecimal is used in many, many files.

sandwwraith (changes requested):
After a discussion, the following points were established:

* Function should accept arbitrary strings and do not validate them with regex. This will allow users to do various json-embeddings, support non-base10 numbers, different notations, etc and will also simplify an invariant for users and save us from potential problems with regex correctness and performance.

* Function should accept a `String?` and return `JsonNull` if a null was passed to it, for consistency with `JsonPrimitive` overloads.
* However, if string literal `"null"` is passed (i.e. `value == "null"`), function should throw `IllegalArgumentException`. This is motivated by the fact that when we serialize such a literal, if effective becomes indistinguishable from `JsonNull` and can't be deserialized back to `JsonLiteral`. We cannot know for sure if this was a user intent to do so or bug in validation. Better to prohibit such cases. IAE message can sound like this: "It is impossible to create a literal unquoted value of 'null'. If you want to create json null literal, use JsonNull object, otherwise, use quoted literal." 

Summarising this all up, I suggest to name the new function 'JsonUnquotedPrimitive'. What do you think?

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -105,6 +126,14 @@ internal class JsonLiteral internal constructor(
     }
 }
 
+/** Used as a marker to indicate the encoder should use `encodeInline()` */
+internal val jsonLiteralNumberDescriptor: SerialDescriptor = JsonLiteralNumber.serializer().descriptor
```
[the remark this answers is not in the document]
aSemy: 1. ✅
2. `InlinePrimitiveDescriptor` is internal to kxs.core, so it can't be accessed in kxs.json. https://github.com/Kotlin/kotlinx.serialization/blob/daa95c79ffadc0eedbbb4a481a00556b78212e43/core/commonMain/src/kotlinx/serialization/internal/InlineClassDescriptor.kt#L28

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -61,6 +63,17 @@ public fun JsonPrimitive(value: Number?): JsonPrimitive {
     return JsonLiteral(value, isString = false)
 }
 
+/**
+ * Creates [JsonPrimitive] from the given number.
+ */
+public fun JsonLiteralNumber(value: String?): JsonPrimitive {
```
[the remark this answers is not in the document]
aSemy: @sandwwraith Regarding the rename, you suggested `JsonUnquotedPrimitive`. However, the provided value might not be a primitive (users could use it to encode JSON arrays or objects), and the because might have quotes it could be a bit misleading - the new function doesn't really care about quotes.

How about `JsonLiteralElement` or `JsonRawElement`?

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -61,6 +63,20 @@ public fun JsonPrimitive(value: Number?): JsonPrimitive {
     return JsonLiteral(value, isString = false)
 }
 
+/**
+ * Creates [JsonPrimitive] from the given number.
+ *
+ * // TODO add more docs once functionality is stabilised.
+ */
+@ExperimentalSerializationApi
+public fun JsonLiteralNumber(value: String?): JsonPrimitive {
```
aSemy, line 72: > However, if string literal "null" is passed (i.e. value == "null"), function should throw IllegalArgumentException. This is motivated by the fact that when we serialize such a literal, if effective becomes indistinguishable from JsonNull and can't be deserialized back to JsonLiteral. We cannot know for sure if this was a user intent to do so or bug in validation. Better to prohibit such cases. IAE message can sound like this: "It is impossible to create a literal unquoted value of 'null'. If you want to create json null literal, use JsonNull object, otherwise, use quoted literal."

@sandwwraith Could you explain more about the above, please? I don't understand how a problem could occur.

I thought maybe it was because serialisation should be symmetric, so all `JsonPrimitive`s should decode to `JsonPrimitive`. But `JsonNull` is a subtype of `JsonPrimitive`, so that seems like it works to me.

formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonRawElementTest.kt
```diff
@@ -0,0 +1,187 @@
+package kotlinx.serialization.json.serializers
+
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.json.*
+import kotlinx.serialization.test.assertFailsWithSerialMessage
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonRawElementTest : JsonTestBase() {
```
aSemy, line 9: I've expanded the tests here, but I got pretty stuck. Some of the failure messages don't make sense to me.

```
Failed with streaming = STREAMING expected: kotlinx.serialization.json.JsonLiteral<"lorem"> but was: kotlinx.serialization.json.JsonLiteral<"lorem">
```

```
expected:<{ ="invalid key"}> but was:<{" "="invalid key"}>
```

```
Failed with streaming = STREAMING expected: kotlinx.serialization.json.JsonLiteral<[1,2,3]> but was: kotlinx.serialization.json.JsonArray<[1,2,3]>
```

Can someone help out? :)

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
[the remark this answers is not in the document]
sandwwraith: It worth changing to `@InternalSerializationApi public fun InlinePrimitiveDescriptor` then

formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonRawElementTest.kt
sandwwraith: Actually, I think these tests don't have to be checked for decoding, because in most cases they would be different.

For example, if you encode `JsonRawElement("[1,2,3]")`, naturally, it appears in JSON as an array and is parsed back as an array, thus the difference. I don't think we should do anything with that, so it's OK here to test encoding.

sandwwraith:
Also, I've just noted that the PR is opened against `master` branch, while all new PRs with code should be opened against `dev`. Can you change the branch please?

formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonRawElementTest.kt
aSemy: Thanks, done!

aSemy:
I've updated the PR to point at `dev`, and merged `dev` into this branch, but it looks like `dev` is missing commits from `master`. Can `master` be merged into `dev`?

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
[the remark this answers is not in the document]
sandwwraith: Well, the 'unquoted' part simply means that we won't put quotes around the value. If one creates `JsonUnquotedPrimitive("\"foo\"")` of course it would be quoted — but that is defined by the data, not the function behavior.

> However, the provided value might not be a primitive 

Yes, that's a valid point. Maybe `JsonUnquotedLiteral` then?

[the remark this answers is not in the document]
aSemy: I'd prefer having 'Raw' because it matches similar APIs in other serialization libraries, and it's a common word ('raw' is mentioned in each of the tickets I linked in the PR description), but that's minor - `JsonUnquotedLiteral` is fine.

I have another thought though. What about providing a function overload for `JsonPrimitive()`?

```kotlin
/**
 * Creates [JsonPrimitive] from the given string.
 *
 *  @param[quoted] If false, the value will not be wrapped in quotes <insert more docs here>
 */
@ExperimentalSerializationApi
public fun JsonPrimitive(value: String?, quoted: Boolean): JsonPrimitive {
    return if (!quoted) JsonPrimitive(value)
    else when (value) {
        null -> JsonNull
        JsonNull.content -> throw JsonEncodingException("It is impossible to create a literal unquoted value of 'null'. If you want to create JSON null literal, use JsonNull object, otherwise, use JsonPrimitive")
        else -> JsonLiteral(value, isString = false, coerceToInlineType = jsonRawElementDescriptor)
    }
}
```

It's a little odd, because `@ExperimentalSerializationApi` means it has to be a separate function rather than adding a new argument with a default to the existing `JsonPrimitive()`, but if this gets promoted to stable, then the function could be merged.


```kotlin
public fun JsonPrimitive(value: String?, quoted: Boolean = true): JsonPrimitive {
  // as above
}
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
qwwdfsad: Hi!

Our concern is not about `null` and resulting `JsonNull`, but `"null"` *string* literal. 

The idea is the following -- with this PR, we enable bidirectional _unquoted_ string literal, meaning that we can both write them to `JsonElement` and parse them back.

The problem is simple -- in unquoted mode, both `null` and `"null"` have *the same* serialized form (`{"key": null}`) and cannot be distinguished when read back. It means that we have the following potential options:

* Allow both literals for writing (creating programmatically), but both of them as `JsonNull`. This may lead to an inconsistent data model, but it is consistent with `lenient` mode (https://github.com/Kotlin/kotlinx.serialization/issues/1600)
* Prohibit `"null"` literals as indistinguishable when reading it back. The possible workaround is to use quoted mode for such literals.

We'd like to stick with the second option, but we are open for discussion

aSemy: Sorry, I don't really follow. I'm asking so I can figure out how to write a unit test for this, because at the moment if I disable the exception

```kotlin
@ExperimentalSerializationApi
@Suppress("FunctionName")
public fun JsonUnquotedLiteral(value: String?): JsonPrimitive {
    return when (value) {
        null -> JsonNull
//        JsonNull.content -> throw JsonEncodingException("It is impossible to create a literal unquoted value of 'null'. If you want to create JSON null literal, use JsonNull object, otherwise, use JsonPrimitive")
        else -> JsonLiteral(value, isString = false, coerceToInlineType = jsonUnquotedLiteralDescriptor)
    }
}
```

none of the test fail, e.g. https://github.com/Kotlin/kotlinx.serialization/commit/d00bff2bbda69a17083480ec9daaaad5a54d2b42 (except for the test that is checking that `"null" throws an exception).

I think there are 3 options

1. `"null"` → `JsonNull`
2. `"null"` → `throw JsonEncodingException`
3. `"null"` → `JsonLiteral("null", isString = false)`

You say Option 1 can cause problems, Option 2 will prevent problems, but I think Option 3 is fine, and should work as expected.

Would 3 work?


```kotlin
@ExperimentalSerializationApi
@Suppress("FunctionName")
public fun JsonUnquotedLiteral(value: String?): JsonPrimitive {
    return when (value) {
        null -> JsonNull
        else -> JsonLiteral(value, isString = false, coerceToInlineType = jsonUnquotedLiteralDescriptor)
    }
}
```

qwwdfsad: Here is the test that expected to "fail" (if one replaces `println` with actual asserts):

```
@Test
fun testThatNullAndNullStringAreDistinguishable() {
    val withNull = buildJsonObject { put("key", JsonNull) }
    val withNullString = buildJsonObject { put("key", JsonUnquotedLiteral("null")) }
    println(withNull == withNullString) // 'false' as expected

    val withNullSerilized = Json.encodeToString(withNull)
    val withNullStringSerialized = Json.encodeToString(withNullString)

    println(withNullSerilized == withNullStringSerialized) // Should be 'false', but is 'true

    val deserialized1 = Json.decodeFromString<JsonElement>(withNullSerilized)
    val deserialized2 = Json.decodeFromString<JsonElement>(withNullStringSerialized)
    println(deserialized1 == deserialized2) // Should be 'false', but is 'true'
}
```

We have two **semantically different** json objects, yet they are serialized to the same JSON string and, when parsed, are coerced into one json object, effectively losing information.

While looking harmless, debugging such an error in an enterprise-level software, especially where `JsonElement` is used indirectly when read (e.g. by polymorphic serializer or some man-in-the-middle that invokes `JsonDecoder.decodeJsonElement`)

aSemy: Thanks for the test! I've added a bit more logging to go over why I'm confused...

(I disabled the `value == "null" -> throw Exception` for these tests)

### JsonObject.toString()

```kotlin
val withNull = buildJsonObject { put("key", JsonNull) }
val withNullString = buildJsonObject { put("key", JsonUnquotedLiteral("null")) }
println(
    """
    withNull:                   $withNull
    withNullString:             $withNullString
    withNull == withNullString: ${withNull == withNullString}
    """.trimIndent()
)
```

```text
withNull:                   {"key":null}
withNullString:             {"key":null}
withNull == withNullString: false
```

Which seems wrong to me... They look the same. I think it makes sense that `JsonUnquotedLiteral("null") == JsonNull`.


### Json.encodeToString 

```kotlin
val withNullEncoded = Json.encodeToString(JsonElement.serializer(), withNull)
val withNullStringEncoded = Json.encodeToString(JsonElement.serializer(), withNullString)
println(
    """
    withNullEncoded:                          $withNullEncoded
    withNullStringEncoded:                    $withNullStringEncoded
    withNullEncoded == withNullStringEncoded: ${withNullEncoded == withNullStringEncoded}
    """.trimIndent()
)
```

```text
withNullEncoded:                          {"key":null}
withNullStringEncoded:                    {"key":null}
withNullEncoded == withNullStringEncoded: true
```

This makes sense to me, they look the same. I don't see why it should be false. I'd expect that `JsonUnquotedLiteral("null")` produces outputs `null`.

### Json.decodeToString 

```kotlin
val withNullDecoded = Json.decodeFromString(JsonElement.serializer(), withNullEncoded)
val withNullStringDecoded = Json.decodeFromString(JsonElement.serializer(), withNullStringEncoded)

println(
    """
    withNullDecoded:                          $withNullDecoded
    withNullStringDecoded:                    $withNullStringDecoded
    withNullDecoded == withNullStringDecoded: ${withNullDecoded == withNullStringDecoded}
    """.trimIndent()
)
```

```text
withNullDecoded:                          {"key":null}
withNullStringDecoded:                    {"key":null}
withNullDecoded == withNullStringDecoded: true
```

Again, this also makes sense. They're both the same JSON content.

---

I think it's just difficult for me to imagine how this situation could happen. `JsonUnquotedLiteral()` is akin to disabling type-checking using `dynamic` in Kotlin/JS, so putting a minor safety net in for `"null"` seems unusual. The horse has already left the stable, as they say. 

I imagine that debugging issues basically be asking "are you using `JsonUnquotedLiteral()`? You probably shouldn't!" But you're the maintainers, not me! So if an exception on `"null"` makes it easier for you, then it makes sense.

Anyway, to wrap it up, I'll happily admit that I don't need to understand this! And I really want this PR ready as soon as possible, so I'll try and adapt the test you posted, and update the KDoc on `JsonUnquotedLiteral()`.

aSemy:
Okay, I think I've addressed all the comments, so hopefully there's just minor points to fix! 

I've added some Knit documentation, but I ran of energy towards the end so please give it a proof-read.

formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonUnquotedLiteralTest.kt
```diff
@@ -0,0 +1,137 @@
...
+import kotlinx.serialization.test.assertFailsWithSerialMessage
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonUnquotedLiteralTest : JsonTestBase() {
+
+    private fun assertUnquotedLiteralEncoded(inputValue: String) = parametrizedTest { mode ->
+        val unquotedElement = JsonUnquotedLiteral(inputValue)
+
+        assertEquals(
```
sandwwraith, line 14: Given that this assert is not dependent on the mode, it will be executed multiple times without any changes (currently there are 4 modes, IIRC). I suggest just moving it outside of `parametrizedTest` lambda

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -76,13 +80,52 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
 @Suppress("FunctionName", "UNUSED_PARAMETER") // allows to call `JsonPrimitive(null)`
 public fun JsonPrimitive(value: Nothing?): JsonNull = JsonNull
 
+/**
+ * Create a [JsonPrimitive] from the given string, without surrounding it in quotes.
+ *
+ * **Note:** Creating a literal unquoted value of `null` (as in, `value == "null"`) is forbidden. If you want to create
```
sandwwraith, line 86: I'd move this note (and without the word 'Note') down, under 'Be aware that'... because it would likely be a rare event

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -76,13 +80,52 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
...
+ * JSON null literal, use [JsonNull] object, otherwise, use [JsonPrimitive].
+ *
+ * This function is provided for encoding raw JSON values that cannot be encoded using the [JsonPrimitive] functions.
+ * For example,
+ *
+ * * precise numeric values (avoiding floating-point precision errors associated with [Double] and [Float]),
+ * * large numbers,
+ * * or complex JSON objects.
+ *
+ * Be aware that it is possible to create invalid JSON using this function!
```
sandwwraith, line 96: It's better to keep documentation tone neutral, without exclamation marks

```suggestion
 * Be aware that it is possible to create invalid JSON using this function.
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -76,13 +80,52 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
 @Suppress("FunctionName", "UNUSED_PARAMETER") // allows to call `JsonPrimitive(null)`
 public fun JsonPrimitive(value: Nothing?): JsonNull = JsonNull
 
+/**
+ * Create a [JsonPrimitive] from the given string, without surrounding it in quotes.
```
sandwwraith, line 84:
```suggestion
 * Creates a [JsonPrimitive] from the given string, without surrounding it in quotes.
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -76,13 +80,52 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
...
+ * For example,
+ *
+ * * precise numeric values (avoiding floating-point precision errors associated with [Double] and [Float]),
+ * * large numbers,
+ * * or complex JSON objects.
+ *
+ * Be aware that it is possible to create invalid JSON using this function!
+ *
+ * @see JsonPrimitive This is the preferred method for encoding JSON primitives.
+ * @throws JsonEncodingException If `value == "null"`
```
sandwwraith, line 99:
```suggestion
 * @throws JsonEncodingException if `value == "null"`
```

because pop-ups in IDEA render these as the single sentence

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -76,13 +80,52 @@ public fun JsonPrimitive(value: String?): JsonPrimitive {
...
+ * This function is provided for encoding raw JSON values that cannot be encoded using the [JsonPrimitive] functions.
+ * For example,
+ *
+ * * precise numeric values (avoiding floating-point precision errors associated with [Double] and [Float]),
+ * * large numbers,
+ * * or complex JSON objects.
+ *
+ * Be aware that it is possible to create invalid JSON using this function!
+ *
+ * @see JsonPrimitive This is the preferred method for encoding JSON primitives.
```
sandwwraith, line 98:
```suggestion
 * @see JsonPrimitive is the preferred method for encoding JSON primitives.
```

sandwwraith (commented):
I like your write-up! Only small comments left to fix.

Also, the build indeed gives knit errors, so rerun it `./gradlew knit` and check afterward (`./gradlew knitCheck`). Ping me if there are any problems.

aSemy:
I saw that knit was on 0.3.0, I bumped it to the latest: 0.4.0.

docs/json.md
```diff
@@ -25,6 +25,9 @@ In this chapter, we'll walk through features of [JSON](https://www.json.org/json
   * [Types of Json elements](#types-of-json-elements)
   * [Json element builders](#json-element-builders)
   * [Decoding Json elements](#decoding-json-elements)
+  * [Encoding raw, literal Json content (experimental)](#encoding-raw-literal-json-content-experimental)
+    * [Serializing large decimal numbers](#serializing-large-decimal-numbers)
+    * [Using `JsonUnquotedLiteral` to encode `"null"` is forbidden](#using-jsonunquotedliteral-to-encode-"null"-is-forbidden)
```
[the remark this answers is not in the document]
sandwwraith: It seems the issue is not fixes in 0.4.0 :( Perhaps you can rephrase it to avoid `"null"` mention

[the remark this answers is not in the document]
aSemy: Yeah, I'll try and rephrase it. I've just made an issue & PR for Knit https://github.com/Kotlin/kotlinx-knit/issues/45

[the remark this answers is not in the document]
aSemy: It's fixed now 👍

sandwwraith (approved):
Great job, thanks for implementing this important feature!

sandwwraith:
No, we don't usually publish snapshots. The next release is going to be around Kotlin 1.8.0 release, in a month or two

Issue #1895, opened 2022-04-05
Annotating property type with @Serializable does not override default serialzer


Lukellmann:
### Describe the bug

The default serializer of a serializable class can be overwritten by annotating the property that should be serialized with `SomeOtherSerializer` with `@Serializable(with = SomeOtherSerializer::class)`.

Since the `Serializable` annotation has the targets `PROPERTY`, `CLASS` and `TYPE`, I expected that this should also work when annotating the type of the property instead of the property itself.

However it seems like this is only true for classes that are not serializable by default and always require specifying a serializer.

The reason I want to annotate the type instead of the property is for using typealiases with it. The project that I want to use this for needs to serialize `Instant`s in different formats and to reduce repetition and improve readability I wanted to replace
```kotlin
@Serializable
data class Data(
    @Serializable(with = InstantInEpochSecondsSerializer::class)
    val instant: Instant
)
```
with
```kotlin
typealias InstantInEpochSeconds = @Serializable(with = InstantInEpochSecondsSerializer::class) Instant

@Serializable
data class Data(val instant: InstantInEpochSeconds)
```

### To Reproduce
The issue can be reproduced by the following code (requiring kotlinx-datetime as a dependency for the `Instant` class)

```kotlin
// the Instant class is annotated with @Serializable(with = InstantIso8601Serializer::class)

@Serializable
data class InstantDefault(val instant: Instant)

@Serializable
data class InstantProperty(
    @Serializable(with = InstantInEpochSecondsSerializer::class)
    val instant: Instant,
)

@Serializable
data class InstantType(
    val instant: @Serializable(with = InstantInEpochSecondsSerializer::class) Instant,
)


// the Duration class is not annotated with @Serializable

@Serializable
data class DurationProperty(
    @Serializable(with = DurationInSecondsSerializer::class)
    val duration: Duration,
)

@Serializable
data class DurationType(
    val duration: @Serializable(with = DurationInSecondsSerializer::class) Duration,
)


inline fun <reified T> printJson(value: T) = println(Json.encodeToString(value))

fun main() {
    val instant = Instant.fromEpochSeconds(10)
    val duration = 10.seconds

    // expected {"instant":"1970-01-01T00:00:10Z"}, got {"instant":"1970-01-01T00:00:10Z"}
    printJson(InstantDefault(instant))

    // expected {"instant":10}, got {"instant":10}
    printJson(InstantProperty(instant))

    // this is wrong:
[...]

PR #2048, opened 2022-10-03
Promote typealiases as a global configuration means

sandwwraith:


docs/serializers.md
```diff
@@ -810,6 +811,60 @@ fun main() {
 
 <!--- TEST --> 
 
+### Specifying serializer globally using typealias
+
+kotlinx.serialization tends to be an always-explicit framework when it comes to serialization strategies: normally,
```
qwwdfsad, line 816:
```suggestion
kotlinx.serialization tends to be the always-explicit framework when it comes to serialization strategies: normally,
```

docs/serializers.md
```diff
@@ -810,6 +811,60 @@ fun main() {
...
+they should be explicitly mentioned in `@Serialiazble` annotation. Therefore, we do not provide any kind of global serializer
+configuration (except for [context serializer](#contextual-serialization) mentioned later).
+
+However, in projects with big amount of files and classes in them it may be inconvenient to specify `@file:UseSerializers`
```
qwwdfsad, line 820:
```suggestion
However, in projects with a large number of files and classes, it may be too cumbersome to specify `@file:UseSerializers`
```

docs/serializers.md
```diff
@@ -810,6 +811,60 @@ fun main() {
...
+every time, especially for classes like `Date` or `Instant` that have fixed strategy of serialization across the project.
```
qwwdfsad, line 821:
```suggestion
every time, especially for classes like `Date` or `Instant` that have a fixed strategy of serialization across the project.
```

docs/serializers.md
```diff
@@ -810,6 +811,60 @@ fun main() {
...
+For such cases, it is possible to specify serializers using `typealias`es, as they preserve annotations, including serialization ones:
```
qwwdfsad, line 822:
```suggestion
For such cases, it is possible to specify serializers using `typealias`es, as they preserve annotations, including serialization-related ones:
```

docs/serializers.md
```diff
@@ -810,6 +811,60 @@ fun main() {
...
+<!--- INCLUDE
+import java.util.Date
+import java.text.SimpleDateFormat
+  
+object DateAsLongSerializer : KSerializer<Date> {
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Date", PrimitiveKind.LONG)
```
qwwdfsad, line 828: We tend to recommend using `serialName` as a unique identifier. Here we do have two serializer with the same identifier

formats/json-tests/commonTest/src/kotlinx/serialization/SerializableOnPropertyTypeAndTypealiasTest.kt
```diff
@@ -0,0 +1,94 @@
...
+        val t = TesterWithDefault(WithDefault("a"), WithDefault("b"), WithDefault("c"), WithDefault("d"))
+        assertJsonFormAndRestored(
+            TesterWithDefault.serializer(),
+            t,
+            """{"b1":{"s":"a"},"b2":"b#","b3":"c#","b4":"d#"}"""
+        )
+    }
+
+    @Test
+    @Ignore // TODO: Unignore in 1.8.20
```
qwwdfsad, line 85: Could you please clarify why it is not working now?

docs/serializers.md
sandwwraith: It's not visible to user btw, but OK

formats/json-tests/commonTest/src/kotlinx/serialization/SerializableOnPropertyTypeAndTypealiasTest.kt
sandwwraith: It's a bug that affects only classes with overridden class serializer: https://github.com/Kotlin/kotlinx.serialization/issues/1895#issuecomment-1103453268

qwwdfsad (approved):
Great job!

Issue #2035, opened 2022-09-14
Incorrect deserialization of sealed class


pablobaldez:
**Describe the bug**
I'm using the kotlin serialization properties (not the json) and I'm getting `MissingFieldException` error when I try to decode a sealed class from a map
**To Reproduce**
I have a map
```kotlin
// my sealed classes 
@Serializable
@Keep
sealed class Odds {
    abstract val value: Long
}

@SerialName("EUROPEAN")
@Serializable
@Keep
data class European(override val value: Long) : Odds()

@SerialName("AMERICAN")
@Serializable
@Keep
data class American(override val value: Long) : Odds()


val map = mapOf("type" to "AMERICAN", "value" to 1)
val odds: Odds = Properties.decodeFromMap(map)
```

This is the stracktrace I have
```
Caused by: kotlinx.serialization.MissingFieldException: Field 'value' is required for type with serial name 'AMERICAN', but it was missing
                                                                                         	at kotlinx.serialization.internal.PluginExceptionsKt.throwMissingFieldException(PluginExceptions.kt:20)
                                                                                         	at br.com.mob1st.bet.features.competitions.domain.American.<init>(Bet.kt:36)
                                                                                         	at br.com.mob1st.bet.features.competitions.domain.American$$serializer.deserialize(Bet.kt:36)
                                                                                         	at br.com.mob1st.bet.features.competitions.domain.American$$serializer.deserialize(Bet.kt:36)
                                                                                         	at kotlinx.serialization.encoding.Decoder$DefaultImpls.decodeSerializableValue(Decoding.kt:257)
                                                                                         	at kotlinx.serialization.internal.TaggedDecoder.decodeSerializableValue(Tagged.kt:178)
                                                                                         	at kotlinx.serialization.internal.TaggedDecoder.decodeSerializableValue(Tagged.kt:207)
                                                                                         	at kotlinx.serialization.internal.TaggedDecoder$decodeSerializableElement$1.invoke(Tagged.kt:280)
                                                                                         	at kotlinx.serialization.internal.TaggedDecoder.tagBlock(Tagged.kt:297)
                                                                                         	at kotlinx.serialization.internal.TaggedDecoder.decodeSerializableElement(Tagged.kt:280)
                                                                                         	at kotlinx.serialization.encoding.CompositeDecoder$DefaultImpls.decodeSerializableElement$default(Decoding.kt:533)
[...]

PR #2052, opened 2022-10-10
Fix incorrect behavior while deserializing maps to sealed classes

rodrigovedovato:
This is an initial attempt at fixing an incompatibility between the deserialization of sealed classes when using json vs map

fix: #2035 

By reusing some of the stuff from [StreamingJsonDecoder](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt) I was able to succesfully deserialize a map that has the same structure as the test json.

Am I in the right path?

### DIFF
--- a/formats/json-tests/jvmTest/src/kotlinx/serialization/SealedClassSerializationFromJsonTest.kt
+++ b/formats/json-tests/jvmTest/src/kotlinx/serialization/SealedClassSerializationFromJsonTest.kt
@@ -1,0 +1,33 @@
+package kotlinx.serialization
+
+import kotlinx.serialization.json.Json
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+
+class SealedClassSerializationFromJsonTest {
+    @Serializable
+    sealed class BaseClass {
+        abstract val firstProperty: Long
+        abstract val secondProperty: String
+    }
+
+    @SerialName("FIRSTCHILD")
+    @Serializable
+    data class FirstChild(override val firstProperty: Long, override val secondProperty: String) : BaseClass()
+
+    @SerialName("SECONDCHILD")
+    @Serializable
+    data class SecondChild(override val firstProperty: Long, override val secondProperty: String) : BaseClass()
+    
+    @Test
+    fun testJsonSerialization() {
+        val json = """{"type": "FIRSTCHILD", "firstProperty": 1, "secondProperty": "one"}"""
+
+        val instance: BaseClass = Json.decodeFromString(json)
+
+        assertIs<FirstChild>(instance)
+        assertEquals(instance.firstProperty, 1)
+        assertEquals(instance.secondProperty, "one")
+    }
+}
--- a/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
+++ b/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
@@ -89,6 +89,22 @@ public sealed class Properties(
             return structure(descriptor).also { copyTagsTo(it) }
         }
 
+        final override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
+            val type = map["type"]?.toString()
+            var actualSerializer: DeserializationStrategy<out Any>? = null
+
+            if (type != null && deserializer is AbstractPolymorphicSerializer<*>) {
+                actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)
+            }
+
+            if (actualSerializer == null) {
+                throw RuntimeException()
+            }
+
+            @Suppress("UNCHECKED_CAST")
+            return actualSerializer.deserialize(this) as T
+        }
+
         final override fun decodeTaggedValue(tag: String): Value {
             return map.getValue(tag)
         }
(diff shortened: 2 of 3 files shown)


formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -89,6 +89,22 @@ public sealed class Properties(
...
+        final override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
+            val type = map["type"]?.toString()
+            var actualSerializer: DeserializationStrategy<out Any>? = null
+
+            if (type != null && deserializer is AbstractPolymorphicSerializer<*>) {
+                actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)
+            }
+
+            if (actualSerializer == null) {
+                throw RuntimeException()
```
sandwwraith, line 101: You can use `findPolymorphicSerializer` without 'orNull'. It will throw a `SerializationException` with a proper message for you.

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -89,6 +89,22 @@ public sealed class Properties(
...
 
+        final override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
+            val type = map["type"]?.toString()
+            var actualSerializer: DeserializationStrategy<out Any>? = null
+
+            if (type != null && deserializer is AbstractPolymorphicSerializer<*>) {
+                actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)
+            }
+
+            if (actualSerializer == null) {
```
sandwwraith, line 100: Btw, this approach breaks existing code (check failing tests on CI): you never use the original `deserializer` if it's not a polymorphic one

sandwwraith (changes requested):
Hi, thanks for the PR. Yeah, I think that's the right way to solve the problem; it just needs more polishing.

Please also try to fix this problem for serializing to map, so this functionality won't be half-baked.

formats/json-tests/jvmTest/src/kotlinx/serialization/SealedClassSerializationFromJsonTest.kt
```diff
@@ -0,0 +1,33 @@
+package kotlinx.serialization
+
+import kotlinx.serialization.json.Json
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+
+class SealedClassSerializationFromJsonTest {
```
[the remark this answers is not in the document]
rodrigovedovato: Yeah, I think so. I created this one to make sure that the result was exactly the same, so this will be removed in the final version

rodrigovedovato:
Hey, @sandwwraith! I think I managed to do everything you suggested, but I have a doubt concerning the serialization process.

While it was possible to make it work by creating another method especifically for that, I think the best option would be using the build-in `encodeSerializableValue`. However, AbstractPolymorphicSerializer's upper bound is stricter than the one `encodeSerializableValue`, which allows null values, making it impossible to call `findPolymorphicSerializerOrNull` using the value directly. I tried changing `encodeSerializableValue` to work exclusively with non-nullable values, but it didn't go well.

Do you think there's a better approach to solve this?

sandwwraith:
It seems that JSON just ignores nullability: https://github.com/Kotlin/kotlinx.serialization/blob/93a06df4bee36dc43b7d906ca395b0ac0d3229f3/formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L26

It makes sense, because AbstractPolymorphicSerializer does not support nullable values (they're handled mainly via `NullableSerializer`). So, if `deserializer is AbstractPolymorphicSerializer`, you can safely use `value as Any`

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -89,6 +101,26 @@ public sealed class Properties(
             return structure(descriptor).also { copyTagsTo(it) }
         }
 
+        final override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
+            val type = map["type"]?.toString()
+            if (deserializer is AbstractPolymorphicSerializer<*>) {
+                var actualSerializer: DeserializationStrategy<out Any>? = null
+
+                if (type != null) {
```
sandwwraith, line 109: Actually, `type == null` is also a valid case if someone defined [polymorphic default serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#default-polymorphic-type-handler-for-deserialization) — it may return serializer even if type wasn't specified.

Also, I do not see the reasons to check if `actualSerializer == null` and the call `findPolymorphicSerializer` again just to throw exception. 

These four lines can be simplified to `actualSerializer = deserializer.findPolymorphicSerializer(this, type)`. Am I right?

(double-checking in Json happens due to the fact that first time we look _optimistically_ hoping that type is the first key in object and we wouldn't have to read whole object as tree into memory just to find type token)

sandwwraith (changes requested):

formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
rodrigovedovato: Yeah, you're totally right. I did not know you could define your own polymorphic serializer, so it makes no sense checking if type is null

Thanks for the tips!

sandwwraith (approved):
Thanks a lot for your contribution!

rodrigovedovato:
Glad I could help, @sandwwraith!

PR #2061, opened 2022-10-13
Fix minor documentation and comment typos

aSemy:
Very minor points. I'm mostly committing so my IDE will show fewer warnings.

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
@@ -268,8 +268,8 @@ internal open class StreamingJsonDecoder(
 
     override fun decodeBoolean(): Boolean {
         /*
-         * We prohibit non-true/false boolean literals at all as it is considered way too error-prone,
-         * but allow quoted literal in relaxed mode for booleans.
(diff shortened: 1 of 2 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -268,7 +268,7 @@ internal open class StreamingJsonDecoder(
 
     override fun decodeBoolean(): Boolean {
         /*
-         * We prohibit non true/false boolean literals at all as it is considered way too error-prone,
+         * We prohibit non-true/false boolean literals at all as it is considered way too error-prone,
```
sandwwraith, line 271: Here it would actually change the meaning. The original comment "non true/false" means "any literal that is not equal to 'true' or 'false'", while changed versions read as "We prohibit non-true OR false boolean literals".

You can rephrase this to "We prohibit any boolean literal that is not strictly 'true' or 'false'"

sandwwraith (changes requested):

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
aSemy: good spot, thanks. Your rephrasing is much more clear.

sandwwraith (approved):

PR #2062, opened 2022-10-13
Update test assertions to have clickable assertion failures

aSemy:
This makes test failures easier to see, because when the test fails IntelliJ will generate a 'click to see difference' link that opens the IntelliJ comparison editor, with the two strings provided.

https://stackoverflow.com/q/10934743/4161471


sandwwraith (approved):
Thanks!

PR #2068, opened 2022-10-19
Consolidate exception message and improve it

### LINKED ISSUE #2066: `kotlinx.serialization.SerializationException` message can be misleading/confusing
**Describe the bug**

I recently tried setting up kotlinx.serialization in a project as part of using [graphql-kotlin](https://opensource.expediagroup.com/graphql-kotlin/docs/) in order to replace the jackson serializer we were using.  I unfortunately didn't read the setup instructions very thoroughly (and the setup instructions don't fully apply to us since we use bazel instead of gradle).  After adding the necessary package dependencies, ensuring `@Serializable` annotations were on the data classes and trying to use the kotlinx serializer, I started getting errors like this: 

```
kotlinx.serialization.SerializationException: Serializer for class 'Result' is not found.
```

This was incredibly confusing: The `Result` class _was_ marked with `@Serializable`, but the error message made it seem like the annotation was missing.  I didn't (yet) understand that I had to configure a kotlin compiler plugin and the error message did nothing to point me in that direction.  I've since figured out how to setup the the compiler plugin and now have it working, but the misleading error message was quite unhelpful.

So, my request is to consider improving the `kotlinx.serialization.SerializationException` so that if the type _is_ annotated with `@Serializable` it doesn't suggest that you add it.  Instead could it suggest checking to see if the compiler plugin is setup correctly.

Separate from that, having some bazel instructions in the README would be nice.  I wound up following this example:

https://github.com/bazelbuild/rules_kotlin/tree/48cc40371b071f0c11da4493cfe46617eac4da9b/examples/plugin/src/serialization

...so maybe the README could just link there if y'all don't want to maintain the necessary bazel instructions.

**To Reproduce**

Try using this library with `@Serializable` annotations but without the compiler plugin setup.  The resulting exception is quite confusing.

**Expected behavior**

The exception should ideally point the user in the direction of what they did wrong instead of misleading them.

**Environment**
 - Kotlin version: 16.20
 - Library version: 1.4.1
 - Kotlin platforms: JVM
 - Gradle version: N/A--I was using bazel (I believe this same issue would happen on any gradle version if you didn't setup the compiler plugin, though)
 - IDE version (if bug is related to the IDE): N/A
 - Other relevant context: N/A

qwwdfsad:
* Note that guide tests are not yet fixed

Fixes #2066

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
@@ -103,7 +103,7 @@ internal fun KType.kclass() = when (val t = classifier) {
     is KClass<*> -> t
     is KTypeParameter -> {
         error(
-            "Captured type paramerer $t from generic non-reified function. " +
+            "Captured type parameter $t from generic non-reified function. " +
                     "Such functionality cannot be supported as $t is erased, either specify serializer explicitly or make " +
                     "calling function inline with reified $t"
         )
--- a/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
@@ -93,7 +93,7 @@ internal fun KClass<*>.serializerNotRegistered(): Nothing {
 }
 
 internal fun KClass<*>.notRegisteredMessage(): String = "Serializer for class '${simpleName}' is not found.\n" +
-        "Please ensure that class is marked as '@Serializable' and that serialization plugin is applied."
+        "Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied."
 
 internal expect fun KClass<*>.platformSpecificSerializerNotRegistered(): Nothing
 
--- a/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
@@ -149,16 +149,6 @@ internal expect fun <T> createParametrizedCache(factory: (KClass<Any>, List<KTyp
 
 internal expect fun <T : Any, E : T?> ArrayList<E>.toNativeArrayImpl(eClass: KClass<T>): Array<E>
 
-/**
- * Checks whether the receiver is an instance of a given kclass.
- *
- * This check is a replacement for [KClass.isInstance] because on JVM it requires kotlin-reflect.jar in classpath (see KT-14720).
- *
- * On JS and Native, this function delegates to aforementioned [KClass.isInstance] since it is supported there out-of-the-box;
- * on JVM, it falls back to `java.lang.Class.isInstance`, which causes difference when applied to function types with big arity.
- */
-internal expect fun Any.isInstanceOf(kclass: KClass<*>): Boolean
-
 internal inline fun <T, K> Iterable<T>.elementsHashCodeBy(selector: (T) -> K): Int {
     return fold(1) { hash, element -> 31 * hash + selector(element).hashCode() }
 }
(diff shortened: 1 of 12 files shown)


qwwdfsad:
I'll fix guide tests if you are ok with the wording

core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
```diff
@@ -85,25 +85,30 @@ internal inline fun <T> SerializationStrategy<*>.cast(): SerializationStrategy<T
...
 
 @Suppress("UNCHECKED_CAST")
 internal fun KType.kclass() = when (val t = classifier) {
     is KClass<*> -> t
     is KTypeParameter -> {
-        error("Captured type paramerer $t from generic non-reified function. " +
-                "Such functionality cannot be supported as $t is erased, either specify serializer explicitly or make " +
-                "calling function inline with reified $t")
+        error(
+            "Captured type paramerer $t from generic non-reified function. " +
```
pdvrieze, line 106: `parameter` is spelled incorrectly

core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
```diff
@@ -85,25 +85,30 @@ internal inline fun <T> SerializationStrategy<*>.cast(): SerializationStrategy<T
...
-    throw SerializationException(
-        "Serializer for class '${simpleName}' is not found.\n" +
-            "Mark the class as @Serializable or provide the serializer explicitly."
-    )
+    throw SerializationException(notRegisteredMessage())
 }
 
+internal fun KClass<*>.notRegisteredMessage(): String = "Serializer for class '${simpleName}' is not found.\n" +
+        "Please ensure that class is marked as '@Serializable' and that serialization plugin is applied.\n" +
+        "It is also possible to specify serializer explicitly with '${simpleName}.serializer(typeArgs)'"
```
sandwwraith, line 97: We probably should also mention external non-default serializers here (they can only be used explicitly).

Also, the same message would be shown if use `someModule.serializer()` and no serializer is registered in the module.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -29,10 +28,7 @@ internal actual fun <T : Any, E : T?> ArrayList<E>.toNativeArrayImpl(eClass: KCl
 internal actual fun KClass<*>.platformSpecificSerializerNotRegistered(): Nothing = serializerNotRegistered()
 
 internal fun Class<*>.serializerNotRegistered(): Nothing {
-    throw SerializationException(
-        "Serializer for class '${simpleName}' is not found.\n" +
-                "Mark the class as @Serializable or provide the serializer explicitly."
-    )
+    throw SerializationException(this::class.notRegisteredMessage())
```
sandwwraith, line 31: Looks like this is an error: `this::class` is a `Class::class`

sandwwraith (changes requested):

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
qwwdfsad: Good catch

core/commonMain/src/kotlinx/serialization/internal/Platform.common.kt
qwwdfsad: I've reverted even the last line -- it looks alien and requires an additional explanation for cases like:
`It is also possible to specify serializer explicitly with 'Any.serializer(typeArgs)'`. Also, it doesn't seem that it can be reasonably explained **for a newcomer** why it is possible to specify serializer explicitly when it's not found (???).

The same story with modules -- all the phrasing and corner cases I've found are not worth it

sandwwraith (approved):

PR #2073, opened 2022-10-24
Add support for deserialization Duration in HOCON format

alexmihailov:
Currently **kotlin.time.Duration** by default supports deserialization from **ISO-8601-2** format. But **HOCON** uses a different [format](https://github.com/lightbend/config/blob/main/HOCON.md#duration-format).
Made changes so that when deserializing **kotlin.time.Duration** from **HOCON**, by default use the format specified in the [documentation](https://github.com/lightbend/config/blob/main/HOCON.md#duration-format).

### DIFF
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -22,9 +22,9 @@ import kotlinx.serialization.modules.*
  * [Config] object represents "Human-Optimized Config Object Notation" —
  * [HOCON][https://github.com/lightbend/config#using-hocon-the-json-superset].
  *
- * [Duration] objects decodes using "HOCON duration format" -
+ * [Duration] objects are decoded using "HOCON duration format" -
  * [Duration format][https://github.com/lightbend/config/blob/main/HOCON.md#duration-format]
- * [Duration] objects encodes does not currently support duration HOCON format and encodes to ISO-8601-2.
+ * [Duration] objects encoding does not currently support duration HOCON format and uses standard Duration serializer which produces ISO-8601-2 string. 
  *
  * @param [useConfigNamingConvention] switches naming resolution to config naming convention (hyphen separated).
  * @param serializersModule A [SerializersModule] which should contain registered serializers
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -5,10 +5,13 @@
 package kotlinx.serialization.hocon
 
 import com.typesafe.config.*
+import kotlin.time.*
 import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
 import kotlinx.serialization.encoding.CompositeDecoder.Companion.DECODE_DONE
+import kotlinx.serialization.hocon.internal.SuppressAnimalSniffer
 import kotlinx.serialization.internal.*
 import kotlinx.serialization.modules.*
 
@@ -18,6 +21,10 @@ import kotlinx.serialization.modules.*
  *
  * [Config] object represents "Human-Optimized Config Object Notation" —
  * [HOCON][https://github.com/lightbend/config#using-hocon-the-json-superset].
+ *
+ * [Duration] objects decodes using "HOCON duration format" -
+ * [Duration format][https://github.com/lightbend/config/blob/main/HOCON.md#duration-format]
+ * [Duration] objects encodes does not currently support duration HOCON format and encodes to ISO-8601-2.
  *
  * @param [useConfigNamingConvention] switches naming resolution to config naming convention (hyphen separated).
  * @param serializersModule A [SerializersModule] which should contain registered serializers
@@ -86,6 +93,18 @@ public sealed class Hocon(
 
         private fun getTaggedNumber(tag: T) = validateAndCast<Number>(tag)
 
+        @SuppressAnimalSniffer
+        protected fun <E> decodeDurationInHoconFormat(tag: T): E {
+            @Suppress("UNCHECKED_CAST")
+            return getValueFromTaggedConfig(tag) { conf, path ->
+                try {
+                    conf.getDuration(path).toKotlinDuration()
+                } catch (e: ConfigException) {
+                    throw SerializationException("Value at $path cannot be read as kotlin.Duration because it is not a valid HOCON duration value", e)
+                }
+            } as E
+        }
+
         override fun decodeTaggedString(tag: T) = validateAndCast<String>(tag)
 
         override fun decodeTaggedBoolean(tag: T) = validateAndCast<Boolean>(tag)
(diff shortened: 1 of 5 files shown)


formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconDurationDeserializerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+class HoconDurationDeserializerTest {
+
+    @Serializable
+    data class Simple(val d: Duration)
+
+    @Serializable
+    data class Nullable(val d: Duration?)
+
+    @Serializable
+    data class Complex(val i: Int, val s: Simple, val n: Nullable, val l: List<Simple>, val ln: List<Nullable>, val f: Boolean)
```
sandwwraith, line 22: I'd also add `List<Duration>` and `Map<String, Duration>` to this test. AFAIR, `Map<Duration, Duration>` is prohibited?

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconDurationDeserializerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        assertEquals(2, obj.ln.size)
+        assertNull(obj.ln[0].d)
+        assertEquals(6.hours, obj.ln[1].d!!)
+        assertEquals(6, obj.i)
+        assertTrue(obj.f)
+    }
+
+    @Test
+    fun testThrowsWhenNotTimeUnitHocon() {
+        assertThrows(SerializationException::class.java) {
```
sandwwraith, line 67: There's `assertFailsWith` in `kotlin.test` for that

sandwwraith (changes requested):
Overall the PR is good, but I have several thoughts:

- Hocon format also supports encoding to config (`Hocon.encodeToConfig`). Do you plan to support Duration in encoding, too? If so, how will it be encoded? Milliseconds only?
- Documentation should be updated to reflect this change. Since we do not have a separate HOCON entry in https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serialization-guide.md, KDoc update should do the job.
- Does it make sense to add a flag to restore old behavior? If one wants now to serialize Duration in ISO8601, they would have to resort to cumbersome workarounds like making a custom serializer that delegates to duration serializer but uses a different descriptor. However, I'm not well aware of hocon usages and if no one does ISO Duration serialization, I'm fine with not having flag.

ihostage:
Hi, Leonid @sandwwraith! 

> * Hocon format also supports encoding to config (`Hocon.encodeToConfig`). Do you plan to support Duration in encoding, too? If so, how will it be encoded? Milliseconds only?

Yes, we discussed this with Alex @alexmihailov and decided to implement this in separate PR later. You are right, it will be something like we implemented this case In Config4K 
https://github.com/config4k/config4k/blob/20294740a3735851e20287bdf38874b2bc1f6b63/src/main/kotlin/io/github/config4k/Extension.kt#L97
In this case, it's hard to come up with something striking better 🤷‍♂️ 

> * Does it make sense to add a flag to restore old behavior? If one wants now to serialize Duration in ISO8601, they would have to resort to cumbersome workarounds like making a custom serializer that delegates to duration serializer but uses a different descriptor. However, I'm not well aware of hocon usages and if no one does ISO Duration serialization, I'm fine with not having flag.

I think it's not worth it. Typesafe Config as HOCON reference implementation doesn't support ISO8601 and I've never seen that someone use this format for Duration 😄 I think we will be able to add a flag by request in the future if I'm wrong and it's really needed for someone.

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconDurationDeserializerTest.kt
alexmihailov: Made changes to support `List<Duration>` and `Map<String, Duration>`. `Map<Duration, Duration>` also works.

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -128,6 +131,13 @@ public sealed class Hocon(
         private fun composeName(parentName: String, childName: String) =
             if (parentName.isEmpty()) childName else "$parentName.$childName"
 
+        @SuppressAnimalSniffer
+        private fun Config.decodeDurationInHoconFormat(path: String) : Duration = try {
+            getDuration(path).toKotlinDuration()
+        } catch (e: ConfigException) {
+            throw SerializationException(e)
```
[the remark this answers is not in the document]
alexmihailov: Fixed

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -138,19 +148,24 @@ public sealed class Hocon(
         }
 
         override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
-            if (deserializer !is AbstractPolymorphicSerializer<*> || useArrayPolymorphism) {
-                return deserializer.deserialize(this)
-            }
-
-            val config = if (currentTagOrNull != null) conf.getConfig(currentTag) else conf
+            return when {
+                deserializer == Duration.serializer() -> {
```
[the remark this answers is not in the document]
alexmihailov: Fixed

formats/hocon/src/test/kotlin/kotlinx/serialization/hocon/HoconDurationDeserializerTest.kt
alexmihailov: Fixed

sandwwraith:
> I think it's not worth it. Typesafe Config as HOCON reference implementation doesn't support ISO8601 and I've never seen that someone use this format for Duration

OK, no problem

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -19,6 +22,10 @@ import kotlinx.serialization.modules.*
  * [Config] object represents "Human-Optimized Config Object Notation" —
  * [HOCON][https://github.com/lightbend/config#using-hocon-the-json-superset].
  *
+ * [Duration] objects decodes using "HOCON duration format" -
```
sandwwraith, line 25:
```suggestion
 * [Duration] objects are decoded using "HOCON duration format" -
```

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -19,6 +22,10 @@ import kotlinx.serialization.modules.*
...
+ * [Duration format][https://github.com/lightbend/config/blob/main/HOCON.md#duration-format]
+ * [Duration] objects encodes does not currently support duration HOCON format and encodes to ISO-8601-2.
```
sandwwraith, line 27:
```suggestion
 * [Duration] objects encoding does not currently support duration HOCON format and uses standard Duration serializer which produces ISO-8601-2 string. 
```

sandwwraith (commented):
Apart from minor documentation rewording, I do not have suggestions. However, I kindly ask you to submit encoding PR before the next release (which should happen in about a month), so the format continues to look consistent.

alexmihailov:
> Apart from minor documentation rewording, I do not have suggestions. However, I kindly ask you to submit encoding PR before the next release (which should happen in about a month), so the format continues to look consistent.


For encoding, I already have a solution, I will submit a PR soon

sandwwraith (approved):
Thank you, great job!

Issue #2075, opened 2022-10-24
Support Output/Input stream for protobuf format 


GeorgePap-719:
**What is your use-case and why do you need this feature?**

I'm using springboot and RSocket APIs with protobuf messages ( using kotlinx.serialization impl) and currently to support it i'm using my own custom spring-kotlin-encoder "glue-code". I would like to provide the first step to make it work out of the box ( i guess ktor also could use it?)  for simple and  [delimited messages](https://developers.google.com/protocol-buffers/docs/techniques?hl=en#streaming) 

Imho: I think the first step would be to provide jvm encoders that take input as io-stream.  

**Describe the solution you'd like**

Probably similar APIs like json format has:

`ProtoBuf.encodeToStream`

`ProtoBuf.decodeFromStream`

And for delimited messages:

`ProtoBuf.encodeDelimitedMessagesToStream`

`ProtoBuf.decodeDelimitedMessagesFromStream`

bishiboosh:
I would be quite interested by this to be able to use it Android Datastore. Can we take inspiration from what's been done in [Wire](https://github.com/square/wire) maybe ?

GeorgePap-719:
Ofc, feel free to take anything you like.

ShreckYe:
A vote for this.

One question: why don't you open a pull request or draft pull request for this? it can make reviewing changes easier.

And an immature suggestion: can we get it all multiplatform with [Coroutines channels](https://kotlinlang.org/docs/channels.html) or [kotlinx-io](https://github.com/Kotlin/kotlinx-io)? With channels we get suspending/non-blocking streaming support too. kotlinx-io provides the `Sink` and `Source` APIs though I am not sure how stable it is.

GeorgePap-719:
Hey @ShreckYe, what is your use-case for this feature? Can you explain it here, please.

> can we get it all multiplatform with [Coroutines channels](https://kotlinlang.org/docs/channels.html) or [kotlinx-io](https://github.com/Kotlin/kotlinx-io)? With channels we get suspending/non-blocking streaming support too. kotlinx-io provides the Sink and Source APIs though I am not sure how stable it is.

Usually these kind of low-level APIs are blocking for performance reasons. However, I can provide APIs which return a  `Sequence`, similar to what `Json` format provides.

ShreckYe:
[...]

Issue #1073, opened 2020-09-16
JSON: Provide non-blocking pull-parsing API capable of working with partially read documents


elizarov:
The use-case is parsing very large and incrementally incoming JSON arrays of objects. Like having `Flow<ByteArray>`/`Flow<ByteBuffer>` at the input and `Flow<Data>` at the output. To make it possible, there needs to an alternative low-level JSON parsing API in a non-blocking "pull" style. For performance reasons, it would be helpful to have a ready-to-use API to skip JSON values without parsing them. This way, after reading an array-opening `[` token, the inner JSON objects can be simply skipped to their end and, as soon as there are enough data bytes in the input, the resulting chunk of raw bytes from the input can be then fed to regular high-performance blocking parser to deserialize the object. 

Note, that "skipping without parsing" API to get raw bytes of a JSON object is needed for another use-case. See #1058 for details.

albertocavalcante:
Is there an update on this issue? Thanks

sandwwraith:
@albertocavalcante There's a `decodeToSequence` function https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-to-sequence.html but it uses blocking `InputStream` api

PR #2082, opened 2022-10-29
Add ProtoBuf I/O Stream support on jvm (#2075)

GeorgePap-719:
Add implementations for streaming support in Protobuf format supporting simple and delimited messages.

### DIFF
--- a/formats/protobuf/api/kotlinx-serialization-protobuf.api
+++ b/formats/protobuf/api/kotlinx-serialization-protobuf.api
@@ -1,3 +1,12 @@
+public final class kotlinx/serialization/protobuf/JvmStreamsKt {
+	public static final field DEFAULT_MESSAGE_MAX_SIZE I
+	public static final fun decodeDelimitedMessages (Lkotlinx/serialization/protobuf/ProtoBuf;Lkotlinx/serialization/DeserializationStrategy;Ljava/io/InputStream;I)Ljava/util/List;
+	public static synthetic fun decodeDelimitedMessages$default (Lkotlinx/serialization/protobuf/ProtoBuf;Lkotlinx/serialization/DeserializationStrategy;Ljava/io/InputStream;IILjava/lang/Object;)Ljava/util/List;
+	public static final fun decodeFromStream (Lkotlinx/serialization/protobuf/ProtoBuf;Lkotlinx/serialization/DeserializationStrategy;Ljava/io/InputStream;)Ljava/lang/Object;
+	public static final fun encodeDelimitedToStream (Lkotlinx/serialization/protobuf/ProtoBuf;Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Ljava/io/OutputStream;)V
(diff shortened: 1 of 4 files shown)


qwwdfsad (commented):
Hi, thanks for the effort! 

Unfortunately, we cannot accept your contribution in its current form (you can use https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md for a more predictable result in the future).

In short, the first glance found the following issues:

* At the moment, when there is more than a single way to encode/decode data, we should test both of them **in each test**
* There is a new concept, decode/encode "delimited", but in JSON we already have a different approach. We prefer to be consistent within our library, so such API definitely should be discussed beforehand.
* The API that fails at the moment when the message is greater than some pre-defined constant is just a time bomb waiting to explode in other people's production, which is a no-go
* Streaming API implies that both encoding and decoding of data are streamed. An implementation that reads/writes to `ByteArray` whole message and then passes it as is to the underlying IO stream does not pass this criterion

GeorgePap-719:
Hello, thanks for the feedback much appreciate it!

yes i used the Contributing.md as base line for the pr, but since i had a draft in mind i figured i would open a pr for faster feedback not sure if it's ok or not :) 

Is it alright if i continue to work this contribution based on your feedback or should i just also close the related issue?

qwwdfsad:
The issue is fine -- let's gather some feedback on whether there is a demand for it.

I suggest addressing the most important point, streaming, in the issue, though it's unclear whether it is possible to do so at all due to weird protobuf nature when it's impossible to write a message unless knowing its varint-encoded size in advance

GeorgePap-719:
May i ask a stupid question? How we gather feedback? :P

Ok, i will work on streaming.

About demand not sure if this counts but:

My use case was to provide support for protobuf with kotlin's impl in spring, for example to be able to impl something like this:
https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/codec/protobuf/ProtobufEncoder.java

The encoder when the stream is through flux is using delimited messages. So in order to provide a compatible solution i guess the kotlin's one would need to have delimited messages (i think).

Also, recently they provided a simple integration kotlin's protobuf impl here:
https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/codec/protobuf/KotlinSerializationProtobufEncoder.java

But they state the encoding stream is not yet supported and they have linked the #1073

qwwdfsad:
>How we gather feedback? :P

Mostly wait for people to pass by and leave their comments about the feature request. Common sense also plays its role here :) 

The use case is perfectly valid; what puzzles me is how it is possible to make an actual streaming protobuf without reading all to byte array first

GeorgePap-719:
After some research it seems that there is a way calculate the serialized size without encoding the message. Java's implementation does it, the only "trick" to it is that the function is generated by the compiler plugin. 

The actual implementation seems simple (not sure from compiler's pov tho), at first at glance at least.

I have almost-zero experience on compilers but i will give it a try to produce a draft.

PR #2086, opened 2022-11-05
Improve README to provide a link to a Bazel example.

myronmarston:
This resolves one aspect of #2066.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Kotlin serialization consists of a compiler plugin, that generates visitor code 
   * [Android](#android)
   * [Multiplatform (Common, JS, Native)](#multiplatform-common-js-native)
   * [Maven](#maven)
+  * [Bazel](#bazel)
 
 <!--- END -->
 
@@ -335,3 +336,9 @@ Add dependency on serialization runtime library:
     <version>${serialization.version}</version>
 </dependency>
 ```
+
+### Bazel
+
+To setup the Kotlin compiler plugin for Bazel, follow [the
+example](https://github.com/bazelbuild/rules_kotlin/tree/master/examples/plugin/src/serialization)
+from the `rules_kotlin` repository.


README.md
```diff
@@ -335,3 +335,9 @@ Add dependency on serialization runtime library:
     <version>${serialization.version}</version>
 </dependency>
 ```
+
+### Bazel
+
+To setup the Kotlin compiler plugin for Bazel, follow [the
+example](https://github.com/bazelbuild/rules_kotlin/tree/v1.6.0/examples/plugin/src/serialization)
```
sandwwraith, line 342: Are you sure that fixing `v1.6.0` tag is a good idea? E.g. there is an update in `master` that removes kotlin-reflect from test dependencies.

sandwwraith (changes requested):

README.md
myronmarston: As a general rule, I _always_ include a SHA or tag in a github code link rather than linking to a file on the master or main branch. Otherwise, the link might stop working at some point.  For example, if the maintainers of that repo reorganize the directory structure at some point such that in `master` a file or directory at that path no longer exists...then the link would become a 404.

Also, if someone follows this link and wants to see the latest version of the directory, it's easy for them to replace `v1.6.0` with `master` in their URL to see it.  But the opposite is not true: if we put `master` in the URL, it's not easy for someone to guess what version we were looking at when we wrote this to know what tag or SHA to put in the URL to see what we had in mind when we linked.

All that said: this is linking to an `examples` directory that I expect to be pretty stable, and I think it's reasonable here to link to `master` to optimize for linking folks to the current version of the example.

Either way, let me know if you want me to change it.  It's your repo and I'm happy to defer to whatever you prefer!

README.md
```diff
@@ -335,3 +335,9 @@ Add dependency on serialization runtime library:
     <version>${serialization.version}</version>
 </dependency>
 ```
+
+### Bazel
```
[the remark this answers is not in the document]
myronmarston: (But I'm going to wait for a reply to the other comment thread below so that if I'm changing both things I do it all at once.)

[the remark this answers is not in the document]
myronmarston: Done.

README.md
myronmarston: Updated to `master`.

sandwwraith (approved):
Thanks!

PR #2100, opened 2022-11-18
Introduce CharArray caching for InputStream decoding

qwwdfsad:


qwwdfsad:
JacksonComparisonBenchmark.kotlinFromStream: 260 ops/ms -> 460 ops/ms (NB: it's ASCII-only)

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
```diff
@@ -48,8 +52,12 @@ internal fun <T> Json.decodeToSequenceByReader(
     format: DecodeSequenceMode = DecodeSequenceMode.AUTO_DETECT
 ): Sequence<T> {
     val lexer = ReaderJsonLexer(reader)
-    val iter = JsonIterator(format, this, lexer, deserializer)
-    return Sequence { iter }.constrainOnce()
+    try {
+        val iter = JsonIterator(format, this, lexer, deserializer)
+        return Sequence { iter }.constrainOnce()
+    } finally {
+        lexer.release()
```
sandwwraith, line 59: I may be wrong, but we can't do `release` here because the sequence is not yet evaluated, so a buffer is still needed

qwwdfsad: I'll double-check, thanks

qwwdfsad: You a right: added stress test for each mode and fixed the issue

formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharArrayPool.kt
```diff
@@ -13,20 +17,37 @@ internal object CharArrayPool {
         System.getProperty("kotlinx.serialization.json.pool.size").toIntOrNull()
     }.getOrNull() ?: 1024 * 1024 // 2 MB seems to be a reasonable constraint, (1M of chars)
 
-    fun take(): CharArray {
+    protected fun take(size: Int): CharArray {
```
shanshin, line 20: I think it's better to take the size in the class constructor rather than in the function call.

In fact, this is the case for each instance of the pool right now.
Another reason is a potential error if I call like this
```
val a = take(100)
release(a)
val b = take(100)
release(b)
val c = take(200)
c[101] // !!!
```
The API does not advise in any way that you should not do this.

If you add a size to the constructor, then you can also add a check for the size of the array inside the `releaseImpl`

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
```diff
@@ -47,6 +51,7 @@ internal fun <T> Json.decodeToSequenceByReader(
     deserializer: DeserializationStrategy<T>,
     format: DecodeSequenceMode = DecodeSequenceMode.AUTO_DETECT
 ): Sequence<T> {
+    // Note: no explicit release, as the sequence are lazy and thrown away in an arbitrary manner
```
shanshin, line 54: Do I understand correctly that this function permanently "steals" an array from the pool and if `decodeToSequenceByReader` and `decodeByReader` are called in the application, then in fact this is equivalent to creating a new instance of the array for each call to `decodeToSequenceByReader`?

formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharArrayPool.kt
qwwdfsad: This is a good remark for any general-purpose API, but in this situation, we are dealing with well-tailored internal pools for one specific purpose. Also, it will force us to introduce a few more bytecodes -- for class with ctor, for outer class that will store the instance etc.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
qwwdfsad: Correct

formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharArrayPool.kt
shanshin: In the byte code, only a new field with the int type will be added (the parameter will be transferred from the `take` method to the `init` method), and in the child class, only the value from the parameter will be setted to the child constructor.

I think this is a good tradeoff so that if someone else refactor the code after a while, they don't add an unobvious bug.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
shanshin: Maybe in this case it is better not to use the pool for `decodeToSequenceByReader` so that not affects `decodeByReader`?
At the same time, in any case, each call to `decodeToSequenceByReader` will require the creation of a new array (possibly in another thread), so the total amount of work does not change from this.

formats/json-tests/jvmTest/src/kotlinx/serialization/json/JsonConcurrentStressTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                }
+                val bais = ByteArrayInputStream(baos.toByteArray())
+                assertEquals(values, Json.decodeToSequence(bais, Foo.serializer()).toList())
+            }
+        }
+    }
+
+    private fun getRandomString() = (1..Random.nextInt(0, charset.length)).map { charset[it] }.joinToString(separator = "")
+
+    private fun doTest(iterations: Int, block: (JsonTestingMode) -> Unit) {
```
sandwwraith, line 67: `JsonTestingMode` seems to be not used anywhere.

I'd also add test for `decodeFromStream`

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
sandwwraith: +1. Looks like 'taking and never returning' in decodeToSequence may negatively affect decodeFromStream

sandwwraith: Alternatively, you may try to add `lexer.release()` to `JsonIterator.hasNext()` implementation. This may solve the problems at the first sight but may require thorough testing.

sandwwraith: Although it would still steal buffer if the resulting sequence is used together with things like `.take()`/`limit`, but it's still better than current solution?

formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharArrayPool.kt
sandwwraith: Given that this is a protected method, I don't see that it is necessary for now, but the idea overall is a valid concern.

sandwwraith (changes requested):

formats/json-tests/jvmTest/src/kotlinx/serialization/json/JsonConcurrentStressTest.kt
qwwdfsad: It is used in the test bodies, streams are tested as well

formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharArrayPool.kt
qwwdfsad: The method is indeed protected and sued as a **base** for the implementation.
None of the implementations are prone to the potential bug described above and I prefer to avoid overcomplications in a 50 lines of fully internal code that has a single purpose

formats/json-tests/jvmTest/src/kotlinx/serialization/json/JsonConcurrentStressTest.kt
sandwwraith: Ah, I haven't noticed because it has the default name `it`. Maybe add a name for better readability?

sandwwraith (approved):

shanshin (approved):

formats/json-tests/jvmTest/src/kotlinx/serialization/json/JsonConcurrentStressTest.kt
qwwdfsad: Done

PR #2105, opened 2022-11-21
Provide killswitch for JPMS & improve building.md

sandwwraith:


buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -19,6 +19,8 @@ object Java9Modularity {
     @JvmStatic
     @JvmOverloads
     fun Project.configureJava9ModuleInfo(multiRelease: Boolean = true) {
+        val disableJPMS = this.rootProject.extra.has("disableJPMS")
```
qwwdfsad, line 22: or `Idea.isActive`?

qwwdfsad (approved):

buildSrc/src/main/kotlin/Java9Modularity.kt
sandwwraith: good idea

PR #2109, opened 2022-11-24
Fix incompatible targets in guide

Tapchicoma:
Fix for the change introduced in Kotlin 1.8.0 via this issue: https://youtrack.jetbrains.com/issue/KT-54993


qwwdfsad (approved):

qwwdfsad:
We can just disable "Create merge commit" button on GH if you find it appropriate

Issue #209, opened 2018-09-07
Parsing case insensitive enums


joserobjr:
How can we parse case insensitive enums [like we do on Jackson](https://stackoverflow.com/a/44217670/804976)?

Example:
```kotlin
import kotlinx.serialization.Serializable
import kotlinx.serialization.json.JSON
import kotlin.test.Test
import kotlin.test.assertEquals

enum class SampleEnum {
    FOO, BAR
}

@Serializable
data class SampleData(val a: String, val b: SampleEnum)

class TestCase {
    private val expected = SampleData("value", SampleEnum.FOO)

    @Test
    fun testUpperCase() {
        assertEquals(expected,
                JSON.unquoted.parse("{a:value,b:FOO}"))
    }

    @Test
    fun testLowerCase() {
        assertEquals(expected,
                JSON.unquoted.parse("{a:value,b:foo}"))
    }
}
```

Result:
```
No enum constant SampleEnum.foo
java.lang.IllegalArgumentException: No enum constant SampleEnum.foo
	at java.lang.Enum.valueOf(Enum.java:238)
	at kotlinx.serialization.SerializationKt.enumFromName(Serialization.kt:28)
	at kotlinx.serialization.json.JSON$JsonInput.readEnumValue(JSON.kt:304)
	at kotlinx.serialization.internal.EnumSerializer.load(BuiltinSerializers.kt:115)
	at kotlinx.serialization.internal.EnumSerializer.load(BuiltinSerializers.kt:112)
	at kotlinx.serialization.KInput.readSerializableValue(Serialization.kt:245)
	at kotlinx.serialization.ElementValueInput.readSerializableElementValue(Serialization.kt:440)
	at SampleData$$serializer.load(sample.kt)
	at SampleData$$serializer.load(sample.kt:11)
	at kotlinx.serialization.KInput.read(Serialization.kt:211)
	at kotlinx.serialization.json.JSON.parse(JSON.kt:42)
	at TestCase.testLowerCase(sample.kt:30)
```

henningBunk:
@sandwwraith it does not seem that #33 gets implemented. Would this issue have a bigger chance on it's own?

sandwwraith:
@henningBunk Perhaps, but I hope we would be able to implement #33 in one of the following releases

mecoFarid:
What @JakeWharton implied in #33 is "this probably won't happen"

mimoccc:
kotlinx.serialization.json.JsonNames

[...]

PR #2111, opened 2022-11-29
Provide support for JsonNamingStrategy 

### LINKED ISSUE #33: Fields name policy converters
A very common case of serialization is when API/JSON naming policy conflicts with your code naming, for example, snake case quite popular naming style for JSON fields, but for Kotlin it's camel case. Also, format naming style can be incompatible with JVM (for example field names with dashes).
Gson provides feature to solve that: https://google.github.io/gson/apidocs/com/google/gson/FieldNamingPolicy.html (implementation of FieldNamingStrategy)

In general, it's questionable feature, Moshi doesn't provide it and suggests to use the same case for your model classes.

But it can be a big blocker if you want to migrate to kotlinx.serialization, because you have 3 choices: 
1. Use `@SerialName` for each field. Tedious and error-prone and requires a lot of refactoring
2. Rename all your data models to match API name convention. Tedious, requires even more refactoring, violates Kotlin code style guides and backward incompatible.
3. Write own implementation of JSON (or any other format) that provides such feature, looks like a reasonable solution, but now you have one more copy JSON implementation in your project (because JSON implementation now a part of kotlinx.serialization) with different name, but the same API

Possible solutions:
1. Add naming policy feature to JSON implementation. Some drawbacks: doesn't work for other formats, debatable nature of this feature, required not for all use cases.
2. Provide universal API that allows defining name converters. A user can provide own implementation, so not necessary to include it to core fo kotlinx.serialization, works for all formats without additional effort. By default just uses field names as is, so current behavior.
Possible problems: If you use more than one serialization format, different formats probably require different strategies, so we still should integrate naming policy to format implementation. Also, you sometimes want to use different naming policy for particular requests (from different API), so looks like Serialization format still should provide API to register and use naming policies.

sandwwraith:
to be used in Json for enums and properties (configured separately)

Provide basic implementations: AllLowercase and SnakeCase

Fixes #33

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -7,6 +7,7 @@ package kotlinx.serialization.json
 import kotlinx.serialization.*
 import kotlinx.serialization.json.internal.*
 import kotlinx.serialization.modules.*
+import kotlinx.serialization.descriptors.*
 import kotlin.native.concurrent.*
 
 /**
@@ -314,7 +315,29 @@ public class JsonBuilder internal constructor(json: Json) {
     public var useAlternativeNames: Boolean = json.configuration.useAlternativeNames
 
     /**
+     * Specifies [JsonNamingStrategy] that should be used for all properties in classes for serialization and deserialization.
+     *
+     * `null` by default.
+     *
+     * This strategy is applied for all entities that have [StructureKind.CLASS].
+     */
+    public var namingStrategyForProperties: JsonNamingStrategy? = json.configuration.namingStrategyForProperties
+
+    /**
+     * Specifies [JsonNamingStrategy] that should be used for all enum instances for serialization and deserialization.
+     *
+     * `null` by default.
+     *
+     * This strategy is applied for all entities that have [SerialKind.ENUM].
+     */
+    public var namingStrategyForEnums: JsonNamingStrategy? = json.configuration.namingStrategyForEnums
+
+    /**
      * Module with contextual and polymorphic serializers to be used in the resulting [Json] instance.
+     *
+     * @see SerializersModule
+     * @see Contextual
+     * @see Polymorphic
      */
     public var serializersModule: SerializersModule = json.serializersModule
 
@@ -340,7 +363,8 @@ public class JsonBuilder internal constructor(json: Json) {
             encodeDefaults, ignoreUnknownKeys, isLenient,
             allowStructuredMapKeys, prettyPrint, explicitNulls, prettyPrintIndent,
             coerceInputValues, useArrayPolymorphism,
-            classDiscriminator, allowSpecialFloatingPointValues, useAlternativeNames
+            classDiscriminator, allowSpecialFloatingPointValues, useAlternativeNames,
+            namingStrategyForProperties, namingStrategyForEnums
         )
     }
 }
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
@@ -1,0 +1,120 @@
+package kotlinx.serialization.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+
+
+/**
+ * Represents naming strategy — a transformer for serial names in a [Json] format.
+ * Transformed serial names are used for both serialization and deserialization.
+ * Actual transformation happens in the [serialNameForJson] function.
+ * A naming strategy is always applied globally in the Json configuration builder
+ * (see [JsonBuilder.namingStrategyForEnums] and [JsonBuilder.namingStrategyForProperties]).
+ * However, it is possible to apply additional filtering inside the transformer using the `descriptor` parameter in [serialNameForJson].
+ *
+ * Original serial names are never used after transformation, so they are ignored in a Json input.
+ * If the original serial name is present in the Json input but transformed is not,
+ * [MissingFieldException] still would be thrown. If one wants to preserve the original serial name for deserialization,
+ * one should use the [JsonNames] annotation.
+ *
+ * ### Common pitfalls in conjunction with other Json features
+ *
+ * * Due to the nature of kotlinx.serialization framework, naming strategy transformation is applied to all properties regardless
+ * of whether their serial name was taken from the property name or provided by @[SerialName] annotation.
+ * Effectively it means one cannot avoid transformation by explicitly specifying the serial name.
+ *
+ * * Collision of the transformed name with any other (transformed) properties serial names or any alternative names
+ * specified with [JsonNames] will lead to a deserialization exception.
+ *
+ * * Naming strategies do not transform serial names of the classes used for the polymorphism, as they always should be specified explicitly.
+ * Values from [JsonClassDiscriminator] or global [JsonBuilder.classDiscriminator] also are not altered.
+ *
+ * ### Controversy about using global naming strategies
+ *
+ * Global naming strategies have one key trait that makes them a debatable and controversial topic:
+ * They are very implicit. It means that by looking only at the definition of the class,
+ * it is impossible to say which names it will have in the serialized form.
+ * As a consequence, naming strategies are not friendly to refactorings. Programmer renaming `myId` to `userId` may forget
+ * to rename `my_id`, and vice versa. Generally, any tools one can imagine work poorly with global naming strategies:
+ * Find Usages/Rename in IDE, full-text search by grep, etc. For them, the original name and the transformed are two different things;
+ * changing one without the other may introduce bugs in many unexpected ways.
+ * The lack of a single place of definition, the inability to use automated tools, and more error-prone code lead
+ * to greater maintenance efforts for code with global naming strategies.
+ * However, there are cases where usage of naming strategies is inevitable, such as interop with existing API or migrating a large codebase.
+ * Therefore, one should carefully weigh the pros and cons before considering adding global naming strategies to an application.
+ */
+@ExperimentalSerializationApi
+public fun interface JsonNamingStrategy {
+    /**
+     * Accepts an original [serialName] (defined by property name in the class or [SerialName] annotation) and returns
+     * a transformed serial name which should be used for serialization and deserialization.
+     *
+     * Besides string manipulation operations, it is also possible to implement transformations that depend on the [descriptor]
+     * and its element (defined by [elementIndex]) currently being serialized.
+     * It is guaranteed that `descriptor.getElementName(elementIndex) == serialName`.
+     * For example, one can choose different transformations depending on [SerialInfo]
+     * annotations (see [SerialDescriptor.getElementAnnotations]) or element optionality (see [SerialDescriptor.isElementOptional]).
+     *
+     * Note that some of the invocations of this function may be cached for performance reasons.
+     * Caching strategy is an implementation detail and shouldn't be assumed as a part of the public API contract, as it may be changed in future releases.
+     * Therefore, it is essential for this function to be pure: it should not have any side effects, and it should
+     * return the same String for a given [descriptor], [elementIndex], and [serialName], regardless of the number of invocations.
+     */
+    public fun serialNameForJson(descriptor: SerialDescriptor, elementIndex: Int, serialName: String): String
+
+    /**
+     * Contains basic, ready to use naming strategies.
+     */
+    @ExperimentalSerializationApi
+    public companion object Builtins {
+
+        /**
+         * A strategy that applies [String.lowercase] to the input serial name. The descriptor parameter is not used.
+         *
+         * Given that [String.lowercase] supports one-to-many and many-to-one character mapping, the new serial name's length can differ from the original's.
+         */
+        @ExperimentalSerializationApi
+        public val AllLowercase: JsonNamingStrategy = object : JsonNamingStrategy {
+            override fun serialNameForJson(descriptor: SerialDescriptor, elementIndex: Int, serialName: String): String =
+                serialName.lowercase()
+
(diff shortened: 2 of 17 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -313,8 +314,30 @@ public class JsonBuilder internal constructor(json: Json) {
...
+    public var namingStrategyForProperties: JsonNamingStrategy? = json.configuration.namingStrategyForProperties
+
+    /**
+     * Specifies [JsonNamingStrategy] that should be used for all enum instances for serialization and deserialization.
+     *
+     * `null` by default.
+     *
+     * This strategy is applied for all entities that have [SerialKind.ENUM].
+     */
+    public var namingStrategyForEnums: JsonNamingStrategy? = json.configuration.namingStrategyForEnums
```
whyoleg, line 333: just interested: why to have 2 separate properties, as it's fair easy to handle this case inside `JsonNamingStrategy` like:
```kotlin
JsonNamingStrategy { descriptor, elementIndex, serialName -> 
  if(descriptor.kind == SerialKind.ENUM) serialName
  else JsonNamingStrategy.SnakeCase.serialNameForJson(descriptor, elementIndex, serialName)
}
```
And even introduce something like (name TBD):
```kotlin
fun JsonNamingStrategy.ifEnum(enumStrategy: JsonNamingStrategy): JsonNamingStrategy
```
if this use case is very often.

sandwwraith: I think it would be easier for users to configure them without going deep into descriptors and kinds since most of the strategies simply perform string operations.

But this is a draft design, maybe it's possible to come up with something that's better

pdvrieze: My main consideration with the separate options is that there are many more SerialKinds that one may want to support a naming strategy for. Doing it this way doesn't quite scale.

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -224,18 +224,21 @@ private open class JsonTreeDecoder(
...
+            if (baseName in value.keys) return baseName
+        }
         // Slow path
-        val alternativeNamesMap =
-            json.schemaCache.getOrPut(desc, JsonAlternativeNamesKey, desc::buildAlternativeNamesMap)
-        val nameInObject = value.keys.find { alternativeNamesMap[it] == index }
-        return nameInObject ?: mainName
+        val deserializationNamesMap = json.deserializationNamesMap(descriptor)
+        val nameInObject = value.keys.find { deserializationNamesMap[it] == index }
+        val fallbackName = strategy?.serialNameForJson(descriptor, index, descriptor.getElementName(index)) // Key not found exception should be thrown with transformed named, not original
```
shanshin, line 240: Useless invoke if `nameInObject != null`, so it can be optimized `nameInObject?.let { return it }`

formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Original serial names are never used after transformation, so they are ignored in a Json input.
+ * If the original serial name is present in the Json input but transformed is not,
+ * [MissingFieldException] still would be thrown. If one wants to preserve the original serial name for deserialization,
+ * one should use the [JsonNames] annotation.
+ *
+ * ### Common pitfalls in conjunction with other Json features
+ *
+ * * Due to the nature of kotlinx.serialization framework, naming strategy transformation is applied to all properties regardless
+ * of whether their serial name was taken from the property name or provided by @[SerialName] annotation.
```
shanshin, line 23: This is probably very subjective, but it seems very unobvious to me that the value from the `SerialName` is also transformed.
In addition, the user does not have any opportunity to fix the field name rigidly (for example, some kind of naming exception due to legacy).

If  we leave it like this, I think it would be good to correct the KDoc for the `SerialName` so that it would not be a surprise to the user that the annotation does not always specify the final name of the field.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -11,46 +12,66 @@ import kotlinx.serialization.json.*
...
-                        "${getElementName(getValue(name))} in ${this@buildAlternativeNamesMap}"
+                        "${getElementName(getValue(name))} in ${this@buildDeserializationNamesMap}"
             )
         }
         this[name] = index
     }
 
-    var builder: MutableMap<String, Int>? = null
+    val builder: MutableMap<String, Int> =
+        mutableMapOf() // can be not concurrent because it is only read after creation and never written to?
```
shanshin, line 29: Yes, local builder may be not thread-safe

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: After some consideration and discussions, we've decided that enums parsing are an entirely different feature from naming strategy: while strategy changes **names**, enum parsing should affect **values**. Besides, there are some cases when strategy for enums is not enough: e.g. you want to parse both `Foo` and `foo` into `FOO`, like Jackson's CASE_INSENSETIVE_ENUMS.

Therefore, enum support would be removed entirely from this PR. I'll reopen #209 for it.

sandwwraith:
After some consideration and discussions, we've decided that enums parsing are an entirely different feature from naming strategy: while strategy changes names, enum parsing should affect values. Besides, there are some cases when strategy for enums is not enough: e.g. you want to parse both Foo and foo into FOO, like Jackson's CASE_INSENSETIVE_ENUMS.

Therefore, enum support would be removed entirely from this PR. I'll reopen https://github.com/Kotlin/kotlinx.serialization/issues/209 for it.

sandwwraith:
@qwwdfsad @shanshin It's ready for review

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -11,46 +12,71 @@ import kotlinx.serialization.json.*
...
+        strategy?.let { builder.putOrThrow(it.serialNameForJson(this, i, getElementName(i)), i) }
     }
-    return builder ?: emptyMap()
+    return builder.ifEmpty { emptyMap() }
 }
 
+internal fun Json.deserializationNamesMap(descriptor: SerialDescriptor): Map<String, Int> =
+    schemaCache.getOrPut(descriptor, JsonDeserializationNamesKey) { descriptor.buildDeserializationNamesMap(this) }
+
+internal fun SerialDescriptor.serializationNamesMap(json: Json, strategy: JsonNamingStrategy): Array<String> = json.schemaCache.getOrPut(this, JsonSerializationNamesKey) {
```
shanshin, line 46: `serializationNamesMap` returning `List`?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -11,46 +12,71 @@ import kotlinx.serialization.json.*
...
-            builder!!.putOrThrow(name, i)
+            builder.putOrThrow(name, i)
         }
+        strategy?.let { builder.putOrThrow(it.serialNameForJson(this, i, getElementName(i)), i) }
     }
-    return builder ?: emptyMap()
+    return builder.ifEmpty { emptyMap() }
 }
 
+internal fun Json.deserializationNamesMap(descriptor: SerialDescriptor): Map<String, Int> =
```
shanshin, line 43: There may be it worth in the name or comment to explicitly indicate that these are alternative names (not those that match the name of the property)?

formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
```diff
@@ -0,0 +1,107 @@
...
+         *
+         * **Note on cases:** Whether a character is in upper case is determined by the result of [Char.isUpperCase] function.
+         * Lowercase transformation is performed by [Char.lowercaseChar], not by [Char.lowercase],
+         * and therefore does not support one-to-many and many-to-one character mappings.
+         * See the documentation of these functions for details.
+         */
+        @ExperimentalSerializationApi
+        public val SnakeCase: JsonNamingStrategy = object : JsonNamingStrategy {
+            override fun serialNameForJson(descriptor: SerialDescriptor, elementIndex: Int, serialName: String): String =
+                buildString(serialName.length * 2) {
```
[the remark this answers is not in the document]
pdvrieze: In some way the `serialName` parameter is redundant. In addition, it is not clear whether this property refers to the serial name at the useSite (`serialDescriptor.getElementName(index)`) or the type (`serialDescriptor.getElementDescriptor(index).serialName`).

I would say that if the parameter is retained, the parameter should be named such to make this clear.

formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
```diff
@@ -0,0 +1,107 @@
...
+     * It is guaranteed that `descriptor.getElementName(elementIndex) == serialName`.
+     * For example, one can choose different transformations depending on [SerialInfo]
+     * annotations (see [SerialDescriptor.getElementAnnotations]) or element optionality (see [SerialDescriptor.isElementOptional]).
+     *
+     * Note that invocations of this function are cached for performance reasons.
+     * Caching strategy is an implementation detail and shouldn't be assumed as a part of the public API contract, as it may be changed in future releases.
+     * Therefore, it is essential for this function to be pure: it should not have any side effects, and it should
+     * return the same String for a given [descriptor], [elementIndex], and [serialName], regardless of the number of invocations.
+     */
+    public fun serialNameForJson(descriptor: SerialDescriptor, elementIndex: Int, serialName: String): String
```
pdvrieze, line 63: The serialName parameter could possibly be renamed to `elementSerialName` as there is also a type serial name. (The documentation here, especially the guarantee in code, make the semantics clear though. The parameter still does present a redundant parameter (as it can be retrieved trivially).

formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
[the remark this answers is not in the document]
sandwwraith: @shanshin I'm a bit reluctant to provide such corner-case handling because it will make documentation for this state machine much harder to understand. Although it may be reduced to just several examples?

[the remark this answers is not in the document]
sandwwraith: @pdvrieze A whole strategy is about properties serial names, so I think it won't be confused with classes

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
sandwwraith: Well, its a mapping from int to string, technically) Although it may be renamed to smth like `indices`

formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt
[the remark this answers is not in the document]
shanshin: Maybe it's better to describe not how the string is transformed character by character, but to describe the rules where the underscore is inserted?
eg 
- between characters when a lowercase letter is followed by a capital letter
- between two capital letters, if after the second letter there is a lowercase
- between characters when a non-letter char is followed by a capital letter

the resulting string is converted to lowercase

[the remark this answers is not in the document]
sandwwraith: I've tried to do this in new docs, check them our

shanshin (approved):

Angel-Ponce:
How can i use this feature?
You probably need to drop an example when pushed new features, in my opinion the main documentation page it's a little bit poor

sschuberth:
> How can i use this feature?

Like:

    import kotlinx.serialization.json.JsonNamingStrategy

    val json = Json {
        ignoreUnknownKeys = true
        namingStrategy = JsonNamingStrategy.SnakeCase
        serializersModule = module
    }

Angel-Ponce:
> > How can i use this feature?
> 
> Like:
> 
> ```
> import kotlinx.serialization.json.JsonNamingStrategy
> 
> val json = Json {
>     ignoreUnknownKeys = true
>     namingStrategy = JsonNamingStrategy.SnakeCase
>     serializersModule = module
> }
> ```

I don't have idea of where to use the `val json = ...` and what is `serializersModule = module` 
I want to attach this `Json` configuration to my server

`
 val app: HttpHandler = ApplicationLambda(System.getenv())
        val localLambda = app.asServer(SunHttp(env("APP_PORT").toInt())).start()
`

Angel-Ponce:
> > How can i use this feature?
> 
> Like:
> 
> ```
> import kotlinx.serialization.json.JsonNamingStrategy
> 
> val json = Json {
>     ignoreUnknownKeys = true
>     namingStrategy = JsonNamingStrategy.SnakeCase
>     serializersModule = module
> }
> ```

Thank you so much!

sandwwraith:
All the reference (including this example) is located in our guide: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#global-naming-strategy

itsivag:
why is it still marked Experimental? when it is going to be marked prod?

sandwwraith:
I expect to stabilize it soon, as there are almost no open issues about it

PR #2112, opened 2022-11-29
Improve hint when sealed class is not @Serializable

alshain:
We've had some cases where we weren't sure why serialization for sealed polymorphic subclasses errored out.

The problem was that we just missed adding `@Serializable` onto the base class.


sandwwraith:
Great idea! Since this exception message is used as an example in the guide, you need to update `docs/polymorphism.md` and re-generate knit tests (`./gradlew knit`)

sandwwraith:
Closing in favor of #2180

PR #2117, opened 2022-12-03
The idea of optimizing JsonToJavaStreamWriter

slavonnet:
The string buffer can be used to get the character code and perform the necessary calculations. Copying to a separate char buffer is an overhead and adds an additional loop pass. The string itself can act as an already formed buffer.

The number of castes, cycles and time variables has been reduced.

### DIFF
--- a/formats/json/jvmMain/src/kotlinx/serialization/json/internal/JvmJsonStreams.kt
+++ b/formats/json/jvmMain/src/kotlinx/serialization/json/internal/JvmJsonStreams.kt
@@ -17,20 +17,23 @@ internal class JsonToJavaStreamWriter(private val stream: OutputStream) : JsonWr
     }
 
     override fun write(text: String) {
-        text.forEach(::writeChar)
+        text.codePoints().forEachOrdered{
+            writeUtf8CodePoint(it)
+        }
     }
 
     override fun writeQuoted(text: String) {
         writeUtf8CodePoint('"'.code)
-        text.forEach {
-            val ch = it.code
+        text.codePoints().forEachOrdered { ch ->
             if (ch < ESCAPE_MARKERS.size) {
                 when (val marker = ESCAPE_MARKERS[ch]) {
                     0.toByte() -> {
                         writeUtf8CodePoint(ch)
                     }
                     1.toByte() -> {
-                        ESCAPE_STRINGS[ch]?.forEach(::writeChar)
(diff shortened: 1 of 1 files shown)


shanshin:
Thanks for the PR


Current `dev` branch
```
Benchmark                                   Mode  Cnt     Score    Error   Units
JacksonComparisonBenchmark.kotlinToStream  thrpt   10  1038.831 ± 15.626  ops/ms
```
changes from PR
```
Benchmark                                   Mode  Cnt    Score   Error   Units
JacksonComparisonBenchmark.kotlinToStream  thrpt   10  567.184 ± 4.038  ops/ms
```

Please look at this [commit](https://github.com/Kotlin/kotlinx.serialization/commit/333f9ffb6febfaf880c8dc73a650447543374dff). It explains the motivation for using an intermediate char array.

One of the slow parts in your change is `text.forEach`, because inside `.get(index)` is used, which works slowly for an arbitrary `String` type.

shanshin:
Anyway, we need to copy the explanatory comments from `JsonToStringWriter.kt` to `JvmJsonStreams.kt` so that when reading the code, it does not seem that the char array is nothing more than legacy unoptimal code.

slavonnet:
@shanshin thx. I will fix this.

qwwdfsad:
I suggest to stop working on this PR. This code is quite performance-sensitive, each of its parts is written with performance **and corresponding mental model, measurements and experiments**, just changing some parts because they seemed to be not as fast as they could be is likely counter-productive.

Internal documentation is indeed lacking

slavonnet:
String already has a byte buffer inside, which is allocated and written to memory. It is redundant to copy the same information to a separate charter buffer in order to go through it and apply changes, if there is an opportunity to reach the buffer in String.

In fact, now in the operation code:
- Copying to the char buffer
- Changing the char buffer and sometimes changing the size
- Passing by character by result and writing by character to the final buffer
- Resetting the final buffer to the stream

If character-by-character reading and processing are already going on in the middle, why an additional copy operation to the char buffer? I would understand if you copied with ranges what you don't need to change, but after all, the search is going character by character.

You can iterate through the buffer inside the String and immediately write to the temporary buffer. We need to get to "byte[] value". It seems like its codePoints-Int Stream receives as a source and reads it similarly to the code passing through the buffer char.

I wrote through a translator, I hope it turned out clearly

PR #2125, opened 2022-12-13
Added a factory that creates an enum serializer with annotations on the class

### LINKED ISSUE #2121: SerialInfo annotations on enum classes are not saved in 1.7.20+
**Describe the bug**

SerialInfo annotations on enum classes (not members!) were saved up to Kotlin 1.7.10, but starting from 1.7.20 they're lost.

**To Reproduce**
```
@SerialInfo
@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
annotation class A(val i: Int)

@A(1)
@Serializable
enum class Baz {
    @A(2) X, @A(3) Y
}

@Test
fun enumTest() {
    val d = Baz.serializer().descriptor
    println(d.annotations)
    println(d.getElementAnnotations(0))
    println(d.getElementAnnotations(1))
}
```

**Expected behavior**

On 1.7.10 test printed:
```
[kotlinx.serialization.A$Impl@65b5b5ed]
[kotlinx.serialization.A$Impl@795eddda]
[kotlinx.serialization.A$Impl@c6bf8d9]
```

On 1.7.20 test prints:
```
[]
[kotlinx.serialization.A$Impl@6bc85459]
[kotlinx.serialization.A$Impl@64a67067]
```

Reason: 

In 1.7.20, factory function https://github.com/Kotlin/kotlinx.serialization/blob/9658ea77b6acba798f7ea614aa27406c7fa20800/core/commonMain/src/kotlinx/serialization/internal/Enums.kt#L61 is used instead of generated `$serializer` for enums. It doesn't have slots for class-level annotations.

shanshin:
Resolves #2121

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/Enums.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/Enums.kt
@@ -56,7 +56,7 @@ internal fun <T : Enum<T>> createSimpleEnumSerializer(serialName: String, values
 }
 
 /**
- * The function is preserved for backward compatibility with previously compiled enum classes.
+ * The function has a bug (#2121) and should not be used by new (1.8.20+) plugins. It is preserved for backward compatibility with previously compiled enum classes.
  */
 @OptIn(ExperimentalSerializationApi::class)
 @InternalSerializationApi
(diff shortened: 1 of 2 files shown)


core/commonMain/src/kotlinx/serialization/internal/Enums.kt
```diff
@@ -56,6 +55,9 @@ internal fun <T : Enum<T>> createSimpleEnumSerializer(serialName: String, values
     return EnumSerializer(serialName, values)
 }
 
+/**
+ * The function is preserved for backward compatibility with previously compiled enum classes.
```
sandwwraith, line 59:
```suggestion
 * The function has a bug (#2121) and should not be used by new (1.8.20+) plugins. It is preserved for backward compatibility with previously compiled enum classes.
```

sandwwraith (approved):

PR #2136, opened 2022-12-22
Add stable hashCode() calculation to PrimitiveSerialDescriptor

### LINKED ISSUE #2135: Make PrimitiveSerialDescriptor.hashCode() stable
**Describe the bug**
`PrimitiveSerialDescriptor` class doesn't have custom hashCode() implemented, so, it can change between different JVMs vendors, versions, servers, between different app run etc.
Plus, we can change hashCode generation strategy (`-XX:hashCode` JVM option) – this will change hashCode value too.

I think it is good idea to add stable hashCode() implementation to `PrimitiveSerialDescriptor`

vgv:
Fix #2135

### DIFF
--- a/core/commonTest/src/kotlinx/serialization/internal/PrimitiveSerialDescriptorTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/internal/PrimitiveSerialDescriptorTest.kt
@@ -1,0 +1,28 @@
+package kotlinx.serialization.internal
+
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotSame
+
+class PrimitiveSerialDescriptorTest {
+
+    @Test
+    fun testEqualsImplemented() {
+        val first = PrimitiveSerialDescriptor("test_name", PrimitiveKind.LONG)
+        val second = PrimitiveSerialDescriptor("test_name", PrimitiveKind.LONG)
+
+        assertNotSame(first, second)
+        assertEquals(first, second)
+    }
+
+    @Test
+    fun testHashCodeStability() {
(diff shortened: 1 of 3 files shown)


sandwwraith:
You're right; all descriptors should have meaningful equals. Can you please also add a test?

vgv:
I've added tests

core/commonTest/src/kotlinx/serialization/internal/PrimitiveSerialDescriptorTest.kt
```diff
@@ -0,0 +1,28 @@
+package kotlinx.serialization.internal
```
sandwwraith, line 1: We don't usually put tests in the `internal` package, as it is not descriptive. There are already `SerialDescriptorBuilderTest` and `SerialDescriptorEqualityTest` that look like a proper fit for these test functions. Alternatively, you can just put `PrimitiveSerialDescriptorTest` near them.

sandwwraith (changes requested):
Thank you! Just move tests out of internal package, please

core/commonTest/src/kotlinx/serialization/internal/PrimitiveSerialDescriptorTest.kt
vgv: I've checked `SerialDescriptorBuilderTest` and `SerialDescriptorEqualityTest` 
For me, it seems that separate `PrimitiveSerialDescriptorTest` is better than adding test into one of these classes, so I just moved `PrimitiveSerialDescriptorTest` out of `internal` package.

sandwwraith (approved):

sandwwraith:
This should have been opened against `dev`, but I haven't noticed :( Cherry-picked from master to dev manually.

qwwdfsad:
It's better to merge master into dev instead to avoid duplicate commits while still being in the same repository state eventually

Issue #887, opened 2020-07-05
Provide standard a (de-)serializer for Kotlin's Regex class


sschuberth:
**What is your use-case and why do you need this feature?**

I'd like to (de-)serialize `kotlin.text.Regex` as part of a `data class`.

**Describe the solution you'd like**

With `kotlin.text.Regex` being a Kotlin class, it would be nice if (de-)serialization worked out of the box.

ypawar159:
I would also like this implemented.
It's better to compile the regex at deserilaization and use that object multiple times than compiling the regex each time required.

I have tried the following but didn't work. Learning kotlin coming from python so don't know what is wrong here.

```kotlin
object RegexSerializer : KSerializer<Regex> {
    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Regex", PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: Regex) {
        encoder.encodeString(value.pattern)
    }

    override fun deserialize(decoder: Decoder): Regex {
        return Regex(decoder.decodeString())
    }
}
```

sschuberth:
I was using the same code [here](https://github.com/sschuberth/stan/blob/cb0ddd1656e58382371be3f41aa00bbe6ee74938/lib/src/main/kotlin/utils/RegexSerializer.kt), and it actually worked for me.

ypawar159:
Thanks for the reference code.
I haven't got my code working yet.
Have you used it for nested property?

sschuberth:
Not sure how "nested" it should be, but I've used it for files like [this one](https://github.com/sschuberth/stan/blob/cb0ddd1656e58382371be3f41aa00bbe6ee74938/lib/src/main/resources/config.json#L15).

sandwwraith:
@ypawar159 Have you applied your serializer to `Regex` property as in the guide: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#specifying-serializer-on-a-property ?

ypawar159:
I was using it for a complete class with reference from other part of the documentation.
Will try with this approach as well and post results here.

ypawar159:
@sandwwraith Annotating the specific field in the data class worked successfully. Thanks for the reference.
Then I tried to do the same for the `List<Regex>` field but it fails to compile. Anything specific to be done for the list of it?

sandwwraith:
[...]

PR #2137, opened 2022-12-23
Added support for the `kotlin.text.Regex` class as built-in

alexmihailov:
Issue #887
 [PR](https://github.com/JetBrains/kotlin/pull/5060) for compiler plugin

### DIFF
--- a/core/api/kotlinx-serialization-core.api
+++ b/core/api/kotlinx-serialization-core.api
@@ -189,6 +189,7 @@ public final class kotlinx/serialization/builtins/BuiltinSerializersKt {
 	public static final fun serializer (Lkotlin/jvm/internal/LongCompanionObject;)Lkotlinx/serialization/KSerializer;
 	public static final fun serializer (Lkotlin/jvm/internal/ShortCompanionObject;)Lkotlinx/serialization/KSerializer;
 	public static final fun serializer (Lkotlin/jvm/internal/StringCompanionObject;)Lkotlinx/serialization/KSerializer;
+	public static final fun serializer (Lkotlin/text/Regex$Companion;)Lkotlinx/serialization/KSerializer;
 	public static final fun serializer (Lkotlin/time/Duration$Companion;)Lkotlinx/serialization/KSerializer;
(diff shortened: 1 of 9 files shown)


sandwwraith:
I'll repeat here the question by @mcpiroman from compiler PR: "Shouldn't options also be serialized? I don't think they are included in the pattern string."

alexmihailov:
I really forgot about `RegexOption`. In this case, I propose to select two fields for serialization: `pattern` and `options`. Using the composite serializer:
```
{pattern:"^(.+)@(\\S+)$", options:[]}
```
I will propose to make the options field optional:
```
{pattern:"^(.+)@(\\S+)$"}
```
@sandwwraith Leonid, what do you think ?

sandwwraith:
I've studied several different usages of `KSerializer<Regex>` as well as Jackson's approach to this. I've found that while most people serialize regex as string, some indeed need `options`, which may cause problems when parties do not agree on the representation (see, for example, https://github.com/FasterXML/jackson-module-kotlin/issues/210 and https://github.com/FasterXML/jackson-module-kotlin/issues/269 — the former issue is about parsing `{pattern, options}` object while the latter is about parsing `patter` string which was broken by fixing the former.....). The same problem exists for serializing, too: https://github.com/FasterXML/jackson-databind/issues/1959. 

Please tell us what use case for serializing regexes you have and what format works better for it. It may add something to the perspective. However, for now, I think that we shouldn't add any Regex serializer as a standard one — given that there's apparently no standard, universally accepted, and unambiguous representation of it. It's also not a very popular feature.

alexmihailov:
In my case I use Regex without options in Hocon.  I'm using the [config4k](https://github.com/config4k/config4k) library that support Regex from pattern string, but I want to move to kotlinx-serialization and have a built-in serializer for Regex.

sandwwraith:
Given that your usage is Hocon-only, and there's a controversy about representing Regex in Json, I think we won't provide a built-in Regex serializer. It's very easy to provide a custom one where it's needed.

PR #2094, opened 2022-11-09
Support encode/decode java.time.Duration and ConfigMemorySize in HOCON

alexmihailov:
- Many use cases require serialization support for `java.time.Duration`. It is not possible to implement a custom serializer in user code, since private methods from `kotlinx.serialization.hocon.Hocon` and `kotlinx.serialization.hocon.AbstractHoconEncoder` are required. To do this, I implemented the `JDurationSerializer` serializer, which can be used in user code.

- Hocon supports the memory size format. [Documentation](https://github.com/lightbend/config/blob/main/HOCON.md#size-in-bytes-format).
In the `com.typesafe.config` library, the `ConfigMemorySize` class is responsible for this functionality. There is no way in user code to implement a serializer for this class. So I've implemented a `ConfigMemorySizeSerializer` that allows to serialize `ConfigMemorySize`.

- The `com.typesafe.config` library can create configuration using Java Bean notation. The `com.typesafe.config.ConfigBeanFactory.create` method is used for this. To support this functionality, I've implemented a `JBeanSerializer` that can only be used to decode configuration into a Java Bean class.

This PR depends on the previous [PR](https://github.com/Kotlin/kotlinx.serialization/pull/2080)

PR #2126, opened 2022-12-16
Support encode/decode java.time.Period in HOCON

alexmihailov:
Hocon supports the period format. [Documentation](https://github.com/lightbend/config/blob/main/HOCON.md#period-format).

Depends on: #2094

PR #2139, opened 2022-12-26
Support encode/decode nested com.typesafe.config.Config in HOCON

alexmihailov:
Depends on: #2094


sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/pull/2126#issuecomment-1442173168

PR #2140, opened 2022-12-26
Support encode/decode com.typesafe.config.ConfigValue in HOCON

alexmihailov:
Depends on: #2094, #2139


sandwwraith:
https://github.com/Kotlin/kotlinx.serialization/pull/2126#issuecomment-1442173168

PR #2146, opened 2023-01-05
Update to Kotlin 1.8.0

sandwwraith:


qwwdfsad (changes requested):
README should also be updated

sandwwraith:
I don't think so, as README is usually updated on release and indicates a dependency one is recommended to use with the latest released version

Moreover, IIRC, we do not update READMEs on RC releases?

qwwdfsad:
Then it's LGTM

Issue #1581, opened 2021-07-02
Hcon: error while trying to decode MAP + sealed class


SchweinchenFuntik:
additions to  #1046

```kotlin
package khocon

import com.typesafe.config.ConfigFactory
import kotlinx.serialization.ExperimentalSerializationApi
import kotlinx.serialization.Serializable
import kotlinx.serialization.hocon.Hocon
import kotlinx.serialization.hocon.decodeFromConfig

@Serializable
sealed class Node

@Serializable
data class Root(val name: String): Node()

@Serializable
data class Simple(val a: String)


private val configSimpleClass = """
    key1 {
       a = "simple"
    }
    
    key2 {
       a = "simple"
    }
    """

private val configSealedClass = """
    key1 {
       type = "khocon.Root"
       name = "seale"
    }
    
    key2 {
       type = "khocon.Root"
       name = "sealed"
    }
    """

private val configSealedClassSimple = """
    {
   type = "khocon.Root"
   name = "sealed"
   }
    """

@OptIn(ExperimentalSerializationApi::class)
fun main() {
    val simpleClassConfig = ConfigFactory.parseString(configSimpleClass)
    val sealedClassConfigSimple = ConfigFactory.parseString(configSealedClassSimple)
    val sealedClassConfig = ConfigFactory.parseString(configSealedClass)
    val hocon = Hocon {
      // useArrayPolymorphism = true // error val sealedSimple = hocon.decodeFromConfig<Node>(sealedClassConfigSimple)
    }
    val simple = hocon.decodeFromConfig<Map<String, Simple>>(simpleClassConfig) // Work
    println(simple)
    val sealedSimple = hocon.decodeFromConfig<Node>(sealedClassConfigSimple)
    println(sealedSimple)
    val sealed = hocon.decodeFromConfig<Map<String, Node>>(sealedClassConfig)
    println(sealed)
}
```

**output**
```
Exception in thread "main" java.lang.ClassCastException: class com.typesafe.config.impl.SimpleConfigObject cannot be cast to class com.typesafe.config.ConfigList (com.typesafe.config.impl.SimpleConfigObject and com.typesafe.config.ConfigList are in unnamed module of loader 'app')
	at kotlinx.serialization.hocon.Hocon$MapConfigReader.beginStructure(Hocon.kt:200)
	at kotlinx.serialization.internal.AbstractPolymorphicSerializer.deserialize(AbstractPolymorphicSerializer.kt:130)
[...]

PR #2149, opened 2023-01-10
fix: Hocon polymorphic serialization

LichtHund:
Hello!

I know this PR is not exactly what you guys would do to fix the issue with polymorph on HOCON, but I made a quick patch to make it work for us.

The problem is that currently Polymorph kinds are treated as Lists and tries to cast it which won't work since Config will keep it as an object.
Another issue is that the `getTag` will add a `value.tag` as the parent when trying to get the correct tag from the config (hence why the ugly and not great boolean "poly").

That said I would love this issue fixed, so please give me suggestions on how to properly fix it and I'd be more than happy to modify the PR to be up to standards!

This fix refers to #1581.


LichtHund:
I realized it needs to be dev branch, I'll re-do it.

Issue #932, opened 2020-07-24
Serializer for Nothing


sandwwraith:
A `KSerializer<Nothing?>` should serve the same purposes that `Nothing` in type system: indicate that a particular item should never be serialized. There are several use-cases for it, mainly connected with generics:
https://github.com/Kotlin/kotlinx.serialization/issues/607
https://github.com/Kotlin/kotlinx.serialization/issues/393
https://github.com/Kotlin/kotlinx.serialization/issues/614

**Describe the solution you'd like**
A `NothingSerializer` that can be used in `@Serializable(NothingSerializer)` on types. Or `NoSerializer` serializer. Or `@NoSerializer` annotation as a shorthand.

andylamax:
Can this issue at least get some love?

qwwdfsad:
It is fixed already, the only question is when it will be released. @shanshin knows for sure if it's 1.7.20 or 1.8.0

shanshin:
Target Kotlin release `1.8.0`

ShreckYe:
Is this issue fixed for the JS target?

By running the code below on the JS target

```
class TestData<T>(val value: T)
        serializer<TestData<Nothing?>>()
        Json.encodeToString(TestData(null))
```

I get:

> UnsupportedOperationException: There's no native JS class for Nothing type

sandwwraith:
I see you even filed an issue about that: https://youtrack.jetbrains.com/issue/KT-51333/KJS-some-KType-equals-Nothings-KType-throws-an-exception-breaking-its-symmetry
I think it should be fixed on Kotlin's side, because we have proper BUILTINS_SERIALIZERS map logic.

PR #2150, opened 2023-01-10
Added SerialDescriptor object for NothingSerializer

shanshin:
Improvement of #932

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/NothingSerialDescriptor.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/NothingSerialDescriptor.kt
@@ -12,7 +12,7 @@ import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.descriptors.SerialKind
 import kotlinx.serialization.descriptors.StructureKind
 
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/internal/NothingSerialDescriptor.kt
```diff
@@ -0,0 +1,34 @@
...
+@file:OptIn(ExperimentalSerializationApi::class)
+
+package kotlinx.serialization.internal
+
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.SerialKind
+import kotlinx.serialization.descriptors.StructureKind
+
+public object NothingSerialDescriptor : SerialDescriptor {
```
qwwdfsad, line 15: Serialization doesn't expose its internal types.

Instead, the type should be internal and the only new public signature should be `fun NothingSerialDescriptor(): SerialDescriptor`

qwwdfsad (changes requested):

sandwwraith (approved):

PR #2151, opened 2023-01-10
fix: Hocon polymorphic serialization

LichtHund:
Hello!

I know this PR is not exactly what you guys would do to fix the issue with polymorph on HOCON, but I made a quick patch to make it work for us.

The problem is that currently Polymorph kinds are treated as Lists and tries to cast it which won't work since Config will keep it as an object.
Another issue is that the getTag will add a value.tag as the parent when trying to get the correct tag from the config (hence why the ugly and not great boolean "isPolymorph").

That said I would love this issue fixed, so please give me suggestions on how to properly fix it and I'd be more than happy to modify the PR to be up to standards!

This fix refers to #1581.

### DIFF
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -145,7 +145,7 @@ public sealed class Hocon(
 
     }
 
-    private inner class ConfigReader(val conf: Config, private val poly: Boolean = false) : ConfigConverter<String>() {
+    private inner class ConfigReader(val conf: Config, private val isPolymorph: Boolean = false) : ConfigConverter<String>() {
         private var ind = -1
 
         override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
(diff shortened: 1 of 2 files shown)


LichtHund:
I ended up forgetting about this PR. 😅 
Added a test with a collection and multiple entries of the same sealed type. The test fails on the current version but passes on this fix. We have been using this fix for well over a year now and has been working pretty well!
If there is any improvements though let me know and I'll change it asap!

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
```diff
@@ -161,8 +161,12 @@ public sealed class Hocon(
         private fun composeName(parentName: String, childName: String) =
             if (parentName.isEmpty()) childName else "$parentName.$childName"
 
-        override fun SerialDescriptor.getTag(index: Int): String =
-            composeName(currentTagOrNull.orEmpty(), getConventionElementName(index, useConfigNamingConvention))
+        override fun SerialDescriptor.getTag(index: Int): String {
+            return if (!isPolymorph) composeName(
+                currentTagOrNull.orEmpty(),
+                getConventionElementName(index, useConfigNamingConvention)
+            ) else getElementName(index)
```
sandwwraith, line 168: Shouldn't we also apply `getConventionElementName` here?

sandwwraith (changes requested):
Hi, thanks for the fix! I do not have objections besides minor comments that should be easy to fix.

formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
LichtHund: Didn't think of that, yeah it definitely should!

sandwwraith (approved):
Thank you!

PR #2156, opened 2023-01-14
Implement addAll functions for JsonArrayBuilder

### LINKED ISSUE #2127: `JsonArrayBuilder` - add convenience function for adding multiple elements

When building a JSON array I'd like to add multiple strings.

Currently this must be done one at a time, using [`JsonArrayBuilder.add(value: String?)`](
https://github.com/Kotlin/kotlinx.serialization/blob/67e92598d548331f7f5c55d0f2b80cb9deaa5bd7/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt#L157-L162)

```kotlin
val alphaDependencies = listOf("a", "b", "c")

buildJsonObject {
  put("name", "alpha")
  putJsonArray("dependencies") {

    // currently have to manually define a for-loop
    alphaDependencies.forEach { dep -> 
      add(dep)
    }
  }
}.toString()
```

It's a surprise that `buildJsonArray()` doesn't have `addAll()`, because [`buildList()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-list.html) does have `addAll()`.

```kotlin
val x = listOf('b', 'c')

val y = buildList() {
    addAll(x)
}
```

It would be convenient to have an `addAll(elements: Collection<String>)` helper function that would accept a list

I'm proposing four new functions. 

1 function in [`JsonArrayBuilder`](https://github.com/Kotlin/kotlinx.serialization/blob/67e92598d548331f7f5c55d0f2b80cb9deaa5bd7/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt#L125) for adding multiple `JsonElement`s.

```kotlin
public class JsonArrayBuilder @PublishedApi internal constructor() {
    
    /**
     * Adds the given JSON [elements] to a resulting JSON array.
     *
     * @return `true` if the list was changed as the result of the operation.
     */
    public fun addAll(elements: Collection<JsonElement>): Boolean {
      return content.addAll(elements)
    }
}
```

And 3 extension functions, for adding strings, numbers, booleans:

```kotlin
/** Adds the given string [values] to a resulting JSON array. */
public fun JsonArrayBuilder.addAll(values: Collection<String?>): Boolean =
  addAll(values.map(::JsonPrimitive))

/** Adds the given boolean [values] to a resulting JSON array. */
public fun JsonArrayBuilder.addAll(values: Collection<Boolean?>): Boolean =
  addAll(values.map(::JsonPrimitive))

/** Adds the given numeric [values] to a resulting JSON array. */
public fun JsonArrayBuilder.addAll(values: Collection<Number?>): Boolean =
  addAll(values.map(::JsonPrimitive))
```

Example usage

```kotlin
buildJsonObject {
  put("name", "alpha")
  putJsonArray("dependencies") {
    // no more need for a for loop
    addAll(alphaDependencies)
  }
}.toString()
```

aSemy:
Fix #2127

* new `addAll()` method on `JsonArrayBuilder`
* new extension functions for adding booleans, numbers, strings - either as a list, or using `vararg`

Should the new functions be marked with `@ExperimentalSerializationApi`?

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
@@ -6,7 +6,9 @@
 package kotlinx.serialization.json
 
 import kotlinx.serialization.ExperimentalSerializationApi
-import kotlin.contracts.*
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
 import kotlin.jvm.JvmName
 
 /**
@@ -142,6 +144,7 @@ public class JsonArrayBuilder @PublishedApi internal constructor() {
      *
      * @return `true` if the list was changed as the result of the operation.
      */
+    @ExperimentalSerializationApi
     public fun addAll(elements: Collection<JsonElement>): Boolean =
         content.addAll(elements)
 
@@ -196,38 +199,45 @@ public fun JsonArrayBuilder.addJsonArray(builderAction: JsonArrayBuilder.() -> U
     add(buildJsonArray(builderAction))
 
 /**
- * Adds the given JSON [element] and [elements] to a resulting JSON array.
+ * Adds the given JSON [elements] to a resulting JSON array.
  *
  * @return `true` if the list was changed as the result of the operation.
  */
-public fun JsonArrayBuilder.addAll(element: JsonElement, vararg elements: JsonElement): Boolean {
-    return addAll(listOf(element) + elements)
+@ExperimentalSerializationApi
+public fun JsonArrayBuilder.addAll(vararg elements: JsonElement): Boolean {
+    return addAll(elements.toList())
 }
 
 /** Adds the given string [values] to a resulting JSON array. */
 @JvmName("addAllStrings")
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(values: Collection<String?>): Boolean =
     addAll(values.map(::JsonPrimitive))
 
 /** Adds the given boolean [values] to a resulting JSON array. */
 @JvmName("addAllBooleans")
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(values: Collection<Boolean?>): Boolean =
     addAll(values.map(::JsonPrimitive))
 
 /** Adds the given numeric [values] to a resulting JSON array. */
 @JvmName("addAllNumbers")
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(values: Collection<Number?>): Boolean =
     addAll(values.map(::JsonPrimitive))
 
 /** Adds the given string [values] to a resulting JSON array. */
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(vararg values: String?): Boolean =
     addAll(values.toList())
 
 /** Adds the given boolean [values] to a resulting JSON array. */
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(vararg values: Boolean?): Boolean =
     addAll(values.toList())
 
 /** Adds the given numeric [values] to a resulting JSON array. */
+@ExperimentalSerializationApi
 public fun JsonArrayBuilder.addAll(vararg values: Number?): Boolean =
     addAll(values.toList())
 
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
@@ -142,9 +142,8 @@ public class JsonArrayBuilder @PublishedApi internal constructor() {
      *
      * @return `true` if the list was changed as the result of the operation.
      */
-    public fun addAll(elements: Collection<JsonElement>): Boolean {
-        return content.addAll(elements)
-    }
+    public fun addAll(elements: Collection<JsonElement>): Boolean =
+        content.addAll(elements)
 
     @PublishedApi
     internal fun build(): JsonArray = JsonArray(content)
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
@@ -226,20 +226,5 @@ public fun JsonArrayBuilder.addAll(values: Collection<Boolean?>): Boolean =
 public fun JsonArrayBuilder.addAll(values: Collection<Number?>): Boolean =
     addAll(values.map(::JsonPrimitive))
 
-/** Adds the given string [values] to a resulting JSON array. */
-@ExperimentalSerializationApi
-public fun JsonArrayBuilder.addAll(vararg values: String?): Boolean =
-    addAll(values.toList())
-
-/** Adds the given boolean [values] to a resulting JSON array. */
-@ExperimentalSerializationApi
-public fun JsonArrayBuilder.addAll(vararg values: Boolean?): Boolean =
-    addAll(values.toList())
-
-/** Adds the given numeric [values] to a resulting JSON array. */
-@ExperimentalSerializationApi
-public fun JsonArrayBuilder.addAll(vararg values: Number?): Boolean =
-    addAll(values.toList())
-
 @DslMarker
 internal annotation class JsonDslMarker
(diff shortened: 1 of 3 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -186,6 +195,41 @@ public fun JsonArrayBuilder.addJsonObject(builderAction: JsonObjectBuilder.() ->
...
+public fun JsonArrayBuilder.addAll(vararg values: String?): Boolean =
+    addAll(values.toList())
+
+/** Adds the given boolean [values] to a resulting JSON array. */
+public fun JsonArrayBuilder.addAll(vararg values: Boolean?): Boolean =
+    addAll(values.toList())
+
+/** Adds the given numeric [values] to a resulting JSON array. */
+public fun JsonArrayBuilder.addAll(vararg values: Number?): Boolean =
+    addAll(values.toList())
```
aSemy, line 233: I considered adding a similar function to `add(null)`

https://github.com/Kotlin/kotlinx.serialization/blob/33104957873448189eb242413646b58883623e0a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt#L117-L119

Something like 

```kotlin
@Suppress("UNUSED_PARAMETER") // allows to call `addAll(null)`
public fun JsonArrayBuilder.addAll(values: Collection<Nothing?>): Boolean = addAll(List(values.size) {JsonNull})
```

But I decided against it. It doesn't seem that useful. How often are there lists of type `Nothing`?

Interesting I discovered that Kotlin doesn't allow `vararg values: Nothing?`

image

so that's not an option.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -186,6 +195,41 @@ public fun JsonArrayBuilder.addJsonObject(builderAction: JsonObjectBuilder.() ->
...
     add(buildJsonArray(builderAction))
 
+/**
+ * Adds the given JSON [element] and [elements] to a resulting JSON array.
+ *
+ * @return `true` if the list was changed as the result of the operation.
+ */
+public fun JsonArrayBuilder.addAll(element: JsonElement, vararg elements: JsonElement): Boolean {
+    return addAll(listOf(element) + elements)
+}
```
Goooler, line 205:
```suggestion
public fun JsonArrayBuilder.addAll(vararg elements: JsonElement): Boolean {
    return addAll(elements.asList())
}
```

Goooler:
1. Can we support more `Iterable`, `Sequence`, `Array` as the `elements`' types like [MutableCollection.addAll](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-collection/#extension-functions).
2. Should we add `putAll` for `JsonObjectBuilder`.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -136,6 +137,14 @@ public class JsonArrayBuilder @PublishedApi internal constructor() {
         return true
     }
 
+    /**
+     * Adds the given JSON [elements] to a resulting JSON array.
+     *
+     * @return `true` if the list was changed as the result of the operation.
+     */
+    public fun addAll(elements: Collection<JsonElement>): Boolean =
```
sandwwraith, line 145: It's good this function's signature matches one from `MutableList` — we can add superclass later without breaking anything 👍🏻

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -186,6 +195,41 @@ public fun JsonArrayBuilder.addJsonObject(builderAction: JsonObjectBuilder.() ->
 public fun JsonArrayBuilder.addJsonArray(builderAction: JsonArrayBuilder.() -> Unit): Boolean =
     add(buildJsonArray(builderAction))
 
+/**
+ * Adds the given JSON [element] and [elements] to a resulting JSON array.
+ *
+ * @return `true` if the list was changed as the result of the operation.
+ */
+public fun JsonArrayBuilder.addAll(element: JsonElement, vararg elements: JsonElement): Boolean {
```
sandwwraith, line 203: Why did you use `(element, elements)` structure? I do not see any possible overload resolution conflicts.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -186,6 +195,41 @@ public fun JsonArrayBuilder.addJsonObject(builderAction: JsonObjectBuilder.() ->
...
+public fun JsonArrayBuilder.addAll(values: Collection<Boolean?>): Boolean =
+    addAll(values.map(::JsonPrimitive))
+
+/** Adds the given numeric [values] to a resulting JSON array. */
+@JvmName("addAllNumbers")
+public fun JsonArrayBuilder.addAll(values: Collection<Number?>): Boolean =
+    addAll(values.map(::JsonPrimitive))
+
+/** Adds the given string [values] to a resulting JSON array. */
+public fun JsonArrayBuilder.addAll(vararg values: String?): Boolean =
```
sandwwraith, line 223: I'm not so sure about adding these functions, as stdlib's `buildList/MutableList` do not have them either. It's nice addition to DSL though, but they can be easily replaced with `addAll(listOf(...))` (and upcoming collections literals in future)

sandwwraith (changes requested):
Hi, thanks for your PR and sorry it took so long to process it. I think we can accept this functionality after some refining. As you correctly said, all new functions should be marked with `@ExperimentalSerializationApi`. Can you also please rebase your branch on the latest `dev`?

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
aSemy: I see what you mean, but I think they make using the JSON Builder DSL noticeably easier and less busy.

I suspect that these functions would be made redundant when (if?) Kotlin gets type-unions. Then only one function would be required...

```kotlin
public fun JsonArrayBuilder.addAll(vararg values: String? | Boolean? | Number?): Boolean 
```

And it would allow for adding a mix of values, e.g. `addAll(1, "a", false)`

But yes, collection literals would also help.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
aSemy: 404 reason not found. I've removed it.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
sandwwraith: I understand your reasoning, but we still would like to have them removed, as other places (like JsonArray constructor or stdlib) do not feature them at all, and it doesn't make really big difference.

aSemy: Agreed, it's not a big difference. I've removed the varargs util functions.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
```diff
@@ -186,6 +198,33 @@ public fun JsonArrayBuilder.addJsonObject(builderAction: JsonObjectBuilder.() ->
 public fun JsonArrayBuilder.addJsonArray(builderAction: JsonArrayBuilder.() -> Unit): Boolean =
     add(buildJsonArray(builderAction))
 
+/**
+ * Adds the given JSON [elements] to a resulting JSON array.
+ *
+ * @return `true` if the list was changed as the result of the operation.
+ */
+@ExperimentalSerializationApi
+public fun JsonArrayBuilder.addAll(vararg elements: JsonElement): Boolean {
```
sandwwraith, line 207: Same for `vararg` here

aSemy: Oops, I missed that one. It's now removed.

sandwwraith: Also `apiDump`

sandwwraith (changes requested):

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
aSemy: hey @sandwwraith, sorry I forgot this! Would you mind updating it? You should have permissions.

sandwwraith: Ok, no problems

PR #2172, opened 2023-01-27
KDoc: Transient properties must have default values

aSemy:
The [`@Transient` KDoc](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-transient/) indicates that the transient properties _should_ have default values, 

https://github.com/Kotlin/kotlinx.serialization/blob/0b01b537b05121e5c72696b415268f3a340ae9a7/core/commonMain/src/kotlinx/serialization/Annotations.kt#L165

but in practice I get a compilation error when a default value is missing.

```kotlin
@Serializable
data class ContainsJsonString constructor(val name: String, @Transient val jsonString: String)
```

image

This PR changes 'should' to 'must', which better describes the behaviour.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/core/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -162,7 +162,7 @@ public annotation class Required
 
 /**
  * Marks this property invisible for the whole serialization process, including [serial descriptors][SerialDescriptor].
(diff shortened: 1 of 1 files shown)


sandwwraith:
Makes sense, thank you!

sandwwraith (approved):

PR #2176, opened 2023-01-27
Added support for null values for nullable enums in lanient mode

### LINKED ISSUE #2170: Unexpected MissingFieldException when decoding explicit null value from json for nullable enum
**Describe the bug**
When a property is set to `null` in the json and the model property is nullable, parsing the property to null should always work.   
No coercing or leniency is needed since it directly maps accurately to the model.

However, setting `isLenient` OR `coerceInputValues` to `true` causes an unexpected `MissingFieldException` even though the field is clearly and accurately defined in the json.

**To Reproduce**

This test will unexpectedly fail with a thrown `MissingFieldException`:

```kotlin

import org.junit.Test

class SerializationTest {

    @Test
    fun `model with nullable enum and explicit json null value does not throw MissingFieldException`() {
        // Arrange: Json with a valid null value
        val jsonWithExplicitNullValue = """
            {
                "color": null
            }
        """.trimIndent()

        // Arrange: Serializer that is as forgiving as possible
        val jsonSerializerLenient = Json {
            isLenient = true // NOTE: Test will pass if this is false
            coerceInputValues = true // NOTE: OR test will pass if `isLenient` is true and this is false
            ignoreUnknownKeys = true
            prettyPrint = true
            allowStructuredMapKeys = true
        }

        // Act: Convert json to model
        val model = jsonSerializerLenient.decodeFromString<Model>(jsonWithExplicitNullValue)

        // Assert: Model converted and value is null like in the json
        assert(model.color == null)
    }

    @Serializable
    data class Model(
        val color: Color?
    )

    enum class Color {
        Red,
        Blue,
        Green
    }
}
```

**Expected behavior**

**Environment**
 - Kotlin version: 1.8.0
 - Library version: 1.5.0-RC
 - Attached is a full reproducible project: [missing-field-exception-reproduce-project.zip](https://github.com/Kotlin/kotlinx.serialization/files/10513057/missing-field-exception-reproduce-project.zip)

shanshin:
Fixed #2170

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -99,11 +99,16 @@ internal fun SerialDescriptor.getJsonNameIndexOrThrow(json: Json, name: String, 
 internal inline fun Json.tryCoerceValue(
     elementDescriptor: SerialDescriptor,
     peekNull: () -> Boolean,
+    tryPeekNull: () -> Boolean,
     peekString: () -> String?,
     onEnumCoercing: () -> Unit = {}
 ): Boolean {
     if (!elementDescriptor.isNullable && peekNull()) return true
     if (elementDescriptor.kind == SerialKind.ENUM) {
+        if (elementDescriptor.isNullable && tryPeekNull()) {
+            return false
+        }
+
         val enumValue = peekString()
             ?: return false // if value is not a string, decodeEnum() will throw correct exception
         val enumIndex = elementDescriptor.getJsonNameIndex(this, enumValue)
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -98,14 +98,13 @@ internal fun SerialDescriptor.getJsonNameIndexOrThrow(json: Json, name: String, 
 @OptIn(ExperimentalSerializationApi::class)
 internal inline fun Json.tryCoerceValue(
     elementDescriptor: SerialDescriptor,
-    peekNull: () -> Boolean,
-    tryPeekNull: () -> Boolean,
+    peekNull: (consume: Boolean) -> Boolean,
     peekString: () -> String?,
     onEnumCoercing: () -> Unit = {}
 ): Boolean {
-    if (!elementDescriptor.isNullable && peekNull()) return true
+    if (!elementDescriptor.isNullable && peekNull(true)) return true
     if (elementDescriptor.kind == SerialKind.ENUM) {
-        if (elementDescriptor.isNullable && tryPeekNull()) {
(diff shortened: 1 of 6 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -246,8 +246,11 @@ internal abstract class AbstractJsonLexer {
      * Tries to consume `null` token from input.
      * Returns `true` if the next 4 chars in input are not `null`,
      * `false` otherwise and consumes it.
+     *
+     * If [moveForward] is true and if the next 4 chars in input are `null` then
+     * current position in lexer move forward on these 4 chars.
      */
-    fun tryConsumeNotNull(): Boolean {
+    fun tryConsumeNotNull(moveForward: Boolean = true): Boolean {
```
sandwwraith, line 253: Maybe rename `moveForward` to `consume` to align with other naming? 

Although it seems that function itself is not named properly: it says 'consumeNotNull,' but it consumes only nulls

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -99,11 +99,16 @@ internal fun SerialDescriptor.getJsonNameIndexOrThrow(json: Json, name: String,
 internal inline fun Json.tryCoerceValue(
     elementDescriptor: SerialDescriptor,
     peekNull: () -> Boolean,
+    tryPeekNull: () -> Boolean,
```
sandwwraith, line 102: I'd merge these two lambdas to lambda with boolean param because they have the same implementation for tree decoders and differ only in boolean parameter for string decoder. You can add parameter name for lambda to readability: `peekNull: (consume: Boolean) -> Boolean`

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -245,24 +245,27 @@ internal abstract class AbstractJsonLexer {
     /**
      * Tries to consume `null` token from input.
      * Returns `true` if the next 4 chars in input are not `null`,
```
sandwwraith, line 247:
```suggestion
     * Returns `false` if the next 4 chars in input are not `null`,
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -245,24 +245,27 @@ internal abstract class AbstractJsonLexer {
...
-     * `false` otherwise and consumes it.
+     * `false` otherwise and consumes it if [doConsume] is `true`.
```
sandwwraith, line 248:
```suggestion
     * `true` otherwise and consumes it if [doConsume] is `true`.
```

sandwwraith (approved):

Issue #834, opened 2020-05-08
Serialize class that hold item with generic class fails with JsonParametricSerializer


Slavvkko:
**Describe the bug**
I have data class that holds an item generic class, deserialization works well with `JsonParametricSerializer`, but serialization fails with this error:
`kotlinx.serialization.SerializationException: class Item (Kotlin reflection is not available) is not registered for polymorphic serialization in the scope of class Item (Kotlin reflection is not available)`

**To Reproduce**
Data classes and custom serialization
```
@Serializable
data class Container(
    @Serializable(ItemSerializer::class)
    val item: Item<out BaseData>
)

@Serializable
data class Item<T : BaseData>(
    val type: Int,
    val data: T
)

@Serializable
data class DataType1(val text: String) : BaseData()

@Serializable
data class DataType2(val text: String) : BaseData()

@Serializable
abstract class BaseData

object ItemSerializer : JsonParametricSerializer<Item<out BaseData>>(Item::class) {
    override fun selectSerializer(element: JsonElement): KSerializer<out Item<out BaseData>> {
        return when (val type = element.jsonObject.getPrimitive("type").int) {
            1 -> Item.serializer(DataType1.serializer())
            2 -> Item.serializer(DataType2.serializer())
            else -> throw Exception("Unknown type: $type")
        }
    }
}
```
Usage
```
        val str = """{"item":{"type":1,"data":{"text":"DataType1"}}}"""
        val json = Json(JsonConfiguration.Stable)

        val container = json.parse(Container.serializer(), str)
        println(container)
        println(json.stringify(Container.serializer(), container))
```

**Environment**
 - Kotlin version: 1.3.72
 - Library version: 0.20.0
 - Kotlin platforms: Android

sandwwraith:
Indeed, `JsonParametricSerializer` lacks a method to select a serializer during serialization. I thought the standard one should be enough, but for such generic nesting case, it is impossible to know it in advance

Slavvkko:
After some testing I have this custom JsonParametricSerializer:
```
abstract class MyJsonParametricSerializer<T : Any>(descriptorName: String) : KSerializer<T> {
    override val descriptor: SerialDescriptor = PrimitiveDescriptor(descriptorName, PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: T) {
        selectSerializer(value).serialize(encoder, value)
    }

    override fun deserialize(decoder: Decoder): T {
[...]

PR #2189, opened 2023-02-10
Improved docs for JsonContentPolymorphicSerializer

shanshin:
Relates #834

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonContentPolymorphicSerializer.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonContentPolymorphicSerializer.kt
@@ -60,7 +60,7 @@ import kotlin.reflect.*
  * ```
  *
  * @param T A root type for all classes that could be possibly encountered during serialization and deserialization.
- * Must be an interface, abstract or sealed class
+ * Must be non-final class or interface.
  * @param baseClass A class token for [T].
  */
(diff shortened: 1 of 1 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/JsonContentPolymorphicSerializer.kt
```diff
@@ -55,11 +55,12 @@ import kotlin.reflect.*
...
  *
- * Json.parse(PaymentSerializer, """{"amount":"1.0","date":"03.02.2020"}""")
- * Json.parse(PaymentSerializer, """{"amount":"2.0","date":"03.02.2020","reason":"complaint"}""")
+ * Json.decodeFromString(PaymentSerializer, """{"amount":"1.0","date":"03.02.2020"}""")
+ * Json.decodeFromString(PaymentSerializer, """{"amount":"2.0","date":"03.02.2020","reason":"complaint"}""")
  * ```
  *
- * @param T A root class for all classes that could be possibly encountered during serialization and deserialization.
+ * @param T A root type for all classes that could be possibly encountered during serialization and deserialization.
+ * Must be an interface, abstract or sealed class
```
sandwwraith, line 63: I think we also support `open` classes? Maybe it's better to write just `non-final class or interface`.

sandwwraith (approved):

PR #2191, opened 2023-02-13
add function for removing entry in `JsonObjectBuilder`

TheFruxz:
Adds `JsonObjectBuilder.remove(key: String)` to remove an entry from a `JsonObjectBuilder` instance.

My first PR to this repo, so I hope I did everything right. :D

### DIFF
--- a/formats/json/api/kotlinx-serialization-json.api
+++ b/formats/json/api/kotlinx-serialization-json.api
@@ -324,6 +324,7 @@ public final class kotlinx/serialization/json/JsonObjectBuilder {
 	public fun <init> ()V
 	public final fun build ()Lkotlinx/serialization/json/JsonObject;
 	public final fun put (Ljava/lang/String;Lkotlinx/serialization/json/JsonElement;)Lkotlinx/serialization/json/JsonElement;
+	public final fun remove (Ljava/lang/String;)Lkotlinx/serialization/json/JsonElement;
 }
 
 public final class kotlinx/serialization/json/JsonObjectSerializer : kotlinx/serialization/KSerializer {
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementBuilders.kt
@@ -68,6 +68,14 @@ public class JsonObjectBuilder @PublishedApi internal constructor() {
      */
     public fun put(key: String, element: JsonElement): JsonElement? = content.put(key, element)
 
+    /**
+     * Remove the JSON Element associated with the given [key].
+     *
+     * Returns the previous value associated with [key], or `null` if the key was not present.
+     */
+    @ExperimentalSerializationApi
+    public fun remove(key: String): JsonElement? = content.remove(key)
+
     @PublishedApi
     internal fun build(): JsonObject = JsonObject(content)
 }


TheFruxz:
Firstly, I would expect from a builder class, that it can remove stuff, if there is no technological barrier against it.
But it is also a good thing to have, if you have an external API adding a bunch of stuff to it (or iterative based operations) because then you can remove specific stuff after it.

For example:
I have a reader, reading a file or user input, which reads line per line. Then a line after the first line is telling something that the first line is obsolete or not required. Then I would have to remove something, but with this builder, this is currently not possible.

Another example here would be, that you want to have an identifier at the beginning of the JSON, but it is not required in all cases, if the content added later would lead to a conflict-free store.

Every of these cases would lead to much cleaner code, if the builder has the ability to simply remove entries after they got added.

I would appreciate feedback if I am wrong about the functionality.

sandwwraith:
@qwwdfsad Given this request and #2156 , I think it is worth considering that builders should implement `MutableMap/List` just like regular `JsonElement`s implement `Map/List`

qwwdfsad:
`Mutable*` is another side of the story -- we could do that, and it would probably be a reasonable thing to do after (after a rehearsal of our design notes regarding stdlib).

 The issue with the proposed PR is that it doesn't solve any real-world problems, only theoretical ones.

>Firstly, I would expect from a builder class, that it can remove stuff, if there is no technological barrier against it.

Unfortunately, this is different from how we are inclined to think about API addition. We need a compelling reason to **add** stuff, not the other way around, and the score of the reasoning starts at negative values.

It would be nice to provide builders that comply with such expectations and where `remove*` functions are added deliberately (e.g. not `buildList {}` where it just comes with the `MutableList` contract), and the overall prior art overview.

For example, `Guava`, which is arguably one of the most used libraries, doesn't provide `remove*` with their builders (https://github.com/google/guava/issues/3596) , Scala's [builder](https://www.scala-lang.org/api/2.12.1/scala/collection/mutable/Builder.html) doesn't seem to have one and so on

aSemy:
> Could you please provide a rationale and the reasoning for the proposed API?
What are practical use-cases?

One reason is that there's not a clear way of how to achieve this at the moment. Without a clear helper method, it's easy to 

1. come up with complicated workarounds that aren't easy to read, or scale
   ```kotlin
   val obj = buildJsonObject {
     put("a", 1)
     put("b", 2)
     put("c", 3)
   }
   println(obj)
   
   println(
     // if I'm in a rush I might quickly hack this, which is confusing
     // and incurs technical/documentation debt
     JsonObject(obj.toMap().run { minus("b") })
   )

   println(
     // if I had more time, this is a little more clear
     JsonObject(obj.toMap() - "b")
     // but it requires some deeper understanding how Kotlin/KxS works,
     // more than should be necessary for a basic operation
   )
   ```
2. accidentally write code that looks correct, but introduces bugs

   ```kotlin
   println(
     // this returns a MutableMap, but I wanted JsonObject
     obj.toMutableMap().minus("b")
   }
   ```
2. or (for Kotlin / KxS beginners) get confused and give up!
4. It would also help understanding if `JsonObject` shared as much functionality as possible with `MutableMap`, because learning one helps with learning to use the other.

> I think it is worth considering that builders should implement `MutableMap`/`List` just like regular `JsonElements` implement `Map`/`List`

I agree, `JsonObject` should implement `MutableMap` and `JsonArray` should implement `MutableList`.

aSemy:
I've created a new issue to take this proposal further: https://github.com/Kotlin/kotlinx.serialization/issues/2308

PR #2193, opened 2023-02-14
Add GHA config for CI

Goooler:


sandwwraith:
kotlinx.serialization builds use [Jetbrains' public TeamCity server](https://teamcity.jetbrains.com/project/KotlinTools_KotlinxSerialization?mode=builds), so there is no need for GitHub actions in this project.

Goooler:
Just seeing.

Issue #2199, opened 2023-02-20
Polymorphic using both explicit definition and sealed implicit lookup


hrach:
I have a library that exposes the interface `Destination` and an API accepting this `Destination` - I need the library to be able to process all destinations with KotlinX.Serialization.

I have a user-land app that defines multiple destinations and joins them via a sealed interface.

```kotlin
package library

interface Destination

package app

@Serializable
sealed interface AppDestination : Destination {
   @Serializable
   object One : AppDestination

   @Serializable
   object Two : AppDestination
}
```

To be able to pass e.g. `AppDestination.One` to API accepting `Destination`, I must explicitly wire them together:

```kotlin
serializersModule = SerializersModule {
	polymorphic(Destination::class) {
		subclass(serializer<AppDestination.One>())
		subclass(serializer<AppDestination.Two>())
	}
}
```

I'd like to be able to utilize the sealed interface and just call the subclass for `AppDestination`

```kotlin
serializersModule = SerializersModule {
	polymorphic(Destination::class) {
		subclass(serializer<AppDestination>())
	}
}
```

Use case: https://github.com/kiwicom/navigation-compose-typed

sandwwraith:
Related: #1865

Issue #1865, opened 2022-02-10
Combining `SealedClassSerializer`s


Whathecode:
**What is your use-case and why do you need this feature?**

I have a base, non-sealed, interface, from which sealed classes extend. At runtime (this is a generic framework), I need to be able to create a serializer which can handle all subclasses of multiple sealed classes that extend from this common interface.

My initial expectation was that I could instantiate a new `SealedClassSerializer` and pass all `subclasses` and `subclassSerializers` from the other serializers as constructor parameters. However, I found out that after instantiating `SealedClassSerializer` this information is no longer accessible (it is stored in a private `class2Serializer` field).

As an example, my specific use case: I have an `IntegrationEvent` interface and application services which each define the events they emit by defining a sealed class which extends from this interface. Application services can subscribe to events from other services. Serializing _any_ event received from dependent services thus requires me to be able to deserialize any of the events defined as subclasses for each of the sealed event classes.

P.s. I understand `SealedClassSerializer` is an internal serialization API; I find myself accessing internal APIs quite frequently since I'm using `kotlinx.serialization` in quite elaborate use cases. I hope sharing these may inspire some of the internal APIs to be stabilized and made public, since I would argue these are valid use cases. For now, I don't mind using them as unstable APIs.

**Describe the solution you'd like**

Could `subclasses` and `subclassSerializers`, or a meaningful map thereof, be made public as readonly fields? Once you have access to this class, I don't think information hiding makes sense. In addition, maybe a factory method can be added which takes multiple other sealed class serializers to combine them.

I currently hack this using JVM reflection as workaround:

```
// Get event serializer capable of serializing published events, as well as events the service subscribes to.
val eventSerializers = dependentServices.plus( serviceKlass ).map { getEventSerializer( it ) }
// HACK: There is no public accessor to get the subclass serializers. There probably should b.
val subclassesField = SealedClassSerializer::class.java
    .getDeclaredField( "class2Serializer" )
    .apply { isAccessible = true }
val allSubclassSerializers = eventSerializers.flatMap {
    @Suppress( "UNCHECKED_CAST" )
    val serializers = subclassesField.get( it )
        as Map<KClass<IntegrationEvent<*>>, KSerializer<IntegrationEvent<*>>>
    serializers.toList()
}.toMap()
eventSerializer = SealedClassSerializer(
    IntegrationEvent::class.qualifiedName!!,
    IntegrationEvent::class,
[...]

PR #2201, opened 2023-02-20
Sealed modules

pdvrieze:
This is an implementation for #2199 and #1865. This implementation just modifies the `subclass` method in the builder. It does (internally) expose the registered children in `SealedSerializer`. It is not possible to trigger this behaviour from manually written serializers (instead of `SealedSerializer`).

This does also remove the prohibition of registering sealed classes, but note that it will register the children of the sealed class, but not the sealed parent itself. The code does allow for overlaps (by ignoring the registration in this case) to accommodate overlapping hierarchies (although this can be abused in some ways) - it does mean that sealed child serializers can be manually overridden by registering the sub-serializer before the sealed parent.


pdvrieze:
I'll update the pull request later when I have time. While this introduces "changed" behaviour the current (without this request) implementation is not actually valid.

As sealed classes (and interfaces) are abstract they don't actually make valid serializers in the polymorphic case.

For serialization the value with never be directly of the sealed type, for deserialization it will never create a sealed instance.

The only challenge with changing existing behaviour is when existing code does register the base for polymorphic serialization, this could have 2 consequences:
- The children may already be registered, this will cause duplicate registration exceptions
- If the children weren't registered they would now be, perhaps making certain code paths work accidentally.

As the current code doesn't (appear to) throw an exception in case sealed parent is registered it may be best to use a separate function name. I'll update this in the pull request.

Whathecode:
@sandwwraith At a glance, no. #1865 is a different use case.

> At **runtime** (this is a generic framework), I need to be able to create a serializer which can handle all subclasses of multiple sealed classes that extend from this common interface.

I don't have the types available at compile time, thus I can't register them in a serial module as you would normally do.

My use case is a framework in which the app building on top of it links in additional types, through reflection, which extend from an interface defined in the framework. The [framework then supports serializing aggregations of such extending classes](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/jvmMain/kotlin/dk/cachet/carp/common/application/ApplicationServiceInfo.kt#LL161C32-L161C49).

I haven't looked at whether I could somehow create a `SerializersModule` with dynamically retrieved types through reflection, which perhaps is possible? I would then need to expose that module instead of a serializer to the caller. If that's possible, I _guess_ this feature could be used to achieve something similar? But, I currently don't have time to look into this.

pdvrieze:
I've updated the pull request by a split out function (although the tests that were deleted - and no longer are did verify that passing a sealed type serializer should fail for subclass).
@Whathecode If you have the serializer dynamically then you could use that to register all children to the module. Of course this only works if the base sealed type is serializable and you are not pulling all kinds of class loader shenanigans (that are not transparent to the sealed serializer).

pdvrieze:
@Whathecode The thing this pull request does is support the case where there is a polymorphic base type `P`, a sealed base `B : P` and leaf types `L1 : B`, `L2: B`, `L9: B`, etc.. The pull request allows one to, rather than register all leafs (L1…L9) independently as polymorphic child of `P`, "register" the sealed parent as child of `P` and all the actually serializable leaves are then registered automatically as serializers for the base type `P` (the sealed parent is by definition abstract so not can not/should not be registered itself).

pdvrieze:
I've rebased this upon the current dev, and am happy to make needed changes. For clarity this version does not reuse the function name, but has a separate function instead (the original one would not work validly on sealed classes).

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +19,38 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
     private val baseSerializer: KSerializer<Base>? = null
 ) {
-    private val subclasses: MutableList<Pair<KClass<out Base>, KSerializer<out Base>>> = mutableListOf()
+    private val subclasses: MutableMap<KClass<out Base>, KSerializer<out Base>> = mutableMapOf()
     private var defaultSerializerProvider: ((Base) -> SerializationStrategy<Base>?)? = null
     private var defaultDeserializerProvider: ((String?) -> DeserializationStrategy<Base>?)? = null
 
+
+    /**
+     * Registers the child serializers for the sealed [subclass] [serializer] in the resulting module under the [base class][Base].
```
sandwwraith, line 28: Documentation here does not mention that `Base` has to be sealed.

Also, it would be more helpful if documentation contained the example in which scenario this function has to be used. I think #2199 has a good example with

```kotlin
interface Base

@Serializable
sealed interface Sub 

serializersModule {
  polymorphic(Base::class) {
     subclassesOf<Sub>()
  }
}  
```

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +19,38 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     */
+    public inline fun <reified T : Base> subclassesOf(): Unit =
```
sandwwraith, line 30: All new functions should be marked with `@ExperimentalSerializationApi`

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +19,38 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+        subclassesOf(serializer<T>())
+
+
+    /**
+     * Registers the subclasses of the given class as subclasses of the outer class. This currently requires `baseClass`
+     * to be sealed.
+     */
+    public fun <T: Base> subclassesOf(serializer: KSerializer<T>) {
```
sandwwraith, line 38: After thinking about it for a while, I came to the idea that `subclassesOf` is a misleading name. It implies that we can extract all subclasses for an arbitrary class, which we can't do — and as we all know, not every person will be reading documentation to find out that we are able to handle only sealed classes. So more appropriate name would be `sealedSubclassesOf` or `subclassesOfSealed` or something in that direction.

sandwwraith (commented):
While this does not completely solve #1865, it is a viable workaround, so I'm in favor of adding such function after addressing all the comments

rocketraman:
Any updates on this PR?

rocketraman:
Bump on this PR... reading through the comments above the changes necessary to get this to the finish line seem pretty trivial -- some docs, an experimental annotation, and a function rename.

pdvrieze:
My apologies for not progressing this earlier, I completely missed this.

I have created an update for this (and rebased it). This update renames the function to `subclassesOfSealed` (per suggestion, it makes more sense than the other suggestion).

I did add a check that (grand)children are not themselves open polymorphic at module creation time (rather than leaving it to the validation in the format creation - as done in Json). This was chosen over allowing it, but ignoring these children (requiring those to be still registered the normal way). This latter option looks confusing.

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
     private val baseClass: KClass<Base>,
     private val baseSerializer: KSerializer<Base>? = null
 ) {
-    private val subclasses: MutableList<Pair<KClass<out Base>, KSerializer<out Base>>> = mutableListOf()
+    private val subclasses: MutableMap<KClass<out Base>, KSerializer<out Base>> = mutableMapOf()
```
sandwwraith: There are some various checks in the module builder (https://github.com/Kotlin/kotlinx.serialization/blob/d15dfeec76b6c39cee65f9b32614bbf14cd5b08d/core/commonMain/src/kotlinx/serialization/modules/SerializersModuleBuilders.kt#L199) for duplicating classes/names. By changing list of pairs to map we allow silent overwrite of keys, so I think leaving list here is better

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
     private var defaultSerializerProvider: ((Base) -> SerializationStrategy<Base>?)? = null
     private var defaultDeserializerProvider: ((String?) -> DeserializationStrategy<Base>?)? = null
 
+
+    /**
+     * Registers the child serializers for the sealed [subclass] [serializer] in the resulting module under
+     * the [base class][Base]. Please note that type `T` must be sealed, if not a runtime error will
```
sandwwraith: IMO "type `T` has to be sealed and have standard serializer" is slightly better version, since user may declare a custom serializer for sealed type.

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     * be thrown at registration time. This function is a convenience function for the version that
+     * receives a serializer.
+     *
+     *
+     * Example:
+     * ```kotlin
```
sandwwraith: IIRC lang tags do not work with Dokka, but that has to be re-checked

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+    @ExperimentalSerializationApi
+    public fun <T: Base> subclassesOfSealed(serializer: KSerializer<T>) {
+        // Note that the parameter type is `KSerializer` as `SealedClassSerializer` is an internal type
+        // not available to users
+        require(serializer is SealedClassSerializer) {
+            "subClassesOf only supports automatic adding of subclasses of sealed types."
+        }
+        for ((subsubclass, subserializer) in serializer.class2Serializer.entries) {
+            // This error would be caught by the Json format in its validation, but this is format specific
+            require (subserializer.descriptor.kind != PolymorphicKind.OPEN) {
```
sandwwraith, line 95: Did you not check `PolymorphicKind.SEALED` intentionally? I have intuition that this indeed should work:
```kotlin
interface I

@Serializable sealed interface A: I
@Serializable sealed class B: A
@Serializable class Impl: B()

serializersModule {
    polymorphic(I::class) {
        subclassesOfSealed<A>() // should serialize Impl
    }
}
```

IIRC plugin already flattens hierarchy, so A.class2Serializer will include Impl.serializer().

Please add the corresponding test for that and mention in the documentation that subclasses of sealed subclasses are also collected.

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     * ```
+     *
+     * It is an error if th
+     */
+    @ExperimentalSerializationApi
+    public fun <T: Base> subclassesOfSealed(serializer: KSerializer<T>) {
+        // Note that the parameter type is `KSerializer` as `SealedClassSerializer` is an internal type
+        // not available to users
+        require(serializer is SealedClassSerializer) {
+            "subClassesOf only supports automatic adding of subclasses of sealed types."
```
sandwwraith:
```suggestion
subclassesOfSealed only supports automatic adding of subclasses of sealed types with standard serializers.
```

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -19,15 +20,87 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     * class Sub1: Sub
+     *
+     * serializersModule {
+     *   polymorphic(Base::class) {
+     *      subclassesOfSealed(Sub.serializer())
+     *   }
+     * }
+     * ```
+     *
+     * It is an error if th
```
sandwwraith: Incomplete comment

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -116,3 +189,10 @@ public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.
  */
 public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.subclass(clazz: KClass<T>): Unit =
     subclass(clazz, serializer())
+
+/**
+ * Registers the child serializers for the sealed class [T] in the resulting module under the [base class][Base].
+ */
+@ExperimentalSerializationApi
+public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.subclassesOfSealed(clazz: KClass<T>): Unit =
+    subclassesOfSealed(clazz.serializer())
```
sandwwraith: should be just `serializer()` since you already have `reified T`. I feel like this overload may be not necessary, though. But we may include it for completeness

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        @SerialName("Baz")
+        data class Baz(val baz: String) : Foo()
+    }
+
+    @Serializable
+    sealed class FooOpen: FooBase() {
+        @Serializable
+        data class Bar(val bar: Int) : FooOpen()
+        @Serializable
+        abstract class Baz: FooOpen()
```
sandwwraith, line 48: As I said, please add identical test where this class is sealed and has some inheritors

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import kotlinx.serialization.modules.*
+import kotlinx.serialization.test.*
+import kotlin.test.*
+
+class PolymorphicSealedChildTest {
+
+    @Serializable
+    data class FooHolder(
+        val someMetadata: Int,
+        val payload: List<@Polymorphic FooBase>
```
sandwwraith: `@Polymorphic` is not necessary as abstract classes are polymorphic by default

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    fun testSaveSealedClassesList() {
+        assertStringFormAndRestored(
+            """{"someMetadata":42,"payload":[
+            |{"type":"Bar","bar":1},
+            |{"type":"Baz","baz":"2"}]}""".trimMargin().replace("\n", ""),
+            FooHolder(42, listOf(Foo.Bar(1), Foo.Baz("2"))),
+            FooHolder.serializer(),
+            json,
+            printResult = true
```
sandwwraith: This can be removed

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
pdvrieze: Adapted (ignoring the ability of a custom serializer to mimick sealed serializer sufficiently)

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
pdvrieze: This code intends to guard against the issue where you have:

```kotlin
interface I
@Serializable sealed interface A : I
@Serializable abstract class B: A // or open
@Serializable class Impl(): B()
```
In such case `Impl` would have to be registered independently against base class `I`. I don't think it would be valid to allow this to be resolved by resolving with `B` as base class instead. I'll make sure to have the tests include this.

pdvrieze: Note that already fails with the serializer for Baz can't be registered as subclass because it is open. While I could "skip" Baz as open or let the failure trigger later, it may be better to throw it here as.

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
pdvrieze: I've removed the `printResult` bit.

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
pdvrieze: Fixed

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
pdvrieze: Finally, I did test the code without this check. It actually fails on registration (as an open polymorphic class is registered as subclass). Instead it is possible to "skip" these descendants, and have the leaf types registered, but this is rather confusing in behaviour so I think it is better to disallow this.

core/api/kotlinx-serialization-core.api
```diff
@@ -1336,6 +1336,7 @@ public final class kotlinx/serialization/modules/PolymorphicModuleBuilder {
 	public final fun default (Lkotlin/jvm/functions/Function1;)V
 	public final fun defaultDeserializer (Lkotlin/jvm/functions/Function1;)V
 	public final fun subclass (Lkotlin/reflect/KClass;Lkotlinx/serialization/KSerializer;)V
+	public final fun subclassesOfSealed (Lkotlinx/serialization/KSerializer;)V
```
sandwwraith, line 1339: Api dump looks out of date

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
sandwwraith: I agree it should fail on open classes

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -23,6 +24,79 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
     private var defaultSerializerProvider: ((Base) -> SerializationStrategy<Base>?)? = null
     private var defaultDeserializerProvider: ((String?) -> DeserializationStrategy<Base>?)? = null
 
+
+    /**
+     * Registers the child serializers for the sealed [subclass] [serializer] in the resulting module under
```
sandwwraith:
```suggestion
     * Registers the child serializers for the sealed type [T] in the resulting module under
```

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -23,6 +24,79 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     * the [base class][Base]. Please note that type `T` has to be sealed and have a standard serializer, if not a runtime error will
+     * be thrown at registration time. This function is a convenience function for the version that
```
sandwwraith:
```suggestion
     * be thrown at registration time. 
     *  If one of [T]'s subclasses is a sealed serializable class on its own, its subclasses are registered recursively as well. If one of [T]'s subclasses is an open polymorphic class, an [IllegalArgumentException] is thrown.
     * 
     * This function is a convenience function for the version that
```

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
sandwwraith: There are no [sublcass] nor [serializer]

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -23,6 +24,79 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
...
+     * ```
+     */
+    @ExperimentalSerializationApi
+    public inline fun <reified T : Base> subclassesOfSealed(): Unit =
+        subclassesOfSealed(serializer<T>())
+
+
+    /**
+     * Registers the subclasses of the given class as subclasses of the outer class. This currently
+     * requires `serializer` to be sealed. If not a runtime error will be thrown at registration time.
```
sandwwraith: Please align documentation with version above (with my comments)

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -116,3 +190,10 @@ public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.
  */
 public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.subclass(clazz: KClass<T>): Unit =
     subclass(clazz, serializer())
+
+/**
+ * Registers the child serializers for the sealed class [T] in the resulting module under the [base class][Base].
+ */
+@ExperimentalSerializationApi
+public inline fun <Base : Any, reified T : Base> PolymorphicModuleBuilder<Base>.subclassesOfSealed(clazz: KClass<T>): Unit =
+    subclassesOfSealed(clazz.serializer())
```
sandwwraith:
```suggestion
    subclassesOfSealed(serializer<T>())
```

It is a more correct version.
Although given it behaves exactly the same as the overload without `clazz`, I think we can remove it.

docs/polymorphism.md
```diff
@@ -414,6 +414,13 @@ fun main() {
 
 > Note: On Kotlin/Native, you should use `format.encodeToString(PolymorphicSerializer(Project::class), data))` instead due to limited reflection capabilities.
 
+### Registering sealed children as subclasses
+A sealed parent interface or class can be used to directly register all its children using `subclassesOfSealed`.
```
sandwwraith, line 419: If you add documentation, please also add examples for it (and run `./gradlew knit` afterwards so it can be transformed to tests).
However, given ongoing migration to kotlinlang.org website, I think we can skip this section for now. I'll add it manually later to the new format.

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    @Test
+    fun testOpenGrandchildIsInvalid() {
+
+        lateinit var openJson: Json
+        val e = assertFailsWith<IllegalArgumentException> {
+            openJson = Json {
+                serializersModule = SerializersModule {
+                    polymorphic(FooBase::class) {
+                        subclassesOfSealed<FooOpen>()
+                        fail("This code should be unreachable as the previous operation fails")
```
sandwwraith: There's no need for `fail`, `assertFailsWith` will throw an exception if code completes succesfully. Please change back to previous version

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphicSealedChildTest.kt
```diff
@@ -0,0 +1,125 @@
...
+                        subclass(FooOpen.BazChild1::class)
+                        subclass(FooOpen.BazChild2::class)
+                    }
+                }
+            }
+
+            fail("Unreachable code that would represent the usage if valid (which is isn't per policy)")
```
sandwwraith: Same here

docs/polymorphism.md
pdvrieze: I have added example code (and made the knit test work)

core/api/kotlinx-serialization-core.api
pdvrieze: oops

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
```diff
@@ -23,6 +24,86 @@ public class PolymorphicModuleBuilder<in Base : Any> @PublishedApi internal cons
     private var defaultSerializerProvider: ((Base) -> SerializationStrategy<Base>?)? = null
     private var defaultDeserializerProvider: ((String?) -> DeserializationStrategy<Base>?)? = null
 
+
+    /**
+     * Registers the child serializers for the sealed  type [T] in the resulting module under
```
sandwwraith:
```suggestion
     * Registers the child serializers for the sealed type [T] in the resulting module under
```

extra space

docs/polymorphism.md
```diff
@@ -414,6 +415,67 @@ fun main() {
...
+This will expose all children that would be available when serializing the parent directly, but now as sealed. Please
```
sandwwraith:
```suggestion
This will allow serializing the children using open polymorphism without the need to register each one individually. 
```

docs/polymorphism.md
```diff
@@ -414,6 +415,67 @@ fun main() {
...
+note that this is will remain open serialization, and the sealed parent serializer will not be used in serialization.
```
sandwwraith: I'm not sure this line is needed. What do you mean by "the sealed parent serializer will not be used in serialization." — non-concrete classes are never serialized to json; they're used only as types in the program.

docs/polymorphism.md
```diff
@@ -414,6 +415,67 @@ fun main() {
...
+In addition, it is not valid if the hierarchy contains open (not sealed) polymorphic children (this will result in
```
sandwwraith: IMO it's better to reuse KDoc wording:

If one of the type's subclasses is a sealed serializable class on its own, its subclasses are registered recursively as well. However, if one of the type's subclasses is an open polymorphic class, an [IllegalArgumentException] is thrown. In other words, all children/descendants must be either concrete or sealed.

core/commonMain/src/kotlinx/serialization/modules/PolymorphicModuleBuilder.kt
pdvrieze: Fixed, and did some wrapping of the lines.

docs/polymorphism.md
pdvrieze: Indeed better

docs/polymorphism.md
pdvrieze: Removed it. Looking back at it I intended it to mean that serialization will skip the intermediate [SealedClassSerializer] but custom serializers with `PolymorphicKind.Sealed` are not supported anyway (at least by this function), so this is more of an implementation detail.

docs/polymorphism.md
pdvrieze: You are right. Adjusted.

docs/polymorphism.md
```diff
@@ -414,6 +415,68 @@ fun main() {
 
 > Note: On Kotlin/Native, you should use `format.encodeToString(PolymorphicSerializer(Project::class), data))` instead due to limited reflection capabilities.
 
+### Registering sealed children as subclasses
+A sealed parent interface or class can be used to directly register all its children using `subclassesOfSealed`.
+This will allow serializing the children using open polymorphism without the need to register each one individually. 
+
+If one of the type's subclasses is a sealed serializable class on its own, its subclasses are registered recursively
+as well. However, if one of the type's subclasses is an open polymorphic class, an [IllegalArgumentException] is thrown.
```
sandwwraith: Knit tool is kinda quirky nowadays, so it can't insert references itself. I think using the exception name without the link to the stdlib will be fine.

```suggestion
as well. However, if one of the type's subclasses is an open polymorphic class, an `IllegalArgumentException` is thrown.
```

sandwwraith:
Did you base your branch on `master`? It contains irrelevant commit about KDoc formatting.

sandwwraith:
@pdvrieze can you please drop 74ed0bdb9ceac5a0f3a1f190609e60121335930e from the PR so I can merge it

pdvrieze:
@sandwwraith No idea why that commit keeps being included. Anyway I've adjusted it now.

sandwwraith:
Thanks a lot! I think it will help a lot of people

PR #2213, opened 2023-02-28
Fix MissingFieldException duplication

### LINKED ISSUE #2212: 'at path' is duplicated in MissingFieldException message
**Describe the bug**

Given a JSON array of objects, and one of the objects has a missing field, the error message is repeated.

```
Exception in thread "main" kotlinx.serialization.MissingFieldException: Field 'name' is required for type with serial name 'Cast', but it was missing at path: $.cast[0] at path: $.cast[0] at path: $.cast[0]
```

It looks unusual that the `at path: $.cast[0]` is repeated 3 times. I would expect to see it only once:

```
Field 'name' is required ... but it was missing at path: $.cast[0]
```

Looking at the full stacktrace shows that there is a chain of causes, and each appends the same error message.

```
Exception in thread "main" kotlinx.serialization.MissingFieldException: Field 'name' is required for type with serial name 'Cast', but it was missing at path: $.cast[0] at path: $.cast[0] at path: $.cast[0]
...
Caused by: kotlinx.serialization.MissingFieldException: Field 'name' is required for type with serial name 'Cast', but it was missing at path: $.cast[0] at path: $.cast[0]
...
Caused by: kotlinx.serialization.MissingFieldException: Field 'name' is required for type with serial name 'Cast', but it was missing at path: $.cast[0]
...
Caused by: kotlinx.serialization.MissingFieldException: Field 'name' is required for type with serial name 'Cast', but it was missing
```

**To Reproduce**

```kotlin
fun main() {
  val inputMalformed = """
      {
          "title": "Treasure Island",
          "cast": [
              {
              },
              {
                  "name": "gonzo"
              },
              {
                  "name": "rizzo"
              }
          ]
      }
    """.trimIndent()

  println(Json.decodeFromString<Movie>(inputMalformed))
  // Exception in thread "main" kotlinx.serialization.MissingFieldException:
  // Field 'name' is required for type with serial name 'Cast',
  // but it was missing at path: $.cast[0] at path: $.cast[0] at path: $.cast[0]
}

@Serializable
data class Movie(
  val title: String,
  val cast: List<Cast>,
)

@Serializable
data class Cast(
  val name: String
)
```

**Expected behavior**

**Environment**
 - Kotlin version: 1.7..20
 - Library version: 1.5.0
 - Kotlin platforms: JVM
 - Gradle version: 7.6

qwwdfsad:
Fixes #2212

### DIFF
--- a/formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
+++ b/formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
@@ -1,0 +1,43 @@
+package kotlinx.serialization.json
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.Json.Default.decodeFromString
+import org.junit.*
+import org.junit.Test
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.PrintWriter
+import java.io.StringWriter
+import kotlin.test.*
+
+class MissingFieldExceptionWithPathTest {
+
+    @Test // Repro for #2212
+    fun testMfeIsNotReappliedMultipleTimes() {
+        val inputMalformed = """{"title": "...","cast": [{}]"""
+        try {
+            Json.decodeFromString<Movie>(inputMalformed)
+            fail("Unreacheable state")
+        } catch (e: MissingFieldException) {
+            val writer = StringWriter()
+            val pw = PrintWriter(writer)
+            e.printStackTrace(pw)
+            pw.close()
+            val i1 = writer.toString().indexOf("at path")
+            val i2 = writer.toString().lastIndexOf("at path")
+            assertEquals(i1, i2)
+            assertTrue(i1 != -1)
+        }
+    }
+
+    @Serializable
+    data class Movie(
+        val title: String,
+        val cast: List<Cast>,
+    )
+
+    @Serializable
+    data class Cast(
+        val name: String
+    )
+}
--- a/formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
+++ b/formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
@@ -19,12 +19,9 @@ class MissingFieldExceptionWithPathTest {
             Json.decodeFromString<Movie>(inputMalformed)
             fail("Unreacheable state")
         } catch (e: MissingFieldException) {
-            val writer = StringWriter()
-            val pw = PrintWriter(writer)
-            e.printStackTrace(pw)
-            pw.close()
-            val i1 = writer.toString().indexOf("at path")
-            val i2 = writer.toString().lastIndexOf("at path")
+            val fullStackTrace = e.stackTraceToString()
+            val i1 = fullStackTrace.toString().indexOf("at path")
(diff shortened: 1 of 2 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -87,6 +87,9 @@ internal open class StreamingJsonDecoder(
             return result
 
         } catch (e: MissingFieldException) {
+            // Add "at path" if and only if we've just caught an exception and it hasn't been augmented yet
+            if (e.message!!.contains("at path")) throw e
+            // NB: we could've use some additional flag marker or augment the stacktrace, but it seemed to be as too much of a burden
             throw MissingFieldException(e.missingFields, e.message + " at path: " + lexer.path.getPath(), e)
```
qwwdfsad, line 93: I would prefer `val mfe = .. new mfe with better message ...` to later do `mfe.stackTrace = e.stackTrace`, but unfortunately, it works only on JVM, not sure if we should bother

shanshin, line 93: Can we inherit from class `MissingFieldException` or will it break backward compatibility?

formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Test // Repro for #2212
+    fun testMfeIsNotReappliedMultipleTimes() {
+        val inputMalformed = """{"title": "...","cast": [{}]"""
+        try {
+            Json.decodeFromString<Movie>(inputMalformed)
+            fail("Unreacheable state")
+        } catch (e: MissingFieldException) {
+            val writer = StringWriter()
+            val pw = PrintWriter(writer)
+            e.printStackTrace(pw)
```
shanshin, line 24: `e.stackTraceToString()`?

formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
```diff
@@ -0,0 +1,43 @@
...
+            pw.close()
+            val i1 = writer.toString().indexOf("at path")
```
shanshin, line 26: We can use `message`, can't we?

qwwdfsad: Not really, because then we'll only check for the most outer exception. I wanted to check exactly the whole exception chain

formats/json-tests/jvmTest/src/kotlinx/serialization/json/MissingFieldExceptionWithPathTest.kt
qwwdfsad: Haven't known about that, nice!

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
qwwdfsad: We can, but it's a whole new class for the sake of very minor thing, I believe this one is more than enough

shanshin (approved):

PR #2219, opened 2023-03-04
Fix #2218: NoSuchMethodError when parsing a JSON stream on Java 8

### LINKED ISSUE #2218: java.lang.NoSuchMethodError when parsing a JSON stream on Java 8
**Describe the bug**

I suspect the JSON format to be prone to this issue: https://github.com/plasma-umass/doppio/issues/497#issuecomment-334740243

The problem seems to be in the `CharsetReader`'s constructor, on the line ＃23:

https://github.com/Kotlin/kotlinx.serialization/blob/6dbb8a9d839fb817ece7566f991fb450a226ea31/formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharsetReader.kt#L18-L24

It seem that this class is compiled with a higher language level and, running on Java 8, causing the issue.

I've `javap`-ed the `CharsetReader` class (from the 1.5.0 versio of the library) and although I see the its class version is 52 (Java 8) it seems to be not enough for this nasty bug. The bytecode seems to use the `ByteBuffer`, which is incorrect :point_down: 

image

It [seems](https://github.com/plasma-umass/doppio/issues/497#issuecomment-353394008) that a proper compiler option needs to be passed during the compilation process. The other way to solve this is [using](https://github.com/plasma-umass/doppio/issues/497#issuecomment-334740243) a manual cast.

**To Reproduce**

Please, refer to this minimal reproducible example: https://github.com/madhead/kotlinx-serialization-json-java8. Follow its README.

Basically, to reproduce the issues one needs to call `Json.decodeFromStream` on Java 8. As a result, he gets an exception like this:

```
Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
        at kotlinx.serialization.json.internal.CharsetReader.<init>(CharsetReader.kt:23)
        at kotlinx.serialization.json.internal.JavaStreamSerialReader.<init>(JvmJsonStreams.kt:258)
        at kotlinx.serialization.json.JvmStreamsKt.decodeFromStream(JvmStreams.kt:59)
        at me.madhead.kotlinx.serialization.json.java8.AppKt.main(App.kt:13)
        at me.madhead.kotlinx.serialization.json.java8.AppKt.main(App.kt)
```

**Expected behavior**

No exception is thrown, the stream is parsed correctly.

**Environment**

 - Kotlin platforms: JVM
 - JVM: 8.x
    ```
    openjdk version "1.8.0_322"
    OpenJDK Runtime Environment (Temurin)(build 1.8.0_322-b06)
    OpenJDK 64-Bit Server VM (Temurin)(build 25.322-b06, mixed mode)
    ```
 - Kotlin version: 1.8.10
 - Library version: 1.5.0

madhead:
Creating a test for this one would be tricky as it would require compiling the library with Java 9+ and running a single test on Java 8.

Instead of that one could use the minimal reproducible example repository (https://github.com/madhead/kotlinx-serialization-json-java8) to validate a fix locally :sweat_smile:


formats/json/jvmMain/src/kotlinx/serialization/json/internal/CharsetReader.kt
```diff
@@ -20,7 +20,7 @@ internal class CharsetReader(
             .onMalformedInput(CodingErrorAction.REPLACE)
             .onUnmappableCharacter(CodingErrorAction.REPLACE)
         byteBuffer = ByteBuffer.wrap(ByteArrayPool8k.take())
-        byteBuffer.flip() // Make empty
```
shanshin, line 23: Please additionally write in the comments why an explicit cast is needed here.

Problem with the return type.
In `Buffer`, the `flip` method was final, so it returned `Buffer'.

In subsequent versions of Java, the method was opened, and was overridden for `ByteBuffer`, thereby it began to return `ByteBuffer`

You can see this if you look at what javac compiles `byteBuffer.flip();`
for java 8 `INVOKEVIRTUAL java/nio/ByteBuffer.flip ()Ljava/nio/Buffer;`
for java 11 `INVOKEVIRTUAL java/nio/ByteBuffer.flip ()Ljava/nio/ByteBuffer;`

madhead: Done. A comment added.

shanshin (approved):
Thanks for the research!

PR #2226, opened 2023-03-11
json-okio: rename `target` variables to `sink`

BenWoodworth:
The Okio codebase names Sink/BufferedSink variables "sink", so this aligns the naming convention

okio-sink


shanshin:
Thanks!

PR #2227, opened 2023-03-12
Add support to decode numeric literals containing an exponent

### LINKED ISSUE #2078: Decoding an integer value containing an exponent
I recently came across a situation where [a JSON minifier](https://pkg.go.dev/github.com/tdewolff/minify/v2@v2.9.17/json) minified an integer value with 3 or more trailing zeros using an exponent. E.g. 1000 minified to 1e3. It appears that the JSON spec allows this, but kotlinx.serialization currently doesn't handle it.

As a side note, I have verified that Swift’s implementation of [JSONSerialization](https://developer.apple.com/documentation/foundation/jsonserialization) (MacOS 12.6, Swift 5.7) handles this situation correctly i.e. decodes 1e3 to 1000. 

Example code: 
The first statement in main() works as expected. The second statement throws `kotlinx.serialization.json.internal.JsonDecodingException`.

```

@Serializable
data class SomeData(val count: Int)

fun main() {
    println(Json.decodeFromString<SomeData>("""{ "count": 1000 }"""))
    println(Json.decodeFromString<SomeData>("""{ "count": 1e3 }"""))
}
```

Expected behavior: Should be able to decode integer/long values containing an exponent. 

Environment:
 - Kotlin version: 1.7.20
 - Library version: 1.4.1
 - Kotlin platforms: JVM (I didn’t check the behavior on other platforms)
 - Gradle version: 7.4.2
 - IDE version: IntellijIDEA 2022.2.3
 - MacOS 12.6

xBaank:
Closes #2078

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
@@ -24,12 +24,6 @@ class JsonExponentTest : JsonTestBase() {
     }
 
     @Test
-    fun testExponentDecodingTruncatedDecimal() = parametrizedTest {
-        val decoded = Json.decodeFromString<SomeData>("""{ "count": -1E-1 }""", it)
-        assertEquals(0, decoded.count)
-    }
-
-    @Test
     fun testExponentDecodingPositiveDouble() = parametrizedTest {
         val decoded = Json.decodeFromString<SomeDataDouble>("""{ "count": 1.5E1 }""", it)
         assertEquals(15.0, decoded.count)
@@ -39,6 +33,12 @@ class JsonExponentTest : JsonTestBase() {
     fun testExponentDecodingNegativeDouble() = parametrizedTest {
         val decoded = Json.decodeFromString<SomeDataDouble>("""{ "count": -1e-1 }""", it)
         assertEquals(-0.1, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingErrorTruncatedDecimal() = parametrizedTest {
+        assertFailsWithSerial("JsonDecodingException")
+        { Json.decodeFromString<SomeData>("""{ "count": -1E-1 }""", it) }
     }
 
     @Test
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
@@ -1,0 +1,82 @@
+package kotlinx.serialization.json
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.test.assertFailsWithSerial
+import kotlinx.serialization.test.assertFailsWithSerialMessage
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonExponentTest : JsonTestBase() {
+    @Serializable
+    data class SomeData(val count: Long)
+    @Serializable
+    data class SomeDataDouble(val count: Double)
+
+    @Test
+    fun testExponentDecodingPositiveInteger() {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": 23e11 }""")
+        assertEquals(2300000000000, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingNegativeInteger() {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": -10E1 }""")
+        assertEquals(-100, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingTruncatedDecimal() {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": -1E-1 }""")
+        assertEquals(0, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingPositiveDouble() {
+        val decoded = Json.decodeFromString<SomeDataDouble>("""{ "count": 1.5E1 }""")
+        assertEquals(15.0, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingNegativeDouble() {
+        val decoded = Json.decodeFromString<SomeDataDouble>("""{ "count": -1e-1 }""")
+        assertEquals(-0.1, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingErrorExponent() {
+        assertFailsWithSerialMessage("JsonDecodingException", "Unexpected symbol 'e' in numeric literal")
+        { Json.decodeFromString<SomeData>("""{ "count": 1e-1e-1 }""") }
+    }
+
+    @Test
+    fun testExponentDecodingErrorExponentDouble() {
+        assertFailsWithSerialMessage("JsonDecodingException","Failed to parse type 'double' for input '1e-1e-1'")
+        { Json.decodeFromString<SomeDataDouble>("""{ "count": 1e-1e-1 }""") }
+    }
+
+    @Test
+    fun testExponentOverflowDouble() {
+        assertFailsWithSerialMessage("JsonDecodingException","Unexpected special floating-point value Infinity")
+        { Json.decodeFromString<SomeDataDouble>("""{ "count": 10000e10000 }""") }
+    }
+
+    @Test
+    fun testExponentUnderflowDouble() {
+        assertFailsWithSerialMessage("JsonDecodingException", "Unexpected special floating-point value -Infinity")
+        { Json.decodeFromString<SomeDataDouble>("""{ "count": -100e2222 }""") }
+    }
+
+    @Test
+    fun testExponentOverflowLong() {
+        assertFailsWithSerialMessage("JsonDecodingException","Numeric value overflow")
+        { Json.decodeFromString<SomeData>("""{ "count": 10000e10000 }""") }
+    }
+
+    @Test
+    fun testExponentUnderflowLong() {
+        assertFailsWithSerialMessage("JsonDecodingException","Numeric value overflow")
+        { Json.decodeFromString<SomeData>("""{ "count": -10000e10000 }""") }
+    }
+
+}
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
@@ -1,8 +1,7 @@
 package kotlinx.serialization.json
 
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.decodeFromString
-import kotlinx.serialization.test.assertFailsWithSerialMessage
+import kotlinx.serialization.test.*
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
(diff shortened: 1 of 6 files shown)


sandwwraith:
Hi, thanks for your PR, and sorry it has taken so long to get to it. I see you have some other commits beside yours, did you start your branch off `master`? Please rebase it on `dev`. Regarding contents: this is really nice addition IMO, as json.org spec indeed permits exponent numbers. However, to fully support them, you also need to take into account `JsonElement` parsing — e.g. conversion from `JsonElement` to a serializable class. (`decodeFromJsonElement<SomeData>(parseToJsonElement(string))` As far as I see, it uses simple `String.toLong()` conversion and will also fail on numbers with exponent. To test behavior in all available modes, you may use `parameterizedTest` function from `JsonTestBase` — see its usages.

sandwwraith (changes requested):
.

xBaank:
> I see you have some other commits beside yours, did you start your branch off master? Please rebase it on dev.

you mean creating another pull request to the dev branch with the master changes? or just rebase master into my branch with newest changes?

sandwwraith:
Your branch (with your-only commits) should start from the latest dev. You can force-push it to use the same pull request

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    fun testExponentDecodingTruncatedDecimal() = parametrizedTest {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": -1E-1 }""", it)
+        assertEquals(0, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingPositiveDouble() = parametrizedTest {
+        val decoded = Json.decodeFromString<SomeDataDouble>("""{ "count": 1.5E1 }""", it)
```
xBaank, line 34: Should cases like this one be supported when decoding into a long | int too? and if so, should it be truncated, rounded...?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
```diff
@@ -71,3 +71,32 @@ internal fun String.toBooleanStrictOrNull(): Boolean? = when {
     this.equals("false", ignoreCase = true) -> false
     else -> null
 }
+
+private val INT_NUMBERS_CHARS = arrayOf('1', '2', '3', '4', '5', '6', '7', '8', '9', '0', 'e', 'E', '+', '-')
+internal fun String.toLongJson(): Long {
+    return toLongOrNull() ?: toDouble()
+        .takeIf { all { it in INT_NUMBERS_CHARS } }
```
xBaank, line 78: Same here

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
xBaank, line 1: I had to change this as there was a weird edge case on js, when encoding DOUBLE.MIN_VALUE (4.9E-324) gets rounded to 5E-324, so when decoding to long it would be 0 instead of DOUBLE.MIN_VALUE
image

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
sandwwraith: I think this can be left as is in order not to complicate the code. We can always add it later, if there's a request to do so. Original 'parsing exponent to Long' request was about minifying, and I hope minifier just don't insert random dots in the number.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
sandwwraith: Sorry, I don't quite get where your comment points at

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
```diff
@@ -117,16 +117,18 @@ private object JsonLiteralSerializer : KSerializer<JsonLiteral> {
...
 
         // most unsigned values fit to .longOrNull, but not ULong
         value.content.toULongOrNull()?.let {
             encoder.encodeInline(ULong.serializer().descriptor).encodeLong(it.toLong())
             return
         }
 
-        value.doubleOrNull?.let { return encoder.encodeDouble(it) }
-        value.booleanOrNull?.let { return encoder.encodeBoolean(it) }
+        value.content.toDoubleOrNull()?.let { return encoder.encodeDouble(it) }
```
sandwwraith, line 130: `doubleOrNull` simply calls `toDoubleOrNull`, why this change is required?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
```diff
@@ -71,3 +71,32 @@ internal fun String.toBooleanStrictOrNull(): Boolean? = when {
...
+        ?.takeIf { it < Long.MAX_VALUE && it > Long.MIN_VALUE }
+        ?.toLong()
+    ?: throw NumberFormatException("Invalid number format: $this")
+}
+internal fun String.toLongJsonOrNull(): Long? {
+    return toLongOrNull() ?: toDoubleOrNull()
+        ?.takeIf { all { it in INT_NUMBERS_CHARS } }
```
sandwwraith, line 85: I think such code poses hidden problem: `all` iterates all symbols and `it in` iterates all chars in the `INT_NUMBERS_CHARS` array, so it results in a quadratic complexity. Moreover, it is just hard to read. I see several possibilities here:

1. To simplify condition, I think it is possible just to check `'.' !in it`. All other non-digit and non-E signs should be prohibited by `toDoubleOrNull()`, I guess?
2. A harder, but more correct change would be extract parsing logic from `consumeNumericLiteral` or to create `StringJsonLexer` in place so we wouldn't have two different code paths doing the same thing. 

I think it is possible to also extract these `takeIf`s to a separate function since it doesn't matter if you use `toDouble` or `toDoubleOrNull` given that exception is anyway thrown at the end if needed.

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -630,6 +657,18 @@ internal abstract class AbstractJsonLexer {
             ++current
         }
         currentPosition = current
+
+        fun calculateExponent(exponentAccumulator: Long, isExponentPositive: Boolean): Double = when (isExponentPositive) {
+            false -> 10.0.pow(-exponentAccumulator.toDouble())
+            true -> 10.0.pow(exponentAccumulator.toDouble())
+        }
+
+        if(hasExponent) {
```
sandwwraith, line 666: [nit] formatting: space before opening brace

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
xBaank: Basically DOUBLE.MIN_VALUE is 4.9E-324 in jvm but 5E-324 on js so this test was failing on js

https://github.com/Kotlin/kotlinx.serialization/blob/c01b19fd1ee57a7d3e42eec58d7ba333b20095d1/formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt#L111

because the implementation i made wtih longOrNull allows 5E-324 to be encoded i changed so exponential numbers are delegated to `toDoubleOrNull`.

https://github.com/Kotlin/kotlinx.serialization/blob/a1ed53bbce49f147f961f59ac17caefe53b1f655/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt#L120-L122

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
xBaank: Just to for consistency but its true that this change is not needed at all.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
xBaank: I think  `StringJsonLexer` is the best solution

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -315,6 +321,18 @@ public val JsonPrimitive.contentOrNull: String? get() = if (this is JsonNull) nu
 private fun JsonElement.error(element: String): Nothing =
     throw IllegalArgumentException("Element ${this::class} is not a $element")
 
+private inline fun <T> mapExceptionsToNull(f : () -> T) : T? {
+    return try { f()}
```
sandwwraith, line 325: formatting: space before brace.

I recommend using IntelliJ IDEA auto-formatter, as it honors Kotlin's style guide

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
```diff
@@ -117,16 +117,18 @@ private object JsonLiteralSerializer : KSerializer<JsonLiteral> {
             return encoder.encodeInline(value.coerceToInlineType).encodeString(value.content)
         }
 
-        value.longOrNull?.let { return encoder.encodeLong(it) }
+        //use content instead of longOrNull as it can process exponential notation,
```
sandwwraith, line 120:
```suggestion
        // use .content instead of .longOrNull as latter can process exponential notation,
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt
```diff
@@ -117,16 +117,18 @@ private object JsonLiteralSerializer : KSerializer<JsonLiteral> {
...
+        //and it should be delegated to double when encoding.
```
sandwwraith, line 121:
```suggestion
        // and it should be delegated to double when encoding.
```

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -630,6 +657,18 @@ internal abstract class AbstractJsonLexer {
...
         currentPosition = current
+
+        fun calculateExponent(exponentAccumulator: Long, isExponentPositive: Boolean): Double = when (isExponentPositive) {
+            false -> 10.0.pow(-exponentAccumulator.toDouble())
+            true -> 10.0.pow(exponentAccumulator.toDouble())
+        }
+
+        if (hasExponent) {
+            val doubleAccumulator  = accumulator.toDouble() * calculateExponent(exponentAccumulator, isExponentPositive)
+            if(doubleAccumulator > Long.MAX_VALUE || doubleAccumulator < Long.MIN_VALUE) fail("Numeric value overflow")
```
sandwwraith, line 668: fornatting: space after if

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonExponentTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    fun testExponentDecodingNegative() = parametrizedTest {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": -10E1 }""", it)
+        assertEquals(-100, decoded.count)
+    }
+
+    @Test
+    fun testExponentDecodingTruncatedDecimal() = parametrizedTest {
+        val decoded = Json.decodeFromString<SomeData>("""{ "count": -1E-1 }""", it)
```
sandwwraith, line 28: I've just realized that it is incorrect to truncate such values. That's why the problem with ` 4.9E-324` happens in the first place - it is too small. We should not allow such values to be parsed to `Long`, or any value with non-zero fractional part after applying exponent. I think the easiest way to check it would be with `doubleAccumulator.floor() == doubleAccumulator`, as in here: https://github.com/Kotlin/kotlinx.serialization/blob/e2092a41693c79686e787ccd975579aabb0b2412/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt#L154

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -315,6 +321,18 @@ public val JsonPrimitive.contentOrNull: String? get() = if (this is JsonNull) nu
...
+    catch (e : JsonDecodingException) { null }
+    catch (e : IndexOutOfBoundsException) { null }
```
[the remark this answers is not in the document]
xBaank: Yes, this test was failing on TREE mode, hasChar on consumeNumericLiteral doesn't update when the char is - | + | e | E
https://github.com/Kotlin/kotlinx.serialization/blob/cfce5d8bf4e24fd6a1caf2cef9f62dd689614ce5/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonParserFailureModesTest.kt#L134-L144

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -255,23 +255,32 @@ public val JsonElement.jsonNull: JsonNull
...
+        return result.toInt()
+    }
 
 /**
  * Returns content of the current element as int or `null` if current element is not a valid representation of number
  */
-public val JsonPrimitive.intOrNull: Int? get() = content.toIntOrNull()
+public val JsonPrimitive.intOrNull: Int?
+    get() =
+        mapExceptionsToNull { StringJsonLexer(content).consumeNumericLiteral().toInt().toLong().toInt() }
```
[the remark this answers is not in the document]
xBaank: I don't know why I did that

sandwwraith (approved):
Great job, thank you!

PR #2232, opened 2023-03-15
Add polymorphicAllSuperClasses for SerializersModuleBuilder on Jvm

XYZboom:
Add polymorphicAllSuperClasses for SerializersModuleBuilder on Jvm.
On the jvm platform, the generics in the KClass obtained by reflection will be erased (KClass<*>). We can assert that the KClass of the parent class must be the superClass of the KClass of the child class, so I think we need a function to receive KClass<*>.
The corresponding unit test has included.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -10,6 +10,11 @@ apply from: rootProject.file("gradle/configure-source-sets.gradle")
 
 kotlin {
     sourceSets {
+        jvmMain {
+            dependencies {
+                implementation kotlin("reflect")
+            }
+        }
         jvmTest {
             dependencies {
                 implementation 'io.kotlintest:kotlintest:2.0.7'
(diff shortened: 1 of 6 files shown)


sandwwraith:
Hi, thanks for the pull request! I do not see a connection between the fact that `KClass` type arguments are erased and the check for `actualClass.isSubclassOf(baseClass)`. Can you elaborate on that, please? 

Another proposed function, `polymorphicAllSuperClasses,` looks useful, but unfortunately, we can't accept it, as it requires `kotlin-reflect` as the dependency. Avoiding this dep is one of the crucial features of kotlinx.serialization. Besides, this function is relatively small and can be copy-pasted into any project if needed, and dependency on reflect is acceptable.

XYZboom:
Thank you for your response! When using a KClass object with erased generics as a parameter for the `registerPolymorphicSerializer` function, the compiler cannot recognize the specific type of the erased generics and thus cannot compile successfully. Therefore, I hope there is a function that can support me to make such a call. However, in most cases, the KClass object can be obtained directly using `ClassName::class`. I expect there to be a function that can be used to make batch calls when there are many parent classes in a class that need to be registered for polymorphic serialization.

Issue #1420, opened 2021-04-18
Ignore unknown keys per class / via annotation


sschuberth:
**What is your use-case and why do you need this feature?**

I know [keys can be ignored "globally"](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#ignoring-unknown-keys) on the JSON builder scope. But for safely, and to better document which classes have been implemented completely and which not, I'd prefer to ignore unknown keys per class only.

**Describe the solution you'd like**

I'd prefer an annotation-based solution, similar to Jackson's `@JsonIgnoreProperties(ignoreUnknown = true)` annotation.

codeyash:
any update?

Its very critial for scalability.

ixtli:
Yeah there are many major APIs were forced to use which version their API contracts publicly but still randomly permute certain sub objects.

PR #2235, opened 2023-03-15
Introduced annotation to ignore specific keys for deserialization

nmrsmn:
Created feature to ignore properties per class by introducing an annotation as per request of [#1420](https://github.com/Kotlin/kotlinx.serialization/issues/1420)

Please let me know if this is okay and if I need to solve some issues, thanks for reviewing

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -13,6 +13,7 @@ In this chapter, we'll walk through features of [JSON](https://www.json.org/json
   * [Pretty printing](#pretty-printing)
   * [Lenient parsing](#lenient-parsing)
   * [Ignoring unknown keys](#ignoring-unknown-keys)
+    * [Ignoring specific keys](#ignoring-specific-keys)
   * [Alternative Json names](#alternative-json-names)
   * [Coercing input values](#coercing-input-values)
   * [Encoding defaults](#encoding-defaults)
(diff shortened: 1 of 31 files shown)


sandwwraith:
Hi, this is a quick reminder that I remember about your PR, but haven't thoroughly reviewed it. It's good that you added both tests and documentation! I have several questions, however: 

You mentioned #1420 as the reason for implementing this feature. However, your PR does a different thing: while #1420 suggests a per-class annotation `@JsonIgnoreProperties` that ignore all unknown keys, you suggest an annotation with `vararg keys: String` that requires providing a pre-defined set of properties names' to ignore. I'm not sure this covers the use-case fully, as one may not know the unknown keys' names in advance (because they're, well, unknown). You should open a discussion in #1420 before continuing your work to ask if this direction also fits the purpose.

Additionally, you use `descriptor.annotations.filterIsInstance<JsonIgnoreProperties>().any` on every unknown key. This likely will have some performance impact. I can't predict if it will be big or small, so I say doing some benchmarking is a must. We already have `benchmarks` module you can use. In case annotation retrieval will be too slow, there are some options: 
1. Try to perform some non-asymptotic optimizations like for `@JsonClassDiscriminator` (https://github.com/Kotlin/kotlinx.serialization/blob/93a06df4bee36dc43b7d906ca395b0ac0d3229f3/formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L82). At least it is possible to skip annotation retrieval at all when `ignoreUnknownFields` flag is already globally set.
1. Use `DescriptorSchemaCache` that is already used for features like `@JsonNames` and `JsonNamingStrategy` - see their implementations and benchmark.

sandwwraith:
So this PR doesn't cover it

(although it can be easily adapted to do it)

nmrsmn:
Thanks for the review. I totally misread the purpose of the issue.. Although it is probably an easy fix (which I'll do), it's not the functionality I need. 

I'm working with a backend that sends JSON back that contains one property (called `metadata`) that I want to ignore (since it only contains metadata about the structure of the models, which clutters the models and which we don't use / need).. I want to have my model validated except for the `metadata` part.. 

I guess I need to create my own issue for that, adapt this PR to solve #1420 and then create a new PR for solving the new issue

sandwwraith:
I'll close this for now then

PR #2243, opened 2023-03-21
Replace search.maven.org with central.sonatype.com

DRSchlaubi:
As search.maven.org got deprecated, the old url doesn't work anymore


sandwwraith:
The search.maven.org link works fine for me. (and the new one is really slow). Do you have any links to a statement that search.maven.org is deprecated?

hrach:
https://central.sonatype.org/faq/what-happened-to-search-maven-org/

sandwwraith:
I see. It's still just a redirect though, not complete deprecation

sandwwraith (approved):

DRSchlaubi:
> The search.maven.org link works fine for me. (and the new one is really slow). Do you have any links to a statement that search.maven.org is deprecated?

It always redirects to https://central.sonatype.com/?smo=true no matter where you come from, so you always land on the main page

sandwwraith:
@DRSchlaubi It doesn't redirect if you opt-out from it in the pop-up 🙃

DRSchlaubi:
Yeah that's right, however a lot of people probably won't do that

PR #2248, opened 2023-03-25
Replace `runCatching-map-getOrDefault` in caching

stevenlele:
The `runCatching-map-getOrDefault` logic is too complicated and takes a detour in mind.

A simple `try-catch` returning `true` or `false` should do the trick.


shanshin:
Thanks!

PR #2253, opened 2023-03-30
Yakovlev/wasm target

igoriakovlev:
This is review for wasm target support
THIS IS NOT INTENDED TO MERGE, ONLY FOR REVIEW PURPOSES

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -72,3 +72,7 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.Kotlin2JsCompile.class).configu
         }
     }
 }
+
+tasks.named("jvmApiBuild") {
+    it.enabled = false
+}
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -64,3 +64,11 @@ tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
 }
 
 Java9Modularity.configureJava9ModuleInfo(project)
+
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.Kotlin2JsCompile.class).configureEach {
+    if (it.name.equals("compileTestDevelopmentExecutableKotlinWasm") || it.name.equals("compileTestProductionExecutableKotlinWasm")) {
+        kotlinOptions {
+            freeCompilerArgs += "-Xwasm-enable-array-range-checks"
+        }
+    }
+}
--- a/core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
+++ b/core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
@@ -69,8 +69,8 @@ data class ArraysUmbrella(
 
 @SharedImmutable
 val umbrellaInstance = TypesUmbrella(
-    Unit, true, 10, 20, 30, 40, 50.1f, 60.1, 'A', "Str0", Attitude.POSITIVE, IntData(70),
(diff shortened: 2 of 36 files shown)


core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
```diff
@@ -69,8 +69,8 @@ data class ArraysUmbrella(
 
 @SharedImmutable
 val umbrellaInstance = TypesUmbrella(
-    Unit, true, 10, 20, 30, 40, 50.1f, 60.1, 'A', "Str0", Attitude.POSITIVE, IntData(70),
```
igoriakovlev, line 72: These fixes here just because wasm double parser slightly differs from other's targets.

formats/json-tests/build.gradle.kts
```diff
@@ -34,8 +34,8 @@ kotlin {
         val commonTest by getting {
             dependencies {
                 api(project(":kotlinx-serialization-json"))
-                api(project(":kotlinx-serialization-json-okio"))
-                implementation("com.squareup.okio:okio:${property("okio_version")}")
+                //api(project(":kotlinx-serialization-json-okio"))
```
igoriakovlev, line 37: okio for wasm is not ready yet

core/build.gradle
```diff
@@ -71,4 +71,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.Kotlin2JsCompile.class).configu
             freeCompilerArgs += "-Xwasm-enable-array-range-checks"
         }
     }
+}
+
+tasks.named("jvmApiBuild") {
```
igoriakovlev, line 76: api checks cannot be performed because of kotlin.metadata is not up to date

core/wasmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -0,0 +1,71 @@
...
+        else -> null
+    }
+
+@Suppress("DEPRECATION_ERROR")
+internal actual fun <T : Any> KClass<T>.compiledSerializerImpl(): KSerializer<T>? =
+    this.constructSerializerForGivenTypeArgs()
+
+
+internal actual fun <T> createCache(factory: (KClass<*>) -> KSerializer<T>?): SerializerCache<T> {
+    return object: SerializerCache<T> {
```
sandwwraith, line 45: Is WASM single-threaded? If not, we may need some kind of concurrent solution here

igoriakovlev: Yes, Wasm has the same single-threaded model like JS (at least for now)

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
igoriakovlev: Yes, we hope to fix it later.

core/build.gradle
sandwwraith: You can also try to pick this commit: https://github.com/Kotlin/kotlinx.serialization/commit/8c1a81e7a701964eb38b30592e4fd339386852af

igoriakovlev: AFAIR, There is the problem with ABI changed when kotlin 1.9 emits new method `entries` for enums. So the test data should be fixed or this feature should be disabled.

sandwwraith:
@igoriakovlev Can this be closed? If you want a new preview/review, you can rebase your branch on latest dev

sandwwraith:
Superseded by #2410

PR #2255, opened 2023-04-02
Properties Format: Support wrapped sealed classes

EdwarDDay:
The properties format didn't support serialization of nested sealed classes before. This should be fixed with these changes.

### DIFF
--- a/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
+++ b/formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
@@ -58,6 +58,7 @@ public sealed class Properties(
             if (serializer is AbstractPolymorphicSerializer<*>) {
                 val casted = serializer as AbstractPolymorphicSerializer<Any>
                 val actualSerializer = casted.findPolymorphicSerializer(this, value as Any)
+                encodeTaggedString(nested("type"), actualSerializer.descriptor.serialName)
 
                 return actualSerializer.serialize(this, value)
             }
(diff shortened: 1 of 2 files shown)


formats/properties/commonMain/src/kotlinx/serialization/properties/Properties.kt
```diff
@@ -102,7 +103,7 @@ public sealed class Properties(
         }
 
         final override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
-            val type = map["type"]?.toString()
+            val type = map[nested("type")]?.toString()
```
daniel-jasinski, line 106: I would put this statement under the if condition below. Most types are not polymorphic, and here we will have an unnecessary creation of a nested key.

EdwarDDay: Good idea, thank you, I changed it.

sandwwraith:
Hi, thanks for the PR! Can you please elaborate on what 'wrapped' means? You used lists in test, but it seems change should work also with sealed classes contained in other classes, as in the test in here: https://github.com/Kotlin/kotlinx.serialization/pull/2262/files#diff-bec5521c216f61f12d3d730fecbe158051faf72330a89ec11de2fc6f90b85e16R54

Can you please also add this test to a test set and describe 'wrapped' in more detail in commit message? Thank you!

EdwarDDay:
Yeah, I can do this. Maybe "nested" would be a better wording then "wrapped". But I just wanted to express properties with a sealed class type - or in the case of lists or maps it would be the sealed class type as the type parameter.

sandwwraith (approved):
I think it will work with regular abstract classes (given proper polymorphic module), but I can edit commit message myself.

Thanks for your contribution!

sandwwraith:
Next time, please open the PR against `dev` branch.

PR #2262, opened 2023-04-04
Fix polymorphic serialization in Properties format

daniel-jasinski:
When encoding polymorphic types in Properties format there is a missing "type" descriptor for the concrete class being encoded.
When decoding the "type" descriptor was only correctly read for top-level polymorphic deserialization.


EdwarDDay:
Looks like we fixed the same thing https://github.com/Kotlin/kotlinx.serialization/pull/2255 XD

daniel-jasinski:
@EdwarDDay "No force on earth can stop an idea whose time has come" ;)

daniel-jasinski:
@EdwarDDay Our PRs are essentially the same so I am closing mine. I just added a comment with a minor tweak on your PR.

PR #2268, opened 2023-04-11
Fix incorrect json decoding iterator's .hasNext() behavior on array-wrapped inputs:

### LINKED ISSUE #2267: When using iterator from decodeToSequence, iterator.hasNext() throws when read all items and more
When using iterator created from Json.decodeToSequence<T> method and reading past all items, iterator.hasNext() will throw.
See the repro case where there are two elements in the list, while code tries to read 3 of them.

**To Reproduce**
```kotlin
val source = "[1, 2]"
val input = source.byteInputStream()
input.use {
    val iterator = Json.decodeToSequence<Int>(it).iterator()
    val values = iterator.asSequence().take(3).toList()
    var b = iterator.hasNext()  // throws
}
```

Exception is something like: Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Expected end of the array ']', but had 'EOF' instead at path: $

**Expected result**

`iterator.hasNext()` returns `false`.

**Environment**
 - Kotlin version: 1.8.0
 - Library version: 2.2.4
 - Kotlin platforms: JVM
 - Gradle version: 7.4.2

sandwwraith:
hasNext() shouldn't throw an exception if it was called more than once after the stream has ended.

Fixes #2267

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
@@ -68,7 +68,7 @@ private class JsonIteratorArrayWrapped<T>(
     private val deserializer: DeserializationStrategy<T>
 ) : Iterator<T> {
     private var first = true
-    private var sawClosingBrace = false
+    private var finished = false
 
     override fun next(): T {
         if (first) {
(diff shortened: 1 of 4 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
```diff
@@ -84,6 +85,7 @@ private class JsonIteratorArrayWrapped<T>(
      */
     override fun hasNext(): Boolean {
         if (lexer.peekNextToken() == TC_END_LIST) {
+            sawClosingBrace = true
```
shanshin, line 88: Perhaps it would be better to add the `finished` flag to check it at the very beginning of the `hasNext` function? like `if (finished) return false`

The difference is not fundamental, but it may be a little easier to read.

shanshin (approved):

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonIterator.kt
sandwwraith: Yeah, that makes sense

PR #2272, opened 2023-04-14
KeyValueSerializer: Fix missing call to endStructure()

pavel-vasin:
The sequential decoding path was missing it. Use decodeStructure() to avoid this error.


pavel-vasin:
Rebased and added a test.

sandwwraith (approved):
Thank you!

PR #2274, opened 2023-04-16
Fix memory leak caused by invalid KTypeWrapper equality

jooohn:
`KTypeWrapper`'s `equals` method was not implemented correctly.

This bug causes a cache miss every time [here](https://github.com/Kotlin/kotlinx.serialization/blob/99496a8cd20fec74d0dabbb2698d48795b405cc5/core/jvmMain/src/kotlinx/serialization/internal/Caching.kt#L196-L197), which leads to a memory leak by registering a new serialiser every time it is looked up.


shanshin (approved):
Great job, thank you for the investigation!

PR #2277, opened 2023-04-17
@OptIn for experimental WeakReference

qurbonzoda:
Native `WeakReference` was made experimental in an effort to stabilize Native stdlib. As a result UserProjects build of the `kotlinx.serialization` started failing.


sandwwraith:
Why is this closed?

woainikk:
it's closed because this changes won't work for develop branch, so we put commit only into kotlin-community/dev branch

PR #2291, opened 2023-05-02
Implemented obtainment of KSerializer by KClass in SerializersModule

### LINKED ISSUE #2025: Symmetry in serializer API lookup
Kotlin serialization already has these APIs
 * `SerializersModule.getContextual(KClass<T>, List<KSerializer<*>>)`
 * `SerializersModule.serializer(KType)`

I’d like this API:
 * `SerializersModule.serializer(KClass<T>, List<KSerializer<*>>)`

In particular I have a root type (like `LinkedHashSet` or `Pair` or `Map.Entry`) and serializers for the parameter types, and I want a serializer for the composed type.

shanshin:
Resolves #2025

The limitations of this API are the inability to implement stable caching, because serialization runtime does not control the equals function of the received parameters serializers, which can cause a memory leak.

Also, a technical limitation is the inability to create an array serializer.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -71,6 +71,24 @@ public inline fun <reified T> SerializersModule.serializer(): KSerializer<T> {
  */
 public fun serializer(type: KType): KSerializer<Any?> = EmptySerializersModule().serializer(type)
 
+
+/**
+ * Retrieves serializer for the given [type].
+ * **Retrieving an array serializer is not supported**
+ *
+ * If [type] is a parametrized type then in [typeArgumentsSerializers] it is necessary to pass serializers for generic parameters.
+ *
+ * The nullability of returned serializer is specified using the [isNullable].
+ *
+ * Caching on JVM platform is disabled for this function, so it may work slower than overloading with [KType].
+ *
+ * @throws SerializationException if serializer cannot be created (provided [type] or its type argument is not serializable).
+ * @throws SerializationException if [type] is an array class
+ * @throws Exception if size of [typeArgumentsSerializers] does not match the expected generic parameters count
+ */
+@ExperimentalSerializationApi
+public fun serializer(type: KClass<*>, typeArgumentsSerializers: List<KSerializer<*>>, isNullable: Boolean): KSerializer<Any?> = EmptySerializersModule().serializer(type, typeArgumentsSerializers, isNullable)
+
 /**
  * Creates a serializer for the given [type] if possible.
  * [type] argument is usually obtained with [typeOf] method.
@@ -109,12 +127,29 @@ public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
         .platformSpecificSerializerNotRegistered()
 
 
+/**
+ * Retrieves serializer for the given [type] and,
+ * if [type] is not serializable, fallbacks to [contextual][SerializersModule.getContextual] lookup.
+ *
+ * **Retrieving an array serializer is not supported**
+ *
+ * If [type] is a parametrized type then in [typeArgumentsSerializers] it is necessary to pass serializers for generic parameters.
+ *
+ * The nullability of returned serializer is specified using the [isNullable].
+ *
+ * Caching on JVM platform is disabled for this function, so it may work slower than overloading with [KType].
+ *
+ * @throws SerializationException if serializer cannot be created (provided [type] or its type argument is not serializable and is not registered in [this] module).
+ * @throws SerializationException if [type] is an array class
+ * @throws Exception if size of [typeArgumentsSerializers] does not match the expected generic parameters count
+ */
+@ExperimentalSerializationApi
 public fun <T : Any> SerializersModule.serializer(
     type: KClass<T>,
-    serializers: List<KSerializer<*>>,
+    typeArgumentsSerializers: List<KSerializer<*>>,
     isNullable: Boolean
 ): KSerializer<Any?> =
-    serializerByKClassImpl(type as KClass<Any>, serializers as List<KSerializer<Any?>>, isNullable) ?: type
+    serializerByKClassImpl(type as KClass<Any>, typeArgumentsSerializers as List<KSerializer<Any?>>, isNullable) ?: type
         .platformSpecificSerializerNotRegistered()
 
 /**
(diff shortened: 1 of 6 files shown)


core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -244,8 +274,8 @@ private fun KClass<Any>.compiledParametrizedSerializer(serializers: List<KSerial
 
 @OptIn(ExperimentalSerializationApi::class)
 private fun KClass<Any>.builtinParametrizedSerializer(
-    typeArguments: List<KType>,
     serializers: List<KSerializer<Any?>>,
+    classifierProvider: (Int) -> KClassifier?
```
sandwwraith, line 278: Given that this int is always `0`, maybe rename this parameter to `elementClassifierIfArray` and remove Int lambda parameter?

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -156,14 +165,35 @@ private fun SerializersModule.serializerByKTypeImpl(
...
+private fun SerializersModule.serializerByKClassImpl(
+    rootClass: KClass<Any>,
+    typeSerializers: List<KSerializer<Any?>>,
+    isNullable: Boolean
+): KSerializer<Any?>? {
+
+    val serializer = if (typeSerializers.isEmpty()) {
+        rootClass.serializerOrNull() ?: getContextual(rootClass)
+    } else {
+        rootClass.parametrizedSerializerOrNull(typeSerializers) { throw SerializationException("Array serializer can not be create from KClass arguments") }
```
sandwwraith, line 187: This error message is cryptic: it doesn't explain what is 'KClass arguments', as we have only one KClass<Any>; nor it does explain how to fix it

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -108,6 +108,15 @@ public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
     serializerByKTypeImpl(type, failOnMissingTypeArgSerializer = true) ?: type.kclass()
         .platformSpecificSerializerNotRegistered()
 
+
+public fun <T : Any> SerializersModule.serializer(
```
sandwwraith, line 112: * KDoc?
* @ExperimentalSerializationApi?
* Overload without receiver parameter?

Also, `serializers` parameter name looks generic. I suggest being more specific, like `typeArgumentsSerializers`.

sandwwraith (changes requested):

core/commonTest/src/kotlinx/serialization/SerializersModuleTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    fun testUnsupportedArray() {
+        assertFails {
+            serializer(Array::class, listOf(Int.serializer()), false)
+        }
+    }
+
+    @Test
+    fun testContextual() {
```
sandwwraith, line 99: Test for parameterized contextual is missing

core/commonTest/src/kotlinx/serialization/SerializersModuleTest.kt
```diff
@@ -0,0 +1,116 @@
...
+        val m = SerializersModule {
+            contextual<FileContextualType>(FileContextualSerializer)
```
sandwwraith, line 101: It doesn't matter if there is `@file:UseContextualSerialization` or not, because it is a reflective lookup unaffected by pluing. It will always try to search something in the context

shanshin: This is exactly a test for runtime, that there is a visit to this search branch. Perhaps it is redundant

sandwwraith: I mean, there should be no difference in code path for `FileContextualType` and `ContextualType`

core/commonTest/src/kotlinx/serialization/SerializersModuleTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    @Test
+    fun testContextual() {
+        val m = SerializersModule {
+            contextual<FileContextualType>(FileContextualSerializer)
+            contextual<ContextualType>(ContextualSerializer)
+            contextual<ParametrizedContextual<*>>(ParametrizedContextualSerializer as KSerializer<ParametrizedContextual<*>>)
```
sandwwraith, line 109: there's `contextual` that accepts lambda and is able to pass type arguments' serializers: https://github.com/Kotlin/kotlinx.serialization/blob/c6025c9fb6ed517169b4a5516dcd210bc2d7576a/core/commonTest/src/kotlinx/serialization/modules/ContextualGenericsTest.kt#L48

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -108,6 +126,32 @@ public fun SerializersModule.serializer(type: KType): KSerializer<Any?> =
...
+ * @throws SerializationException if serializer cannot be created (provided [type] or its type argument is not serializable and is not registered in [this] module).
+ * @throws SerializationException if [type] is an array class
+ * @throws Exception if size of [typeArgumentsSerializers] does not match the expected generic parameters count
+ */
+@ExperimentalSerializationApi
+public fun <T : Any> SerializersModule.serializer(
+    type: KClass<T>,
+    typeArgumentsSerializers: List<KSerializer<*>>,
+    isNullable: Boolean
+): KSerializer<Any?> =
```
sandwwraith, line 160: By the way, why return type is `KSerializer<Any>` and not `KSerializer<T>`?

sandwwraith (changes requested):
.

core/commonMain/src/kotlinx/serialization/Serializers.kt
shanshin: Made only because of the uniformity with the functions above.

At the same time, I don't think it will be of much use, because unlike `typeOf<...>()`, `*` will be used as parameters and for `serializer(Parametrized::class, listOf(Int.serializer()), false)` it will still be uncomfortable.

sandwwraith: KClass in overload without module should still have T parameter imho, for consistency

sandwwraith: I think it makes sense to leave return type as `KSerializer<Any>` because T may not have information about `isNullable` and `typeParametersSerializers`

Issue #1561, opened 2021-06-18
It's impossible to parse non-string keys (e.g. see CBOR Web Tokens)


wkornewald:
**Describe the bug**

According to the RFC, CBOR Web Tokens use a map with integer keys:

https://datatracker.ietf.org/doc/html/rfc8392

Currently, there is no way to parse such data. Ideally, `@SerialName` would support arbitrary keys (integer, string, etc.), but even parsing into a `Map` didn't seem to work because the implementation assumes that keys can only be strings.

**Environment**
 - Kotlin version: 1.5.10
 - Library version: 1.2.1

JesusMcCloud:
see  #2412

sandwwraith:
I suppose it is supported by now via `@CborLabel`

PR #2298, opened 2023-05-12
CBOR: Support maps that have integers as keys

Gyurme-Mattr:
As mentioned in https://github.com/Kotlin/kotlinx.serialization/issues/1561,

Standards documents like those for CBOR web tokens (https://datatracker.ietf.org/doc/html/rfc8392) have CBORs with maps that have integer keys.

I'd like to add support for integer element names in CBOR maps

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
@@ -249,7 +249,7 @@ internal open class CborReader(private val cbor: Cbor, protected val decoder: Cb
             val knownIndex: Int
             while (true) {
                 if (isDone()) return CompositeDecoder.DECODE_DONE
-                val elemName = decoder.nextString()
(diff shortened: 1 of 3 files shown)


sandwwraith:
The PR overall looks fine, but I'm reluctant to add it because it makes encoding/decoding asymmetric. We do not have any ways to specify non-string serial names (there was an issue for that, but I can't find it), so you use `@SerialName("0")` to denote int serial names and parse them accordingly. However, it is now impossible to write such data back, because `@SerialName("0")` on writing will insert the string "0", not the number. I'm not sure this behavior is acceptable for you and it will definitely be a surprising one who doesn't know about this implementation detail. What I can suggest is to hide this behavior under feature flag in Cbor configuration. The question is, should it enable only parsing strings as numbers during deserialization or should it affect serialization as well. Or maybe it should be an additional annotation like `@ByteString` instead (although two annotations is already quite verbose)

sandwwraith:
I'll close this for now; feel free to reopen when you'll have a proposed solutions for problems I've described

PR #2303, opened 2023-05-14
Introduce new StringProperties object that can encode and decode properties from a string

petitcl:
Introduce a new StringProperties object that can encode and decode properties from a string format. This new object is heavily based on the existing kotlinx.serialization `Properties` object.
- for encoding, it first encodes the input Kotlin object a string map using `Properties` and them formats that string map to a properties file. Some parameters can be passed to customise the encoding format, such as key/value separator, number of spaces before and after the key/value separator, and line separator. 
- for decoding, it first parses the input data as a string map and then decodes it into a Kotlin object using `Properties`. The parsers handles comments, whitespaces, escapes, natural lines vs logical lines. The intent is to support the same format as the Java Properties API ([doc](https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html)). For now, unicode characters are not implemented.

The implementation has been inspired a lot from the [gist](https://gist.github.com/RaphaelTarita/748e02c06574b20c25ab96c87235096d) made by https://github.com/RaphaelTarita linked in this [issue](https://github.com/Kotlin/kotlinx.serialization/issues/1152). I have added many unit tests that show how it is used and how it behaves.

This PR should fix https://github.com/Kotlin/kotlinx.serialization/issues/1152 and [KT-42716](https://youtrack.jetbrains.com/issue/KT-42716).


formats/properties/commonMain/src/kotlinx/serialization/properties/StringProperties.kt
```diff
@@ -0,0 +1,291 @@
...
+) : SerialFormat by properties, StringFormat {
+
+    /**
+     * Encodes properties from the given [value] to a properties String using the given [serializer].
+     * `null` values are omitted from the output.
+     */
+    @ExperimentalSerializationApi
+    public override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String {
+        val map = properties.encodeToMap(serializer, value)
+        val builder = StringBuilder()
```
hfhbd, line 72: I would extract the actual output renderer into another function taking a map overload to encode a giving String map to a Properties file
```@ExperimentalSerializationApi
    public fun <T> encodeToString(map: Map<String, String>): String {
```

formats/properties/commonMain/src/kotlinx/serialization/properties/StringProperties.kt
```diff
@@ -0,0 +1,291 @@
...
+                }
+                if (parsedLine[i] !in separators) {
+                    valueBegin = i
+                    break
+                }
+            }
+
+            result[parsedLine.substring(0, keyEnd)] = parsedLine.substring(valueBegin)
+        }
+        return properties.decodeFromStringMap(deserializer, result)
```
hfhbd, line 123: Same here, I would add an overload to get the input string as `Map<String, String` too.

PR #2306, opened 2023-05-16
Fix outdated buildClassSerialDescriptor documentation

### LINKED ISSUE #1434: Improve documentation for buildClassSerialDescriptor

I'm trying to build a custom `KSerializer` for a class that has a list property:

```kotlin
@Serializable(with = ListingSerializer::class)
data class Listing(
    val after: String?,
    val before: String?,
    val posts: List<Post>
)

object ListingSerializer: KSerializer<Listing> {
    override val descriptor: SerialDescriptor = buildClassSerialDescriptor("Listing") {
        element<String>("after")
        element<String>("before")
       // what to put in here ??
    }

    override fun serialize(encoder: Encoder, value: Listing) {  ...  }
    override fun deserialize(decoder: Decoder): Listing {  ...  }
}
```

The [hand-written composite serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#hand-written-composite-serializer) section for the serializers guide says to use `buildClassSerialDescriptor`. However, it does not say that to use when using collections. The [guide for json](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md) also has a  few snippets where `buildClassSerialDescriptor` is used, but not for collections.

Checking the [documentation for buildClassSerialDescriptor](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-core/kotlinx-serialization-core/kotlinx.serialization.descriptors/build-class-serial-descriptor.html) itself, I'm quite confused because it doesn't show `buildClassSerialDescriptor` but instead `SerialDescriptor`. Am I supposed to use `SerialDescriptor` instead? That's not what it's in the guide, so I'm a bit confused.

Inside it are these three lines:

```kotlin
    element<Long>("_longField") // longField is named as _longField
    element("stringField", listDescriptor<String>())
    element("nullableInt", descriptor<Int>().nullable)
```

This seems similar to what the guide says about `buildClassSerialDescriptor`, but calling them inside it does not work for `listDescriptor` or `descriptor<Int>().nullable`.

In the documentation for `buildClassSerialDescriptor`, there also an example for generic classes. In it, there's `listSerialDescriptor` which is what I used for my case.

```kotlin
override val descriptor: SerialDescriptor = buildClassSerialDescriptor("Listing") {
    element<String>("after")
    element<String>("before")

    // is this better?
    element("things", listSerialDescriptor(Post.serializer().descriptor))
    // or is this better?
    element("things", listSerialDescriptor<Post>())
}
```

My code works great! Although it says it's "experimental" so I'm not sure if it's the recommended way to do it.

I think it would be pretty cool if the guide had an example for a descriptor using a collection, or if the documentation for `buildClassSerialDescriptor` looked a bit more like how it's used in the guide (and with a collection example).

**Environment**
 - Kotlin version: 1.4.32
 - Library version: 1.4.32
 - Kotlin platforms: JVM
 - Gradle version: 6.7
 - IDE version: 2021.1
 - Other relevant context: Java 11, Linux 5.10

sandwwraith:
Fixes #1434

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -5,6 +5,7 @@
 package kotlinx.serialization.descriptors
 
 import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
 import kotlinx.serialization.encoding.*
 import kotlinx.serialization.internal.*
 import kotlin.reflect.*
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -24,22 +25,24 @@ import kotlin.reflect.*
  *     val nullableInt: Int?
  * )
  * // Descriptor for such class:
- * SerialDescriptor("my.package.Data") {
+ * buildClassSerialDescriptor("my.package.Data") {
  *     // intField is deliberately ignored by serializer -- not present in the descriptor as well
  *     element<Long>("_longField") // longField is named as _longField
- *     element("stringField", listSerialDescriptor<String>())
+ *     element("stringField", listSerialDescriptor<String>()) // or ListSerializer(String.serializer()).descriptor
```
qwwdfsad, line 31: By the way, are those intrinsified?

sandwwraith: Given they use `serializer<T>()`, this is likely to happen

PR #2307, opened 2023-05-18
Update releasing docs for the API reference

IgnatBeresnev:
The changes are based on my limited knowledge of how it works, so take it with a grain of salt

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -41,10 +41,7 @@ If review is not required, commit directly to `dev`.
    * Close the repository and wait for it to verify.
    * Release it.
    
-5. Propose the website documentation update:<br>
-    * Set a new value for [`KOTLINX_SERIALIZATION_RELEASE_TAG`](https://github.com/JetBrains/kotlin-web-site/blob/master/.teamcity/BuildParams.kt), creating a pull request in the website's repository.
-    * The website team will be notified about the pull request, review your changes, and merge it to master after all checks pass.
-    * Once the pull request is merged to the main branch, it automatically will trigger the website's update, which will be done within an hour.   
+5. Set a new value for [`KOTLINX_SERIALIZATION_RELEASE_TAG`](https://github.com/JetBrains/kotlin-web-site/blob/master/.teamcity/BuildParams.kt), creating a pull request in the website's repository.
 
 6. Create a new release in [Github releases](https://github.com/Kotlin/kotlinx.serialization/releases). Use created git tag for title and changelog message for body.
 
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -56,3 +53,12 @@ If review is not required, commit directly to `dev`.
    ```
 
 5. Announce new release in [Slack](https://kotlinlang.slack.com).
+
+# API reference documentation
```
IgnatBeresnev, line 57: Could add a link to the [TeamCity build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinSites_KotlinlangTeamcityDsl_KotlinxSerializationBuildApiReference) responsible for this, but it's the internal `buildserver` one, so not sure if it'll be of any use.

Could also add a link to the [configuration of the build trigger](https://github.com/JetBrains/kotlin-web-site/blob/fadd83777b02ea183dfd3bf668abb8fbb0653280/.teamcity/builds/apiReferences/vcsRoots/KotlinxSerialization.kt#L5), but again not sure if it'll be of any use.

Tell me if you think it's worth mentioning either one or both, I will.

nikpachoo (approved):

RELEASING.md
nikpachoo: I don't mind sharing the link to the TeamCity build, it's up to you.

RELEASING.md
```diff
@@ -56,3 +53,12 @@ If review is not required, commit directly to `dev`.
...
+
+The [API reference documentation](https://kotlinlang.org/api/kotlinx.serialization/) is built and deployed automatically
+for every commit in `master`, typically within the same day.
```
sandwwraith, line 60: Please add link to TC so team members can easily find it.

Also, you haven't explained the relation between `master` builds and tag. I think you need to include sentence like 'Despite automatic builds, it is still important to update `KOTLINX_SERIALIZATION_RELEASE_TAG` because it is used to display a version number on the website.'

IgnatBeresnev: At the time of writing it, I didn't actually know why it was needed :sweat_smile:  Tried to capture what I found out :+1:

sandwwraith:
It is OK to merge it to master since it is a documentation change. Build is indeed flaky

nikpachoo:
>  mentioned that he tried to remove the tag, but there were complications.

[@IgnatBeresnev](https://github.com/IgnatBeresnev) [@sandwwraith](https://github.com/sandwwraith), let me explain it, and maybe we can find a solution.

The issue is the following:
- Dokka shows the version in the header based on the version parameter from the gradle.properties
- in the master branch of kotlinx.serialization the version is 1.5.2-SNAPSHOT
- When you build API Reference the version in the header is 1.5.2-SNAPSHOT
- The released version is 1.5.1, and we fixed the version parameter while building on CI.

We use this workaround to fix the inconsistency of the version, but we can remove it, if we find a better way.

PR #2311, opened 2023-05-19
Rename Serializers.kt to SerializersJs.kt to support JS per-file compilation

JSMonk:


sandwwraith:
Can you please describe in more details how JS per-file compilation works and why same filenames are a problem? I want to include this info into the commit message

qwwdfsad:
Also, please ensure that Klib compatibility is not broken

JSMonk:
With the per-file compilation we use near the same approach as JVM target: we compiler the file in path of its package name + file name, so, if we had a `Foo.kt` file under the `org.bar.baz` package, the compiler result of this file will be placed in the build directory by `org/bar/baz/Foo.mjs` path.
I met a clash between two serializer files while I was compiling the space, so there we are.

sandwwraith:
Sorry I didn't notice this in time, can you rebase it on `dev` please?

PR #2314, opened 2023-05-25
Fix typo

Vinccool96:
Fixed a small typo.

In the documentation of `JsonElementSerializer`, we could read "It can only be used by with Json format **an** its input". Replaced it to "It can only be used by with [Json] format **and** its input".


sandwwraith (approved):
It is internal anyway, but thanks

Issue #1931, opened 2022-05-09
Library of JDK type serializers?


mikehearn:
Are there any libraries, community or otherwise, that add serializers for common JDK types like Date, Instant, PublicKey, URI, URL etc? It feels like one should exist by now but there's no mention of such an effort in the docs as far as I can tell.

pdvrieze:
@mikehearn I'd be happy to contribute to one. I might even be interested in seeing how such a library could be made to "autoregister"

sandwwraith:
There isn't one that I'm aware of. We had such plans in the past, but we never implemented it, see the rationale here: https://github.com/Kotlin/kotlinx.serialization/pull/350#discussion_r251471644

mikehearn:
The rationale being that people might want features it doesn't have yet?

pdvrieze:
@mikehearn another issue is that when interacting with external systems (rather than just writing reading its own data) there are different ways to serialize the same type. Especially dates are an issue for that, but other types too.

aSemy:
I'm interested in a common library that would support `BigDecimal` and `BigInteger` (when #2041 is released).

I think it would really help with using KxS, as the learning curve is *very* steep at the beginning. Serialising any non-standard type requires quite a lot of investment to understand how to implement custom serialisers. 

The point raised in https://github.com/Kotlin/kotlinx.serialization/pull/350#discussion_r251471644 (how should `java.io.File` be handled?) could be handled by using the recently added [typealias serializaton method](https://github.com/Kotlin/kotlinx.serialization/blob/1.4.1-release/docs/serializers.md#specifying-serializer-globally-using-typealias)

```kotlin
import kotlinx.serialization.*
import kotlinx.serialization.descriptors.*
import kotlinx.serialization.encoding.*
import java.io.File

/** Encodes a File as a string, using [invariantSeparatorsPath] */
typealias FileAsInvariantPath = @Serializable(with = FileAsInvariantStringSerializer::class) File


object FileAsInvariantStringSerializer : KSerializer<File> {
    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("java.io.File", PrimitiveKind.STRING)

    override fun deserialize(decoder: Decoder): File = File(decoder.decodeString())

    override fun serialize(encoder: Encoder, value: File) = encoder.encodeString(value.invariantSeparatorsPath)
}
```

[...]

PR #2321, opened 2023-06-04
Added LocalDateTimeToStringSerializer

kchung1995:
Added serializer for `java.time.LocalDateTime` to `String`.

I have read some discussions from [issues: Library of JDK type serializers?](https://github.com/Kotlin/kotlinx.serialization/issues/1931) and thought `LocalDateTime` is a good JDK type to start from.
Kotlin itself has `Instant` for expressing a specific instant of time, but JDK's `LocalDateTime`is also widely used.
It would be a good idea to include a serializer for `LocalDateTime`, and any other widely used JDK types.

### DIFF
--- a/core/jvmMain/src/kotlinx/serialization/builtins/LocalDateTimeAsStringSerializer.kt
+++ b/core/jvmMain/src/kotlinx/serialization/builtins/LocalDateTimeAsStringSerializer.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.builtins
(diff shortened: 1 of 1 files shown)


sandwwraith:
Kotlin stdlib itself doesn't have `Instant`, only `Duration`. `Instant` is from `java.time` package (which is available from Java8 and will require additional desugaring setup to work on Android). Not all people use `java.time`; moreover there is `kotlinx.datetime` that provides a lot of serializers out of the box. It will also leave questions like 'why do you have serializer for LocalDateTime, but not for ZonedDateTime'. 

What can I suggest you (or anyone stumbling across this issue) is to create a separate library with serializers for whole `java.time` API. We can ask it to the list of community-supported formats then to increase its visibility.

PR #2322, opened 2023-06-05
[Only For Community Branch] 1.9.20-migration: Remove deprecated K/N targets

projedi:
See https://kotl.in/native-targets-tiers for details.

### DIFF
--- a/gradle/configure-source-sets.gradle
+++ b/gradle/configure-source-sets.gradle
@@ -102,7 +102,7 @@ kotlin {
         }
     }
 
-    def targetsWithoutTestRunners = ["linuxArm32Hfp", "linuxArm64", "mingwX86"]
+    def targetsWithoutTestRunners = ["linuxArm64"]
     configure(targets) {
         // Configure additional binaries to run tests in the background
         if (["macos", "linux", "mingw"].any { name.startsWith(it) && !targetsWithoutTestRunners.contains(name) }) {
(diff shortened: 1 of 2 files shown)


woainikk (approved):

sandwwraith (approved):

qwwdfsad:
It would be nice to target develop for such commits, to reduce the maintenance burden in the future.
YOu can `if` that by Kotlin version from KGP, for example

woainikk:
We discussed it before with @sandwwraith and agreed not to put it into develop now. 
@sandwwraith WDYT?
> It would be nice to target develop for such commits, to reduce the maintenance burden in the future. YOu can `if` that by Kotlin version from KGP, for example

sandwwraith:
@qwwdfsad Given these targets are deprecated only in 1.9.20, we still should be able to develop for them using 1.9.0 in dev.
However, if it is possible to make an `if` on the Kotlin version, it would be more plausible.

projedi:
`if` alone wouldn't be sufficient. Identifiers themselves are now deprecated with an error, so we would have to annotate usages with `@Suppress("DEPRECATION_ERROR")`. Is that more preferable than this PR?

qwwdfsad:
In terms of maintenance, anything targeting `dev` branch is much more preferable than targeting `community-dev` branch. I believe our "Libraries Compilation fix" instruction mentions that

projedi:
Fair enough. Let's try this then: https://github.com/Kotlin/kotlinx.serialization/pull/2325

PR #2327, opened 2023-06-09
Update Dokka to 1.8.20

IgnatBeresnev:
Built the docs locally, nothing looks out of place

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,7 +13,7 @@ kotlin.version.snapshot=1.9.255-SNAPSHOT
 
 junit_version=4.12
 jackson_version=2.10.0.pr1
-dokka_version=1.8.10
+dokka_version=1.8.20
 native.deploy=
 validator_version=0.11.0
 knit_version=0.4.0


sandwwraith:
@IgnatBeresnev Can you please investigate why `knitCheck` task failed? Maybe Knit can't work with new Dokka version and needs an update

IgnatBeresnev:
Update: I started investigating the problem with knit, but couldn't understand what the problem was exactly - something strange is happening inside knit, so more time is needed to debug it. I've switched to other activities for now, but we'll return to it when we have some time available

sandwwraith:
@IgnatBeresnev I think `knit` was updated in master and dev so you may try again

IgnatBeresnev:
@sandwwraith we tried it in coroutines (https://github.com/Kotlin/kotlinx.coroutines/pull/3776#issuecomment-1679835290), but no luck so far :(

whyoleg:
Closed in favour of https://github.com/Kotlin/kotlinx.serialization/pull/2525

PR #2335, opened 2023-06-16
Formats Readme: update json5k, which now supports Kotlin/Native

aSemy:
json5k supports JVM and most native targets

https://github.com/xn32/json5k/blob/bc10a585ff5c0455fa5a7c66eb4780372c628c4b/build.gradle.kts#L21-L43


sandwwraith (approved):
Great!

PR #2341, opened 2023-06-21
Fix Kotlin user projects build

qurbonzoda:
* Removes @SharedImmutable usages
* Suppresses animal sniffer validation error

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -184,6 +184,8 @@ subprojects {
                 case "kotlinx-serialization-hocon":
                     annotationValue = "kotlinx.serialization.hocon.internal.SuppressAnimalSniffer"
                     break
+                case "kotlinx-serialization-protobuf":
+                    annotationValue = "kotlinx.serialization.protobuf.internal.SuppressAnimalSniffer"
             }
             annotation = annotationValue
         }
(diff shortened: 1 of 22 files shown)


sandwwraith:
1. Can you open PR against `dev` please?
2. Am I right that starting with Kotlin 1.8 `@SharedImmutable` does nothing anyway and can be safely removed?

qurbonzoda:
> * Can you open PR against `dev` please?

Done
> * Am I right that starting with Kotlin 1.8 `@SharedImmutable` does nothing anyway and can be safely removed?

`@SharedImmutable` does nothing in the new K/N memory manager, which is the default one since 1.7.20.
The legacy memory manager is removed completely in 1.9.20, and their is no way to switch back to it.
So the answer is yes. `@SharedImmutable` does nothing and can be safely removed.

sandwwraith:
We plan to do a release with Kotlin 1.9.0. If someone uses old memory manager, it means theirs code will break without this annotation?

qurbonzoda:
> We plan to do a release with Kotlin 1.9.0. If someone uses old memory manager

According to Native team user programs will fail at runtime if the references are used by multiple threads.

qwwdfsad:
> If someone uses old memory manager

We don't have to support it anymore

sandwwraith (approved):

PR #2344, opened 2023-06-22
Remove decoder optimization that causes invalid decoding

### LINKED ISSUE #2343: JsonTreeDecoder fails to decode poymorphic class when it contains a reference to a class with same descriptor
**Describe the bug**
When using JsonTreeDecoder with sealed classes, an optimization in  JsonTreeDecoder causes invalid decoding https://github.com/Kotlin/kotlinx.serialization/blob/5fb55ff49df65a3a57f0aca55fe890b8dbe97045/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt#L258-L265

The optimization tries to reuse the decoder, but doesn't consider that the `position` field is already at some value other than 0, and will skip json entries whose indexes are lower than the value.

**To Reproduce**
The reproducer uses a serializer similar to `JsonContentPolymorphicSerializer` to switch from StreamingJsonDecoder to JsonTreeDecoder

```kotlin
import kotlinx.serialization.KSerializer
import kotlinx.serialization.descriptors.SerialDescriptor
import kotlinx.serialization.encoding.Decoder
import kotlinx.serialization.encoding.Encoder
import kotlinx.serialization.json.JsonDecoder
import org.junit.Test
import kotlin.test.assertEquals

class JsonTreeDecoderBugTest {

    @Serializable
    sealed class ExampleSealed

    @Serializable
    data class Example(
        val a: String,
        val b: String,
        val c: Example? = null,
        val d: String?
    ) : ExampleSealed()

    val innerExample = Example(
        "InnerA",
        "InnerB",
        null,
        "InnerC"
    )
    val outerExample = Example(
        "OuterA",
        "OuterB",
        innerExample,
        "OuterC"
    )

    val expectedJson =
        """{
    "type": "com.ionspin.utils.versioning.JsonTreeDecoderBugTest.Example",
    "a": "OuterA",
    "b": "OuterB",
    "c": {
        "a": "InnerA",
        "b": "InnerB",
        "c": null,
        "d": "InnerC"
    },
    "d": "OuterC"
}"""

    val json = Json {
        prettyPrint = true
    }

    @Test
    fun streamingJson() {

        val result = json.encodeToString(outerExample as ExampleSealed)
        assertEquals(expectedJson, result)
    }

    class SwitchToTreeDecoderSerializer(
        val originalSerializer: KSerializer<ExampleSealed>
    ) : KSerializer<ExampleSealed> {
        override fun deserialize(decoder: Decoder): ExampleSealed {
            val jsonDecoder = decoder as JsonDecoder // Skip check
            val tree = jsonDecoder.decodeJsonElement()
            val decoded = jsonDecoder.json.decodeFromJsonElement(originalSerializer, tree)
            return decoded
        }

        override val descriptor: SerialDescriptor = originalSerializer.descriptor

        override fun serialize(encoder: Encoder, value: ExampleSealed) {
            // Irrelevant for the example
            TODO()
        }
    }

    @Test
    fun jsonTreeDecoder() {
        val result = json.decodeFromString(SwitchToTreeDecoderSerializer(ExampleSealed.serializer()), expectedJson)
    }
}
```

Running the `jsonTreeDecoder` causes the following exception:
```
Fields [a, b] are required for type with serial name 'com.ionspin.utils.versioning.JsonTreeDecoderBugTest.Example', but they were missing at path: $
kotlinx.serialization.MissingFieldException: Fields [a, b] are required for type with serial name 'com.ionspin.utils.versioning.JsonTreeDecoderBugTest.Example', but they were missing at path: $
	at app//kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:93)
	at app//kotlinx.serialization.json.Json.decodeFromString(Json.kt:107)
	at app//com.ionspin.utils.versioning.JsonTreeDecoderBugTest.jsonTreeDecoder(JsonTreeDecoderBugTest.kt:86)
```

**Expected behavior**

Decoding is completed successfully

**Environment**
 - Kotlin version: 1.8.20
 - Library version:1.5.1
 - Kotlin platforms: Multiplatform
 - Gradle version: 7.5

ionspin:
This pull request removes the optimization in `JsonTreeDecoder' https://github.com/Kotlin/kotlinx.serialization/blob/5fb55ff49df65a3a57f0aca55fe890b8dbe97045/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt#L258-L265 

It seems this optimization cannot work as `JsonTreeDecoder` has an inner `position` mutable state that is tied to the current object being decoded.  This fixes #2343. #2343 also explains the reasoning in more detail and provides reproducing example.

This pull request also removes the `polyDiscriminator` parameter from the `JsonTreeDecoder` constructor as it is not needed any more.

The original code with the optimization refers to issue #807 . The code example in that issue is not valid any more but with an updated reproducer, it passes when this pull request is applied.

Updated reproducer for #807 
```kotlin

    val json = Json {
        ignoreUnknownKeys = true
    }

    @Serializable
    sealed class Foo {
        @Serializable
        data class Bar(val a: Int) : Foo()
    }
    @Test
    fun testFoo() {
        val jsonElement = json.encodeToJsonElement<Foo>(Foo.Bar(1))
        println(json.decodeFromJsonElement<Foo>(jsonElement))
        println(json.decodeFromJsonElement<Foo>(jsonElement))
    }
```

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
@@ -30,7 +30,7 @@ internal fun <T> Json.readPolymorphicJson(
     element: JsonObject,
     deserializer: DeserializationStrategy<T>
 ): T {
-    return JsonTreeDecoder(this, element, discriminator, deserializer.descriptor).decodeSerializableValue(deserializer)
+    return JsonTreeDecoder(this, element, discriminator).decodeSerializableValue(deserializer)
 }
 
 private sealed class AbstractJsonTreeDecoder(
@@ -189,8 +189,7 @@ private class JsonPrimitiveDecoder(json: Json, override val value: JsonElement) 
 private open class JsonTreeDecoder(
     json: Json,
     override val value: JsonObject,
-    private val polyDiscriminator: String? = null,
-    private val polyDescriptor: SerialDescriptor? = null
+    private val polyDiscriminator: String? = null
 ) : AbstractJsonTreeDecoder(json, value) {
     private var position = 0
     private var forceNull: Boolean = false
@@ -256,11 +255,6 @@ private open class JsonTreeDecoder(
     override fun currentElement(tag: String): JsonElement = value.getValue(tag)
 
     override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
-        /*
-         * For polymorphic serialization we'd like to avoid excessive decoder creating in
-         * beginStructure to properly preserve 'polyDiscriminator' field and filter it out.
-         */
-        if (descriptor === polyDescriptor) return this
         return super.beginStructure(descriptor)
     }
 


sandwwraith:
> The original code with the optimization refers to issue https://github.com/Kotlin/kotlinx.serialization/issues/807 . The code example in that issue is not valid any more but with an updated reproducer, it passes when this pull request is applied.

Unfortunately it does not, as you can see from failing tests. The updated reproducer you're suggesting uses `ignoreUnknownKeys = true`, but it will fail with `ignoreUnknownKeys = false`, because `polyDiscriminator` is needed in `endStructure` to check that there are no unknown keys. And therefore, we should preserve it somehow. I'll take a look at the original issue and investigate what can be done there.

sandwwraith:
Test for #807 is still here btw: https://github.com/Kotlin/kotlinx.serialization/blob/5e8ccad1f70a9457e0ffe6ae6b10a0bd0eaaa618/formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonDeserializePolymorphicTwiceTest.kt#L11

ionspin:
> > The original code with the optimization refers to issue #807 . The code example in that issue is not valid any more but with an updated reproducer, it passes when this pull request is applied.
> 
> Unfortunately it does not, as you can see from failing [tests](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxSerialization_Ko/4202038?expandBuildProblemsSection=true&hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandPull+Request+Details=true). The updated reproducer you're suggesting uses `ignoreUnknownKeys = true`, but it will fail with `ignoreUnknownKeys = false`, because `polyDiscriminator` is needed in `endStructure` to check that there are no unknown keys. And therefore, we should preserve it somehow. I'll take a look at the original issue and investigate what can be done there.

Ah, I did miss that, in the original issue in my project I was using ignoreUnknownKeys and when I saw the unknown keys exception ignored it as a standard operating procedure, sorry about that.

ionspin:
Would returning a new instance solve this, because it would reset the position to 0 while retaining polyDiscriminator
```kotlin
override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
        /*
         * For polymorphic serialization we'd like to avoid excessive decoder creating in
         * beginStructure to properly preserve 'polyDiscriminator' field and filter it out.
         */
        if (descriptor === polyDescriptor) return JsonTreeDecoder(
            json, value, polyDiscriminator, polyDescriptor
        )
        return super.beginStructure(descriptor)
    }
```

sandwwraith:
Yes, I think this approach might work - you can open a new PR for that. I don't see any problems with `currentObject` being protected.

PR #2346, opened 2023-06-23
Fix beginStructure in JsonTreeDecoder when inner structre descriptor is same as outer

ionspin:
Attempt 2 (previous pull request and discussion here #2344) at solving #2343. This time instead of returning the same instance of decoder (with invalid position) a new instance is created, and `polyDiscriminator` and `polyDescriptor` are passed to the new instance.

This way check for unknown keys in `endStructure` can properly filter out polymorphic discriminator (by default "type) from potential unknown keys.

Added a simple test that fails with the original code and passes with the fix, but I'm not sure what to call it or if it's in the right place.

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonTreeDecoderPolymorphicTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonTreeDecoderPolymorphicTest.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.json.polymorphic
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.*
+import kotlin.test.*
+
+class JsonTreeDecoderPolymorphicTest : JsonTestBase() {
+
+    @Serializable
+    sealed class Sealed
+
+    @Serializable
+    data class ClassContainingItself(
+        val a: String,
+        val b: String,
+        val c: ClassContainingItself? = null,
+        val d: String?
+    ) : Sealed()
+
+    val inner = ClassContainingItself(
+        "InnerA",
+        "InnerB",
+        null,
+        "InnerC"
+    )
+    val outer = ClassContainingItself(
+        "OuterA",
+        "OuterB",
+        inner,
+        "OuterC"
+    )
+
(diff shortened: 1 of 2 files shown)


formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonTreeDecoderPolymorphicTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    )
+    val outer = ClassContainingItself(
+        "OuterA",
+        "OuterB",
+        inner,
+        "OuterC"
+    )
+
+    @Test
+    fun testDecodingWhenClassContainsItself() {
```
sandwwraith, line 38: Please use `JsonTestBase` as a base class for test and function `parametrizedTest`. It helps check behavior for all possible json decoding paths, including JsonTreeDecoder (problematic one), as well as String decoder and Stream integration. You can see example of using one here: https://github.com/Kotlin/kotlinx.serialization/blob/5e8ccad1f70a9457e0ffe6ae6b10a0bd0eaaa618/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonReifiedCollectionsTest.kt#L15

sandwwraith (changes requested):
Pull requests with code changes should be opened against `dev`. You have to rebase your commit on dev and then change base in the PR (just changing base results in unrelated commits added). Thank you!

formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonTreeDecoderPolymorphicTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        "OuterA",
+        "OuterB",
+        inner,
+        "OuterC"
+    )
+
+    @Test
+    fun testDecodingWhenClassContainsItself() = parametrizedTest { jsonTestingMode ->
+        val encoded = default.encodeToString(outer as Sealed, jsonTestingMode)
+        val decoded : Sealed = Json.decodeFromString(encoded, jsonTestingMode)
```
sandwwraith, line 40:
```suggestion
        val decoded: Sealed = Json.decodeFromString(encoded, jsonTestingMode)
```

formatting

ionspin:
I also squashed the commits.

sandwwraith (approved):

sandwwraith:
Great job, thanks!

PR #2350, opened 2023-06-26
Properly fix Java 8 API compatibility (#2218)

### LINKED ISSUE #2218: java.lang.NoSuchMethodError when parsing a JSON stream on Java 8

Vampire:
Fixes #2218

There are 7 methods in `ByteBuffer` that have the problem described in #2218 and was attempted to be fixed in #2219.
Unfortunately, the fix in #2219 was incomplete as another of these 7 methods is used in the same class.
Also it could happen anytime and with similar cases, that this happens again and is only recognized very late.

With this PR, the JVM toolchains feature is used to ensure the proper API being compiled against.
As a side-effect I also improved the `module-info.java` compilation that was done "sub-optimally"
by changing configuration of the task during its execution phase, and compiling into the same
output directory as the main compile task and then moving it out in a `doLast` action,
and having a manual task dependency, ...

With the new setup, no explicit task dependencies are necessary, which are bad-practice,
the output directories are right from the start separated, and no manual task dependencies
are necessary anymore as the jar configuration and the compile task arguments both have the
necessary implicit task dependencies automatically and by using a `CommandLineArgumentProvider`
which also has the necessary implicit dependencies, there is also no configuration change at
execution time necessary anymore.

### DIFF
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -3,15 +3,22 @@
  */
 
 import org.gradle.api.*
+import org.gradle.api.file.*
+import org.gradle.api.provider.*
+import org.gradle.api.tasks.*
 import org.gradle.api.tasks.bundling.*
 import org.gradle.api.tasks.compile.*
+import org.gradle.jvm.toolchain.*
 import org.gradle.kotlin.dsl.*
+import org.gradle.language.base.plugins.LifecycleBasePlugin.*
+import org.gradle.process.*
 import org.jetbrains.kotlin.gradle.dsl.*
+import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.*
-import org.jetbrains.kotlin.gradle.plugin.mpp.pm20.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.pm20.util.*
 import org.jetbrains.kotlin.gradle.targets.jvm.*
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 import java.io.*
 
 object Java9Modularity {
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -73,39 +75,42 @@ object Java9Modularity {
...
             source(sourceFile)
-            outputs.file(targetFile)
             classpath = files()
-            destinationDirectory.set(compileTask.destinationDirectory)
-            sourceCompatibility = JavaVersion.VERSION_1_9.toString()
-            targetCompatibility = JavaVersion.VERSION_1_9.toString()
+            destinationDirectory.set(targetDirectory)
+            javaCompiler.set(
+                this@registerCompileModuleTask.the<JavaToolchainService>().compilerFor {
+                    languageVersion.set(JavaLanguageVersion.of(9))
```
ALikhachev, line 86: This makes it harder to build the project on Apple Silicon Mac machines, as the `foojay` plugin provides no JDK 9 toolchains for the Apple Silicon

Perhaps it's more convenient to use Java 11 here and Java 9 as the target.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -73,39 +75,42 @@ object Java9Modularity {
         }
     }
 
-    private fun Project.registerCompileModuleTask(taskName: String, compileTask: KotlinCompile, sourceFile: File, targetFile: File) =
+    private fun Project.registerCompileModuleTask(taskName: String, compileTask: KotlinCompile, sourceFile: File, targetDirectory: Provider<out Directory>) =
         tasks.register(taskName, JavaCompile::class) {
-            // Also add the module-info.java source file to the Kotlin compile task;
-            // the Kotlin compiler will parse and check module dependencies,
-            // but it currently won't compile to a module-info.class file.
-            compileTask.source(sourceFile)
```
ALikhachev, line 81: If you remove this line, please provide an alternative way. It was here to perform validation of the sources in accordance to the `module-info` file, and now nothing performs this validation.

ALikhachev (changes requested):

buildSrc/src/main/kotlin/Java9Modularity.kt
Vampire: That's a valid point. This is fixed now.

buildSrc/src/main/kotlin/Java9Modularity.kt
Vampire: Can you tell me what exactly that did / checked?
What kind of bad change did that capture, so what would make the build fail with it?

sandwwraith: I'm sorry we lack documentation about this. Actually, Kotlin compiler is able to read `module-info.java` file and verify it (the comment says 'the Kotlin compiler will parse and check module dependencies'). It verifies that all package names in `exports` are correct and all `requires` dependencies are present (not so sure about latter, @ALikhachev ?).
What happens here is we add `module-info.java` location to the inputs of `KotlinCompile` task so Kotlin compiler can perform this verification.

Vampire: Hm, I see.
But I'm not sure whether there is a good solution, besides having the normal compile task compile with Java 8 toolchain and having a second set of compile tasks with Java 11 toolchain and jdk-release 9 / jvm-target 9 that is hooked into the `check` task dependencies. :-/

If I just add `-Xjdk-release=1.8` as free compiler arg into the existing (pre-my-PR) setup, there come a looot of errors about classes being in modules that are not depended upon.

If with the new (with-my-PR)  setup I use JVM toolchain 17 (to reproduce the problematic bytecode eventually) and add jdk-release 1.8 / jvm-target 1.8, the correct bytecode is produced but still those errors are not verified as the module infos seem to be ignored.

If I additionally add the `module-info.java` to the `KotlinCompile` task again, I get the same wall of errors I get if I just add "jdk-release 1.8" to the existing setup.

If I remove the "jdk-release" and only keep the "jvm-target", the verifications work again, but the wrong byte-code is produced. *grml*

sandwwraith: Let me try to sum this up: the problem is, when `module-info.java` is in the `KotlinCompile` task inputs AND `jdk-release=1.8` specified to use correct Java API, then there are a lot of errors. What this errors are, exactly? I think this scenario is popular for multi-release JARs and should be somewhat supported without much setup, but I may be wrong. Judging by https://stackoverflow.com/a/55143699/5438558, to solve similar problem in Java, specifying `module-path` is enough.

sandwwraith: However, after giving it a second thought, I'm not so sure we need this verification at all. We have `JavaCompile` task for module-info anyway — it will tell us if some of dependencies are missing and/or exported incorrectly, won't it?

Vampire: No, it wont actually.

The `JavaCompile` task just compiles the `module-info.java` file.
It will complain if there is an error within, like requiring a module that is not present, or exporting a package that does not exist.
But the intended verification happens when compiling the Kotlin files JPMS aware.
So if you for example remove `exports kotlinx.serialization;` from `core`, `formats/json` compilation will fail with the verification for using classes in non-exported packages.
And if you for example remove `requires transitive kotlinx.serialization.core;` from `formats/json`, its compilation will fail with the verification for using classes without requiring the module.

This verification still has some bugs like https://youtrack.jetbrains.com/issue/KT-60583 and https://youtrack.jetbrains.com/issue/KT-60582 for example.

I've added the alternative means for the validation that @ALikhachev requested now.
It needed some work-arounds due to shortcomings in the Kotlin Gradle Plugin and in the `module-info` handling.
But all are reported and documented with code comments.
And from my tests hopefully the same verifications done before are now again done.
Thank you very much @Tapchicoma for your thorough help in getting this set up.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -42,70 +49,146 @@ object Java9Modularity {
...
+                    val targetDirectory = compileKotlinTask.destinationDirectory.map {
+                        it.dir("../${it.asFile.name}Module")
+                    }
 
                     // only configure the compilation if necessary
                     if (sourceFile != null) {
-                        // the default source set depends on this new source set
-                        defaultSourceSet.dependsOn(this)
+                        // register and wire a task to verify module-info.java content
+                        val verifyModuleTask = registerVerifyModuleTask(
```
sandwwraith, line 62: Am I right that this actually will cause the full compilation to happen twice (you've mentioned it somewhere)?
I believe it should be mentioned in the comments then

ALikhachev (approved):
LGTM, thank you.

Would you mind making a minor adjustment by adding a description (the `description` property of the task) to the verification task in addition to settings the group?

buildSrc/src/main/kotlin/Java9Modularity.kt
Vampire: Yes, when running with `check` this is correct.
During pure building it should not happen.
I added that info to the comment now.

Vampire:
> Would you mind making a minor adjustment by adding a description

Not at all, done :-)

sandwwraith:
@Vampire Thanks for all your efforts! I'm terribly sorry for not noticing it sooner, but your PR is based against `master`, while development is usually done in `dev`. At this point, dev is already updated to Kotlin 1.9.0. I've tried to put your commit on top of dev and I got a failed build:

```
> Task :kotlinx-serialization-core:compileKotlinJs FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kotlinx-serialization-core:compileKotlinJs'.
> Failed to query the value of task ':kotlinx-serialization-core:compileKotlinJs' property 'incrementalModuleInfoProvider'.
   > Could not isolate value org.jetbrains.kotlin.gradle.incremental.IncrementalModuleInfoBuildService$Parameters_Decorated@757090b2 of type IncrementalModuleInfoBuildService.Parameters
      > Cannot query the value of task ':kotlinx-serialization-cbor:verifyKotlinJvmModule' property 'ownModuleName' because it has no value available.
        The value of this property is derived from:
          - task ':kotlinx-serialization-cbor:compileKotlinJvm' property 'moduleName'
```

I'm not sure why it happens, as you're setting ownModuleName explicitly. Can you rebase your branch on `dev`, please? Maybe @Tapchicoma could help us again.

Vampire:
No problem, fixed.
I set `ownModuleName` explicitly, yes.
But I set it to `compileTask.moduleName` which is now deprecated and replaced by `compileTask.kotlinOptions.moduleName`.
The deprecated one seems to stay unset and so the `ownModuleName` also stayed unset.
I replaced that deprecated usage in both places and verified the verification still works as intended, so should be fine now hopefully.

sandwwraith (approved):

sandwwraith:
Thanks again!

PR #2356, opened 2023-07-05
Manually apply the `java` plugin

sellmair:
To support Kotlin Multiplatform 1.9.20 and forward together with the animalsniffer plugin

See:
https://github.com/xvik/gradle-animalsniffer-plugin/issues/84 https://youtrack.jetbrains.com/issue/KT-59595


sandwwraith:
Great explanation, thank you!

sandwwraith (approved):

Issue #1560, opened 2021-06-18
CBOR tagged data can't be deserialized


wkornewald:
**Describe the bug**

When you have CBOR data that is tagged, deserialization fails with an exception. With tags I'm referring to section 3.4 of the RFC:

https://datatracker.ietf.org/doc/html/rfc8949#section-3.4

In our case for the CovPass / covid vaccination certificate app some countries use tag 0 to annotate strings as being in date/time format.

The other issue is that we can't decode the COSE data format because the array is tagged with 18:

https://cose-wg.github.io/cose-spec/#rfc.appendix.C.2.1

1. By default, tags should just be ignored during deserialization.
2. We could need an annotation that a certain tag should be used for serialization and optionally also be required during deserialization.
3. For more dynamic tagging something like a special `CborTagged(tag, data)` object would be nice which allows extracting arbitrary tags (or null) during deserialization and setting arbitrary tags during serialization.

**Environment**
 - Kotlin version: 1.5.10
 - Library version: 1.2.1

Gyurme-Mattr:
It looks like the changes in https://github.com/Kotlin/kotlinx.serialization/pull/1614 allows parsing of tagged date strings which is great!

I was wondering if anything was in the works for supporting tags in the other issue mentioned, similar to tag 18 there is also tag 24 (https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.5.1) that is used to indicated embedded CBOR items that causes a similar exception

I'm currently looking at getting tag 24 ignored in my test fork but wondering if some other approach was better 🤔

wkornewald:
Not really. We need to be able to work with the tags, i.e. match them or extract them and set them.

sandwwraith:
I'll leave this open then

JesusMcCloud:
see #2412

PR #2359, opened 2023-07-10
CBOR: Basic Support for Tags

JesusMcCloud:
Introduces basic support for CBOR tags.

Basic means that no semantic checks are performed yet, so you freely use reserved tags (e.g. serialize a cornucopia of fruit and other produce using tags reserved for dates, URIs, MIME messages, negative bignums and so forth) and no error will be thrown.

* New annotations ValueTags and KeyTags to specify tags for keys and values, respectively
   *  Tags can be nested freely as specified in the annotations
   *  By default, tags will be validated upon deserialization. If no annotations are present on a property, validation is omitted. Behaviour can be customized (see below).
   *  By default, tags will be written as annotated. Behaviour can be customized (see below).
* Introduces the following configuration parameters to the Cbor class (names are self-explanatory):
  -  `writeKeyTags`
  -  `writeValueTags`
  -  `verifyKeyTags`
  - `verifyValueTags`

This is the first PR in a series of planned enhancements, but we still consider it valuable on its own (as it also affects https://github.com/Kotlin/kotlinx.serialization/issues/1560).

In the long run, we plan to support semantically correct serialization for all reserved CBOR tags. This includes separate contributions to koltinx.datetime and the native support for self-describing CBOR. The latter of which is planned to enable the deserialization of CBOR structures, while also enabling access to the original underlying raw byte-string. This is a crucial features for conveniently working with (i.e. verifying) signed data. (Which, in turn, is a requirement for a 100% pure Kotlin multiplatform COSE implementation.)
Hence, we want to use this opportunity to also collect early feedback from those who are more familiar with kotlinx.serialization than we are.

(ofc, this should be squashed prior to merging/rebasing)

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
@@ -25,10 +25,10 @@ import kotlinx.serialization.modules.*
  *                       False by default; meaning that properties with values equal to defaults will be elided.
  * @param ignoreUnknownKeys specifies if unknown CBOR elements should be ignored (skipped) when decoding.
  * @param writeKeyTags Specifies whether tags set using the [KeyTags] annotation should be written (or omitted)
- * @param writeValueTags Specifies whether tags set using the [Tagged] annotation should be written (or omitted)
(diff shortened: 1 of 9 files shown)


JesusMcCloud:
Superseded by #2412. closing this one…

PR #2364, opened 2023-07-12
Update to Kotlin 1.9.0:

sandwwraith:
- Apply @ExperimentalNativeApi on WeakReference usages
- Drop Legacy JS compiler as it is deprecated
- Update API Dump with .entries enum properties
- Update yarn.lock

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -27,7 +27,6 @@ kover.enabled=true
 kotlin.mpp.enableCompatibilityMetadataVariant=true
 kotlin.mpp.stability.nowarn=true
 
-kotlin.js.compiler=ir
 kotlin.incremental.multiplatform=true
 
 org.gradle.parallel=true
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -65,10 +65,4 @@ tasks.withType(Jar).named(kotlin.jvm().artifactsTaskName) {
     }
 }
 
-tasks.withType(Kotlin2JsCompile.class).configureEach {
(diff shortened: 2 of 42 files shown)


gradle.properties
```diff
@@ -27,7 +27,7 @@ kover.enabled=true
 kotlin.mpp.enableCompatibilityMetadataVariant=true
 kotlin.mpp.stability.nowarn=true
 
-kotlin.js.compiler=both
+kotlin.js.compiler=ir
```
JakeWharton, line 30: This can be straight-up removed, right? IR is the default and so you don't need to specify this anymore.

sandwwraith: I guess so, but there's no harm in explicitly mentioning that

shanshin (commented):
It seems to me that we should also remove functions related to legacy JS (like `noJsLegacy`, `isJsLegacy` etc), since they are no longer relevant and can only confuse

gradle.properties
qwwdfsad: I'd suggest removing it, it's really tough to keep track of all the properties later on

qwwdfsad (changes requested):
We have quite a few `legacy`-related quirks -- `isLegacyBackend`, `compileTestKotlinJsLegacy` etc. Would be nice to get rid of them.

Also, please file an issue to support SubclassOptInRequired

shanshin (approved):

sandwwraith:
@qwwdfsad

PR #2371, opened 2023-07-19
CBOR: Use Serial Labels

nodh:
Adds the feature of using serial labels, i.e. numbers, as CBOR map keys instead of the usual strings as serial labels. This is probably useful for supporting COSE structures, or anything related like COSE keys, where numbers are used extensively as map keys.

We could also apply it on top of #2359.


JesusMcCloud:
Superseded by #2412. please close this one.

PR #2372, opened 2023-07-19
Fix an example in docs/json.md

agluszak:
The `Response` example had descriptor elements swapped around


sandwwraith:
It seems you need to re-generate test data after your change (see here: https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md#running-the-knit-tool)

PR #2381, opened 2023-07-28
Added annotation for named companion objects

shanshin:
The annotation will be added to the named companion class by the compiler


core/commonTest/src/kotlinx/serialization/SerializersLookupNamedCompanionTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.internal.*
+import kotlin.reflect.*
+import kotlin.test.*
+
+class SerializersLookupNamedCompanionTest {
+    @Serializable
+    class Plain(val i: Int) {
+        @NamedCompanion
```
sandwwraith, line 20: Please also add a test without explicit annotation with assertion that would fail when compiler is actually updated (like it was with `EnumDescriptorTest.assertFullyAnnotated`)

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -64,6 +58,36 @@ internal fun <T: Any> Class<T>.constructSerializerForGivenTypeArgs(vararg args:
...
+    }
+}
+
+private fun <T: Any> Class<T>.findNamedCompanionByAnnotation(): Any? {
+    val companionClass = declaredClasses.firstOrNull { clazz ->
+        clazz.getAnnotation(NamedCompanion::class.java) != null
+    } ?: return null
+
+    return try {
+        val companion = getDeclaredField(companionClass.simpleName)
```
sandwwraith, line 83: I think this can be extracted and unified with `private fun Class<*>.companionOrNull`

sandwwraith (changes requested):

core/commonTest/src/kotlinx/serialization/SerializersLookupNamedCompanionTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.test.runSince
+import kotlin.reflect.*
+import kotlin.test.*
+
+class SerializersLookupNamedCompanionTest {
+    @Serializable
+    class Plain(val i: Int) {
+        companion object Named
```
sandwwraith, line 18: I said 'add', not 'replace' :) You need at least one test with explicit annotation to verify that all new reflective stuff is working now.

sandwwraith (changes requested):

core/commonTest/src/kotlinx/serialization/test/CompilerVersions.kt
```diff
@@ -0,0 +1,180 @@
...
+    var error: Throwable? = null
+    try {
+        test()
+    } catch (e: Throwable) {
+        error = e
+    }
+
+    if (version && platform) {
+        if (error == null) {
+            throw Exception("Exception with type '${T::class.simpleName}' expected for $args")
```
sandwwraith, line 59: I'd rather make these exceptions `AssertionError` to be clear

sandwwraith (approved):

sandwwraith (approved):

PR #2388, opened 2023-08-02
Add annotations from expect entity to actual entity. Required after KT-58551

woainikk:
[KT-58551](https://youtrack.jetbrains.com/issue/KT-58551)

### DIFF
--- a/formats/json/nativeMain/src/kotlinx/serialization/json/internal/FormatLanguage.kt
+++ b/formats/json/nativeMain/src/kotlinx/serialization/json/internal/FormatLanguage.kt
@@ -7,6 +7,16 @@ package kotlinx.serialization.json.internal;
 import kotlinx.serialization.InternalSerializationApi
 
 @InternalSerializationApi
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER,
+    AnnotationTarget.FIELD,
+    AnnotationTarget.VALUE_PARAMETER,
+    AnnotationTarget.LOCAL_VARIABLE,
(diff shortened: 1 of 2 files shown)


shanshin (approved):

formats/json/nativeMain/src/kotlinx/serialization/json/internal/FormatLanguage.kt
```diff
@@ -7,6 +7,16 @@ package kotlinx.serialization.json.internal;
...
+@Target(
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER,
+    AnnotationTarget.FIELD,
+    AnnotationTarget.VALUE_PARAMETER,
+    AnnotationTarget.LOCAL_VARIABLE,
+    AnnotationTarget.ANNOTATION_CLASS
+)
 public actual annotation class FormatLanguage(
```
sandwwraith, line 20: There's an `actual annotation class FormatLanguage` in JS source set, too

Also, what does the new rule says about `actual typealias`? We have this in JVM

sandwwraith (changes requested):
.

formats/json/nativeMain/src/kotlinx/serialization/json/internal/FormatLanguage.kt
woainikk: Fixed for Js source set.
"Annotations with retention SOURCE on expect are not checked if actual is typealias to compiled code." - and there is no compiler errors except error with FormatLanguage

sandwwraith: But `kotlin.Target` has default retention (runtime)

sandwwraith (approved):

PR #2400, opened 2023-08-05
Fix negative valued enums deserializing for protobuf

Dogacel:
[From language guide](https://protobuf.dev/programming-guides/proto2/#enum):

> Enumerator constants must be in the range of a 32-bit integer. Since enum values use [varint encoding](https://protobuf.dev/programming-guides/encoding) on the wire, negative values are inefficient and thus not recommended

Previously, I had the following error when trying to deserialize:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 4
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor.getElementAnnotations(PluginGeneratedSerialDescriptor.kt:137)
	at kotlinx.serialization.protobuf.internal.HelpersKt.extractProtoId(Helpers.kt:74)
	at kotlinx.serialization.protobuf.internal.ProtobufDecoder.findIndexByTag(ProtobufDecoding.kt:84)
	at kotlinx.serialization.protobuf.internal.ProtobufDecoder.decodeTaggedEnum(ProtobufDecoding.kt:183)
	at kotlinx.serialization.protobuf.internal.ProtobufTaggedDecoder.decodeEnum(ProtobufTaggedDecoder.kt:40)
```

Added a test case to verify that solution works.

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
@@ -80,7 +80,7 @@ internal open class ProtobufDecoder(
 
     private fun findIndexByTag(descriptor: SerialDescriptor, protoTag: Int): Int {
         // Fast-path: tags are incremental, 1-based
-        if (protoTag < descriptor.elementsCount) {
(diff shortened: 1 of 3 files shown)


sandwwraith (approved):
Thanks!

Note: It seems it will also make possible to use negative numbers for regular proto fields (which is not allowed per spec), but we can always prohibit it later.

Issue #2401, opened 2023-08-05
Protobuf repeated uint, fixed and enum serialization not working


Dogacel:
**Describe the bug**

Any field in the following message is not being able to serialize:
```proto
syntax = "proto3";

package repeateds;

message RepeatedsMessage {
  repeated uint32 repeated_uint32 = 3;
  repeated uint64 repeated_uint64 = 4;
  repeated fixed32 repeated_fixed32 = 7;
  repeated fixed64 repeated_fixed64 = 8;
  repeated sfixed32 repeated_sfixed32 = 9;
  repeated sfixed64 repeated_sfixed64 = 10;

  repeated NestedEnum repeated_nested_enum = 18;
  repeated ForeignEnum repeated_foreign_enum = 19;

  enum NestedEnum {
    FOO = 0;
    BAR = 1;
    BAZ = 2;
    NEG = -1;  // Intentionally negative.
  }
}

enum ForeignEnum {
  FOREIGN_FOO = 0;
  FOREIGN_BAR = 1;
  FOREIGN_BAZ = 2;
}

```

I think[ the documentation](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-integer-type/index.html) has some conflicting into, according to my tests uints and sfixed's work fine.

I just added `SIGNED` to sfixed and it turned out OK. But when it comes to repeated values, I think just writing
```kotlin
  @ProtoNumber(number = 5)
  @ProtoPacked
  @ProtoType(type = SIGNED)
  public val repeatedSint32: List<Int> = emptyList(),
```
does not work. I haven't tested `Map` completely yet but I think there is gonna be a similar issue there too.

**To Reproduce**

As I am running conformance tests from my public project and it contains lots of lines, here is the link to tests, you can find every offender in tests as commented in source code.

[Link to project tests](https://github.com/Dogacel/kotlinx-protobuf-gen/blob/f9c068bd39dcdefbebbe3dca60c79c1cfac6f42d/app/src/test/kotlin/dogacel/kotlinx/protobuf/gen/RepeatedTest.kt)

**Expected behavior**

All types should serialize / deserialize correctly.

**Environment**
 - Kotlin version: 1.9.0
 - Library version: 1.9.0
 - Kotlin platforms: JVM
 - Gradle version: 7

sandwwraith:
I guess it's because 
```
ProtoType(type = SIGNED)
[...]

PR #2404, opened 2023-08-07
Add protobuf conformance tests v1

Dogacel:
Includes the message type used to run conformance tests by Google.

My intention is to catch and demonstrate some serialization errors & missing features. Commented parts are proven to be non-working. Probably some tests can be improved by using generators and multiple iterations with edge cases.

A part of
- #2400 
- #2401

### DIFF
--- a/formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
+++ b/formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2017-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.protobuf.conformance
+
+import com.google.protobuf_test_messages.proto3.*
+import kotlinx.serialization.*
+import kotlinx.serialization.protobuf.*
+import kotlin.test.*
+
+@Serializable
+data class KTestMessageProto3Oneof(
+    @ProtoNumber(111) val oneofUint32: UInt? = null,
+    @ProtoNumber(112) val oneofNestedMessage: KTestMessagesProto3Message.KNestedMessage? = null,
+    @ProtoNumber(113) val oneofString: String? = null,
+    @ProtoNumber(114) val oneofBytes: ByteArray? = null,
+    @ProtoNumber(115) val oneofBool: Boolean? = null,
+    @ProtoNumber(116) val oneofUint64: ULong? = null,
+    @ProtoNumber(117) val oneofFloat: Float? = null,
+    @ProtoNumber(118) val oneofDouble: Double? = null,
+    @ProtoNumber(119) val oneofEnum: KTestMessagesProto3Enum.KNestedEnum? = null,
+) {
+    init {
+        require(
+            listOf(
+                oneofUint32,
+                oneofNestedMessage,
+                oneofString,
+                oneofBytes,
+                oneofBool,
+                oneofUint64,
+                oneofFloat,
+                oneofDouble,
+                oneofEnum,
+            ).count { it != null } == 1
+        )
+    }
+}
+
+class Proto3OneofTest {
+    @Test
+    fun default() {
+        listOf(
+            KTestMessageProto3Oneof(oneofUint32 = 150u),
+            KTestMessageProto3Oneof(oneofNestedMessage = KTestMessagesProto3Message.KNestedMessage(a = 150)),
+            KTestMessageProto3Oneof(oneofString = "150"),
+            KTestMessageProto3Oneof(oneofBytes = "150".toByteArray()),
+            KTestMessageProto3Oneof(oneofBool = true),
+            KTestMessageProto3Oneof(oneofUint64 = 150uL),
+            KTestMessageProto3Oneof(oneofFloat = 150f),
+            KTestMessageProto3Oneof(oneofDouble = 150.0),
+            KTestMessageProto3Oneof(oneofEnum = KTestMessagesProto3Enum.KNestedEnum.BAR),
+        ).forEach { message ->
+            val bytes = ProtoBuf.encodeToByteArray(message)
+            val restored = TestMessagesProto3.TestAllTypesProto3.parseFrom(bytes)
+            if (restored.hasOneofUint32()) assertEquals(message.oneofUint32, restored.oneofUint32.toUInt())
+            if (restored.hasOneofNestedMessage()) assertEquals(
+                message.oneofNestedMessage?.a,
+                restored.oneofNestedMessage?.a
+            )
+            if (restored.hasOneofString()) assertEquals(message.oneofString, restored.oneofString)
+            if (restored.hasOneofBytes()) assertContentEquals(message.oneofBytes, restored.oneofBytes.toByteArray())
+            if (restored.hasOneofBool()) assertEquals(message.oneofBool, restored.oneofBool)
+            if (restored.hasOneofUint64()) assertEquals(message.oneofUint64, restored.oneofUint64.toULong())
+            if (restored.hasOneofFloat()) assertEquals(message.oneofFloat, restored.oneofFloat)
+            if (restored.hasOneofDouble()) assertEquals(message.oneofDouble, restored.oneofDouble)
+            if (restored.hasOneofEnum()) assertEquals(message.oneofEnum?.name, restored.oneofEnum?.name)
+
+            val restoredMessage = ProtoBuf.decodeFromByteArray<KTestMessageProto3Oneof>(restored.toByteArray())
+            assertEquals(message, restoredMessage.copy(oneofBytes = message.oneofBytes))
+            assertContentEquals(message.oneofBytes, restoredMessage.oneofBytes)
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


Dogacel:
Hi team, would be glad if some review can be done, I am hoping those bugs will be resolved and protobuf module can hopefully promote to stable after all tests are fixed 🙏

shanshin:
Hi,
sorry for the delay and thanks for the PR!
We will review it in the next few days.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
```diff
@@ -0,0 +1,75 @@
...
+                message.oneofNestedMessage?.a,
+                restored.oneofNestedMessage?.a
+            )
+            if (restored.hasOneofString()) assertEquals(message.oneofString, restored.oneofString)
+            if (restored.hasOneofBytes()) assertContentEquals(message.oneofBytes, restored.oneofBytes.toByteArray())
+            if (restored.hasOneofBool()) assertEquals(message.oneofBool, restored.oneofBool)
+            if (restored.hasOneofUint64()) assertEquals(message.oneofUint64, restored.oneofUint64.toULong())
+            if (restored.hasOneofFloat()) assertEquals(message.oneofFloat, restored.oneofFloat)
+            if (restored.hasOneofDouble()) assertEquals(message.oneofDouble, restored.oneofDouble)
+            if (restored.hasOneofEnum()) assertEquals(message.oneofEnum?.name, restored.oneofEnum?.name)
```
pdvrieze, line 68: Rather than checking restored, you should probably check message

sandwwraith:
Just for the clarification of your intentions — do you want this PR to be merged with commented code? Or you want to leave it as a showcase until all issues are fixed?

Dogacel:
> Just for the clarification of your intentions — do you want this PR to be merged with commented code? Or you want to leave it as a showcase until all issues are fixed?

Hi,

My personal preference is to merge those tests, either with comments or mark them as `@Ignored` (which causes some duplicate code that's why I choose to comment). My intention to add them to tests was to showcase what's missing in the code and also make testing easier for developers.

As library maintainers, please let me know how you want to merge this. If you are against keeping comments, I think it is fine to remove them and open issues in GitHub to track. But I think we should merge the non-commented tests anyway because conformance tests are an important indicator to verify functionality.

sandwwraith:
I see. 

> But I think we should merge the non-commented tests anyway because conformance tests are an important indicator to verify functionality.

Yes, this is a great point. I suggest to do the following:

1. Carefully split test data to working and non-working parts (latter in a separate file). It looks to me that amount of duplicated code won't be that big. Non-working parts should be `@Ignore`d, not commented.
2. Create an issue(s) (if there's not one already) that describes missing parts. Issue number should be added as a comment to `@Ignore` annotation. You can create one big issue or several separate, whatever suits you best.

Does that sound plausible to you?

Dogacel:
> I see.
> 
> > But I think we should merge the non-commented tests anyway because conformance tests are an important indicator to verify functionality.
> 
> Yes, this is a great point. I suggest to do the following:
> 
> 1. Carefully split test data to working and non-working parts (latter in a separate file). It looks to me that amount of duplicated code won't be that big. Non-working parts should be `@Ignore`d, not commented.
> 2. Create an issue(s) (if there's not one already) that describes missing parts. Issue number should be added as a comment to `@Ignore` annotation. You can create one big issue or several separate, whatever suits you best.
> 
> Does that sound plausible to you?

Hi,

Yep that sounds plausible. I am currently traveling and also sick so will complete this when I have time and feel better 😆

sandwwraith:
No problems, take your time and get well!

Dogacel:
@sandwwraith Updated the tests and opened three issues:
- #2418 
- #2419 
- #2417 

I believe they all are related.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
```diff
@@ -0,0 +1,75 @@
...
+            KTestMessageProto3Oneof(oneofUint32 = 150u),
+            KTestMessageProto3Oneof(oneofNestedMessage = KTestMessagesProto3Message.KNestedMessage(a = 150)),
+            KTestMessageProto3Oneof(oneofString = "150"),
+            KTestMessageProto3Oneof(oneofBytes = "150".toByteArray()),
+            KTestMessageProto3Oneof(oneofBool = true),
+            KTestMessageProto3Oneof(oneofUint64 = 150uL),
+            KTestMessageProto3Oneof(oneofFloat = 150f),
+            KTestMessageProto3Oneof(oneofDouble = 150.0),
+            KTestMessageProto3Oneof(oneofEnum = KTestMessagesProto3Enum.KNestedEnum.BAR),
+        ).forEach { message ->
```
pdvrieze, line 54: Rather than testing all types in a single function it would make sense to have the inner loop in a separate function and just have separate test functions for the different data members. This would help with testing individual types, and brings more visibility (each actual test function would just call the actual implementation with the message as parameter). It would also allow testing each type even in case of some failure.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3PrimitiveTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        assertEquals(message.optionalSint32, restored.optionalSint32)
+        assertEquals(message.optionalSint64, restored.optionalSint64)
+        assertEquals(message.optionalFixed32, restored.optionalFixed32)
+        assertEquals(message.optionalFixed64, restored.optionalFixed64)
+        assertEquals(message.optionalSfixed32, restored.optionalSfixed32)
+        assertEquals(message.optionalSfixed64, restored.optionalSfixed64)
+        assertEquals(message.optionalFloat, restored.optionalFloat)
+        assertEquals(message.optionalDouble, restored.optionalDouble)
+        assertEquals(message.optionalBool, restored.optionalBool)
+        assertEquals(message.optionalString, restored.optionalString)
```
pdvrieze, line 68: As you are already comparing the full message below these comparisons seem double.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3PrimitiveTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        assertContentEquals(message.optionalBytes, restored.optionalBytes.toByteArray())
+
+        val restoredMessage = ProtoBuf.decodeFromByteArray<KTestMessagesProto3Primitive>(restored.toByteArray())
+        assertEquals(message, restoredMessage.copy(optionalBytes = message.optionalBytes))
```
pdvrieze, line 72: You may add a comment on what you are doing here (compensating for the fact that you didn't provide an equals implementation for the message type).

pdvrieze (commented):
Overall it looks ok to me (not an official dev). I'd say split the individual tests off into separate functions per type. Otherwise I find the code containing a lot of repetitive elements that don't help to make the code clear. Given that the types are local to the tests the API itself isn't really an issue, but perhaps the resulting verbosity isn't ideal (I don't really know how to solve the clarity issue though without adding just as much complexity in other places).

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3PrimitiveTest.kt
Dogacel: I should have compared the fields with the serialized one rather than the restored.

Sometimes message can be serialized and deserialized it might be the same but the serialization format might be different so if deserialized by another library it can be something else. This is most obvious with signed ints (i.e. it appears as 2*int - 1)

My intention was to catch this serialized format check.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
sandwwraith: I agree with that. This would allow to determine what exact instance of `KTestMessageProto3Oneof` caused failure.

Or, If you prefer a different way, you can add an assertion message to each assertion, but this looks much more boilerplate-ish.

sandwwraith (commented):
Looks great! Just fix the minor issues I've mentioned, and we can merge this.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3OneofTest.kt
Dogacel: I tried to reduce boilerplate by creating some verification functions. Let me know how it looks now.

formats/protobuf/jvmTest/src/kotlinx/serialization/protobuf/conformance/Proto3PrimitiveTest.kt
Dogacel: Yep, added where I do this trick.

Dogacel:
> Looks great! Just fix the minor issues I've mentioned, and we can merge this.

Thanks. It took me a bit of time to catch-up but fixed some stuff and ready for another round of review. Thanks!

PR #2405, opened 2023-08-08
Use JavaCompile.destinationDirectory instead of the whole task

### LINKED ISSUE #2398: DuplicateRelativeFileException
```java
Execution failed for task ':app:mergeDemoDebugJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.MergeJavaResWorkAction
   > 2 files found with path 'META-INF/versions/9/previous-compilation-data.bin' from inputs:
      - /Users/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlinx/kotlinx-serialization-core-jvm/1.6.0-RC/dd8fbc282650a6d12858820b8190919055de6bed/kotlinx-serialization-core-jvm-1.6.0-RC.jar
      - /Users/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlinx/kotlinx-serialization-json-jvm/1.6.0-RC/b6556b04359bc6ed73bc758c8b01a2ec478dbead/kotlinx-serialization-json-jvm-1.6.0-RC.jar
     Adding a packagingOptions block may help, please refer to
     https://developer.android.com/reference/tools/gradle-api/8.2/com/android/build/api/dsl/ResourcesPackagingOptions
     for more information
```

sandwwraith:
as a source directory for copying module-info.class to the multi-release jar.

This ensures that no additional files except java compiler output get into jar.

Fixes #2398

### DIFF
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -82,7 +82,7 @@ object Java9Modularity {
                         )
 
                         // add the resulting module descriptor to this target's artifact
-                        artifactTask.from(compileModuleTask) {
+                        artifactTask.from(compileModuleTask.map { it.destinationDirectory }) {
                             if (multiRelease) {
                                 into("META-INF/versions/9/")
                             }


buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -82,7 +82,7 @@ object Java9Modularity {
                         )
 
                         // add the resulting module descriptor to this target's artifact
-                        artifactTask.from(compileModuleTask) {
+                        artifactTask.from(compileModuleTask.map { it.destinationDirectory }) {
```
shanshin, line 85: Does calling `jar` tasks (artifactTask) result in calling `compileModuleTask`?
This should work, but in some versions of Gradle it sometimes worked incorrectly.

sandwwraith: I guess so, it is a default java compilation task.

shanshin: Checked locally, calling `jar` leads to running `compileKotlinJvmModule`

Issue #2399, opened 2023-08-05
Expected quotation mark '"', but had '"' instead at path


nbransby:
I did want to goggle for the exact error message before I posted an issue but I couldn't because it has quotes in it so surrounding it in quotes in google doesn't help!

Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 4054: Expected quotation mark '"', but had '"' instead at path: $.isFork
JSON input: .....07-28T21:06:29Z","updated_at":"2022-11-04T13:28
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.fail(AbstractJsonLexer.kt:584)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$default(AbstractJsonLexer.kt:582)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$kotlinx_serialization_json(AbstractJsonLexer.kt:223)
	at kotlinx.serialization.json.internal.StringJsonLexer.consumeKeyString(StringJsonLexer.kt:88)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.consumeString(AbstractJsonLexer.kt:369)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.consumeStringLenient(AbstractJsonLexer.kt:449)
	at kotlinx.serialization.json.internal.AbstractJsonLexer.skipElement(AbstractJsonLexer.kt:534)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.handleUnknown(StreamingJsonDecoder.kt:250)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeObjectIndex(StreamingJsonDecoder.kt:240)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeElementIndex(StreamingJsonDecoder.kt:175)
	at dev.gitlive.internal.service.Repository$$serializer.deserialize(Repository.kt:8)
	at dev.gitlive.internal.service.Repository$$serializer.deserialize(Repository.kt:8)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:70)
[...]

Issue #2360, opened 2023-07-10
Error on unquoted string gives incorrect position


sandwwraith:
**Describe the bug**

If key or value in Json is not quoted (e.g. one wanted to parse a data in lenient mode, but forgot to turn it on),
the position points to the end of such value, but message mentions the beginning:

**To Reproduce**
```kotlin
@Test
fun lenientIncorrect() {
        val input1 = """{boxed:str}"""
        val input2 = """{"boxed":str}"""
        val ser = serializer<Box<String>>()
        Json.decodeFromString(ser, input2)
}
```

**Expected behavior**

Parsing `input2` gives
```
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 11: Expected quotation mark '"', but had 'r' instead at path: $.boxed
JSON input: {"boxed":str}
```

although message should be 'had s instead' and position 8, to point at the beginning of the string where quotation is expected.

Parsing `input1` gives

```kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 5: Expected quotation mark '"', but had 'd' instead at path: $
JSON input: {boxed:str}
```

Same here: should have offset=1 and message 'had b instead'.

Reason: `consumeStringLenient` [here](https://github.com/Kotlin/kotlinx.serialization/blob/f76d07314ead1db92f27c389494f537c12d3ea7b/formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt#L204) advances position and doesn't roll it back.

**Environment**
 - Kotlin version: 1.8.21
 - Library version: 1.5.1
 - Kotlin platforms: JVM

PR #2406, opened 2023-08-09
Test & fix several exception messages from Json parser

sandwwraith:
To avoid cryptic and incorrect ones, such as `Expected quotation mark '"', but had '"' instead` or `unexpected token: 10`.

Fixes https://github.com/Kotlin/kotlinx.serialization/issues/2360
Fixes https://github.com/Kotlin/kotlinx.serialization/issues/2399


qwwdfsad (approved):
Nice one, I played with failure modes for a while, and it is much more approachable 👍

PR #2410, opened 2023-08-15
Implement wasm target

igoriakovlev:


build.gradle
```diff
@@ -59,7 +59,7 @@ buildscript {
     repositories {
         maven { url 'https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev' }
         // kotlin-dev with space redirector
-        maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
```
sandwwraith, line 62: This repository is required for TC aggregate and nightly builds, do not remove it.

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
```diff
@@ -68,8 +68,8 @@ data class ArraysUmbrella(
 }
 
 val umbrellaInstance = TypesUmbrella(
-    Unit, true, 10, 20, 30, 40, 50.1f, 60.1, 'A', "Str0", Attitude.POSITIVE, IntData(70),
-    null, null, 11, 21, 31, 41, 51.1f, 61.1, 'B', "Str1", Attitude.NEUTRAL, null,
+    Unit, true, 10, 20, 30, 40, 50.5f, 60.5, 'A', "Str0", Attitude.POSITIVE, IntData(70),
```
sandwwraith, line 71: I'd rather avoid changing test data if WASM has some problems with float numbers. You may add some `if`s around here if you need to.

core/wasmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -0,0 +1,71 @@
...
+
+internal actual fun <T> createParametrizedCache(factory: (KClass<Any>, List<KType>) -> KSerializer<T>?): ParametrizedSerializerCache<T> {
+    return object: ParametrizedSerializerCache<T> {
+        override fun get(key: KClass<Any>, types: List<KType>): Result<KSerializer<T>?> {
+            return kotlin.runCatching { factory(key, types) }
+        }
+    }
+}
+
+internal actual fun <T : Any, E : T?> ArrayList<E>.toNativeArrayImpl(eClass: KClass<T>): Array<E> {
```
sandwwraith, line 60: WASM doesn't have non-reified `toTypedArray()`?

formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt
```diff
@@ -6,13 +6,15 @@ package kotlinx.serialization.features
...
 import kotlinx.serialization.json.JsonTestBase
+import kotlinx.serialization.test.*
 import kotlin.test.Test
 
 
 class EmojiTest : JsonTestBase() {
 
     @Test
     fun testEmojiString() {
+        if (isWasm()) return
```
sandwwraith, line 17: Is there an issue with TODO for this test?

formats/json/build.gradle
```diff
@@ -26,6 +26,16 @@ kotlin {
                 api project(":kotlinx-serialization-core")
             }
         }
+        jsMain {
```
sandwwraith, line 29: Why `jsMain` has to be changed?

sandwwraith: Because you moved some files to share them between WASM and JS? Is KMM plugin OK with that? Does import in IDE work correctly?

integration-test/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     repositories {
         mavenLocal()
         mavenCentral()
-        maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
```
sandwwraith, line 10: Same here

sandwwraith (changes requested):

build.gradle
igoriakovlev: Yes, already fixed.

formats/json/build.gradle
igoriakovlev: I do not see any issues to do so. But this commit is not necessary and we can drop it (but we need to copy-paste this files into different targets). Or we can try make new source set and depend on it from target source sets, but then we need to ignore metadata tasks for it.

integration-test/build.gradle
igoriakovlev: Yeah, i already rolled it back. Thanks

core/wasmMain/src/kotlinx/serialization/internal/Platform.kt
igoriakovlev: It has. Fixed.

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
igoriakovlev: It is quite random numbers. The new ones is not better or worse than old ones :)
You think that it is nice to have here?

formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt
sandwwraith: Please take a look

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
sandwwraith: If they're random, why change them?

formats/json/build.gradle
igoriakovlev: Its a bug in IDE. The only thing we can do is copy-paste the code. And this is only inspection bug (the gutter works fine).

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
igoriakovlev: Agggh...because this data is not suitable to wasm due to lack of wasm stdlib implementation. If i will add another dataset only for wasm how can it make things better?

pdvrieze: If the "random" values break something we got lucky that they detect a bug. They shouldn't be changed. Rather the bug fixed, or the test "ignored" if that is best.

formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt
igoriakovlev: This seems to be a bug in okio. I keep this test ingored for wasm and put the [link](https://github.com/square/okio/issues/1333) to the issue nearby.

formats/json/build.gradle
sandwwraith: I wanna see this IDE bug fixed :) So please report it, if it is not created yet.
IDE consistency and correctness directly affects daily work, so yes, I do not want to see any red code in the IDE if it can be compiled correctly. Alternatively, it is possible to use `@Suppress("NO_ACTUAL_FOR_EXPECT")` but I'm not sure that suppressing errors is better than copy pasting several files. @qwwdfsad Maybe you have an opinion on this?

formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt
sandwwraith: Great, thank you!

core/commonTest/src/kotlinx/serialization/BasicTypesSerializationTest.kt
```diff
@@ -173,6 +173,17 @@ class BasicTypesSerializationTest {
         assertNotSame(umbrellaInstance, other)
     }
 
+    @Test
+    fun testKvSerialization() {
+        if (isWasm()) return
```
sandwwraith, line 178: Please add a comment with YouTrack issue for this wasm bug with float numbers

core/commonTest/src/kotlinx/serialization/UmbrellaTypes.kt
sandwwraith: So old ones ARE better, given that they catch bugs, aren't they? :) 

Good idea with two tests!

formats/json/build.gradle
pdvrieze: I have found that in the latest version actuals in a shared module do work, but they require correct dependencies (the shared module must depend on the common module). A case that still breaks is where you want to share code between jvm and android and use (shared) jvm types.

igoriakovlev: I see that this invalid report occurs on current `dev` branch (without wasm target added), so it seems it is not a result of my changes.

igoriakovlev: https://youtrack.jetbrains.com/issue/KTIJ-26668/False-positive-diagnostic-for-missing-actual-declararion-for-expect-object

igoriakovlev: So what do you prefer?

- Copy and Paste (does not work anyway, the diagnostic will appear)
- Directory sharing
- Sourceset dependency

sandwwraith: I'm on `dev` and I don't see this error. Did you reimport project after branch switching?

igoriakovlev: I made new project.

core/commonTest/src/kotlinx/serialization/BasicTypesSerializationTest.kt
sandwwraith: Why did you delete added tests though?

formats/json/build.gradle
sandwwraith: How it is related to serializtion then? With only jsMain, this error doesn't happen.

We'll discuss it on the next week, since now it is time to release 1.6.0 (this PR won't go there anyway)

igoriakovlev: Well, as I see, with my IDEA it does not related at all.
And I see that with any other MPP project (even without wasm).
I dont understand why it so bother you, its a bug and it is reported.

Screenshot 2023-08-17 at 15 55 17

core/commonTest/src/kotlinx/serialization/BasicTypesSerializationTest.kt
igoriakovlev: I delete only one that was originally ok. There is two similar test with similar testdata. One is passing and the second is not.

formats/json/build.gradle
igoriakovlev: I am trying understand the reasons for all of that....if tomorrow new bugs will appear in IDEA, would you rewrite your code to escape this bug or what? Or stop to getting MR's that affected with this bug in editor? What is going on? Why I am wasting my (and yours time). If you are so aware for editor problems with green code, push the IDEA but not codeowners.

formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt
swankjesse: https://youtrack.jetbrains.com/issue/KT-60212

formats/json/build.gradle
sandwwraith: It's not a stopper for the PR, indeed. I just want to find a suitable workaround.

igoriakovlev: Well, I discussed this with a guys from MPP and we seems to be found whats wrong. Its `.idea` directory that contains artefacts that infer with compiler settings. `.idea` directory does not deleted by any recoveries or cache cleaning options of IDEA, that is why diagnostics persists. You can remove `.idea` folder and this error should away.
We can also compare old `.idea` files and new ones and found the cause (i have already deleted this directory so cannot compare it.

sandwwraith: Great find, thanks! I'll take a look at .idea

sandwwraith: I don't know what changed exactly, but the error seems gone for me after switching branches. I guess we can mark it as resolved for now, applying `@Suppress` if it happens again. As I said, I'll merge this PR after 1.6.0 release (tomorrow)

igoriakovlev: Perfect. Thanks. Is it ok for you current approach with intermediate `sourceSet` and ignoring `compileJsWasmMainKotlinMetadata` task as I do for now?

sandwwraith: > ignoring compileJsWasmMainKotlinMetadata task

Can you elaborate on that? I don't see any changes in buildfiles relating to this task

formats/json/build.gradle
```diff
@@ -26,7 +26,22 @@ kotlin {
...
+        wasmMain {
+            dependsOn(sourceSets.jsWasmMain)
+        }
     }
 }
 
 Java9Modularity.configureJava9ModuleInfo(project)
+
+tasks.whenTaskAdded { task ->
+    if (task.name == 'compileJsWasmMainKotlinMetadata') {
```
sandwwraith, line 44: Ah, found it. Can you add a comment here explaining why it is needed?

igoriakovlev: Done

igoriakovlev:
I have added commit to make possible build and test with kotlin 1.9.30+

swankjesse (commented):
LGTM

sandwwraith (approved):

sandwwraith:
Great job! 💪🏻

igoriakovlev:
Thank you all for the review!

PR #2442, opened 2023-09-14
At least mention tomlkt in third party formats.

Peanuuutz:
There was a request (#1818) years ago, but was abandoned. After a lot more development, on the second thought, I'm gonna reopen this request.

At this point, there are several aspects that tomlkt goes apart from ktoml:
1. the data structure: tomlkt tries to resemble official JSON format in regards to heavy usage of TomlElement, while ktoml specifies every data type separately and kind of keeps them as internal implementation.
2. the feature set: tomlkt does implement more features either from TOML or Kotlin itself, so less workaround. Most issues posted there are solvable with tomlkt.
3. the decoding process: tomlkt consumes source file character by character and cares much about operations and memory usage, so it has finer control and IMO is more robust (see [benchmark](https://github.com/Peanuuutz/tomlkt/blob/master/benchmark/src/jmh/kotlin/test/Benchmark.kt)).


pdvrieze:
Looking at your description in the documentation it appears that the library is not a format, but rather a Kotlin binding for TOML. You should clarify how it works in terms of being a serialization format (that doesn't require use of any TOML specific data types).

Peanuuutz:
"how it works" is given in the last part - TomlElement.

sandwwraith (approved):
Thank you! I'm happy to see efforts like this going into developing of community formats

PR #2450, opened 2023-09-21
Implementation of class discriminator mode in Json

### LINKED ISSUE #1247: Allow custom policy for adding a discriminator

sandwwraith:
Split into two commits to decide if we want to merge POLYMORPHIC_AND_INHERITORS right away or postpone it until plugin changes

Fixes #1247

### DIFF
--- a/core/api/kotlinx-serialization-core.api
+++ b/core/api/kotlinx-serialization-core.api
@@ -283,12 +283,14 @@ public abstract interface class kotlinx/serialization/descriptors/SerialDescript
 	public abstract fun isElementOptional (I)Z
 	public abstract fun isInline ()Z
 	public abstract fun isNullable ()Z
+	public abstract fun isPartOfSealedHierarchy ()Z
 }
 
 public final class kotlinx/serialization/descriptors/SerialDescriptor$DefaultImpls {
 	public static fun getAnnotations (Lkotlinx/serialization/descriptors/SerialDescriptor;)Ljava/util/List;
 	public static fun isInline (Lkotlinx/serialization/descriptors/SerialDescriptor;)Z
 	public static fun isNullable (Lkotlinx/serialization/descriptors/SerialDescriptor;)Z
+	public static fun isPartOfSealedHierarchy (Lkotlinx/serialization/descriptors/SerialDescriptor;)Z
 }
 
 public final class kotlinx/serialization/descriptors/SerialDescriptorKt {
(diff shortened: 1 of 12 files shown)


pdvrieze:
I had a look. The current approach is somewhat limited, it doesn't quite allow full introspection by formats, rather it is narrowly useful. What about having a function on the descriptor that returns the descriptors (or even serializers) of the serializable parents (sealed interfaces allow multiple)? With such functionality it might be possible to deal with cases in system such as ktor where the serialization often uses the "wrong" type by automatically using the parent.

qwwdfsad:
I'll skim through public API shape, use-cases and examples. Review of the implementation is left for @shanshin 😄

sandwwraith:
Split in two parts; first one is https://github.com/Kotlin/kotlinx.serialization/pull/2532

PR #2465, opened 2023-10-05
Update MissingFieldException documentation

qwwdfsad:


core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -65,6 +66,12 @@ public open class SerializationException : IllegalArgumentException {
  * [MissingFieldException] is thrown on missing field from all [auto-generated][Serializable] serializers and it
  * is recommended to throw this exception from user-defined serializers.
  *
+ * [MissingFieldException] is constructed from the following properties:
+ * - [missingFields] -- fields that were required for the deserialization but have not been found.
+ *   They are always non-empty and their names match the corresponding names in [SerialDescriptor.elementNames]
+ * - Optional `serialName` -- serial name of the enclosing class that failed to get deserialized.
```
sandwwraith, line 72: I'm not sure this is correct because the only constructor that does not accept `serialName` is intended for plugin usage only and can accept only one missing field.

We also have a constructor with bare `message: String?`. Should we add a note that it shouldn't be used by the user?

pdvrieze: If the constructor should not be used by the user, perhaps it can be marked `Deprecated(hidden)` (or otherwise) / internal + `@PublishedAPI`.

qwwdfsad: >I'm not sure this is correct because the only constructor that does not accept serialName is intended for plugin usage only and can accept only one missing field.

The primary one as well.

>Should we add a note that it shouldn't be used by the user?

I would prefer not to; it's completely reasonable to supply your own exception message with a lot of contextual information, but not a serial name; thus the optionality

sandwwraith: If you're OK with the primary constructor with arbitrary message available to the public, then we can leave this as is.

sandwwraith (approved):

PR #2466, opened 2023-10-06
cbor: check inline value classes if marked as @ByteString (fixes #2187)

### LINKED ISSUE #2187: CBOR `@ByteString` should work with value classes

According to the [documentation](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#byte-arrays-and-cbor-data-types), CBOR's `@ByteString` annotation can be used in data classes to encode or decode a `ByteArray` as a byte string (major type 2). This is fine if you're dealing with CBOR data that can be represented as a data class, but there cases where a data class is not suitable (one of which is covered in https://github.com/Kotlin/kotlinx.serialization/issues/2037).

My use-case is to decode a single raw byte string, without any kind of wrapper. Other primitive types can be serialized directly (e.g. `Cbor.decodeFromByteArray<Int>(data)`) or using a value class:
```
@JvmInline
@Serializable
value class Value(val data: Int)
```
which can then be decoded with `Cbor.decodeFromByteArray<Value>(data)`.

This is not possible with byte strings.

I would like to be able to define a value class as such:
```
@JvmInline
@Serializable
value class Value(@ByteString val data: ByteArray)
```
Currently, if I try using this to decode the byte string `4401020304`, I get the following error: `kotlinx.serialization.cbor.internal.CborDecodingException: Expected start of array, but found 44`.

the-eater:
This fixed inline value classes not being processed as a ByteString, by checking inside `encodeSerializableValue` and `decodeSerializableValue` to see if the current descriptor is marked as inline and with `@ByteString`

this is as far as I understand the most viable way to do it

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
@@ -70,6 +70,8 @@ internal open class CborWriter(private val cbor: Cbor, protected val encoder: Cb
         if (encodeByteArrayAsByteString && serializer.descriptor == ByteArraySerializer().descriptor) {
             encoder.encodeByteString(value as ByteArray)
         } else {
+            encodeByteArrayAsByteString = encodeByteArrayAsByteString || serializer.descriptor.isInlineByteString()
+
             super.encodeSerializableValue(serializer, value)
         }
     }
@@ -278,6 +280,7 @@ internal open class CborReader(private val cbor: Cbor, protected val decoder: Cb
             @Suppress("UNCHECKED_CAST")
             decoder.nextByteString() as T
         } else {
+            decodeByteArrayAsByteString = decodeByteArrayAsByteString || deserializer.descriptor.isInlineByteString()
             super.decodeSerializableValue(deserializer)
         }
     }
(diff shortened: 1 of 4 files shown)


formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -636,6 +639,11 @@ private fun SerialDescriptor.isByteString(index: Int): Boolean {
     return getElementAnnotations(index).find { it is ByteString } != null
 }
 
+private fun SerialDescriptor.isInlineByteString(): Boolean {
+    // inline item classes should only have 1 item
+    return isInline && isByteString(0)
```
sandwwraith, line 644: Yes, this is a correct way to do it

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoding.kt
```diff
@@ -278,6 +280,7 @@ internal open class CborReader(private val cbor: Cbor, protected val decoder: Cb
             @Suppress("UNCHECKED_CAST")
             decoder.nextByteString() as T
         } else {
+            decodeByteArrayAsByteString = decodeByteArrayAsByteString || deserializer.descriptor.isInlineByteString()
```
sandwwraith, line 283: This looks a bit suspicious. If the current descriptor is not a ByteArraySerializer().descriptor and not a value class over `@ByteString ByteArray`, why we should still try to decode it as ByteArray? 

If this is to support cases when value class property is not annotated itself, like this:
```kotlin
@Serializable data class Outer(@ByteString val i: InnerValue)

@Serializable 
@JvmInline value class InnerValue(val b: ByteArray)
```

Then I don't see tests for it.

sandwwraith: > I'm not sure if that is wanted
I'm also unsure because I didn't see any requests for that. At the first look, it sounds reasonable, but I do not use byte strings that often

sandwwraith (approved):
Thank you!

PR #2469, opened 2023-10-11
Fix IllegalAccessException being thrown on an attempt to retrieve ser…

### LINKED ISSUE #2449: Serializing `emptyMap()` throws `IllegalAccessException`
**Describe the bug**

Calling `kotlinx.serialization.serializerOrNull(emptyMap<String, String>().javaClass)` causes

```
Exception in thread "main" java.lang.IllegalAccessException: class kotlinx.serialization.internal.PlatformKt cannot access a member of class kotlin.collections.EmptyMap with modifiers "public static final"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:392)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
	at java.base/java.lang.reflect.Field.checkAccess(Field.java:1102)
	at java.base/java.lang.reflect.Field.get(Field.java:423)
	at kotlinx.serialization.internal.PlatformKt.findObjectSerializer(Platform.kt:161)
	at kotlinx.serialization.internal.PlatformKt.constructSerializerForGivenTypeArgs(Platform.kt:48)
	at kotlinx.serialization.SerializersKt__SerializersJvmKt.reflectiveOrContextual$SerializersKt__SerializersJvmKt(SerializersJvm.kt:169)
	at kotlinx.serialization.SerializersKt__SerializersJvmKt.typeSerializer$SerializersKt__SerializersJvmKt(SerializersJvm.kt:160)
	at kotlinx.serialization.SerializersKt__SerializersJvmKt.serializerByJavaTypeImpl$SerializersKt__SerializersJvmKt(SerializersJvm.kt:108)
	at kotlinx.serialization.SerializersKt__SerializersJvmKt.serializerOrNull(SerializersJvm.kt:98)
	at kotlinx.serialization.SerializersKt__SerializersJvmKt.serializerOrNull(SerializersJvm.kt:55)
```

**To Reproduce**

```kotlin
import kotlinx.serialization.serializerOrNull

fun main() {
    serializerOrNull(emptyMap<String, String>().javaClass)
}
```

**Expected behavior**

Serializer able to serialize `emptyMap()` into `{}` is returned.

**Environment**
 - Kotlin version: 1.9.10
 - Library version: 1.6.0
 - Kotlin platforms: JVM
 - Gradle version: 8.3 
 - Other relevant context: Java Temurin 17.0.4

sandwwraith:
…ializer for some private implementation classes from stdlib.

Fixes #2449

### DIFF
--- a/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -153,7 +153,7 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 }
 
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
-    // Special case to avoid IllegalAccessException on Java11+
+    // Special case to avoid IllegalAccessException on Java11+ (#2449)
     // There are no serializable objects in the stdlib anyway.
     if (this.canonicalName.let { it.startsWith("java.") || it.startsWith("kotlin.") }) return null
     // Check it is an object without using kotlin-reflect
--- a/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -153,6 +153,9 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 }
 
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
+    // Special case to avoid IllegalAccessException on Java11+
+    // There are no serializable objects in the stdlib anyway.
+    if (this.canonicalName.let { it.startsWith("java.") || it.startsWith("kotlin.") }) return null
     // Check it is an object without using kotlin-reflect
     val field =
         declaredFields.singleOrNull { it.name == "INSTANCE" && it.type == this && Modifier.isStatic(it.modifiers) }
--- a/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
+++ b/core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
@@ -155,7 +155,7 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
     // Special case to avoid IllegalAccessException on Java11+ (#2449)
     // There are no serializable objects in the stdlib anyway.
-    if (this.canonicalName.let { it.startsWith("java.") || it.startsWith("kotlin.") }) return null
+    if (this.canonicalName?.let { it.startsWith("java.") || it.startsWith("kotlin.") } != false) return null
     // Check it is an object without using kotlin-reflect
     val field =
         declaredFields.singleOrNull { it.name == "INSTANCE" && it.type == this && Modifier.isStatic(it.modifiers) }
--- a/formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
+++ b/formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
@@ -17,6 +17,7 @@
 package kotlinx.serialization
 
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.test.typeTokenOf
 import org.junit.Test
 import java.util.HashMap
 import java.util.HashSet
@@ -24,7 +25,8 @@ import kotlin.collections.LinkedHashMap
 import kotlin.collections.Map
 import kotlin.collections.hashMapOf
 import kotlin.collections.hashSetOf
-import kotlin.test.assertEquals
+import kotlin.reflect.*
+import kotlin.test.*
 
 
 class JavaCollectionsTest {
@@ -38,7 +40,7 @@ class JavaCollectionsTest {
     )
 
     @Test
-    fun test() {
+    fun testJavaCollectionsInsideClass() {
         val original = HasHashMap("42", hashMapOf(1 to "1", 2 to "2"), hashSetOf(11), LinkedHashMap(), null)
         val serializer = HasHashMap.serializer()
         val string = Json.encodeToString(serializer = serializer, value = original)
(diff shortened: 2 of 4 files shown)


core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -153,6 +153,9 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 }
 
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
+    // Special case to avoid IllegalAccessException on Java11+
+    // There are no serializable objects in the stdlib anyway.
+    if (this.canonicalName.let { it.startsWith("java.") || it.startsWith("kotlin.") }) return null
```
pdvrieze, line 158: Rather than check for particular packages, the problem in the underlying bug is actually that the type itself is private. So what about just using `Modifier.isPublic(modifiers)` rather than a textual check?

pdvrieze: @sandwwraith Package private would still not be correct on "public", but I can see where you are over constraining the context where this function is called from within the package.

formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
```diff
@@ -49,4 +51,55 @@ class JavaCollectionsTest {
...
+    fun testTopLevelSetsAndLists() {
+        // Same reasoning as for maps
+        assertNull(serializerOrNull(emptyList<String>().javaClass))
+        assertNull(serializerOrNull(listOf<String>("a").javaClass))
+        assertNull(serializerOrNull(listOf<String>("a", "b").javaClass))
+        assertNull(serializerOrNull(emptySet<String>().javaClass))
+        assertNull(serializerOrNull(setOf<String>("a").javaClass))
+        assertNull(serializerOrNull(setOf<String>("a", "b").javaClass))
+        assertContains(
+            serializer(typeTokenOf<Set<String>>()).descriptor.serialName,
```
shanshin, line 89: Should we also add a test on KClass for the `serializer(kClass: KClass<*>, typeArgumentsSerializers: List<KSerializer<*>>, isNullable: Boolean)` function?

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -153,6 +153,9 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 }
 
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
+    // Special case to avoid IllegalAccessException on Java11+
+    // There are no serializable objects in the stdlib anyway.
```
shanshin, line 157: Maybe we should add a link to the [corresponding issue](https://github.com/Kotlin/kotlinx.serialization/issues/2449)?

sandwwraith: It'll be in the commit message, but I can add it here too

formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
sandwwraith: It's a debatable question. Currently, `serializer(List::class, listOf(String.serializer()), false)` returns the proper serializer for a list of strings. However, `serializer(emptyList<String>()::class, listOf(String.serializer()), false)` does not because we check strict equality for `List::class`, not subtyping. From an ideological point of view, I'd say that it could return a proper serializer, so returning `null` for now is a bug/implementation detail, not a dedicated design choice. So it's different from `serializerOrNull(emptyMap<String, String>().javaClass)`, where it is specified by design that we return `null` on generic classes (even if their implementation is non-generic).

formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
```diff
@@ -49,4 +51,55 @@ class JavaCollectionsTest {
...
         assertEquals(expected = original, actual = restored)
     }
+
+    @Test
+    fun testTopLevelMaps() {
+        // Returning null here is a deliberate choice: map constructor functions may return different specialized
+        // implementations (e.g., kotlin.collections.EmptyMap or java.util.Collections.SingletonMap)
+        // that may or may not be generic. Since we generally cannot return a generic serializer using Java class only,
+        // all attempts to get map serializer using only .javaClass should return null.
+        assertNull(serializerOrNull(emptyMap<String, String>().javaClass))
```
pdvrieze, line 61: There are two additional tests that should be added here. This adds two additional test cases (both fail with the suggested implementation):
- Getting a serializer from  an anonymous object (`canonicalName` returns `null`) throws an NPE
- Getting a serializer from a local named object throws because it is not built in

```suggestion
    private object namedObject

    @Test
    fun testTopLevelMaps() {
        // Returning null here is a deliberate choice: map constructor functions may return different specialized
        // implementations (e.g., kotlin.collections.EmptyMap or java.util.Collections.SingletonMap)
        // that may or may not be generic. Since we generally cannot return a generic serializer using Java class only,
        // all attempts to get map serializer using only .javaClass should return null.
        assertNull(serializerOrNull(emptyMap<String, String>().javaClass))
        assertNull(serializerOrNull(object {}.javaClass))
        assertNull(serializerOrNull(namedObject.javaClass))
```

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -153,6 +153,9 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
 }
 
 private fun <T : Any> Class<T>.findObjectSerializer(): KSerializer<T>? {
+    // Special case to avoid IllegalAccessException on Java11+ (#2449)
+    // There are no serializable objects in the stdlib anyway.
+    if (this.canonicalName.let { it.startsWith("java.") || it.startsWith("kotlin.") }) return null
```
pdvrieze, line 158: The problem is that it is not accessible (`IllegalAccessException`), if it is not accessible then it is also not an object. See the suggested additions to the test suite (JavaCollectionsTest) that are not fixed by the solution.
```suggestion
    if (!Modifier.isPublic(this.modifiers)) return null
```

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -153,6 +153,9 @@ private fun <T : Any> Class<T>.createEnumSerializer(): KSerializer<T> {
...
     // Check it is an object without using kotlin-reflect
     val field =
         declaredFields.singleOrNull { it.name == "INSTANCE" && it.type == this && Modifier.isStatic(it.modifiers) }
```
pdvrieze, line 161: INSTANCE could also not be public, and therefore the type would not be an object (small chance, but add a check for it being public).
```suggestion
    val field = declaredFields.singleOrNull {
        it.name == "INSTANCE" &&
            it.type == this &&
            Modifier.isStatic(it.modifiers) &&
            Modifier.isPublic(it.modifiers)
    } ?: return null
```

pdvrieze (commented):
I've gone in and added some tests. Using modifiers fixes the original problem, but also addresses some other object cases (tests added).

formats/json-tests/jvmTest/src/kotlinx/serialization/JavaCollectionsTest.kt
sandwwraith: Thanks for pointing out the problem with the anonymous class! Adding a safe call to `canonicalName` is enough to fix it, though.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
sandwwraith: This check will not work correctly for `protected` situation like this one:
```
open class A {
  @Serializable protected object B {}
}
```

Regarding `@Serializable private object X` — you're right; this code will throw an IllegalAccessError now. However, it will indicate problems with users' code instead of masking it behind the `null`. Also, we support `@Serializable private class X` even if it is in a different package. So, probably, it should be supported instead of returning null. Seems like it is out of the scope of this PR anyways.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
sandwwraith: Can you provide an example? IIRC Kotlin doesn't produce INSTANCE outside of Kotlin objects and it's hard to produce such code by yourself

pdvrieze: Basically if someone were to have a Java based singleton such as:
```java
class BreakThings {
    private static BreakThings INSTANCE = BreakThings();

    public static BreakThings getInstance() { return INSTANCE; }
}

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
pdvrieze: Maybe the "best" way would be to just catch the `IllegalAccessException` as it is otherwise really hard to know what is accessible.

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
sandwwraith: I think it's ok to fail or have unspecified behavior when asked for serializer for some java class

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
sandwwraith: It's an adequate approach (although we lose information that is contained in expectation message). However, I should first check if we can support `@Serializable private object X` properly.

shanshin (approved):

PR #2472, opened 2023-10-12
fixed typo

AbhineshJha:
Some words are corrected in right manner and also some grammar mistakes .

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,11 +180,11 @@ dependencies {
 }
 ```
 
->We also provide `kotlinx-serialization-core` artifact that contains all serialization API but does not have bundled serialization format with it
+>We also provide `kotlinx-serialization-core` artifacts that contains all serialization API but does not have bundled serialization format with it
 
 ### Android
 
-By default, proguard rules are supplied with the library.
+By default, ProGuard rules are supplied with the library.
 [These rules](rules/common.pro) keep serializers for _all_ serializable classes that are retained after shrinking,
 so you don't need additional setup.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -216,7 +216,7 @@ com.example.myapplication.HasNamedCompanion2
 
 
 <details>
-<summary>Example of named companion rules for R8 full mode</summary>
+<summary>Example of named companion rules for R8 full-mode</summary>
```
sandwwraith, line 219: 'Full mode' is mentioned in Google documentation without a dash:
https://developer.android.com/build/shrink-code#full-mode
https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md#r8-full-mode

I prefer being consistent with them.

README.md
```diff
@@ -180,11 +180,11 @@ dependencies {
 }
 ```
 
->We also provide `kotlinx-serialization-core` artifact that contains all serialization API but does not have bundled serialization format with it
+>We also provide `kotlinx-serialization-core` artifacts that contains all serialization API but does not have bundled serialization format with it
```
sandwwraith, line 183: While technically, several core artifacts exist (for jvm, native, etc...), ideologically, they are all the same. Moreover, platform artifacts have different coordinates (`kotlinx-serialization-core-jvm`, etc). From a user standpoint, only one `kotlinx-serialization-core` should be used. Therefore, I'd prefer to leave this word in a singular form.

Also, you've had to change `contains` to match `artifacts`.

sandwwraith (changes requested):

AbhineshJha:
ok as your wish !

README.md
AbhineshJha: ok as your wish

PR #2473, opened 2023-10-13
Build improvements: Kotlin 1.9.20-RC, default target hierarchy, bump Okio

sandwwraith:
Updated to Kotlin 1.9.20-RC:
- Kotlin 1.9.0/10 had some blockers in XCode 15 support.
- Simplify WASM setup (it has been changed in 1.9.20).
- Drop obsolete Native targets.

Apply default target hierarchy:
- Greatly simplified buildscript for native targets.
- Supported targets are now runnable from gutter in the IDE.
- Now all possible targets are built by default (without -Pnative.deploy flag), resulting in less confusion for contributors and CI (but longer local `check` times).
- Publishing library with K2 should be possible now, no more incorrect `publishNativeKotlinMetadata` failing task.
- More targets available in integration-test project.

Updated Okio for:
- linuxArm64 target support


gradle/native-targets.gradle
```diff
@@ -2,178 +2,42 @@
...
+        // Tier 2
+        watchosSimulatorArm64()
+        watchosX64()
+        watchosArm32()
+        watchosArm64()
+        tvosSimulatorArm64()
+        tvosX64()
+        tvosArm64()
+        iosArm64()
+        linuxArm64()
```
shanshin, line 30: Is `linuxArm64`  now supported in IO?

shanshin (approved):

igoriakovlev (approved):

gradle/native-targets.gradle
```diff
@@ -2,178 +2,42 @@
...
-
-        if (!project.ext.singleTargetMode) {
-            configure(project.ext.nativeMainSets) {
-                dependsOn nativeMain
-            }
-
-            configure(project.ext.nativeTestSets) {
-                dependsOn nativeTest
-            }
+        targetHierarchy.default()
```
whyoleg, line 11: 1. targetHierarchy.default is deprecated in 1.9.20 AFAIK
2. there is no need to apply default hierarchy starting from 1.9.20 - so no need to call it here :)

sandwwraith: 1. There's no deprecation message during the build
2. It is not applied by default because there is a custom shared js+wasm source set:
```
w: The Default Kotlin Hierarchy Template was not applied to 'project ':kotlinx-serialization-json'':
Explicit .dependsOn() edges were configured for the following source sets:
[jsMain, jsWasmMain, wasmJsMain]
```

whyoleg: 1. 
image
2. hm, I see it now. In this case yes, it should be applied directly via `applyDefaultHierarchyTemplate` (stable API from 1.9.20)

sandwwraith: It seems that deprecation notices aren't shown in Groovy

whyoleg: `applyDefaultHierarchyTemplate` is available only starting from 1.9.20

sandwwraith: Thanks for the info!

PR #2474, opened 2023-10-13
[Kotlin 2.0] Tests for fixes for #KT-62522 and #KT-62215

sandwwraith:
JVM works since 1.8.0; Native should work with 2.0.


shanshin (approved):

sandwwraith:
Wasm tests are failing because of https://youtrack.jetbrains.com/issue/KT-73830, moving to draft again....

sandwwraith:
I think I'll move problematic `TypedSealedClass` into kotlinx-serialization itself instead of integration test, to catch problems earlier.

Issue #2221, opened 2023-03-06
More lenient JSON parser or seamless JSON5 integration


qwwdfsad:
We have a steady stream of reports that all is about more permissive parsing: 

* Trailing commas (#1812)
* Commas and comments
* Multiline strings
* Optional commas
* Optional quotation

The requirements altogether resemble [JSON5](https://json5.org/) format (requested as #797), and we already have semi ad-hoc `lenient` mode that is an ad-hoc solution to JSON5.

We should decide whether we want to support a separate JSON5 format, gradually transform `lenient` mode into `JSON5` or introduce `json5Mode` into our default `Json` format (that also can have a default `object Json5`) counterpart.

This issue is rather an umbrella for gathering feedback, so if you have any specific requests, whether they are covered by JSON5 or you would like them to be part of `lenient` parser, please let us know

aSemy:
Supporting JSON5 would be great! It's a much nicer format than JSON, and I think it's quite fitting for Kotlin: comparing JSON5 to JSON is like comparing Kotlin to Java.

Here are some notes I have, after having working with JSON5 in https://github.com/adamko-dev/json5-kotlin and https://github.com/xn32/json5k

#### Remove JSON?

What about removing JSON completely? JSON5 is a superset of JSON. If KxS implemented JSON5, then this would be supplant JSON, and JSON could be supported by setting the correct configuration for JSON5 encoding.

Even if JSON isn't removed, then JSON5 and JSON both have the same primitive/array/object types.

Would it be possible to share the JsonElement classes (similar to the proposal in #1573?)
Or would converter functions (`fun JsonObject.toJson5Object()`, `fun Json5Array.toJsonArray()` be possible/required?

#### Comments

Can KxS support comments?

I'd like to see an option to ignore them when encoding, or decoding. When decoding, decoding them as optional, additional properties #1978 might work?

See also: 
* https://github.com/xn32/json5k/issues/6
* https://github.com/adamko-dev/json5-kotlin/issues/7
* #1970

Issue #797, opened 2020-04-11
JSON5 support


acanthite1855:
**What is your use-case and why do you need this feature?**
It would be nice to have JSON5 support to serialize/deserialize configs and other data files that may be written/read by humans. It has quite useful features:
- No quotes for keys when they're not needed
- Multi-line strings
- Trailing comas
- Single- and Multi-line comments.
- and some [more](https://json5.org/)

**Describe the solution you'd like**
Well, pretty much the same as for JSON but form JSON5. I don't know what else to say here. All details on the format are on [json5.org](https://json5.org/)

sandwwraith:
Side note: trailing commas and unquoted literals are now somewhat supported by `isLenient = true` setting

MarcelReiter:
Side note: IntelliJ has native support (e.g. syntax highlighting) for Json5 ([link](https://www.jetbrains.com/help/idea/json.html)), which would make this a cool feature for Android developers especially

lisonge:
now  How is it going ？

aSemy:
I'm also interested in this. For now, I've converted an existing JSON5 Java library to Kotlin and added some Kotlinx Serialization interop, so it can convert to/from JSON5/JSON. Please let me know if it's useful, and you're able to help make it more stable (I haven't tested it significantly) 

https://github.com/adamko-dev/json5-kotlin

AmibeSkyfy16:
Hi, any news about that ?
Im trying to use json5, but it seems that kotlinx.serialization dont like comment

```json
{
  /*
   * The day when players will be able to do damage to themselves. default -> 2
   */
  "dayOfAuthorizationOfThePvP": 2,

  /*
   * The day when players can get into the nether. default -> 4
   */
  "dayOfAuthorizationOfTheEntryInTheNether": 4,

  /*
   * If players can use enderpearl to assault bases or not
   */
  "allowEnderPearlAssault": false
}
```

Caused by: java.lang.RuntimeException: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 11: Expected semicolon ':', but had ' ' instead at path: $
JSON input: {
  /*
   * The day when players will b.....
	at ch.skyfy.jsonconfig.example3.config.Configs.<clinit>(Configs.kt:25)
	... 90 more
Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 11: Expected semicolon ':', but had ' ' instead at path: $

[...]

PR #2480, opened 2023-10-17
Add a flag to allow parser to accept trailing commas.

### LINKED ISSUE #1812: Support trailing commas in Json lenient mode
The original use-case is to parse-out handwritten configuration JSON where the use of trailing commas is just a convenience (e.g. for prettier `git diff`), but we would also like to hear other use-cases.

sandwwraith:
This is one of the popular community requests and one of the main reasons people ask for Json5 support. Implementing this flag separately will allow alleviating large paint points quickly without a need to wait for full Json5 support.

Fixes #1812
Relates to: #797, #2221

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonErrorMessagesTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonErrorMessagesTest.kt
@@ -6,6 +6,7 @@
 package kotlinx.serialization.json
 
 import kotlinx.serialization.*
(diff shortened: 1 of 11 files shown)


sandwwraith:
To discussion: #1812 says 'support trailing commas in lenient mode', but I'm not sure this is a good idea to 'Just put everything non-complaint under lenient'. Probably we should leave it only for quotation and even rename it to `relaxedQuotes` or something.

shanshin (approved):

PR #2481, opened 2023-10-17
Improve polymorphic deserialization optimization:

### LINKED ISSUE #2478: Detect missing deserializer before hitting JSON "slow path" under polymorphism
**Describe the bug**

When attempting to decode data that contains unknown types (e.g. generated by a newer version of our application), the StreamingJsonDecode abandons streaming and tries to build the entire (huge) structure in-memory using JsonObject etc, which would ultimately fail anyway because the deserializer simply isn't available.

I think it'd be possible to error out earlier and not fall back to the non-streaming approach in such a case.

I've tracked this down to [this fragment in StreamingJsonDecode](https://github.com/Kotlin/kotlinx.serialization/blob/ba38b7099e502c16efd30386d18174e3a7aefb3a/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt):
```
val discriminator = deserializer.descriptor.classDiscriminator(json)
            val type = lexer.peekLeadingMatchingValue(discriminator, configuration.isLenient)
            var actualSerializer: DeserializationStrategy<Any>? = null
            if (type != null) {
                actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)
            }
            if (actualSerializer == null) {
                // Fallback if we haven't found discriminator or serializer
                return decodeSerializableValuePolymorphic<T>(deserializer as DeserializationStrategy<T>)
            }
```

Suppose `type != null` because we did find the discriminator, but `actualSerializer == null` because the serializer isn't known in the current application version. Is there even a point to try to fallback to the "slow path"? 

It'd be nice if you could error out right then and there before entering the [slow-path](https://github.com/Kotlin/kotlinx.serialization/blob/ba38b7099e502c16efd30386d18174e3a7aefb3a/formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L58).

Are there cases where it's required to enter this code-path when you find the discriminator but don't find a deserializer? From my reading of the code, you always error out [here](https://github.com/Kotlin/kotlinx.serialization/blob/ba38b7099e502c16efd30386d18174e3a7aefb3a/formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L67):

```kt
val actualSerializer = deserializer.findPolymorphicSerializerOrNull(this, type)
        ?: throwSerializerNotFound(type, jsonTree)
 ```

### PR DESCRIPTION
**Pull request by sandwwraith**
Previously, when discriminator was found as the first key in Json, but there was no deserializer for it, we still fell back to a slow path with JsonTree. It was actually meaningless because a slow path always throws exception when a serializer is not found. Such behavior led to unnecessary memory pressure & consumption in exceptional cases (see linked ticket for details).

Also make polymorphic deserialization exception messages more meaningful and make them more consistent with serialization ones.

Also fix behavior when the actual discriminator value is JsonNull (it should be treated as missing, not as "null" string).

Fixes #2478

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
@@ -133,4 +133,17 @@ class PolymorphismWithAnyTest: JsonTestBase() {
             json = json
         )
     }
+
+    @Test
+    fun testTypeKeyLastInInput() = parametrizedTest { mode ->
+        val json = Json { serializersModule = (baseAndDerivedModuleAtAny + BaseAndDerivedModule) }
+        val input = """{"data":{"a":{"id":1,"s":"foo","type":"kotlinx.serialization.PolyDerived"}},
+                |"polyBase":{"id":1,"s":"foo","type":"kotlinx.serialization.PolyDerived"}}""".trimMargin().lines().joinToString(
+            "")
+        val data = MyPolyDataWithPolyBase(
+            mapOf("a" to PolyDerived("foo")),
+            PolyDerived("foo")
+        )
+        assertEquals(data, json.decodeFromString(MyPolyDataWithPolyBase.serializer(), input, mode))
+    }
 }
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
@@ -5,13 +5,13 @@
 package kotlinx.serialization.features
 
 import kotlinx.serialization.*
-import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.*
 import kotlinx.serialization.modules.*
 import kotlinx.serialization.modules.plus
 import kotlinx.serialization.test.assertStringFormAndRestored
 import kotlin.test.*
 
-class PolymorphismWithAnyTest {
+class PolymorphismWithAnyTest: JsonTestBase() {
 
     @Serializable
     data class MyPolyData(val data: Map<String, @Polymorphic Any>)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt [hunk 1]
```diff
@@ -71,19 +71,20 @@ internal open class StreamingJsonDecoder(
...
-            }
 
-            discriminatorHolder = DiscriminatorHolder(discriminator)
             @Suppress("UNCHECKED_CAST")
-            val result = actualSerializer.deserialize(this) as T
-            return result
+            val actualSerializer = try {
+                    deserializer.findPolymorphicSerializer(this, type)
+                } catch (it: SerializationException) { // Wrap SerializationException into JsonDecodingException to preserve position, path, and input.
+                    val message = it.message!!.substringBefore('\n').removeSuffix(".")
```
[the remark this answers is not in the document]
sandwwraith: You mean example of the message we're trying to parse? Sure

qwwdfsad (approved):
Nice one

formats/json-tests/commonTest/src/kotlinx/serialization/features/PolymorphismWithAnyTest.kt
```diff
@@ -117,17 +120,17 @@ class PolymorphismWithAnyTest {
     @Test
     fun testBindModules() {
         val json = Json { serializersModule = (baseAndDerivedModuleAtAny + BaseAndDerivedModule) }
-        assertStringFormAndRestored(
+        assertJsonFormAndRestored(
             expected = """{"data":{"a":{"type":"kotlinx.serialization.PolyDerived","id":1,"s":"foo"}},
```
shanshin, line 124: Perhaps we should also add a test where a slow path will be implicitly used: for example, `{"id":1,"s":"foo", "type":"kotlinx.serialization.PolyDerived"}}`

sandwwraith: added

PR #2483, opened 2023-10-18
Kotlin community variables

sandwwraith:
Transfer commits regarding TC setup from `kotlin-community/k2/dev` branch.

After landing these commits in `dev`, branch should be identical to `kotlin-community/k2/dev`, so there will be no need for community branches — K2 Aggregate build will be able to use `dev` directly. Current aggregate build should also use `dev` instead of `kotlin-community/dev` now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,10 @@ buildscript {
     ]
     ext.koverEnabled = property('kover.enabled') ?: true
 
+    def noTeamcityInteractionFlag = rootProject.hasProperty("no_teamcity_interaction")
+    def buildSnapshotUPFlag = rootProject.hasProperty("build_snapshot_up")
+    ext.teamcityInteractionDisabled = noTeamcityInteractionFlag || buildSnapshotUPFlag
+
     /*
     * This property group is used to build kotlinx.serialization against Kotlin compiler snapshot.
     * When build_snapshot_train is set to true, kotlin_version property is overridden with kotlin_snapshot_version.
--- a/gradle/teamcity.gradle
+++ b/gradle/teamcity.gradle
@@ -3,7 +3,7 @@
  */
 
 def teamcitySuffix = project.findProperty("teamcitySuffix")?.toString()
-if (project.hasProperty("teamcity") && !(build_snapshot_train || rootProject.properties['build_snapshot_up'])) {
+if (!teamcityInteractionDisabled && project.hasProperty("teamcity") && !(build_snapshot_train || rootProject.properties['build_snapshot_up'])) {
     // Tell teamcity about version number
     def postfix = (teamcitySuffix == null) ? "" : " ($teamcitySuffix)"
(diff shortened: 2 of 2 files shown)


FenstonSingel:
I am unsure that running TeamCity configurations for user projects straight on top of kotlinx libraries' (and any user projects', for that matter) development branches is a good idea. There have been precedents for migrations that were required for successful TeamCity builds yet couldn't be immediately cherry-picked to the development branch.

A stark example is [KT-57838](https://youtrack.jetbrains.com/issue/KT-57838) and #2277; the corresponding migration commits for 1.9.20 could not be cherry-picked because the `kotlin.experimental.ExperimentalNativeApi` annotation was only introduced in Kotlin 1.9.0 while kotlinx-serialization's development branch was built with Kotlin 1.8.10 at the time.

It means we will have to commit to always building libraries with the newest stable Kotlin release if we want to run user project builds on top of development branches. Mind that this is but one example that immediately came to my mind; @woainikk might remember more examples and reasons for having separate branches for user projects.

FenstonSingel:
All that said, I don't think I see any significant problems with cherry-picking these particular changes to `dev`. It *will* make maintenance of the `kotlin-community/dev` branch easier. Just wanted to voice my concerns about getting rid of `kotlin-community/dev` altogether.

One thing would be to consider getting rid of 6dd871e18aba1c31557a4cd11b46282a5f40617f, I suppose. That commit was made when we only started setting up the K2 user projects, were overloaded with compilation errors alone, and decided that we would like not to run tests in configurations for K2 user projects just yet. Since that decision has been reverted and user projects' tests are now a part of the K2 quality gates, I think the `-Pdisable_project_tests` flag is no longer needed.

sandwwraith:
Yes, I agree that `kotlin-community` branches still may be needed for fix-ups like #2277. However, these branches should be as close to the `dev` as possible and regularly be updated. For example, the current `kotlin-community/dev` is 5 or 6 commits behind `dev`. 
I'd say we need to restructure our process: community branches should be automatically rebased on top of `dev` every check-out. This PR will allow to implement this approach in future easier.

> One thing would be to consider getting rid of https://github.com/Kotlin/kotlinx.serialization/commit/6dd871e18aba1c31557a4cd11b46282a5f40617f, I suppose.

If it is not needed anymore, I'll drop it from the PR.

FenstonSingel:
Reliable actualization of separate branches for user projects is kind of an open question right now, yeah— well, we do know that it needs to be implemented; we're just trying to get there.

I think we're slowly moving toward automated rebasing at the moment (for example, a prototype TeamCity configuration was implemented for kotlinx-serialization by the infrastructure team, as far as I know). The exact technical decisions and processes around it have yet to be entirely finalized right now, though.

It does make sense to transfer all "infrastructural" commits like the ones in this PR to `dev` in the meanwhile, yes.

woainikk (approved):

PR #2524, opened 2023-12-04
Simplify Dokka source link configuration

whyoleg:
Based on discussion: https://github.com/Kotlin/kotlinx-datetime/pull/323#discussion_r1412147614

Checked locally - content is identical for both approaches


sandwwraith:
Nice job!

PR #2527, opened 2023-12-06
Reorder and filter mavenLocal() for more reproducible builds

sandwwraith:
The mavenLocal() repository has been moved down below and filtered to ensure that it is checked last and only for kotlinx-serialization:1.x.y-SNAPSHOT.

Also replace outdated `stdlib-common` artifact name.

### DIFF
--- a/integration-test/build.gradle
+++ b/integration-test/build.gradle
@@ -5,9 +5,13 @@ buildscript {
     ext.serialization_version = mainLibVersion
 
     repositories {
-        mavenLocal()
         mavenCentral()
         maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
+        mavenLocal() {
+            mavenContent {
+                snapshotsOnly()
+            }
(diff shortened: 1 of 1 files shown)


sandwwraith:
PR is for master because builds are failing there, too. Will cherry-pick to dev manually after review.

integration-test/build.gradle
```diff
@@ -5,9 +5,13 @@ buildscript {
     ext.serialization_version = mainLibVersion
 
     repositories {
-        mavenLocal()
         mavenCentral()
         maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
+        mavenLocal() {
+            mavenContent {
```
shanshin, line 11: The local repository is only used to search for recently builded snapshot serialization artifacts?

shanshin (approved):

integration-test/build.gradle
sandwwraith: Yes

PR #2530, opened 2023-12-15
Do not try to coerce input values for properties 

### LINKED ISSUE #2529: Misleading MissingFieldException on unexpected enum with coerceInputValues
**Describe the bug**
Json configuration option `coerceInputValues = true` is very helpful to future-proof network code, but it has some unfortunate side effects. Some of them are already raised in different issues, but this one is not (AFAIK):

When model has enum field with no default, and unexpected enum value is encountered, confusing `MissingFieldException` is thrown instead of correct exception with helpful error message. This caught me off-guard a couple times, causing time wasted looking for typos in field name, etc.

**To Reproduce**
```kotlin
@Test
fun `GIVEN json with unexpected enum value WHEN deserialize THEN throw meaningful exception`() {
    val jsonWithUnexpectedValue = """
        {
            "color": "SOMETHING_UNEXPECTED"
        }
    """.trimIndent()
    val jsonSerializer = Json {
        coerceInputValues = true
    }
    val throwable = runCatching {
        jsonSerializer.decodeFromString<Model>(jsonWithUnexpectedValue)
    }.exceptionOrNull()

    assertThat(throwable, instanceOf(SerializationException::class.java))
    assertThat(throwable!!.message, containsString("SOMETHING_UNEXPECTED"))
}

@Serializable
data class Model(val color: Color)

enum class Color { RED, GREEN }
```

**Expected behavior**
`kotlinx.serialization.SerializationException: gmk57.Color does not contain element with name 'SOMETHING_UNEXPECTED' at path $.color`
That's what is thrown with `coerceInputValues = false`

**Observed behavior**
`kotlinx.serialization.MissingFieldException: Field 'color' is required for type with serial name 'gmk57.Model', but it was missing at path: $`

**Environment**
 - Kotlin version: 1.9.10
 - Library version: 1.6.2

I guess the root cause is the same as in #1947, but the issue we face is a bit different.

sandwwraith:
that do not have default values

Trying so leads to confusing errors about missing values despite a json key actually present in the input.

Fixes #2529

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -287,7 +287,7 @@ public class JsonBuilder internal constructor(json: Json) {
     public var prettyPrintIndent: String = json.configuration.prettyPrintIndent
 
     /**
-     * Enables coercing incorrect JSON values to the default property value in the following cases:
+     * Enables coercing incorrect JSON values to the default property value (if exists) in the following cases:
      *   1. JSON value is `null` but the property type is non-nullable.
      *   2. Property type is an enum type, but JSON value contains unknown enum member.
      *
--- a/docs/basic-serialization.md
+++ b/docs/basic-serialization.md
@@ -534,7 +534,7 @@ the `null` value to it.
 
 ```text
 Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 52: Expected string literal but 'null' literal was found at path: $.language
-Use 'coerceInputValues = true' in 'Json {}' builder to coerce nulls to default values.
+Use 'coerceInputValues = true' in 'Json {}' builder to coerce nulls if property has a default value.
 ```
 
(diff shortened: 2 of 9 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -287,7 +287,7 @@ public class JsonBuilder internal constructor(json: Json) {
     public var prettyPrintIndent: String = json.configuration.prettyPrintIndent
 
     /**
-     * Enables coercing incorrect JSON values to the default property value in the following cases:
+     * Enables coercing incorrect JSON values to the default property value (if exists) in the following cases:
```
shanshin, line 290: M.b. `if specified`?

shanshin (approved):

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: I think `if specified` may be incorrectly read as `if coerceInputValues = true is specified`

PR #2531, opened 2023-12-17
Adding kebab-case naming strategy

Kantis:
1-to-1 copy of the snake-case implementation, but replaced all underscore with dashes.

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamingStrategyTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamingStrategyTest.kt
@@ -103,6 +103,52 @@ class JsonNamingStrategyTest : JsonTestBase() {
         }
     }
 
+    @Test
+    fun testKebabCaseStrategy() {
+        fun apply(name: String) =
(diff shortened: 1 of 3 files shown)


sandwwraith:
If it is the same algorithm, as you said, I'd prefer moving an implementation to some internal function instead of duplicating the code.

Also, I'd like to ask whether you are going to use this strategy yourself and for what kind of application it is required. When we studied the matter, it looked like the majority (>90%) of use cases were SnakeCase alone. That's why we decided not to add others.

Kantis:
@sandwwraith I extracted the algorithm in last commit. And yes, I interact with a json REST API which uses kebab casing for object properties, so I have the duplicated algorithm in my code. Figured I could try to submit it instead. If you're hesitant we could let it dwell for a few months and see if there's support for including it..

sandwwraith:
> If you're hesitant we could let it dwell for a few months and see if there's support for including it..

No, that's not needed, I'm OK with adding your PR. 

You need to run `./gradlew apiDump` to get new signatures added to API file. (see here: https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md#updating-the-public-api-dump)

sandwwraith (approved):
Sorry, it seems I've missed the email with the last commit. Thank you for your efforts!

PR #2533, opened 2023-12-19
Slightly improve error messages thrown from serializer<T>() function

sandwwraith:
and mention intrinsics to keep them in sync.


shanshin (approved):

shanshin:
Can we somehow rule out similar problems in the future?
Maybe we should add error tests in the plugin for each intrinsic?

sandwwraith:
There are already error tests in the compiler for star projections and (now) non-reified type arguments. I do not remember having other type-related errors.

PR #2550, opened 2024-01-22
Minor, move decodeDuration logic to a function reference

udalov:
This is needed because once we enable indy lambdas by default (KT-45375), `@SuppressAnimalSniffer` annotation stops working, and the `animalsnifferMain` task reports several errors about `java.time.Duration` being used.

There are multiple ways to workaround this issue, for example we could annotate the lambda with `@JvmSerializableLambda` or compile the whole module with `-Xlambdas=class`, but I chose to use a simple function reference instead, since we don't generate those via invokedynamic yet (KT-45658), and it doesn't make the code any more difficult.

### DIFF
--- a/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
+++ b/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt
@@ -109,11 +109,13 @@ public sealed class Hocon(
 
         private fun getTaggedNumber(tag: T) = validateAndCast<Number>(tag)
 
+        @Suppress("UNCHECKED_CAST")
+        protected fun <E> decodeDuration(tag: T): E =
+            getValueFromTaggedConfig(tag, ::decodeDurationImpl) as E
+
         @SuppressAnimalSniffer
-        protected fun <E> decodeDuration(tag: T): E {
-            @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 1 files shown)


sandwwraith:
I take it that R8 supports indy lambdas? Ok then.  Please add the same comment to the code.

Also, you've deleted the import and forgot to delete actual usage of annotation.

sandwwraith (changes requested):
.

udalov:
Maybe my commit message isn't clear. The annotation is still needed. If we compile this with without the annotation, `animalsnifferMain` will complain about `java.time.Duration` not available in older Android API. And extracting it to a function reference is required because `animalsnifferMain` does not handle indy lambdas correctly.

R8 supports indy lambdas, but it is irrelevant here.

Import was optimized automatically because there's a `*`-import from the whole package already.

sandwwraith:
> Import was optimized automatically because there's a *-import from the whole package already.

And yet, all builds fail with `Compilation error: :kotlinx-serialization-hocon:compileKotlin: e: file:///mnt/agent/work/b2fef8360e1bcf3d/formats/hocon/src/main/kotlin/kotlinx/serialization/hocon/Hocon.kt:115:10 Unresolved reference: SuppressAnimalSniffer`. This looks like a compiler bug, so explicit import may be a necessary workaround.

udalov:
Yeah, sorry. 🤦 It was stupid of me not to check compilation one last time after optimizing imports.

PR #2561, opened 2024-02-03
Fix serializing nulls for a property of a parameterized type with a nullable upper bound with Protobuf

ShreckYe:


ShreckYe:
The checks on Linux failed because "Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)". Is there a way to rerun it?

ShreckYe:
Here is some brief information for this PR.

See such a class definition with a type parameter `T` which has a nullable upper bound:

```kotlin
@Serializable
data class Box<T>(val value: T)
```

`ProtoBuf.encodeToByteArray(Box(null))` fails with an exception:

```
kotlinx.serialization.SerializationException: 'null' is not allowed for not-null properties
```

While for JSON, `Json.encodeToString(Box(null))` runs successfully with the result `{"value":null}`.

pdvrieze:
It looks like what you are trying to do is to omit the box if the member is empty. Is that not more of a case for either a custom serializer or a value class (which formats can handle specially). In terms of your tests, you at least need to test for correct behaviour if there are more members than just 1.
Beyond that, this behaviour would need to be configurable (and not default) as you are making an incompatible change to the format.

ShreckYe:
> It looks like what you are trying to do is to omit the box if the member is empty.

No I don't think this is what I am doing. I mean it's probably a bug that, when the upper bound of a type of a property is nullable instead of the type itself, serializing a `null` fails for Protobuf. I will change the verb in the subject of this PR to "fix".

pdvrieze:
> > It looks like what you are trying to do is to omit the box if the member is empty.
> 
> No I don't think this is what I am doing. I mean it's probably a bug that, when the upper bound of a type of a property is nullable instead of the type itself, serializing a `null` fails for Protobuf. I will change the verb in the subject of this PR to "fix".

I was looking at the test data. The data is empty when the member is null, and not empty if there is a member. This looks like what happens for inline structs. Looking back it may be because you directly serialize the struct so the box is not tagged (as it is the outer type).

ShreckYe:
> The data is empty when the member is null, and not empty if there is a member.

I think this is the correct behavior of Protobuf here. In the Protobuf Java library, [writeTo()](https://github.com/protocolbuffers/protobuf/blob/5b32936822e64b796fa18fcff53df2305c6b7686/java/core/src/main/java/com/google/protobuf/AbstractMessageLite.java#L56-L62) serializes a message just by serializing all its fields, and if all its optional fields are empty, the result is an empty message. There is a [writeDelimitedTo()](https://github.com/protocolbuffers/protobuf/blob/5b32936822e64b796fa18fcff53df2305c6b7686/java/core/src/main/java/com/google/protobuf/AbstractMessageLite.java#L65-L74) which writes the message size in varint encoding, but it seems it's not supported here in this library yet.

If you run the code below on the `dev` branch the result is `0`, which means the result is also an empty string.

```kotlin
class NullableIntBox(val value: Int?)
println(ProtoBuf.encodeToHexString(NullableIntBox(null)).length)
```

ShreckYe:
@cliffred Cool. I also played with the compiler plugin for a while and later found that this could be resolved without changing the code in the compiler plugin. I think both are viable solutions.

cliffred:
Me too, it's probably a good idea to deal with this within the individual encoders, therefore I also submitted a PR to the MongoDB BSON kotlinx encoder to fix it also within that encoder.

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufTypeParameterTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+class ProtobufTypeParameterTest {
+    @Serializable
+    data class Box<T>(val value: T)
+
+    @Serializable
+    data class ExplicitNullableUpperBoundBox<T : Any?>(val value: T)
+
+    @Serializable
+    data class ExplicitNullableUpperNullablePropertyBoundBox<T : Any?>(val value: T?)
```
sandwwraith, line 18: Do you mind also adding test for `class Foo<T: Any>(val t: T)` to check that non-nullable type parameters still reject `null`s?

sandwwraith (commented):
Thanks for your PR! I believe this is the correct way to fix this in the encoder.

sandwwraith:
I've also taken a look at the PR by @cliffred. It seems to be the correct approach to my intuition, but I should re-check if it doesn't break anything.

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufTypeParameterTest.kt
ShreckYe: I'm sorry I am not sure how to do this. I could add some tests for `class Foo<T: Any>(val t: T)` but it's impossible to set a `null` for the member `t`, unless I go for some hacks like calling the constructor in Java and passing a `null` argument (I am not completely sure this works though).

sandwwraith: Yeah, you're right, that's impossible for Kotlin's type system. Never mind :)

sandwwraith (approved):

PR #2566, opened 2024-02-09
Added annotation to keep generated serializer when a custom one is specified

### LINKED ISSUE #1169: Keep access to plugin generated serializer when mark class with @Serializable(with = MySerializer)
### What is your use-case and why do you need this feature?
When we write custom serializer for class, which contains plugin generated serializer inside, we cannot use `with` parameter of `@Serializable` because on this case, plugin serializer will not be generated. 

In this case we have to manually use custom serializer in all case when we want to use our class serialization, because we cannot use `@Serializable(with = MySerializer)` instead.

In my case I'm trying to deserialize enum with generated serializer and if I get exception I catch it and return default value. It also can be used for some data transformations during decoding or encoding.

### Describe the solution you'd like
It would be nice to have ability to generate and use plugin serializer when use class annotation `@Serializable(with = MySerializer)`. For example, we can add extra parameter `generateSerializer` for `@Serializable` or use one more annotation to let plugin know, that we still need serializer. Generated serializer can have another name, for example `generatedSerializer()`.

shanshin:
If the @KeepGeneratedSerializer annotation is specified on the class, and a custom serializer is specified, then keep to generate a plugin serializer.

Resolves #1169

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/core/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -325,6 +325,23 @@ public annotation class UseSerializers(vararg val serializerClasses: KClass<out 
 public annotation class Polymorphic
 
 /**
+ * Instructs the serialization plugin to generate serializer automatically generate implementation of [KSerializer]
+ * for the current class if a custom serializer is specified at the same time `@Serializable(with=SomeSerializer::class)`.
+ *
+ * Automatically generated serializer is available via `generatedSerializer()` function in companion object of serializable class.
+ *
+ * Generated serializers allow to use custom serializers on classes from which other serializable classes are inherited.
+ *
+ * Used only with the [Serializable] annotation.
+ *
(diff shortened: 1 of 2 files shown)


core/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -324,6 +324,23 @@ public annotation class UseSerializers(vararg val serializerClasses: KClass<out
 //@Retention(AnnotationRetention.RUNTIME) // Runtime is the default retention, also see KT-41082
 public annotation class Polymorphic
 
+/**
+ * Instructs the serialization plugin to generate serializer automatically generate implementation of [KSerializer]
```
sandwwraith, line 328:
```suggestion
 * Instructs the serialization plugin to keep automatically generated implementation of [KSerializer]
```

sandwwraith (approved):

PR #2587, opened 2024-03-03
Support using an integer as the class/case discriminator in polymorphic serialization

ShreckYe:
As for now, this library serializes the class/type of a subclass using a `serialName` in polymorphic serialization, which can be overridden with the `@SerialName` annotation. I'd like to propose supporting using a number and requiring subclasses to use numbers in polymorphic serialization, as it shortens the size of the serialized message greatly, especially in a binary format such as Protobuf.

When choosing to use a binary format such as Protobuf, one of the biggest concerns is shortening the size of serialized messages. In such a case serializing a fully qualified name of a subclass by default doesn't make much sense, as the binary format such as Protobuf compared to a readable string format as JSON reduces the serialized size of other fields by more than 50%, while a class discriminator takes the same size and takes up the majority of the message size, which defies the whole point. An alternative might be to override the serial name with a short name or a single letter, but IMO, still, the message is longer and it's less elegant than using a number.

### DIFF
--- a/core/api/kotlinx-serialization-core.api
+++ b/core/api/kotlinx-serialization-core.api
@@ -69,6 +69,7 @@ public final class kotlinx/serialization/PolymorphicSerializer : kotlinx/seriali
 public final class kotlinx/serialization/PolymorphicSerializerKt {
 	public static final fun findPolymorphicSerializer (Lkotlinx/serialization/internal/AbstractPolymorphicSerializer;Lkotlinx/serialization/encoding/CompositeDecoder;Ljava/lang/String;)Lkotlinx/serialization/DeserializationStrategy;
 	public static final fun findPolymorphicSerializer (Lkotlinx/serialization/internal/AbstractPolymorphicSerializer;Lkotlinx/serialization/encoding/Encoder;Ljava/lang/Object;)Lkotlinx/serialization/SerializationStrategy;
+	public static final fun findPolymorphicSerializerWithNumber (Lkotlinx/serialization/internal/AbstractPolymorphicSerializer;Lkotlinx/serialization/encoding/CompositeDecoder;Ljava/lang/Integer;)Lkotlinx/serialization/DeserializationStrategy;
 }
 
 public abstract interface annotation class kotlinx/serialization/Required : java/lang/annotation/Annotation {
@@ -79,6 +80,7 @@ public final class kotlinx/serialization/SealedClassSerializer : kotlinx/seriali
 	public fun <init> (Ljava/lang/String;Lkotlin/reflect/KClass;[Lkotlin/reflect/KClass;[Lkotlinx/serialization/KSerializer;[Ljava/lang/annotation/Annotation;)V
 	public fun findPolymorphicSerializerOrNull (Lkotlinx/serialization/encoding/CompositeDecoder;Ljava/lang/String;)Lkotlinx/serialization/DeserializationStrategy;
 	public fun findPolymorphicSerializerOrNull (Lkotlinx/serialization/encoding/Encoder;Ljava/lang/Object;)Lkotlinx/serialization/SerializationStrategy;
+	public fun findPolymorphicSerializerWithNumberOrNull (Lkotlinx/serialization/encoding/CompositeDecoder;Ljava/lang/Integer;)Lkotlinx/serialization/DeserializationStrategy;
 	public fun getBaseClass ()Lkotlin/reflect/KClass;
 	public fun getDescriptor ()Lkotlinx/serialization/descriptors/SerialDescriptor;
 }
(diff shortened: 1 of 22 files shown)


ShreckYe:
# More details to discuss

## Current API design

The current design is like this:

```kotlin
@Serializable
@UseSerialPolymorphicNumbers
sealed class Sealed {
    @Serializable
    @SerialPolymorphicNumber(Sealed::class, 1)
    object Case : Sealed()
}
```

```kotlin
Json.encodeToString<Sealed>(Sealed.Case)
Json.decodeFromString<Sealed>("""{"type":1}""")
```

One marks the classes with `@UseSerialPolymorphicNumbers` and `@SerialPolymorphicNumber`, when the message is serialized it will be `{"type":1}` in JSON, for example.

## Alternatives of implementation details and designs

The implemented code is a prototype. Here are improvements and alternatives to be discussed.

### Alternative names for "serial polymorphic number"

There can be alternative names for the word "serial polymorphic number". For example, these are what I can think of:

1. discriminator number
1. polymorphic number
1. polymorphic serial number
1. subclass number
1. case number

###  Functions

There are 2 properties `useSerialPolymorphicNumbers` and `serialPolymorphicNumberByBaseClass` and a function `getSerialPolymorphicNumberByBaseClass` which depend on `annotations` added to the `SerialDescriptor` interface, which are also overridden in `CommonSerialDescriptor` for lazy caching. If this brings about unintended breaking changes to the existing core API, I can refactor them to be extension functions in a `Polymorphic.kt` file like the one in `kotlinx.serialization.json.internal`.

The added function names may have better alternatives too.

## Some more possible unimplemented APIs

### An independent version of  the `@SerialPolymorphicNumber` annotation

We can also add a version of  the `@SerialPolymorphicNumber` annotation which is independent of the base class:

```kotlin
@Serializable
@UseSerialPolymorphicNumbers
sealed class Sealed {
    @Serializable
    @IndependentSerialPolymorphicNumber(1)
    object Case : Sealed()
}
```

### Complements to the annotations in the `SerializersModule` DSL

In addition to the annotations, there can be completion APIs to set them in the `SerializersModule` DSL, to ensure this requirement for all data, or for external classes which can't be marked with annotations. So we can add the following APIs:

1. `allUseSerialPolymorphicNumbers` property in `SerializersModuleBuilder` to require all polymorphic serialization to use numbers, in case there are missing ones.
1. `useSerialPolymorphicNumbers` property in `PolymorphicModuleBuilder` as a complement to `@UseSerialPolymorphicNumbers`.
1. An additional `serialPolymorphicNumber` argument in `PolymorphicModuleBuilder.subclass` as a complement to `@SerialPolymorphicNumber`.

Example:

```kotlin
SerializersModule {
    allUseSerialPolymorphicNumbers = true
    polymorphic(Project::class) {
        useSerialPolymorphicNumbers = true
        subclass(OwnedProject::class, serialPolymorphicNumber = 1)
        defaultDeserializer { BasicProject.serializer() }
    }
}
```

### Other uncompleted tasks

There are some KDocs I left empty with `TODO` since the APIs are not finalized yet.

The mardown docs in Kotlin Serialization Guide are also not updated yet.

pdvrieze:
There is a lot of added complexity and special casing in this code. The same outcome could be achieved with a custom implementation of a polymorphic serializer (that maps between numbers and serial names), that can be marked as serializer for the base type. There are then only a few additions needed from serialization itself:
- A nicer way to enumerate all subtypes/subserializers for a base type (this is possible now, but not quite elegant)
- Some way for the serialization plugin to enforce (or warn about) requirements on subtypes (such as the need to specify an annotation).

Note also that your numbers system is a challenge as the numbers need to be unique (possibly within the entire program) if they are to completely replace serial names (on types).

recursive-rat4:
>Alternative names for "serial polymorphic number"

Number is ambiguous here, I typically use `UByte`.

ShreckYe:
@pdvrieze Thanks for the reply.

> The same outcome could be achieved with a custom implementation of a polymorphic serializer (that maps between numbers and serial names), that can be marked as serializer for the base type.

This is indeed the approach I am currently adopting in my projects. However, I think it's more verbose and separates the discriminator numbers from the annotation configurations (like `@ProtoNumber`). So I think supporting this with annotations still looks more elegant and provides a better separation of concerns.

> * A nicer way to enumerate all subtypes/subserializers for a base type (this is possible now, but not quite elegant)

This looks good, but I think it doesn't help that much if such information is only available at runtime and such exhaustiveness checks are not available at compile time.

> * Some way for the serialization plugin to enforce (or warn about) requirements on subtypes (such as the need to specify an annotation).

Yeah this looks like a general and elegant way.

> Note also that your numbers system is a challenge as the numbers need to be unique (possibly within the entire program) if they are to completely replace serial names (on types).

In the current implementation, the numbers are only unique in the scope of one certain base class. That's why a `baseClass` argument needs to be passed to `@SerialPolymorphicNumber`.

ShreckYe:
@recursive-rat4 Yeah perhaps "integer" is a better word. I still think using an `Int` or a `UInt` is a better idea here. Though it's unlikely that a parent class has more than 256 subclasses, using an `Int`/`UInt` covers such cases and is as fast as a `Byte`/`UByte` because of our 32-bit/64-bit CPU architectures.

PR #2588, opened 2024-03-03
Should `findPolymorphicSerializer...` functions that return `SerializationStrategy` be renamed to `findPolymorphicDeserializer...`?

ShreckYe:
These functions return `DeserializationStrategy`s. It's a bit misleading. The word "deserializer" is used in other places such as `defaultDeserializer` to refer to a `DeserializationStrategy`.

If breaking compatibility is an issue the old ones can be kept as aliases and marked as deprecated.

These are minor possible improvement I noticed while working on another PR. Just close this PR if I got this wrong.

### DIFF
--- a/core/api/kotlinx-serialization-core.api
+++ b/core/api/kotlinx-serialization-core.api
@@ -70,7 +70,7 @@ public final class kotlinx/serialization/PolymorphicSerializer : kotlinx/seriali
 }
 
 public final class kotlinx/serialization/PolymorphicSerializerKt {
-	public static final fun findPolymorphicSerializer (Lkotlinx/serialization/internal/AbstractPolymorphicSerializer;Lkotlinx/serialization/encoding/CompositeDecoder;Ljava/lang/String;)Lkotlinx/serialization/DeserializationStrategy;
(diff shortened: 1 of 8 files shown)


pdvrieze:
Generally, there are 3 concepts: `SerializationStrategy`, `DeserializationStrategy` and `Serializer` that combines both. Serializer is generally used as name to refer to either. In this case the changed name could be used, but as this is public API I don't think it is worth it to try to rename it (the original names need to be retained anyway).

ShreckYe:
@pdvrieze Got it. Thanks for clarifying.

sandwwraith:
> but as this is public API I don't think it is worth it to try to rename it (the original names need to be retained anyway)

The declarations in question is marked with `@InternalSerializationApi`, so in theory, we could change it without preserving the original name. However, I do not see much profit in doing so just for change's sake. The term 'serializer' is used in the codebase quite freely, referring either to `KSerializer`, `SerializationStrategy`, or `DeserializationStrategy`. It is clear from the return type that we're getting `DeserializationStrategy` here.

PR #2589, opened 2024-03-03
Fix all misuses of the article "the" before a class name enclosed in backticks

ShreckYe:
This is similar to #2562. I found there are more of them.


sandwwraith (approved):
Thank you!

PR #2629, opened 2024-04-15
added decodeFromStringOrNull

amitgcse:
The function will provide a way to try to serialize a given JSON into an object and return null if the deserialization fails.

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -94,6 +94,13 @@ public sealed class Json(
      */
     public inline fun <reified T> decodeFromString(@FormatLanguage("json", "", "") string: String): T =
             decodeFromString(serializersModule.serializer(), string)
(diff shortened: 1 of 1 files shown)


sandwwraith:
Hi! Since this is your first contribution, I recommend you familiarize yourself with contributing guidelines (https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md), especially with "Follow the Kotlin Coding Conventions" and "If you introduce any new public APIs" sections.

Regarding the proposed API itself: in my opinion, it promotes bad practices. There are a limited number of cases where you want to drop the (any!) exception entirely and replace it with `null`, and you should do it with a full understanding of the downsides of this approach. Typically, exceptions should be logged and/or shown to the user. Also, catching any exception instead of just `SerializationException` or `IllegalStateException` may hide potential flaws or bugs in the application. 
Therefore, any exception handling should be performed by the user, not by the serialization library. It is pretty easy to write such an extension function yourself.

PR #2633, opened 2024-04-15
feat: Make SerialDescriptorForNullable.original public

### LINKED ISSUE #2631: Make accessible the original descriptor when wrapped into nullable
I'm currently working for the avro4k library that is the implementation of avro format with this powerful kotlin serialization library. 

We need to generate schemas usually more complex than just based on descriptors' kinds. To do that, it was previously done by creating a custom class implementing a `SerialDescriptor` and add custom methods to generate the schemas according to a custom logic. The main issue with it is that, just for nullable descriptor (with `descriptor.nullable` done by the plugin), with need to use kotlin reflection to get back the original descriptor because it is a private field. To bypass it, we had to make a complex system that discovers custom annotations and generate a schema according those annotations. 

But still, the simplest way would be to just access to the original descriptor. It allows us to decouple the schema generation system from the custom logics inferred by the custom serializers.

Here is the example of what 

Just make public the `original` field inside the `SerialDescriptorForNullable` data class.

A second option would be to have the same extension method as `public val SerialDescriptor.capturedKClass: KClass<*>?`:
```kotlin
public val SerialDescriptor.unwrapOriginalDescriptor: SerialDescriptor
    get() = when (this) {
        is SerialDescriptorForNullable -> this.original
        else -> this
    }
```

I can do the PR if needed

Chuckame:
Closes #2631

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -224,6 +224,24 @@ public val SerialDescriptor.nullable: SerialDescriptor
     }
 
 /**
+ * Returns non-nullable serial descriptor for the type if this descriptor has been auto-generated (plugin
+ * generated descriptors) or created with `.nullable` extension on a descriptor or serializer.
+ *
+ * Otherwise, returns this.
+ *
+ * It may return nullable descriptor if this descriptor has been created manually as nullable by directly implementing SerialDescriptor interface.
+ *
+ * @see SerialDescriptor.nullable
+ * @see KSerializer.nullable
+ */
+@ExperimentalSerializationApi
+public val SerialDescriptor.nonNullOriginal: SerialDescriptor
+    get() = when (this) {
+        is SerialDescriptorForNullable -> original
+        else -> this
+    }
+
+/**
  * Builder for [SerialDescriptor] for user-defined serializers.
  *
  * Both explicit builder functions and implicit (using reified type-parameters) are present and are equivalent.
--- a/core/api/kotlinx-serialization-core.api
+++ b/core/api/kotlinx-serialization-core.api
@@ -308,6 +308,7 @@ public final class kotlinx/serialization/descriptors/SerialDescriptorsKt {
 	public static synthetic fun buildClassSerialDescriptor$default (Ljava/lang/String;[Lkotlinx/serialization/descriptors/SerialDescriptor;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lkotlinx/serialization/descriptors/SerialDescriptor;
 	public static final fun buildSerialDescriptor (Ljava/lang/String;Lkotlinx/serialization/descriptors/SerialKind;[Lkotlinx/serialization/descriptors/SerialDescriptor;Lkotlin/jvm/functions/Function1;)Lkotlinx/serialization/descriptors/SerialDescriptor;
 	public static synthetic fun buildSerialDescriptor$default (Ljava/lang/String;Lkotlinx/serialization/descriptors/SerialKind;[Lkotlinx/serialization/descriptors/SerialDescriptor;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lkotlinx/serialization/descriptors/SerialDescriptor;
+	public static final fun getNonNullOriginal (Lkotlinx/serialization/descriptors/SerialDescriptor;)Lkotlinx/serialization/descriptors/SerialDescriptor;
 	public static final fun getNullable (Lkotlinx/serialization/descriptors/SerialDescriptor;)Lkotlinx/serialization/descriptors/SerialDescriptor;
 	public static final fun listSerialDescriptor (Lkotlinx/serialization/descriptors/SerialDescriptor;)Lkotlinx/serialization/descriptors/SerialDescriptor;
 	public static final fun mapSerialDescriptor (Lkotlinx/serialization/descriptors/SerialDescriptor;Lkotlinx/serialization/descriptors/SerialDescriptor;)Lkotlinx/serialization/descriptors/SerialDescriptor;
--- a/core/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
+++ b/core/commonTest/src/kotlinx/serialization/SerialDescriptorBuilderTest.kt
@@ -94,4 +94,44 @@ class SerialDescriptorBuilderTest {
         assertTrue(descriptor.isNullable)
         assertEquals("my.Simple?", descriptor.serialName)
     }
+
+    @Test
+    fun testNonNullOriginal() {
+        listOf(
+            buildClassSerialDescriptor("my.Simple") {},
+            Boolean.serializer().descriptor,
+            String.serializer().descriptor,
+            ListSerializer(Int.serializer()).descriptor,
+        ).forEach { originalDescriptor ->
+            // Unwrapping original descriptor when it is not nullable should return the same descriptor (no-op operation)
+            assertSame(originalDescriptor.nonNullOriginal, originalDescriptor)
+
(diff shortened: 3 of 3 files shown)


sandwwraith (changes requested):
You should also do a couple more things:

1. Update public API dump (see https://github.com/Kotlin/kotlinx.serialization/blob/master/CONTRIBUTING.md#updating-the-public-api-dump)
2. Add some tests for the change — that `x.nullable.unwrapNullable === x`, that for `@Serializable class X(val s: String?)`, `val n = X.serializer().descriptor.getElementDescriptor(0)` results in `n.isNullable == true` and `n.unwrapNullable == String.serializer().descriptor`, for your use case, etc.

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -223,6 +223,17 @@ public val SerialDescriptor.nullable: SerialDescriptor
         return SerialDescriptorForNullable(this)
     }
 
+/**
+ * Returns the serial descriptor previously transformed to nullable with [SerialDescriptor.nullable] back
+ * to its original form.
+ */
+@ExperimentalSerializationApi
+public val SerialDescriptor.unwrapNonNull: SerialDescriptor
```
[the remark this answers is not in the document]
Chuckame: I hesitated. Done

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -223,6 +223,17 @@ public val SerialDescriptor.nullable: SerialDescriptor
         return SerialDescriptorForNullable(this)
     }
 
+/**
+ * Returns the serial descriptor previously transformed to nullable with [SerialDescriptor.nullable] back
```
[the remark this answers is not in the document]
Chuckame: Yeah, writing this doc made my brain blurry 😄 I like your doc. Done

Chuckame:
Following the contributing guide, I've rebased onto `dev` (but the PR is on `master` ⚠️), tests have been added, and doc+knit+apiDump executed @sandwwraith ! 🚀

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -223,6 +223,20 @@ public val SerialDescriptor.nullable: SerialDescriptor
...
 
+/**
+ * Returns non-nullable serial descriptor for the type, if [this] is a nullable descriptor created automatically or
+ * with `.nullable` extensions on a descriptor or serializer. Otherwise, returns [this].
+ *
+ * @see SerialDescriptor.nullable
+ * @see KSerializer.nullable
+ */
+@ExperimentalSerializationApi
+public val SerialDescriptor.unwrapNullable: SerialDescriptor
```
sandwwraith, line 234: After looking at the api dump, I've realized that `unwrap` is actually a verb, and its counter-intuitive to have a verb in property name (https://kotlinlang.org/docs/coding-conventions.html#choose-good-names). So it should be either

1. A function `SerialDescriptor.unwrapNullable()`
or
2. A property with a noun, e.g. `unwrappedNullable`.

I'm not sure what is better. Doing (1) will bring asymmetry with `.nullable`. (2) is confusing, because it is not clear what 'unwrapped' is since we do not have 'wrapped' anywhere else. Besides, 'unwrappedNullable' sounds like some flavor of a nullable thing, but in reality, returned descriptor is non-nullable.

Chuckame: You're right, I'm feeling the same. I would prefer `unwrappedNullable` to not have this asymmetry. I'll commit now. If you feel not comfortable, could we have a third person reviewing this to help chosing a better name ?

sandwwraith: @shanshin or @qwwdfsad may help here, I guess

Chuckame: Or, we can do `unwrappedOriginal` and also unwrap `WrappedSerialDescriptor.original` (I've looked at all the SerialDescriptors and only SerialDescriptorForNullable and WrappedSerialDescriptor would have sense of unwrapping the original descriptor)

Chuckame: I like the idea, but it would mean also renaming the current data class  `WrappedSerialDescriptor` to something else to allow creation of the interface. What do you think @sandwwraith ?

sandwwraith: > the route of introducing a WrappedSerialDescriptor interface would make a lot of sense, it would also future proof the design to support other wrapping descriptors

IMO this is way out of scope of this PR.

sandwwraith: WDYT of `nonNullableOriginal` name? We can add to the documentation that this property "does not affect descriptors that were created as nullable by directly implementing SerialDescriptor interface."

Chuckame: New doc, is it okay for you @sandwwraith ?
```

 * Returns non-nullable serial descriptor for the type, only if this descriptor has been auto-generated (plugin
 * generated descriptors) or with `.nullable` extension on a descriptor or serializer.
 *
 * Otherwise, returns this.
 * 
 * It may return nullable descriptor if this descriptor has been created manually as nullable by directly implementing SerialDescriptor interface.
 *
```

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -223,6 +223,24 @@ public val SerialDescriptor.nullable: SerialDescriptor
         return SerialDescriptorForNullable(this)
     }
 
+/**
+ * Returns non-nullable serial descriptor for the type, only if this descriptor has been auto-generated (plugin
+ * generated descriptors) or with `.nullable` extension on a descriptor or serializer.
```
sandwwraith, line 228:
```suggestion
 * generated descriptors) or created with `.nullable` extension on a descriptor or serializer.
```

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -223,6 +223,24 @@ public val SerialDescriptor.nullable: SerialDescriptor
         return SerialDescriptorForNullable(this)
     }
 
+/**
+ * Returns non-nullable serial descriptor for the type, only if this descriptor has been auto-generated (plugin
```
sandwwraith, line 227:
```suggestion
 * Returns non-nullable serial descriptor for the type, if this descriptor has been auto-generated (plugin
```

sandwwraith (approved):
Yep, looks good to me

sandwwraith:
@Chuckame You need to update apiDump with tool: `getNonNullOriginal` alphabetically goes before `getNullable`

Chuckame:
@sandwwraith it's ready, apiDump done, doc updated. When do you think it could be released ?

sandwwraith:
@Chuckame We usually do releases around Kotlin's release. Given that 2.0.0 is just around the corner, I think it will be in about a month.

Chuckame:
Ok thanks, so I'll wait this PR release before releasing on my side.

sandwwraith:
Thank you!

Chuckame:
Can I access to a snapshot build until the release ?

sandwwraith:
Unfortunately, we don't do snapshot builds. You can build it yourself and publish it to your maven local (https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/building.md) if you want to test the changes with your library

PR #2638, opened 2024-04-17
Documentation improvement: initial Json samplyification

qwwdfsad:


sandwwraith:
Should we use a `@sample` tag or something? How can we show to users what `println` in our samples actually prints?

qwwdfsad:
Discussed it internally that `@sample` will be introduced later when IDE part will be figured

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -27,29 +27,48 @@ import kotlin.native.concurrent.*
...
+ * data class Data(val id: Int, val data: String, val extensions: JsonElement)
  *
- * val json = Json
- * val instance = DataHolder(42, "some data", buildJsonObject { put("additional key", "value") }
+ * val json = Json { prettyPrint = true }
+ * val instance = Data(42, "some data", buildJsonObject { put("key", "value") })
  *
- * // Plain StringFormat usage
- * val stringOutput: String = json.encodeToString(instance)
+ * // Plain Json usage: returns '{"id": 42, "some data", "extensions": {"key": "value" } }'
```
sandwwraith, line 35: Given that you've enabled `prettyPrint = true` for a `json` instance, it will print a different result. I think it is possible here to use a default `Json` or provide another configuration flag in case you need to show instance creation. (But we need to show that we have a default `Json` instance, too).

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -265,21 +402,31 @@ public class JsonBuilder internal constructor(json: Json) {
...
+     * class Key(val type: String, val opens: String)
+     * val pretty = Json { prettyPrint = true }
+     * /*
+     *  * Prints
+     *  * {
+     *  *     "type": "keycard",
+     *  *     "opens": "secret door"
+     *  * }
+     *  */
+     * println(pretty.encodeToString(Pair("keycard", "secret door")))
```
sandwwraith, line 420:
```suggestion
     * println(pretty.encodeToString(Key("keycard", "secret door")))
```

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -295,22 +442,12 @@ public class JsonBuilder internal constructor(json: Json) {
      */
     public var coerceInputValues: Boolean = json.configuration.coerceInputValues
```
sandwwraith, line 443: I think this setting would really benefit from examples

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -349,37 +478,38 @@ public class JsonBuilder internal constructor(json: Json) {
...
+     * With this feature enabled, it is no longer possible to decode enum values that have the same name in a lowercase form.
      * The following code will throw a serialization exception:
-     *
      * ```
-     * enum class BadEnum { Bad, BAD }
-     * val j = Json { decodeEnumsCaseInsensitive = true }
-     * j.decodeFromString<Box<BadEnum>>("""{"boxed":"bad"}""")
+     * enum class CaseSensitiveEnum { One, ONE }
+     * val json = Json { decodeEnumsCaseInsensitive = true }
+     * // Fails with SerializationException: ambiguous enum decoding
```
sandwwraith, line 503: We do not have such exception text as 'ambiguous enum decoding', do we?

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -295,22 +442,12 @@ public class JsonBuilder internal constructor(json: Json) {
...
-     * `false` by default.
-     *
-     * This option can only be used if [classDiscriminatorMode] in a default [ClassDiscriminatorMode.POLYMORPHIC] state.
-     */
-    public var useArrayPolymorphism: Boolean = json.configuration.useArrayPolymorphism
-
     /**
      * Name of the class descriptor property for polymorphic serialization.
-     * "type" by default.
+     * `"type"` by default.
```
sandwwraith, line 447: If you want to use code-like formatting, then I think we should omit quotes, since value is not quoted:

```suggestion
     * `type` by default.
```

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -27,29 +27,48 @@ import kotlin.native.concurrent.*
  * Example of usage:
  * ```
  * @Serializable
- * class DataHolder(val id: Int, val data: String, val extensions: JsonElement)
+ * data class Data(val id: Int, val data: String, val extensions: JsonElement)
  *
- * val json = Json
- * val instance = DataHolder(42, "some data", buildJsonObject { put("additional key", "value") }
+ * val json = Json { allowTrailingComma = true }
```
sandwwraith, line 32:
```suggestion
 * val json = Json { ignoreUnknownKeys = true }
```

sandwwraith (approved):

PR #2642, opened 2024-04-18
Improve test for JVM intrinsics:

sandwwraith:
Old one used time-based approach. It was a good indicator, but in rare circumstances it may have been flaky and produced incorrect results. Flaky time-based tests are problematic for large builds, such as Kotlin's Aggregate build.

New approach is based on cache presence and should not give incorrect results.

See also #KTI-1726


core/jvmTest/src/kotlinx/serialization/CachingTest.kt
```diff
@@ -43,4 +43,21 @@ class CachingTest {
 
         assertEquals(1, factoryCalled)
     }
+
+    @Serializable
+    class Target
```
shanshin, line 48: There are no other branches of intrinsic, like value classes, e.t.c?

It may not be enough to check only the intrinsic of a regular class `class Target`

shanshin (approved):

core/jvmTest/src/kotlinx/serialization/CachingTest.kt
sandwwraith: There aren't any branches depending on class type. Either intrinsics are triggered and replacing `serializer<T>()` with _something_, or they are not. We have different branches for caches though (parametrized vs non-parametrized), but that is not what being tested here.

Issue #1533, opened 2021-06-04
Can't decode String to ByteArray


ccc-sseylani:
**Describe the bug**
Hi there, 
I am facing a really weird issue.
I am fetching a JSON file from Api (with ktor). But I am getting this issue
I am expecting to receive a list of Asset class but it keeps complaining that
 """
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 15: Expected start of the array '[', but had ':' instead
JSON input: {"asset":[{"id":"AzWSKY5\/Eeu06AJCwKhABQ==",".....
"""
Json format is in right format, and I can easily get the desired  result with GSON and Retrofit.
Is this a bug with kotlin serialization?

**To Reproduce**
Attach a code snippet or test data if possible.

<img width="1262" alt="image" src="https://user-images.githubusercontent.com/75082292/120816712-01d00480-c51f-11eb-82c6-a200a9d4741c.png">

<img width="1186" alt="image" src="https://user-images.githubusercontent.com/75082292/120816823-1e6c3c80-c51f-11eb-8464-b1f55f50d012.png">

<img width="890" alt="image" src="https://user-images.githubusercontent.com/75082292/120816909-35129380-c51f-11eb-9a18-56ee4659fc41.png">

**Expected behavior**

Accept the Json file and convert it to AppData object.

**Environment**
 - Kotlin version: [1.5.0] 
 - Library version: [1.5.0]
 - Kotlin platforms: [e.g. JVM, JS, Native or their combinations]
 - Gradle version: [e.g. 4.10]
 - IDE version (if bug is related to the IDE) [e.g. IntellijIDEA 2019.1, Android Studio 3.4]
 - Other relevant context [e.g. OS version, JRE version, ... ]

sandwwraith:
What version do you use? There's no `1.6.0` version

ccc-sseylani:
> What version do you use? There's no `1.6.0` version

sorry, KTOR is 1.6.0. Library version is 1.5.0

sandwwraith:
There's no 1.5.0 either

ccc-sseylani:
> There's no 1.5.0 either
It was 1.0.0
just updated to 1.2.1
Still same error
<img width="1179" alt="image" src="https://user-images.githubusercontent.com/75082292/120826986-faadf400-c528-11eb-9087-b7b3e73d4e6e.png">
[...]

PR #2644, opened 2024-04-18
Add how to encode/decode Base64 formats example

GeorgePap-719:
I propose to add examples on how we can encode/decode Base64 formats, based on the demand from users
(see [#1533](https://github.com/Kotlin/kotlinx.serialization/issues/1533)).

**A bit on why I don't recommend adding `ByteArrayAsBase64Serializer` as a builtin serializer**

**tl;dr** To Provide this "feature" as a library, we have to make some trade-offs and the resulting
API might not be worth it. It is simpler to be solved in the user's codebase directly choosing their use-case. 

There are two variants of the Base64 RFC in the wild that are considered the default.
For example, see [Jackson's default implementation](https://github.com/FasterXML/jackson-core/blob/4d5028685ba30dd44ff81537790d2c25dd1339a3/src/main/java/com/fasterxml/jackson/core/Base64Variants.java#L106) where it uses a variation of Mime Base64. It seems Jackson "treats" the Base64 Mime RFC as the most standard. On the other hand, Kotlin's and Java's implementations consider the [RFC-4648](https://www.rfc-editor.org/rfc/rfc4648#section-4) as the default one. Based on Jackson's heavy usage by the community (e.g. spring uses it as the default serializer) this might lead to unexpected compatibility issues.

**Disadvantages:**

- We have to provide a thorough test-suite to catch any potential compatibility issues.
- We will have to decide which variation to provide as the default, as most users and frameworks
  will use it by default.
- It's not clear-cut if we should also provide serializers for all the variations, but in case we
  do, it might not be worth it for the increased API surface (since it can be easily solved in
  user's codebase).

**Other libraries:** Researching other serialization libraries such as `moshi` and `gson` looks
like neither of them provide Base64 integration out of the box. However, `moshi` provides an example
on how to create a serializer to encode/decode Base64 formats.

A scenario where I see value in adding a `ByteArrayAsBase64Serializer`, and potentially all its
variations provided by Kotlin's stdlib, is when we have use cases where other formats except `Json` require it.


sandwwraith (changes requested):
I agree with your reasoning not to provide a built-in `ByteArrayAsBase64Serializer`. A documentation section for it would be a nice addition!

docs/json.md
```diff
@@ -510,6 +511,82 @@ Consult their documentation for details.
...
+  val value = Value(string.toByteArray())
+  val encoded = Json.encodeToString(value)
+  val decoded = Json.decodeFromString<Value>(encoded)
+  println(decoded.base64Input.decodeToString())
+}
+```
+
+> You can get the full code [here](../guide/example/example-json-13.kt)
+
+Note that some serializers use different RFC for base64 encoding by default. For example, Jackson uses a variant of
```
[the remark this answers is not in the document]
GeorgePap-719: Good point.

GeorgePap-719:
Heads up: I added a small advice on specifying a global serializer using type alias, see last commit.

docs/json.md
```diff
@@ -21,6 +21,7 @@ In this chapter, we'll walk through features of [JSON](https://www.json.org/json
   * [Allowing special floating-point values](#allowing-special-floating-point-values)
   * [Class discriminator for polymorphism](#class-discriminator-for-polymorphism)
   * [Class discriminator output mode](#class-discriminator-output-mode)
+  * [Base64](#base64)
```
sandwwraith, line 24: Is there any particular reason for you to decide that it should be placed here in the list of features? Usually we add new entries to the end of the list, i.e. after 'Global naming strategy' in this case.

GeorgePap-719: Followed your advice from last [attempt](https://github.com/Kotlin/kotlinx.serialization/pull/1676#discussion_r725494382)

sandwwraith: I see.

> after 'class discriminator' section

And that was the last section in that list at the time :) So I would kindly ask you to move it a little bit down to the end of 'Json configuration' feature list. It is not ideal (since it is not actually related to `Json {}` config), but I can't think right now of a better place

sandwwraith (approved):
Thank you for all your efforts!

PR #2648, opened 2024-04-22
Refine exception messages in case of deserializing data from JsonElement.

### LINKED ISSUE #2630: JsonDecoder: decodeSerializableValuePolymorphic - Include currentTag in cast error
`JsonDecodingException: Expected class kotlinx.serialization.json.JsonObject (Kotlin reflection is not available) as the serialized body of F, but had class kotlinx.serialization.json.JsonLiteral (Kotlin reflection is not available)`

This error message is not very helpful without the property name, but the name is available: `currentTag`.

To reproduce:
```kotlin
@Serializable
sealed interface F {
  @Serializable
  @SerialName("a")
  data class A(val f: F): F
}

Json.decodeFromString(F.serializer(), """{"type":"a", "f":"wrong-input"}""")
```

`JsonDecodingException: Serial name f expected class kotlinx.serialization.json.JsonObject (Kotlin reflection is not available) as the serialized body of F, but had class kotlinx.serialization.json.JsonLiteral (Kotlin reflection is not available)`

sandwwraith:
Such a code path is often used when we cannot find a type discriminator as a first key in Json (for example, if json input is invalid, and we get a string instead of an object). In such cases, we should display a nice error message.

Also, add a tag stack — the equivalent of a Json path — to most of the error messages. Note that it is far from ideal since changing between string and tree decoders (such happens in polymorphism) won't preserve stack or path correctly. Yet, it is the best we can do for now.

Fixes #2630

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonErrorMessagesTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonErrorMessagesTest.kt
@@ -6,11 +6,79 @@
 package kotlinx.serialization.json
 
 import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
 import kotlinx.serialization.test.*
 import kotlin.test.*
 
 
 class JsonErrorMessagesTest : JsonTestBase() {
+
+    @Serializable
+    @SerialName("app.Failure")
+    sealed interface Failure {
+        @Serializable
+        @SerialName("a")
+        data class A(val failure: Failure) : Failure
+    }
+
+    @Test
+    fun testPolymorphicCastMessage() = parametrizedTest { mode ->
+        checkSerializationException({
+            default.decodeFromString(
+                Failure.serializer(),
+                """{"type":"a", "failure":"wrong-input"}""",
+                mode
+            )
+        }, {
+            assertContains(
+                it,
+                "Expected JsonObject, but had JsonLiteral as the serialized body of app.Failure at element: \$.failure"
+            )
+        })
+    }
+
+    @Test
+    fun testPrimitiveInsteadOfObjectOrList() = parametrizedTest { mode ->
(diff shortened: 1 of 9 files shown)


formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
```diff
@@ -68,7 +68,7 @@ private open class DynamicInput(
     }
 
     override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
-        return decodeSerializableValuePolymorphic(deserializer)
+        return decodeSerializableValuePolymorphic(deserializer) { renderTagStack() }
```
shanshin, line 71:
```suggestion
        return decodeSerializableValuePolymorphic(deserializer, ::renderTagStack)
```

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonErrorMessagesTest.kt
```diff
@@ -6,12 +6,80 @@
...
+                assertContains(message, "Unexpected JSON token at offset 10: Expected numeric literal at path: \$.boxed")
+        })
+
+        checkSerializationException({
+            default.decodeFromString(Box.serializer(String.serializer()), """{"boxed": {"x":"y"}}""", mode)
+        }, { message ->
+            if (mode == JsonTestingMode.TREE)
+                assertContains(message, "Expected JsonPrimitive, but had JsonObject as the serialized body of string at element: \$.boxed")
+            else
+                assertContains(message, "Expected beginning of the string, but got { at path: \$.boxed")
```
shanshin, line 79: `Unexpected JSON token at offset 10: ` ?

shanshin (approved):

PR #2661, opened 2024-05-03
Stabilize explicitNulls feature:

### LINKED ISSUE #2586: coerceInputValues=true and explicitNulls=false should decode unknown enum value without default to null
**Describe the bug**
We updated from 1.6.2 to 1.6.3 and now nullable enums without explicit null default don't get decoded anymore, leading to an exception.

**To Reproduce**
```
val json = Json {
    ignoreUnknownKeys = true
    isLenient = true
    coerceInputValues = true
    explicitNulls = false
}

@Serializable
data class Outer(
    @SerialName("service")
    val service: Service?,
)

@Serializable
enum class Service {
    @SerialName("card")
    CARD,
}

val serviceJson = """{"service":"test"}"""
json.decodeFromString<Outer>(serviceJson)
```

SerializationException: org.example.Service does not contain element with name 'test' at path $.service

**Expected behavior**
We were expecting that these configs (taken from the documentation) get "added up" together:
* coerceInputValues = true:
* * Enables coercing incorrect JSON values to the default property value in the following cases:
* * Property type is an enum type, but JSON value contains unknown enum member
* explicitNulls = false:
* * When this flag is disabled [...] during decoding, the absence of a field value is treated as null for nullable properties without a default value.

In other words: For a nullable enum without default value, set its value to "null" if the JSON value is an unknown enum member.

**Other**
Apart from combining these two config properties, I wouldn't know how we can get automatic nulls? We would have to set an explicit null to each nullable enum?
Also see #2585.

**Environment**
 - Kotlin version: 1.9.22
 - Library version: Bug introduced in 1.6.3, working in 1.6.2
 - Kotlin platforms: JVM
 - Gradle version: 8.6
 - Other relevant context: OpenJDK 21.0.2

### LINKED ISSUE #2636: Stabilize explicitNulls in Json
It is here from 1.3, we get the overall positive feedback and there is not many reasons to keep it experimental any longer

sandwwraith:
- Bring back its interaction with coerceInputValues flag
- Enhance documentation and add more samples to it
- Remove @ExperimentalSerializationApi

Fixes #2636
Fixes #2586


docs/json.md
```diff
@@ -313,10 +268,94 @@ while nullable properties `version` and `description` are filled with their defa
...
+
+This property only affects decoding. It treats a limited subset of invalid input values as if the
+corresponding property was missing.
+The current list of supported invalid values is:
+
+* `null` inputs for non-nullable types
+* unknown values for enums
+
+If value is missing, it is replaced either with a default property value if it exists,
+or with a `null` if [explicitNulls](#explicit-nulls) flag is set to `false` and a property is nullable.
```
shanshin, line 297: > or with a `null` if [explicitNulls](#explicit-nulls) flag is set to `false` and a property is nullable.

I think we should point out that this only applies to enums

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonCoerceInputValuesTest.kt
```diff
@@ -39,6 +39,16 @@ class JsonCoerceInputValuesTest : JsonTestBase() {
         val e: SampleEnum
     )
 
+    @Serializable
+    data class NullableEnumWithoutDefault(
```
shanshin, line 43: There is already exist class `NullableEnumHolder`

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonCoerceInputValuesTest.kt
```diff
@@ -142,4 +152,23 @@ class JsonCoerceInputValuesTest : JsonTestBase() {
             assertEquals(e2.message, e1.message)
         }
     }
+
+    @Test
+    fun testNullableEnumWithoutDefault() {
+        val j = Json(json) { explicitNulls = false }
+        parametrizedTest { mode ->
+            assertEquals(NullableEnumWithoutDefault(null), j.decodeFromString("{}"))
+            assertEquals(NullableEnumWithoutDefault(null), j.decodeFromString("""{"e":"incorrect"}"""))
```
shanshin, line 161: I think we should add test with `explicitNulls = true` to be sure that the error is thrown on unknown enum value

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -434,9 +458,32 @@ public class JsonBuilder internal constructor(json: Json) {
...
     /**
-     * Enables coercing incorrect JSON values to the default property value (if exists) in the following cases:
+     * Enables coercing incorrect JSON values in the following cases:
+     *
      *   1. JSON value is `null` but the property type is non-nullable.
-     *   2. Property type is an enum type, but JSON value contains unknown enum member.
+     *   2. Property type is an enum type, but JSON value contains an unknown enum member.
+     *
+     * Coerced values are treated as missing; they are replaced either with a default property value if it exists, or with a `null` if [explicitNulls] flag
+     * is set to `false` and a property is nullable.
```
shanshin, line 467: same about enums

docs/json.md
sandwwraith: It is mentioned above, but OK

Issue #2608, opened 2024-03-20
Nested classes in ProtoBuf produce empty array for native release binary


whyoleg:
**Describe the bug**

Nested classes in ProtoBuf produce empty array for native release binary.
No such issue with Cbor or with any other target and Native debug binary. (all cases are provided in repository with reproducer)

**To Reproduce**
```kotlin

@Serializable
data class OuterClass(val data: InnerClass)

@Serializable
data class InnerClass(val data: String)

@OptIn(ExperimentalSerializationApi::class, ExperimentalStdlibApi::class)
class SerializationTest {
    @Test
    fun testProtobufNested() {
        val model = OuterClass(InnerClass("something"))

        val data = ProtoBuf.encodeToByteArray(OuterClass.serializer(), model)
        // `data` is an empty array on Native in release mode
        assertEquals("0a0b0a09736f6d657468696e67", data.toHexString()) 
    }
}
```
Repository with reproducer: https://github.com/whyoleg/kx-serialization-protobuf-reproducer
To reproduce: `./gradlew macosArm64ReleaseTest` or `[TARGET_NAME]ReleaseTest` for other targets, or just `check` to run all tests (to compare behaviour)

**Expected behavior**

Correct serialised value is returned independent to the target or debug/release mode.

**Environment**
 - Kotlin version: 1.9.23 (same for 2.0.0-Beta5)
 - Library version: 1.6.3
 - Kotlin platforms: Native

Note: I can create the same issue in Kotlin YT, as most likely the issue is there, but thought it would be better to create it here to keep track and may be introduce some kind of tests in repository.

sandwwraith:
Thanks for report. It happens for me for both `linuxX64ReleaseTest` on Linux machine and `macosArm64ReleaseTest` on Apple Silicon machine. Looks like the Native issue though.

whyoleg:
just in case: does it make sense additionally run tests in `RELEASE` mode to the library (and possibly other `kotlinx.*` libraries), to be proactive with such bugs?
If you think it could be useful, I'm ready to provide PR.

sandwwraith:
Yes, it makes sense. I'll ask Native team additionally of the best way to do so. We are now refactoring our build scripts (#2654), so you'll have to wait with your PR a little bit :)

PR #2667, opened 2024-05-09
Setup running native tests in release mode

whyoleg:
Proposed here: https://github.com/Kotlin/kotlinx.serialization/issues/2608#issuecomment-2077438394

To run native tests in release mode, separate tasks are created by KGP for additional test runs, f.e `macosArm64ReleaseTest` or `linuxX64ReleaseTest`. By default, they will be run also on `build`, `check` and `allTests` tasks, so they should automatically run on TC.

Note: with Kotlin 1.9.22 the original issue (#2608) doesn't reproduce nor via running `protobuf` tests (which contains tests for inner classes, so I haven't added additional test from original issue), nor via running tests in `integration-test` module (by adding reproducer test there, not included)

But another tests failed there:
* `kotlinx.serialization.json.polymorphic.JsonListPolymorphismTest#testPolymorphicNullableValues`
* `kotlinx.serialization.json.polymorphic.JsonMapPolymorphismTest#testPolymorphicNullableValues`

With the same exception: 
```
kotlinx.serialization.json.internal.JsonDecodingException: Expected class kotlinx.serialization.json.JsonArray as the serialized body of kotlinx.serialization.Polymorphic<InnerBase>, but had class kotlinx.serialization.json.JsonObject
```

Those tests are now finish successfully with Kotlin 2.0.0-RC3.

### DIFF
--- a/buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
@@ -1,4 +1,5 @@
 import org.jetbrains.kotlin.gradle.*
+import org.jetbrains.kotlin.gradle.plugin.mpp.*
 
 /*
  * Copyright 2017-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
@@ -47,6 +48,15 @@ kotlin {
         linuxArm32Hfp()
     }
 
+    // setup tests running in RELEASE mode
+    targets.withType<KotlinNativeTarget>().configureEach {
+        binaries.test(listOf(NativeBuildType.RELEASE))
+    }
+    targets.withType<KotlinNativeTargetWithTests<*>>().configureEach {
+        testRuns.create("releaseTest") {
+            setExecutionSourceFrom(binaries.getTest(NativeBuildType.RELEASE))
+        }
+    }
 }
 
 fun doesNotDependOnOkio(project: Project): Boolean {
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
```diff
@@ -47,6 +48,15 @@ kotlin {
         linuxArm32Hfp()
     }
 
+    // setup tests running in RELEASE mode
+    targets.withType<KotlinNativeTarget>().configureEach {
```
sandwwraith, line 52: From its looks, it replaces all existing targets with the RELEASE build type, so we do not build DEBUG binaries anymore. But I'm not an expert in Kotlin/Native's Gradle.

whyoleg: not really. it just adds `RELEASE` test binary, as by default only `DEBUG` test binary is created.
It can be checked by running old existing tasks like `macosArm64Test` - they will depend on `linkDebug*` tasks.

using `binaries.test(listOf(DEBUG, RELEASE))` will cause error trying to create binary which already exists (with `DEBUG`).

shanshin (approved):

buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
sandwwraith: thnx for explanations

Issue #1310, opened 2021-01-27
Can't access default serializer which is crutial for polymorphism.


krottv:
**Describe the bug**

Currently if we provide @Serializable(with = CustomSerializer) annotation then we no longer have access to original generated serializer. Which is fine in most cases, but it is critical when we deal with polymotphism. The main issue is that we run into dependency cycle.

**To Reproduce**
The following code crashes the app with NPE
```
@Serializable(with = MediaImageSerializer::class)
@SerialName("image")
data class MediaImage(...)

object MediaImageSerializer : JsonTransformingSerializer<MediaImage>(MediaImage.serializer()) 
```

The problem here is that we can't avoid `@Serializable(with = MediaImageSerializer::class)` here. Because the object that contains mediaImage doesn't know its type:

```
@Serializable
class Template(medias: List<@Serializable with [we don't know which serializer to use] Media>)
```

**Expected behavior**

We need to be able to use original generated serializer

**Environment**
 - Kotlin version: 1.4.30-RC
 - Library version: 1.0.1
 - Kotlin platforms: JVM

sandwwraith:
Duplicate of #1169, looks like

As a workaround, you can use `@file:UseSerializers` instead of repeating `@Serializable(MediaImageSerializer::class)`

krottv:
@sandwwraith where should I put @file:UseSerializers? I tried it everywhere, doesn't make any difference, still the dafault one is always used.

sandwwraith:
Fixed by #1169

PR #2669, opened 2024-05-13
Added documentation for @KeepGeneratedSerializer feature

### LINKED ISSUE #2660: Add documentation for @KeepSerializer feature
Its KDoc should be updated, and a section in the guide should be added.

shanshin:
Resolves #2660

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/core/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -343,7 +343,6 @@ public annotation class Polymorphic
  *
  * A compiler version `2.0.20` and higher is required.
  */
-@InternalSerializationApi
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 public annotation class KeepGeneratedSerializer
--- a/core/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/core/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -31,6 +31,9 @@ import kotlin.reflect.*
  * MyAnotherData.serializer() // <- returns MyAnotherDataCustomSerializer
  * ```
  *
+ * To continue generating the implementation of [KSerializer] using the plugin, specify the [KeepGeneratedSerializer] annotation.
+ * In this case, the serializer will be available via `generatedSerializer()` function, and will also be used in the heirs.
+ *
  * For annotated properties, specifying [with] parameter is mandatory and can be used to override
  * serializer on the use-site without affecting the rest of the usages:
  * ```
@@ -64,6 +67,7 @@ import kotlin.reflect.*
  *
  * @see UseSerializers
  * @see Serializer
+ * @see KeepGeneratedSerializer
  */
 @MustBeDocumented
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS, AnnotationTarget.TYPE)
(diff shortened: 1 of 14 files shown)


sandwwraith (changes requested):
I think it is worth adding an example with `JsonTransformingSerializer(MyObj.generatedSerializer())` as well (see https://github.com/Kotlin/kotlinx.serialization/issues/1310) — likely to `json.md#json-transformations` section that will explain how to make a JsonTransformingSerializer a default one while delegating to generated serializer at the same time.

Also, `knitCheck` task fails.

shanshin:
Waiting for `2.0.20` release

docs/json.md
```diff
@@ -1260,6 +1261,51 @@ No class discriminator is added in the JSON output:
 
 <!--- TEST -->
 
+### Extending the behavior of the plugin generated serializer
+In some cases, it may be necessary to add additional serialization logic on top of the plugin generated logic.
+For example, it is necessary to add a preliminary modification of JSON elements or to add processing of unknown values of enums.
```
sandwwraith, line 1266:
```suggestion
For example, to add a preliminary modification of JSON elements or to add processing of unknown values of enums.
```

docs/json.md
```diff
@@ -1260,6 +1261,51 @@ No class discriminator is added in the JSON output:
...
+
+In this case, you should mark the serializable class with the [`@KeepGeneratedSerializer`][KeepGeneratedSerializer] annotation and get the generated serializer using the `generatedSerializer()` function.
```
sandwwraith, line 1268:
```suggestion
In this case, you can mark the serializable class with the [`@KeepGeneratedSerializer`][KeepGeneratedSerializer] annotation and get the generated serializer using the `generatedSerializer()` function.
```

docs/json.md
```diff
@@ -1260,6 +1261,51 @@ No class discriminator is added in the JSON output:
...
+
+An example of the simultaneous use of [JsonTransformingSerializer] and polymorphism:
```
sandwwraith, line 1270:
```suggestion
Here is an example of the simultaneous use of [JsonTransformingSerializer] and polymorphism.
In this example, we use `transformDeserialize` function to rename `basic-name` key into `name` so it matches the `abstract val name` property from the `Project` supertype.
```

sandwwraith (approved):

shanshin:
I think it would be safer to add https://github.com/Kotlin/kotlinx.serialization/pull/2758

core/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -330,13 +334,15 @@ public annotation class Polymorphic
...
  * Automatically generated serializer is available via `generatedSerializer()` function in companion object of serializable class.
  *
- * Generated serializers allow to use custom serializers on classes from which other serializable classes are inherited.
+ * Keeping generated serializers allow to use plugin generated serializer in inheritors even if custom serializer is specified.
+ *
+ * Used only with annotation [Serializable] with the specified argument [Serializable.with], e.g. `@Serializable(with=SomeSerializer::class)`.
  *
- * Used only with the [Serializable] annotation.
+ * Annotation is not allowed on classes involved in polymorphic serialization:
+ * interfaces, sealed classes, abstract classes, classes marked by [Polymorphic].
```
hfhbd, line 342: I don't understand this sentence: `@KeepGeneratedSerializer` is not allowed with any inheritance, but your sample code uses a class with inherits from an interface. (see other comment)

docs/json.md
```diff
@@ -1260,6 +1261,53 @@ No class discriminator is added in the JSON output:
...
+```kotlin
+@Serializable
+sealed class Project {
+    abstract val name: String
+}
+
+@KeepGeneratedSerializer
+@Serializable(with = BasicProjectSerializer::class)
+@SerialName("basic")
+data class BasicProject(override val name: String): Project()
```
hfhbd, line 1281: Also, do you test your sample automatically? If not, can you add this code to your tests in #2758?
Background: I use Kotlin 2.0.20-RC with kx-serial 1.7.1, and I get this error:
```
java.lang.NullPointerException
	at kotlinx.serialization.SealedClassSerializer$special$$inlined$groupingBy$1.keyOf(_Collections.kt:1547)
```
I will also create an issue with a small reproducer.

core/commonMain/src/kotlinx/serialization/Annotations.kt
shanshin: It means that you cannot specify the `@KeepGeneratedSerializer ` interface on the interfaces, sealed/abstract classes, but it's ok to add this annotation add to the heirs

docs/json.md
shanshin: > Also, do you test your sample automatically?

yes

> I will also create an issue with a small reproducer.

Thanks!

PR #2670, opened 2024-05-13
Fix confusing description of JsonPrimitive.content

gyuhyeon:
#### Summary
- clarifies that JsonPrimitive.content will return `"null"` rather than `null`
- include suggestion that `contentOrNull` should be used if the JsonPrimitive could be `JsonNull`
- fix potentially misleading usage in `JsonTreeTest`

-----  

JsonPrimitive.content returns `"null"` when it is of JsonNull type.
The original comment was like below:
```
public sealed class JsonPrimitive : JsonElement() {
    /**
     * Content of given element without quotes. For [JsonNull] this methods returns `null`
     */
    public abstract val content: String
```
This is potentially confusing if the reader hadn't noticed that `content` is of non-nullable type `String`.

This has led to a **very common** usage of the following code:

`elem?.jsonPrimitive?.content`

I doubt most are aware that `object.get("key")?.jsonPrimitive?.content` returns `"null"` as string if the key was explicitly null like `{"key": null}`, and only returns actual `null` when the key does not exist in the json object.

I'm open to corrections, but I cannot think of any scenarios where
`elem?.jsonPrimitive?.content` that returns `"null"` or `null` depending on the serialization policy of the json producer would be beneficial.

`contentOrNull` should always be preferred over `content` if there's a possibility of the value being null (either by missing key, or explicitly set as null - resulting into a JsonNull), and the revised comment reflects that view.

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
@@ -41,8 +41,8 @@ public sealed class JsonPrimitive : JsonElement() {
     public abstract val isString: Boolean
 
     /**
-     * Content of given element without quotes. For [JsonNull] this method returns `"null"`.
-     * [JsonPrimitive.contentOrNull] should be used for [JsonNull] to return `null`.
+     * Content of given element without quotes. For [JsonNull], this method returns a "null" string.
+     * [JsonPrimitive.contentOrNull] should be used for [JsonNull] to get a `null`.
      */
     public abstract val content: String
 
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
@@ -41,7 +41,8 @@ public sealed class JsonPrimitive : JsonElement() {
     public abstract val isString: Boolean
 
     /**
-     * Content of given element without quotes. For [JsonNull] this methods returns `null`
+     * Content of given element without quotes. For [JsonNull] this method returns `"null"`.
+     * [JsonPrimitive.contentOrNull] should be used for [JsonNull] to return `null`.
      */
     public abstract val content: String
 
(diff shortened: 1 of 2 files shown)


pdvrieze:
@gyuhyeon The way JsonPrimitive works is that it is not valid to use the string value directly, without checking that it is a string value. In that case content is the actual content of the string, otherwise it is the string representation of the literal. In other words the representation of `true` and `"true"` in the content property are the same, you need to know whether it is a string to distinguish.

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -41,7 +41,8 @@ public sealed class JsonPrimitive : JsonElement() {
     public abstract val isString: Boolean
 
     /**
-     * Content of given element without quotes. For [JsonNull] this methods returns `null`
+     * Content of given element without quotes. For [JsonNull] this method returns `"null"`.
```
sandwwraith, line 44:
```suggestion
     * Content of given element without quotes. For [JsonNull], this method returns a "null" string.
```

formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -41,7 +41,8 @@ public sealed class JsonPrimitive : JsonElement() {
...
+     * [JsonPrimitive.contentOrNull] should be used for [JsonNull] to return `null`.
```
sandwwraith, line 45:
```suggestion
     * [JsonPrimitive.contentOrNull] should be used for [JsonNull] to get a `null`.
```

sandwwraith (approved):
Well, `.content` couldn't return `null` because it has non-nullable return type. But clarification is good in any case.

gyuhyeon:
review looks good to me, it doesn't hurt to clarify things.
Unlike extensions like `JsonPrimitive.int` and `JsonPrimitive.intOrNull`, `JsonPrimitive.contentOrNull` isn't in adjacent lines with `JsonPrimitive.content` - which is one of the reasons why the antipattern `jsonPrimitive?.content` is so common in codebases.

sandwwraith:
Thank you!

PR #2674, opened 2024-05-15
Setup Kover for validation of each module

### LINKED ISSUE #2634: Increase common coverage and coverage of individual modules
It is necessary to configure Kover in such a way that it is possible to verify the coverage of individual modules.
The main problem is that many modules do not have tests and are being tested by other modules.

After configuring Kover, we need to consider the uncovered sections of the code and figure out how to significantly increase the percentage of coverage.
If the coverage is greatly increased by writing small tests, then such tests must be written. Otherwise, we need to create a separate task for analyzing and increasing coverage

shanshin:
Also, slightly increased coverage

Resolves #2634


build.gradle.kts
```diff
@@ -137,9 +138,45 @@ subprojects {
...
 // == Kover setup ==
-subprojects {
-    if (project.name in uncoveredProjects) return@subprojects
-    apply(plugin = "kover-conventions")
+kover {
+    if (hasProperty("kover.enabled") && property("kover.enabled") != "true") {
+        disable()
+    }
+
+    currentProject {
```
sandwwraith, line 146: Can we extract Kover configuration to a separate file?

shanshin: Sure

formats/json-io/commonTest/src/kotlinx/serialization/json/io/IoTests.kt
```diff
@@ -0,0 +1,70 @@
...
+        val readArray = CharArray(2)
+        assertEquals(1, reader.read(readArray, 0, 1) )
+        assertEquals(1, reader.read(readArray, 1, 1) )
+
+        assertContentEquals(originalChars, readArray)
+    }
+
+
+    @Test
+    fun testEncodingAndDecoding() {
```
sandwwraith, line 36: Theoretically, IO's `encodeToSink/decodeFromSource` should also be covered by `json-tests` module. Isn't this the case?

sandwwraith (approved):

formats/json-io/commonTest/src/kotlinx/serialization/json/io/IoTests.kt
shanshin: In `json-tests`, methods with a different signature are being tested (with explicit transfer of the serializer)

PR #2678, opened 2024-05-16
Prepare 1.7.0-RC release

sandwwraith:


formats/cbor/api/kotlinx-serialization-cbor.klib.api
```diff
@@ -7,17 +7,17 @@
 
 // Library unique name: <org.jetbrains.kotlinx:kotlinx-serialization-cbor>
 final class kotlinx.serialization.cbor/CborBuilder { // kotlinx.serialization.cbor/CborBuilder|null[0]
-    final var encodeDefaults // kotlinx.serialization.cbor/CborBuilder.encodeDefaults|<get-encodeDefaults>(){}[0]
+    final var encodeDefaults // kotlinx.serialization.cbor/CborBuilder.encodeDefaults|{}encodeDefaults[0]
```
sandwwraith, line 10: @fzhinkin FYI updating 1.9.2x -> 2.0.0 changes format of Klib dumps slightly. I assume this is because this file is produced directly by the compiler? It would be nice if the compiler team avoided making such changes if they were not absolutely necessary.

qwwdfsad (approved):

formats/cbor/api/kotlinx-serialization-cbor.klib.api
fzhinkin: Yeap, that's because of the fixed formatting: https://github.com/JetBrains/kotlin/commit/6972ed67d24644cd143861f21e398508d1689b3f

Issue #2662, opened 2024-05-04
Strict mode for CBOR


timmc:
**What is your use-case and why do you need this feature?**

My app needs to be able to sign and verify serialized data, which requires that the parse be unambiguous. The main problem is duplicate map keys. If there are repeated keys, then two recipients may disagree on what the signed data says, which is obviously a problem. :-)

CBOR's spec describes a "strict mode" that forbids repeated map keys and other dangers to canonicalization: https://datatracker.ietf.org/doc/html/rfc7049#section-3.10

**Describe the solution you'd like**

1. Generally support rejection of repeated map keys, as an optional feature. (I'd love for this to be enabled by default, but I understand that there are competing needs here.)
2. Expose this as part of a new strict-mode feature in the `Cbor` implementation.

Note: https://github.com/Kotlin/kotlinx.serialization/issues/1990 covers rejection of duplicate keys, but it wasn't clear if that was meant to be specific to JSON.

timmc:
(Updated with more info from the spec.)

sandwwraith:
Yes, it sounds very reasonable. I do not mind adding it at some point in the future or via contribution.

timmc:
Great, I may try to work up a PR. Where would you recommend getting started? I found [`insertKeyValuePair` in CollectionSerializers](https://github.com/Kotlin/kotlinx.serialization/blob/51cb8e8e556983fc83a565d5f04bb089363453e0/core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt#L265)—but it didn't seem that the tests actually exercise that code path, and I can't find anything that calls `insertKeyValuePair`.

pdvrieze:
@timmc What happens from a format perspective is that it recognizes that this is a collection serializer. Then it will call the update decoder (at some point it was part of DeserializerStrategy, but it is no longer) with the previous list value.That update will create/update the collection with the new value (or not). It would be possible to create a new version of this function that takes a strict argument (with the default implementation of the previous one passing false for strictness).

timmc:
Would you be able to point me to a code location? I'm not finding useful references to `update`, `CollectionSerializer`, etc. I just discovered that `LinkedHashMapSerializer` extends `LinkedHashMap`, but I'm having trouble finding the code that actually puts key-value pairs into a map when deserializing map data.

timmc:
[...]

PR #2681, opened 2024-05-20
Add ability to disallow repeated keys in CBOR

timmc:
Fixes https://github.com/Kotlin/kotlinx.serialization/issues/2662 by adding a `visitKey` method to `CompositeDecoder`; map and set serializers should call this so that decoders have an opportunity to throw an error when a duplicate key is detected. Calls to this method are added to:

- `MapLikeSerializer.readElement` to handle maps
- `CborReader.decodeElementIndex` to handle data classes

A new config option `Cbor.forbidDuplicateKeys` can be set to true to enable this new behavior. This can form the basis of a Strict Mode in the future.


core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -133,3 +133,10 @@ internal constructor(message: String?) : SerializationException(message) {
     // This constructor is used by the generated serializers
     constructor(index: Int) : this("An unknown field for index $index")
 }
+
+/**
+ * Thrown when a map deserializer encounters a repeated map key (and configuration disallows this.)
+ */
+@ExperimentalSerializationApi
+public class DuplicateMapKeyException(key: Any?) :
+    SerializationException("Duplicate keys not allowed in maps. Key appeared twice: $key")
```
timmc, line 142: TODO: Just take a `message` param; I'm probably not going to end up doing assertions on key, and this is going to be annoying to generalize to sets, classes, etc.

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
```diff
@@ -558,6 +558,18 @@ public interface CompositeDecoder {
...
+     * Called after a key has been read.
+     *
+     * This could be a map or set key, or anything otherwise intended to be
+     * distinct within the collection under normal circumstances.
+     *
+     * Implementations might use this as a hook for throwing an exception when
+     * duplicate keys are encountered.
+     */
+    @ExperimentalSerializationApi
+    public fun visitKey(key: Any?) { }
```
timmc, line 572: I'm not happy with this name but couldn't think of anything better -- suggestions very welcome. :-)

formats/cbor/commonMain/src/kotlinx/serialization/cbor/Cbor.kt
```diff
@@ -32,13 +32,14 @@ import kotlinx.serialization.modules.*
...
     internal val encodeDefaults: Boolean,
     internal val ignoreUnknownKeys: Boolean,
+    internal val allowDuplicateKeys: Boolean,
     override val serializersModule: SerializersModule
 ) : BinaryFormat {
 
     /**
      * The default instance of [Cbor]
      */
-    public companion object Default : Cbor(false, false, EmptySerializersModule())
```
timmc, line 41: Is there a convention in this library that the defaults should be false, or is the introduction of a `true`-defaulting setting OK? (Wasn't sure whether to go with allow or forbid in the name.)

timmc: Leaning towards switching this to `forbidDuplicateKeys`, actually.

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborStrictModeTest.kt
```diff
@@ -0,0 +1,20 @@
+package kotlinx.serialization.cbor
+
+import kotlinx.serialization.assertFailsWithMessage
+import kotlinx.serialization.decodeFromByteArray
+import kotlinx.serialization.HexConverter
+import kotlinx.serialization.DuplicateMapKeyException
+import kotlin.test.Test
+
+class CborStrictModeTest {
```
timmc, line 9: TODO: Add a (passing) test for duplicate key detection when deserializing a data class. I have a test locally that fails -- I think I need to add ` visitKey` call to some inlined code somewhere, but I haven't found the appropriate callsite yet.

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
pdvrieze: @xiaozhikang0916 The challenge is that the behaviour needs to be configurable. There is no central configuration mechanism in the library, only formats and `SerialDescriptors`. There is also the need to do this in an ABI/API compatible way. Formats already need to do various things to play nice/support the various library features. This would just be another one. It also makes sense as various formats have their own duplication semantics.

timmc: I initially tried plumbing the information through on the deserialization layer and it went very badly. This is only the start of what would be required: https://github.com/Kotlin/kotlinx.serialization/compare/dev...timmc:timmc/all-strict (although in practice it would be a config object, not a single boolean -- this was just a failed proof of concept.)

formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborStrictModeTest.kt
timmc: I think I need to modify the serialization plugin's code generator in order to do this! Where does that code even live?

timmc: Hmm. I had previously tried modifying `decodeElementIndex` and wasn't getting the results I expected. Maybe I'll take another crack at it.

timmc: Oh! I see, the `decodeElementIndex` in `CborReader`. Got it.

core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
xiaozhikang0916: I mean the duplication check in other formats ( if they need ), would share some common logic, like holding a set and putting keys in it. These common logic can be extracted to a utility class.

------

On the other hand, I am against adding this function in `Decoding` interface. It is implemented by each format with different behaviours, and only for internal using, which means not public, and users writing custom serializers should never call this function.

IMO, you may probably leave some instructions for format developers about how to achieve duplication check, rather than a public default funtion in this base interface, and make it open for users.

pdvrieze: It would be possible to circumvent the MapSerializer if you have a map specific decoder. This decoder could intercept the decoding of the key, and then check there (before the result is returned to the serializer).

timmc: I'm happy to implement whatever is needed, but I don't have a good understanding of what would be acceptable to the maintainers. It sounds like visitKey isn't well-liked, and the options might be:

- A. Each format that wants a duplicate-detection option provides its own replacement for a `MapLikeSerializer`-like serializer, fully reimplementing the logic. (This would be done via a serializers module, right?)
- B. Copy `MapLikeSerializer` to a new utility class `NoDuplicatesMapLikeSerializer` and add the duplicate detection; formats use this in their serializers module if desired.
- C. Like B, but just change `MapLikeSerializer` to make `readElement` non-final so that it can be overridden by classes like `HashMapSerializer` (or copies of them) rather than copying the implementation.
- D. Change the map *builder* to reject duplicate keys (haven't looked into feasibility here)

Which of these seems most palatable? Are there other possibilities?

timmc: One concern I have with anything that involves the serializers module is that it risks interference with application-configured serializers. Each relevant format would have to document warnings like "if you specify an alternative serializer for maps, you'll need to reimplement duplicate key detection". Having it be performed in shared code (i.e. MapLikeSerializer) would at least allow the warning to recommend subclassing that shared code.

PR #2689, opened 2024-05-26
Added link to kotlinx-serialization-smile in the formats list

vooft:
Hi all,

I've built an implementation of the Smile format for kotlinx.serialization, would like to add it to the list of the supported formats.

Thanks!


sandwwraith (approved):
Great, it's always nice to see new libraries!

PR #2705, opened 2024-06-05
Use @PublishedApi annotation on functions called from plugin-generated code

### LINKED ISSUE #2703: `createAnnotatedEnumSerializer` and `createSimpleEnumSerializer` functions must be made ABI-public
These functions:
- https://github.com/Kotlin/kotlinx.serialization/blob/4bf4113c2c6936e841caf2a8670382e701e7fcc2/core/commonMain/src/kotlinx/serialization/internal/Enums.kt#L83
- https://github.com/Kotlin/kotlinx.serialization/blob/4bf4113c2c6936e841caf2a8670382e701e7fcc2/core/commonMain/src/kotlinx/serialization/internal/Enums.kt#L54

are `internal`, but the kotlinx.serialization compiler plugin generates calls to these functions in user code. This is access violation.

Why this is important to us (the Kotlin compiler team): we're trying to [enable visibility checks during IR validation](https://youtrack.jetbrains.com/issue/KT-68784) in the compiler. Basically, these checks try to ensure that e.g. an internal function is only ever called from the same module it's declared in. We're trying to enable these checks in compiler tests, however, the kotlinx.serialization-related tests are failing because of access violations.

We suggest annotating these functions (and potentially other `internal` functions, calls to which may be emitted into user code by the compiler plugin) as `@PublishedApi`, or making them public.

sandwwraith:
Since these functions can be called from other modules, they should have this annotation to satisfy a compiler visibility checker.

Fixes #2703


broadwaylamb:
> to satisfy a compiler visibility checker

Important note: not just for that, but because these functions are effectively public ABI, but nothing in their signature indicate that they are, and that they should be maintained as such.

shanshin (approved):

shanshin:
@sandwwraith , api check failed

PR #2707, opened 2024-06-06
Add integration with kotlinx-io library

### LINKED ISSUE #253: Add integration with Kotlin IO multiplatform library
This will address: 
#66, #204 and #251.

`kotlinx.io.core.Input/Output` should replace `ByteArray`s and `String`s in Binary and String formats, thus allowing to work with streaming API and easily convert content to either byte arrays or strings with desired charsets.

sandwwraith:
Integration is similar to Okio's one with functions `encodeToSink`, `decodeFromSource`, etc.

Fixes #253


sandwwraith:
The main question is how to name module/package:

`kotlinx.serialization.json.kxio` (as now) — a cryptic abbreviation
`kotlinx.serialization.json.kotlinx.io` — duplication of 'kotlinx'
`kotlinx.serialization.json.io` — not clear which IO we are referring to

fzhinkin:
> The main question is how to name module/package:
>
> `kotlinx.serialization.json.kxio` (as now) — a cryptic abbreviation

Personally, I like this one (most likely because I used to shorten the library name to `kx-io`), but indeed it looks a bit cryptic.

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TwitterFeedStreamBenchmark.kt
```diff
@@ -88,4 +115,24 @@ open class TwitterFeedStreamBenchmark {
...
+        // It seems there is no optimal way to reuse `bytes` between benchmark, so we are forced
+        // to write them to buffer every time.
+        // Note that it makes comparison with Jackson and InputStream integration much less meaningful.
+        val b = OkioBuffer()
+        b.write(bytes)
+        return jsonIgnoreUnknwn.decodeFromBufferedSource(MicroTwitterFeed.serializer(), b)
+    }
+
+    @Benchmark
+    fun decodeMicroTwitterKotlinxIoStream(): MicroTwitterFeed {
```
fzhinkin, line 130: For some reason, this particular benchmark shows worse results compared to Okio version. It doesn't block integration in any way, I'll check the root cause later.

Dominaezzz:
> `kotlinx.serialization.json.io` — not clear which IO we are referring to

I think this is best. This library is under `kotlinx` so it's sensible for the default `io` to also be the `kotlinx` one.

Any duplicate mention of `kotlinx`, be it the abbreviation `kx` or the full thing `kotlinx` just seems like a waste of characters imo.

DRSchlaubi:
> I think this is best. This library is under kotlinx so it's sensible for the default io to also be the kotlinx one.

Have to agree with that one, but what about simply including it like kx-datetime supports kx.ser out of the box, or does this run into the issue of compileOnly not working on non-jvm platforms?

sandwwraith:
@DRSchlaubi Yes, kotlinx-datetime actually uses `api` dependency for Js and Native.

sandwwraith:
Let's stick with `kotlinx.serialization.json.io` then.

formats/json-io/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+        val commonMain by getting {
+            dependencies {
+                api(project(":kotlinx-serialization-core"))
+                api(project(":kotlinx-serialization-json"))
+                implementation(libs.kotlinx.io)
+            }
+        }
+        val commonTest by getting {
+            dependencies {
+                implementation(libs.kotlinx.io)
```
shanshin, line 33: is it not inherited from `commonMain`?

shanshin (approved):

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TwitterFeedStreamBenchmark.kt
fzhinkin: That's a perf issue on the kotlinx-io size: an extension function to read code point values is defined only on `Source` and it performs a few actions to fill the inner buffer before decoding the code point. In Okio, that's a member function with a different (and optimized) implementation for `Buffer`. 
kx-io should do the same: https://github.com/Kotlin/kotlinx-io/issues/342.

fzhinkin (approved):

formats/json-io/build.gradle.kts
sandwwraith: I just copied config from `okio` module, guess it's not needed there too

PR #2722, opened 2024-06-20
feat: Add a way to reorder elements during encoding

### LINKED ISSUE #2668: Non-sequential encoding
Currently, a decoder is able of retrieving the element index to decode, that is really powerful when we want to decode encoded data in a different order than the current element order. 

Now, if we need to encode the elements in a different order than the descriptor's one, then we have to make complex implementation to buffer the encoded data, by buffering the encoded data and then really encoding it using `endStructure`. 

For sure it works, but it feels like kotlinx-serialization is missing a part of the sequentiality feature, where we are able to select if we want sequential decoding or not, but not for the encoding part.

By looking at the generated code made by the serialization plugin, it seems quite "simple" to implement non-sequential encoding as exactly the same way as decoding : a method `encodeSequentially` that returns true by default for backward compatibility.

When this method returns false, a method `getElementIndexToEncode` is called unless it returns `-1` (encoding done). For each matching index, it encodes the corresponding element's value.

Side note: I've seen some issues about a way to reorder fields, rejected because this job is "do-able" with the current status of the library. But why `decodeSequentially` and `decodeElementIndex` exists as it is also still possible to implement in sequential-only way?

For sure I know it's also a feature request for the serialization plugin, which I'll do it. 

To finish, having this feature would simplify all the encoders that needs to reorder fields or skip fields, without implementing complex buffering or transformers. It would also make native the `AbstractEncoder.encodeElement(descriptor, index): Boolean` natively handled as the non-encoded element index wouldn't be returned at all (also a little more performant I suppose). For the last drop-the-mic argument, it will make the sequential mode "parallel" as it will be available in encoding and decoding 🚀

Chuckame:
Closes #2668

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
+++ b/core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
@@ -1,0 +1,247 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.encoding
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.modules.*
+
+
+/**
+ * Encodes elements in a user-defined order managed by [mapElementIndex].
+ *
+ * This encoder will replicate the behavior of a standard encoding, but calling the `encode*Element` methods in
+ * the order defined by [mapElementIndex]. It first buffers each `encode*Element` calls in an array following
+ * the given indexes using [mapElementIndex], then when [endStructure] is called, it encodes the buffered calls
+ * in the expected order by replaying the previous calls on the given [compositeEncoderDelegate].
+ *
+ * This encoder is stateful and not designed to be reused.
+ *
+ * @param compositeEncoderDelegate the [CompositeEncoder] to be used to encode the given descriptor's elements in the expected order.
+ * @param encodedElementsCount The final number of elements to encode, which could be smaller than the original descriptor when [mapElementIndex] returns [SKIP_ELEMENT_INDEX] or when the index mapper has returned the same index twice.
+ * @param mapElementIndex maps the element index to a new positional zero-based index.
+ *                          The mapped index just helps to reorder the elements,
+ *                          but the reordered `encode*Element` method calls will still pass the original element index.
+ *                          If this mapper returns [SKIP_ELEMENT_INDEX] or -1, the element will be ignored and not encoded.
+ *                          If this mapper provides the same index for multiple elements,
+ *                          only the last one will be encoded as the previous ones will be overridden.
+ */
+@ExperimentalSerializationApi
+public class ReorderingCompositeEncoder(
+    encodedElementsCount: Int,
+    private val compositeEncoderDelegate: CompositeEncoder,
+    private val mapElementIndex: (SerialDescriptor, Int) -> Int,
+) : CompositeEncoder {
+    private val bufferedCalls = Array<BufferedCall?>(encodedElementsCount) { null }
+
+    public companion object {
+        @ExperimentalSerializationApi
+        public const val SKIP_ELEMENT_INDEX: Int = -1
+    }
+
+    override val serializersModule: SerializersModule
+        // No need to return a serializers module as it's not used during buffering
+        get() = EmptySerializersModule()
+
+    private data class BufferedCall(
+        val originalElementIndex: Int,
+        val encoder: () -> Unit,
+    )
+
+    private fun bufferEncoding(
+        descriptor: SerialDescriptor,
+        index: Int,
+        encoder: () -> Unit
+    ) {
+        val newIndex = mapElementIndex(descriptor, index)
+        if (newIndex != SKIP_ELEMENT_INDEX) {
+            bufferedCalls[newIndex] = BufferedCall(index, encoder)
+        }
+    }
+
+    override fun endStructure(descriptor: SerialDescriptor) {
+        bufferedCalls.forEach { fieldToEncode ->
+            // In case of skipped fields, overridden fields (mapped to same index) or too big [encodedElementsCount],
+            // the fieldToEncode may be null as no element was encoded for that index
+            fieldToEncode?.encoder?.invoke()
+        }
+        compositeEncoderDelegate.endStructure(descriptor)
+    }
+
+    override fun encodeBooleanElement(descriptor: SerialDescriptor, index: Int, value: Boolean) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeBooleanElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeByteElement(descriptor: SerialDescriptor, index: Int, value: Byte) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeByteElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeCharElement(descriptor: SerialDescriptor, index: Int, value: Char) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeCharElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeDoubleElement(descriptor: SerialDescriptor, index: Int, value: Double) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeDoubleElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeFloatElement(descriptor: SerialDescriptor, index: Int, value: Float) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeFloatElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeIntElement(descriptor: SerialDescriptor, index: Int, value: Int) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeIntElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeLongElement(descriptor: SerialDescriptor, index: Int, value: Long) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeLongElement(descriptor, index, value)
+        }
+    }
+
+    override fun <T : Any> encodeNullableSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        serializer: SerializationStrategy<T>,
+        value: T?
+    ) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeNullableSerializableElement(descriptor, index, serializer, value)
+        }
+    }
+
+    override fun <T> encodeSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        serializer: SerializationStrategy<T>,
+        value: T
+    ) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeSerializableElement(descriptor, index, serializer, value)
+        }
+    }
+
+    override fun encodeShortElement(descriptor: SerialDescriptor, index: Int, value: Short) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeShortElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeStringElement(descriptor: SerialDescriptor, index: Int, value: String) {
+        bufferEncoding(descriptor, index) {
+            compositeEncoderDelegate.encodeStringElement(descriptor, index, value)
+        }
+    }
+
+    override fun encodeInlineElement(descriptor: SerialDescriptor, index: Int): Encoder {
+        return BufferingInlineEncoder(descriptor, index)
+    }
+
+    override fun shouldEncodeElementDefault(descriptor: SerialDescriptor, index: Int): Boolean {
+        return compositeEncoderDelegate.shouldEncodeElementDefault(descriptor, index)
+    }
+
+    private inner class BufferingInlineEncoder(
+        private val descriptor: SerialDescriptor,
+        private val elementIndex: Int,
+    ) : Encoder {
+        private var encodeNotNullMarkCalled = false
+
+        override val serializersModule: SerializersModule
+            get() = this@ReorderingCompositeEncoder.serializersModule
+
+        private fun bufferEncoding(encoder: Encoder.() -> Unit) {
+            bufferEncoding(descriptor, elementIndex) {
+                compositeEncoderDelegate.encodeInlineElement(descriptor, elementIndex).apply {
+                    if (encodeNotNullMarkCalled) {
+                        encodeNotNullMark()
+                    }
+                    encoder()
+                }
+            }
+        }
+
+        override fun encodeNotNullMark() {
+            encodeNotNullMarkCalled = true
+        }
+
+        override fun <T : Any> encodeNullableSerializableValue(serializer: SerializationStrategy<T>, value: T?) {
+            bufferEncoding { encodeNullableSerializableValue(serializer, value) }
+        }
+
+        override fun <T> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) {
+            bufferEncoding { encodeSerializableValue(serializer, value) }
+        }
+
+        override fun encodeBoolean(value: Boolean) {
+            bufferEncoding { encodeBoolean(value) }
+        }
+
+        override fun encodeByte(value: Byte) {
+            bufferEncoding { encodeByte(value) }
+        }
+
+        override fun encodeChar(value: Char) {
+            bufferEncoding { encodeChar(value) }
+        }
+
+        override fun encodeDouble(value: Double) {
+            bufferEncoding { encodeDouble(value) }
+        }
+
+        override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+            bufferEncoding { encodeEnum(enumDescriptor, index) }
+        }
+
+        override fun encodeFloat(value: Float) {
+            bufferEncoding { encodeFloat(value) }
+        }
+
+        override fun encodeInt(value: Int) {
+            bufferEncoding { encodeInt(value) }
+        }
+
+        override fun encodeLong(value: Long) {
+            bufferEncoding { encodeLong(value) }
+        }
+
+        @ExperimentalSerializationApi
+        override fun encodeNull() {
+            bufferEncoding { encodeNull() }
+        }
+
+        override fun encodeShort(value: Short) {
+            bufferEncoding { encodeShort(value) }
+        }
+
+        override fun encodeString(value: String) {
+            bufferEncoding { encodeString(value) }
+        }
+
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+ * the given indexes using [mapElementIndex], then when [endStructure] is called, it encodes the buffered calls
+ * in the expected order by replaying the previous calls on a concrete [CompositeEncoder] provided
+ * by [beginStructure].
+ *
+ * @param structureDescriptor descriptor of the structure being encoded
+ * @param beginStructure provides the [CompositeEncoder] to be used to encode the given descriptor's elements in the expected order.
+ * @param mapElementIndex maps the element index to the new reordered index (zero-based). If this mapper provides the same index for multiple elements, only the last one will be encoded as the previous ones will be overridden.
+ */
+@ExperimentalSerializationApi
+internal class ReorderingCompositeEncoder(
```
pdvrieze, line 26: If you leave this internal it is not useful. You probably meant for it to be public.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+ * Encodes composite elements in a specific order managed by [mapElementIndex].
+ *
+ * This encoder will replicate the behavior of a standard encoding, but calling the `encode*Element` methods in
+ * the order defined by [mapElementIndex]. It first buffers each `encode*Element` calls in an array following
+ * the given indexes using [mapElementIndex], then when [endStructure] is called, it encodes the buffered calls
+ * in the expected order by replaying the previous calls on a concrete [CompositeEncoder] provided
+ * by [beginStructure].
+ *
+ * @param structureDescriptor descriptor of the structure being encoded
+ * @param beginStructure provides the [CompositeEncoder] to be used to encode the given descriptor's elements in the expected order.
```
pdvrieze, line 22: This parameter is quite surprising/confusing and can probably be documented more (or renamed). It is used as `beginStructure` in this type's `endStructure`, but from the caller perspective what it does is provide the delegate `CompositeEncoder` that will receive the reordered events.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+ * @param mapElementIndex maps the element index to the new reordered index (zero-based). If this mapper provides the same index for multiple elements, only the last one will be encoded as the previous ones will be overridden.
+ */
+@ExperimentalSerializationApi
+internal class ReorderingCompositeEncoder(
+    structureDescriptor: SerialDescriptor,
+    private val beginStructure: (SerialDescriptor) -> CompositeEncoder,
+    private val mapElementIndex: (SerialDescriptor, Int) -> Int,
+) : CompositeEncoder {
+    // Each time we encode a field, if the next expected schema field index is not the good one, it is buffered until it's the time to encode it
+    private var bufferedCalls = Array<BufferedCall?>(structureDescriptor.elementsCount) { null }
```
pdvrieze, line 32: This cache means you need to create an instance at least per structure descriptor, but more conveniently for every instance. In such case `beginStructure` can just become `private val delegateEncoder: CompositeEncoder`.

Although it may also make sense to have this retained (and used for subStructures, although in such case the element index would be an important parameter)

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    override fun encodeInlineElement(descriptor: SerialDescriptor, index: Int): Encoder {
+        return BufferingInlineEncoder(descriptor, index)
+    }
+
+    override fun shouldEncodeElementDefault(descriptor: SerialDescriptor, index: Int): Boolean {
+        callShouldEncodeElementDefault = true
+        return true
+    }
```
pdvrieze, line 136: I don't think you need/want this. The delegate/final encoder should be happy if it is never called even for the "default" value. As such `callShouldElementDefault` isn't needed. If it were to return false the field is just not written (and kept null/skipped). 

If you wish to have the encoder to force encoding defaults (rather than just delegating it to the composite) just have a property for that. `return forceEncodingDefaults || delegateEncoder.shouldEncodeElementDefault(descriptor, index)`. I'd be very surprised if there is any `CompositeEncoder` that is stateful in this function.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    private inner class BufferingInlineEncoder(
+        private val descriptor: SerialDescriptor,
+        private val elementIndex: Int,
+    ) : Encoder {
+        private var nullable = false
```
pdvrieze, line 142: This name is confusing. It is a state that indicates `encodeNotNullMark` was called and needs to be replayed.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        override fun encodeInline(descriptor: SerialDescriptor): Encoder {
+            invalidCall("encodeInline")
+        }
+
+        private fun invalidCall(methodName: String): Nothing {
+            // This method is normally called by encodeSerializableValue or encodeNullableSerializableValue that is buffered, so we should never go here during buffering as it will be delegated to the concrete CompositeEncoder
+            throw UnsupportedOperationException("$methodName should not be called when reordering fields, as it should be done by encodeSerializableValue or encodeNullableSerializableValue")
+        }
```
pdvrieze, line 226: This is incorrect. It is valid to have an inline/value class have a struct member. That would just result in a nested reordering encoder (delay-created with the result of calling `beginStructure`).

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+        bufferEncoding(descriptor, index) { encodeNullableSerializableElement(descriptor, index, serializer, value) }
+    }
+
+    override fun <T> encodeSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        serializer: SerializationStrategy<T>,
+        value: T
+    ) {
+        bufferEncoding(descriptor, index) { encodeSerializableElement(descriptor, index, serializer, value) }
```
pdvrieze, line 118: This is incorrect. You will need a wrapping encoder so `Encoder.beginStructure` will work correctly. To make that work you also need a wrapping `SerializationStrategy` that captures the delegate encoder (for the wrapping) unless you add a way to determine the encoder in another way.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
```diff
@@ -0,0 +1,240 @@
...
+        bufferEncoding(descriptor, index) { encodeShortElement(descriptor, index, value) }
+    }
+
+    override fun encodeStringElement(descriptor: SerialDescriptor, index: Int, value: String) {
+        bufferEncoding(descriptor, index) { encodeStringElement(descriptor, index, value) }
+    }
+
+    override fun encodeInlineElement(descriptor: SerialDescriptor, index: Int): Encoder {
+        return BufferingInlineEncoder(descriptor, index)
+    }
```
pdvrieze, line 131: Note that the implementation of this should be equivalent to:
```thisEncoder. encodeSerializableElement(MyData. serializer. descriptor, 0, MyInt. serializer(), myInt)```

Including the support of nested structs (nested structs always go through `encodeSerializableElement`)

File: core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt [hunk 9]
```diff
@@ -0,0 +1,240 @@
...
+        }
+        callShouldEncodeElementDefault = false
+    }
+
+    override fun endStructure(descriptor: SerialDescriptor) {
+        encodeBufferedFields(descriptor)
+    }
+
+    private fun encodeBufferedFields(descriptor: SerialDescriptor) {
+        val classEncoder = beginStructure(descriptor)
```
pdvrieze, line 68: This can be called before creating this compositeEncoder in the call to `beginStructure` that returns this reordered one.

pdvrieze (commented):
A good starting point, but some issues, including a lack of proper support for nested structs (or a nested struct to be wrapped in a value class).

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: This is just in draft, for sure it's not perfect as is was meant to be discussed and have something concrete as a starting point 🙂 let's focus on the content first and then discuss about the public API, I hope it's good for you!

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: My main idea was to have a composite encoder just reordering, that way if the user is doing custom logic in the shouldEncodeDefault method, then it will continue working exactly as a standard encoding.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: The beginStructure is called by the serializer from inside the `encodeSerializableValue`, so it should never be called directly AFAIK 🤔

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Yes, this is working like this after reordering, or maybe I misunderstood

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: I don't understand why it may not work. The serializer is not going to be called during the buffering step. FYI I already tested with a data class with a value class in a value class containing a list, and another value class containing a data class, and beginStructure is called only inside the `encodeSerializableValue`

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: For the first part, I suppose you were talking about the supplier that should just be the encoder (no supplier), and I agree.

But for the second part, I don't understand, the buffer is just for the corresponding descriptor, that should not be shared with sub-structure: we are reordering one level of a given descriptor, but not the sub structures.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
pdvrieze: There is no requirement for `shouldEncodeDefault` to be called at all. And per its own definition, there is no guarantee that it is followed by the encoder for that same index (so if you want to keep this, you have to change it to an int with the (original) element index/a negative value for absent)

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
pdvrieze: What I mean is that you have an encoder on which `beginStructure` is called. In the implementation it will create/call the reordering composite encoder (and return it), and perhaps also to the delegate encoder (to be passed to the reordering encoder).

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
pdvrieze: Say we have a structure: `{ a: { b: "foo", c: 42}, d: true }` then property `a` is encoded by the outer serializer (o) invoking `encodeSerializableElement(descriptor, 0, innerSerializer, value.a)`. This is then implemented by getting an `Encoder` instance and calling `innerSerializer.serialize(encoder, value)`. This `InnerSerializer.serialize` function then calls `beginStructure` (for the inner struct) which returns another `CompositeEncoder`.

If you implement this by delegating `encodeSerializableElement` to the non-reordered encoder it gets to decide what `CompositeEncoder` is used for the inner struct. If you interject it is possible for the ordered composite encoder to override it.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
pdvrieze: What about:
```kotlin
@Serializable data class Outer(val a: Middle, val b: Int)

@Serializable value class Middle(val i: Inner)

@Serializable value class Inner(val c: String, d: Boolean)
```
The way I read it, this at least can be implemented using `encodeInlineElement(d, 0)` on the composite encoder for outer. Which would then mean that the inner value is serialized by calling `encodeSerializableValue()` on that (`BufferingInlineEncoder`) encoder.

pdvrieze:
@Chuckame The buffering is basically correct (I've implemented this for XML for quite some time now)

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Oh I get it, if it returns false, then the encoding method is not called so no buffering is done. I would just try to not add another parameter here. I think I'll try to : use the final composite encoder's shouldEncodeDefault directly, or buffer the call.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Will be replaced by directly the encoder (no supplier)

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Has been changed to `compositeEncoderDelegate.shouldEncodeDefault` and implicitly does not buffer the encoding.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Replaced by `compositeEncoderDelegate: CompositeEncoder`

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: renamed to `encodeNotNullMarkCalled`

core/commonTest/src/kotlinx/serialization/encoding/ReorderingCompositeEncoderTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        val first = 42
+        val second = "string"
+        val third = true
+        val value = OuterToBeReordered(Middle(Inner(second, third)), first)
+
+        val encoder = ReorderingEncoder()
+
+        OuterToBeReordered.serializer().serialize(encoder, value)
+
+        assertContentEquals(actual = encoder.encodedValues, expected = listOf(first, second, third))
```
Chuckame, line 25: The test passed, no exception at all and it's encoded in the good order.

Btw, this test won't remains, but I cannot code as I have a big issue on IntelliJ, even with cache invalidation, I have everything red. This is the only project with that issue, but also the only multiplatform project that I have... Still searching, kotlin standard library completion doesn't work (mutableListOf, apply, UnsupportedOperationException), it's a nightmare 😅

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Test has been successful with your example, so I still don't get your point https://github.com/Kotlin/kotlinx.serialization/pull/2722/files#r1648164815

> the inner value is serialized by calling encodeSerializableValue() on that (BufferingInlineEncoder) encoder
No, here are the steps:
- `encodeInlineElement(d, 0)` -> returns `BufferingInlineEncoder` with descriptor of `Middle` and index 0, but does not buffer anything for the moment (this is the role of `BufferingInlineEncoder`)
- `BufferingInlineEncoder.encodeSerializableValue` is called, so nothing more is done, no nested call is done even for `Inner` thanks to the following `bufferEncoding` buffering `compositeEncoderDelegate.encodeInlineElement(descriptor, elementIndex).encodeSerializableValue(serializer, value)`
- then when `endStructure` is called, and replays the elements encoding

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: But as you said, it's calling `encodeSerializableElement` so it won't reach the `innerSerializer` during buffering, but only during the next step when encoding with the CompositeEncoder delegate.

Chuckame:
@pdvrieze I've tried with all the possible combinations, and value classes serializers are not directly using `encodeInlineElement` but `encodeSerializableElement` that is buffered.

Also, I just discovered a bug where `encodeInlineElement` is never called (see #2738)

Let's imagine that this bug doesn't exist, and `encodeInlineElement` is called as expected, `beginStructure` call on the inline's encoder is explicitly forbidden or discouraged as said in the docs:
```kotlin

    /**
     * [...]
     *
     * Calling [Encoder.beginStructure] on returned instance leads to an unspecified behavior and, in general, is prohibited.
     */
    public fun encodeInline(descriptor: SerialDescriptor): Encoder
```
So I would say that this use case shouldn't be handled to follow the docs, as this `beginStructure` call should be done by the structure's serializer.

core/commonMain/src/kotlinx/serialization/encoding/ReorderingCompositeEncoder.kt
Chuckame: Ok I got it. Is it such an expected standard use case to encode inlined structures like this `encoder.encodeInlineElement(desc, idx).beginStructure(...)` ? Shouldn't it be expected to be like the serializer plugin (`encoder.encodeInlineElement(desc, idx).encodeSerializableValue(...)`) where structures are always encoded through `encodeSerializableValue` or `encodeNullableSerializableValue` to allow interception and support nullability ?

Issue #2657, opened 2024-05-02
Performance Comparison: Kotlin Serialization vs Moshi


ngehad:
**Describe the bug**
This issue compares the performance of KotlinX Serialization with Moshi for JSON parsing and serialization operations. The benchmark focuses on time taken (execution time) and the number of objects allocated during the process. Benchmarking was done with Jetpack Microbenchmark, and we tested with small json response and a large one which counts approximately x21 the size of the small one.

TLDR: KotlinX serialization allocates way more objects in the memory than Moshi specially with large json response, and its execution time during serializing json is longer than Moshi.

Looking at the below table, in which we recorded our experiment results, we can see that:
1- With Small Payload (first 2 result rows): Moshi performed executed in slightly less time _when converting serializing json_, while kotlinx serialization executed in half the time of Moshi _when parsing to json_. Also, the number of objects allocated in the memory was slightly higher for serializing and equal for parsing to json.
2- With Larger Response (last 2 rows): KotlinX serialization has jumped to double the execution time of Moshi when serializing json, but it performed better (time-wise) when it parsed to json. On the other hand, KotlinX serialization looks to have allocated way too many objects in the memory compared to moshi for serialization, which seems off to me. (PS. I'm aware that the number of objects allocated doesn't exactly reflect amount of memory used but my concern is about the overhead created for the garbage collector).

<img width="747" alt="ktxserialization issue" src="https://github.com/Kotlin/kotlinx.serialization/assets/167112460/f7b6ad15-1c43-4136-98d2-7a9e5b85fdc6">

Detailed results in json format: [moshi results](https://gist.github.com/ngehad/62b41be7a345b8c74baea6bb8f78d3e4), [kotlinx serialization results](https://gist.github.com/ngehad/7bfa273b9ab6790ca0936b337a6a0d93)

Please share your thoughts if you think we're doing something wrong.

**To Reproduce**
Json Test data: [small json](https://gist.github.com/ngehad/72b10127020cb8ce2faf484ed381c1b4), [larger json](https://gist.github.com/ngehad/c2a415291f6dfd1be034317855f03f29)

1. Setup micro-benchmark library
2. generate model classes from json for both responses
3. Use measureRepeated method from BenchmarkRule object in a test
examples:
`benchmarkRule.measureRepeated {
            Json.decodeFromString<StResponse>(data)
        }`
`benchmarkRule.measureRepeated {
            Json.encodeToString(data)
        }`
[...]

PR #2742, opened 2024-07-11
Introduce the basic Moshi benchmark setup in order to use it for further comparisons

qwwdfsad:
Relates to #2657

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaseline.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaseline.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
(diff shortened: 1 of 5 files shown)


benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaseline.kt
```diff
@@ -0,0 +1,80 @@
...
+    fun moshiToString(): String = jsonAdapter.toJson(pixelEvent)
+
+    @Benchmark
+    fun moshiToOkio() = jsonAdapter.toJson(devNullSink, pixelEvent)
+
+    @Benchmark
+    fun moshiFromString(): DefaultPixelEvent = jsonAdapter.fromJson(pixelEventJson)!!
+
+    @Benchmark
+    fun moshiFromSource(): DefaultPixelEvent = jsonAdapter.fromJson(source.copy())!!
```
shanshin, line 65: For each iteration, we call `copy()`. This is not critical, because in fact, only the creation of a new instance of the `Segment` takes place there.

shanshin (approved):

PR #2754, opened 2024-07-30
Clarify that Json.isLenient flag only affects invalid json such as unquoted strings and does not affect quoted values type mapping.

### LINKED ISSUE #2749: Quoted booleans are being allowed even with `isLenient = false`
**Describe the bug**
I have `isLenient = false` but the serializer is allowing quoted booleans, which I believe goes against [RFC-4627](https://www.ietf.org/rfc/rfc4627.txt).

Example:
```json
{
 "aBoolean": "true"
}
```

The JSON above should not be accepted based on RFC-4627.

The issue was introduces by #2440.

**To Reproduce**
User the following JSON configuration and try to parse a quoted boolean as mentioned above:
```kotlin
Json {
  isLenient = false
}
```

**Expected behavior**
The parser should fail and throw an exception:  `JsonConvertException`

**Environment**
 - Kotlin version: 1.9.20
 - Library version: 1.6.3

sandwwraith:
Fixes #2749

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -121,14 +121,14 @@ fun main() {
 
 > You can get the full code [here](../guide/example/example-json-02.kt).
 
-You get the object, even though all keys of the source JSON, string, and enum values are unquoted:
+You get the object, even though all keys of the source JSON, string and enum values are unquoted:
 
 ```text
 Project(name=kotlinx.serialization, status=SUPPORTED, votes=9000)
 ```
 
 > Note that parsing of quoted numbers or booleans such as `votes: "9000"` to `val votes: Int` is generally allowed by kotlinx.serialization
-> regardless of `isLenient` flag, since such JSON is syntactically valid.
+> regardless of the `isLenient` flag, since such JSON is syntactically valid.
 
 <!--- TEST -->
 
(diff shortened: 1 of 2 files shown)


docs/json.md
```diff
@@ -121,13 +121,15 @@ fun main() {
 
 > You can get the full code [here](../guide/example/example-json-02.kt).
 
-You get the object, even though all keys of the source JSON, string, and enum values are unquoted, while an
-integer is quoted:
+You get the object, even though all keys of the source JSON, string, and enum values are unquoted:
```
shanshin, line 124:
```suggestion
You get the object, even though all keys of the source JSON, string and enum values are unquoted:
```

docs/json.md
```diff
@@ -121,13 +121,15 @@ fun main() {
...
 
 ```text
 Project(name=kotlinx.serialization, status=SUPPORTED, votes=9000)
 ```
 
+> Note that parsing of quoted numbers or booleans such as `votes: "9000"` to `val votes: Int` is generally allowed by kotlinx.serialization
+> regardless of `isLenient` flag, since such JSON is syntactically valid.
```
shanshin, line 131:
```suggestion
> regardless of the `isLenient` flag, since such JSON is syntactically valid.
```

shanshin (approved):

PR #2755, opened 2024-07-31
Update platforms for kaml library

krzema12:
It now supports all platforms.


krzema12:
CC @charleskorn @aSemy

PR #2756, opened 2024-07-31
Bump Kotlin version to 2.0.20-RC

shanshin:


formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonCoerceInputValuesTest.kt
```diff
@@ -49,6 +49,9 @@ class JsonCoerceInputValuesTest : JsonTestBase() {
         isLenient = true
     }
 
+    private val jsonLenient = Json(json) { isLenient = false }
```
sandwwraith, line 52: `jsonNonLenient` sounds more reasonable here

PR #2764, opened 2024-08-07
Introduce @SubclassOptInRequired to codebase and stabilize part of SerialDescriptor's API

sandwwraith:


sandwwraith:
@qwwdfsad, although I'm not sure if we should introduce it now since https://youtrack.jetbrains.com/issue/KTIJ-30831 may lead to a poor UX

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptor.kt
```diff
@@ -275,19 +292,17 @@ public interface SerialDescriptor {
...
      */
-    @ExperimentalSerializationApi
     public fun isElementOptional(index: Int): Boolean
 }
 
 /**
  * Returns an iterable of all descriptor [elements][SerialDescriptor.getElementDescriptor].
  */
-@ExperimentalSerializationApi
 public val SerialDescriptor.elementDescriptors: Iterable<SerialDescriptor>
```
qwwdfsad, line 306: While we are here, let's add a short sample of code?

qwwdfsad: The same goes for all the other "stable" parts

core/commonMain/src/kotlinx/serialization/descriptors/SerialKinds.kt
```diff
@@ -239,7 +234,7 @@ public sealed class StructureKind : SerialKind() {
  * bounded and sealed polymorphism common property: not knowing the actual type statically and requiring
  * formats to additionally encode it.
  */
-@ExperimentalSerializationApi
+@ExperimentalSerializationApi // Intentionally left experimental to sort out things with buildSerialDescriptor(PolymorphicKind.SEALED)
```
qwwdfsad, line 237: Thanks for the comment!

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptor.kt
sandwwraith: I've added examples for `SerialDescriptor` and its builders. I believe that `SerialKind`s don't have to have them

qwwdfsad (approved):
Great job!

One meta-comment: eventually we'll switch to `@sample` that will be auto-runnable from kotlinlang and from IDE. 
For that, you probably may want to change the sample form from
```
someFoo // computes to "kotlin.somefoo"
```

to the following
```
val someFooName = someFoo
print(someFoo) // Prints "kotlin.somefoo"
```

sandwwraith:
Closed in favor of #2827

PR #2768, opened 2024-08-10
Improve readability of proto decode exception message

xiaozhikang0916:
Found it hard to debug when exceptions with message like `Expected wire type 0, but found 1` were got. This PR is about to make some improvement of these messages for better tracability and readability.

1. Wrap definition of wire types with enum class `ProtoWireType` to show both name and id in message,
2. Catch and rethrow any `ProtobufDecodingException` in most `decodeXXX` functions, with proto number and type name in new exception message.

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
@@ -22,7 +22,7 @@ internal enum class ProtoWireType(val typeId: Int) {
     ;
 
     companion object {
-        fun fromTypeId(typeId: Int): ProtoWireType {
+        fun from(typeId: Int): ProtoWireType {
             return ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
         }
     }
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
@@ -12,10 +12,29 @@ import kotlinx.serialization.modules.*
 import kotlinx.serialization.protobuf.*
 
 internal typealias ProtoDesc = Long
-internal const val VARINT = 0
-internal const val i64 = 1
-internal const val SIZE_DELIMITED = 2
-internal const val i32 = 5
+
+internal enum class ProtoWireType(val typeId: Int) {
+    INVALID(-1),
+    VARINT(0),
+    i64(1),
+    SIZE_DELIMITED(2),
+    i32(5),
+    ;
+
+    companion object {
+        fun fromTypeId(typeId: Int): ProtoWireType {
+            return ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
+        }
+    }
+
+    fun wireIntWithTag(tag: Int): Int {
+        return ((tag shl 3) or typeId)
+    }
+
+    override fun toString(): String {
+        return "${this.name}($typeId)"
+    }
+}
 
 internal const val ID_HOLDER_ONE_OF = -2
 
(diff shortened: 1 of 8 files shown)


xiaozhikang0916:
@sandwwraith can you take a look on this PR?

sandwwraith:
@xiaozhikang0916 sure

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -12,10 +12,29 @@ import kotlinx.serialization.modules.*
 import kotlinx.serialization.protobuf.*
 
 internal typealias ProtoDesc = Long
-internal const val VARINT = 0
-internal const val i64 = 1
-internal const val SIZE_DELIMITED = 2
-internal const val i32 = 5
+
+internal enum class ProtoWireType(val typeId: Int) {
```
Goooler, line 16: This could be a value class.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -12,10 +12,29 @@ import kotlinx.serialization.modules.*
...
+    INVALID(-1),
+    VARINT(0),
+    i64(1),
+    SIZE_DELIMITED(2),
+    i32(5),
+    ;
+
+    companion object {
+        fun fromTypeId(typeId: Int): ProtoWireType {
```
Goooler, line 25:
```suggestion
        fun from(typeId: Int): ProtoWireType {
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -12,10 +12,29 @@ import kotlinx.serialization.modules.*
...
+        }
+    }
+
+    fun wireIntWithTag(tag: Int): Int {
+        return ((tag shl 3) or typeId)
+    }
+
+    override fun toString(): String {
+        return "${this.name}($typeId)"
+    }
```
Goooler, line 36: This could be placed into the value class.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
Goooler: Oooops, I know that it's sealed, not suitable for using value class. Nevermind for this.

xiaozhikang0916: > it's sealed, not suitable for using value class.

You got the point

sandwwraith (changes requested):

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/TestFunctions.kt
```diff
@@ -24,9 +25,74 @@ inline fun <reified T : Throwable> assertFailsWithMessage(
...
+    override fun assertFailsWith(vararg message: String) {
+        val exceptionStackSize = exception.exceptionStackSize
+        assertTrue(message.size <= exceptionStackSize, "Expected exception to be assembled by ${message.size} throwable, but it has $exceptionStackSize, is $exception")
+        var index = 0
+        var currentException: Throwable? = exception
+        while (index < message.size) {
+            val currentMessage = message[index]
+            assertNotNull(currentException, "Expected exception to have a cause with message $currentMessage, but it was null")
+            require(currentException != null)
+            assertEquals(currentMessage, currentException.message, "Expected exception to have a cause with message <$currentMessage>, but it was <${currentException.message}> at cause stack ${index + depth}")
```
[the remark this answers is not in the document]
xiaozhikang0916: After diving into `assertEquals` I found that it would append the `Expected <$expected>, actual <$actual>.` string after message passing in. So I would just simplify the input message.

It works in my IntelliJ IDEA 2024.2.1 RC and you can comment out the `Ignore` annotation in the new `TestFunctionTest` to see if it also works in other environments. And since these tests are for checking IDEA behaviours, I would just leave it failing, not wraping another `assertFailWith` with message check.

PR #2787, opened 2024-08-19
Docs: Fix code snippet mark

hfhbd:
Current doc:
Bildschirmfoto 2024-08-19 um 22 17 22


core/commonMain/src/kotlinx/serialization/encoding/Decoding.kt
```diff
@@ -358,7 +358,7 @@ public interface CompositeDecoder {
      * Sequential decoding is a performance optimization for formats with strictly ordered schema,
      * usually binary ones. Regular formats such as JSON or ProtoBuf cannot use this optimization,
      * because e.g. in the latter example, the same data can be represented both as
-     * `{"i": 1, "d": 1.0}`"` and `{"d": 1.0, "i": 1}` (thus, unordered).
+     * `{"i": 1, "d": 1.0}`" and `{"d": 1.0, "i": 1}` (thus, unordered).
```
sandwwraith, line 361: Shouldn't it be without " because there is no opening one?
```
`{"i": 1, "d": 1.0}`
```

PR #2789, opened 2024-08-21
Remove redundandant `toString()`

lunakoly:
The corresponding checker may be promoted to a default one, resulting into a warning in this place. See: KT-69938.

Sorry for the outdated commit title & message. If there are no other changes needed, I'd prefer having the right one during the rebase&merge rather than doing a force push.

### DIFF
--- a/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
+++ b/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicDecoders.kt
@@ -248,7 +248,7 @@ private class DynamicMapInput(
         if (isKey) {
             val value = decodeTaggedValue(tag)
             if (value !is String) return decode(tag)
-            return value.toString().cast() ?: throwIllegalKeyType(tag, value, type)
+            return value.cast() ?: throwIllegalKeyType(tag, value, type)
         }
         return decode(tag)
     }


sandwwraith:
Hm, there is no 'redundant toString()' warning even in K1 IDE, but OK

sandwwraith (approved):

sandwwraith:
Ah, I noticed it too late, but all PRs should be against `dev`. You can cherry-pick to `kotlin-community/dev` later manually. Can you open PR with this commit to `dev`, please?

lunakoly:
Sure: https://github.com/Kotlin/kotlinx.serialization/pull/2790

lunakoly:
Isn't such a workflow with `dev` and `kotlin-community/dev` a bit cumbersome? On TC we run `kotlin-community/dev` against Kotlin aggregate builds, so one would still have to work on `kotlin-community/dev` first (and keep it during the MR to `kotlin.git`), but then rebase the changes onto `dev` to cherry-pick them back into `kotlin-community/dev`. Or am I missing something?

sandwwraith:
Ideally, there shouldn't be any `kotlin-community/dev` branches and both tests with Kotlin master (in libraries) and Kotlin's Aggregate build should be run against `dev`.

PR #2793, opened 2024-08-25
Update Annotations.kt, fixed typo: RgpPixel -> RgbPixel (p->b)

asnov:
It was RgpPixel but it should be RgbPixel as the class with this name was declared before:
`class RgbPixel(val red: Short, val green: Short, val blue: Short)`


sandwwraith (approved):
Thank you!

PR #2806, opened 2024-09-02
Clarify example for SerializationException

sandwwraith:


core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -22,12 +22,14 @@ import kotlinx.serialization.descriptors.*
...
- * class Foo(...) {
+ * class User(val age: Int, val name: String) {
  *     init {
- *         required(age > 0) { ... }
+ *         require(age > 0) { ... }
  *         require(name.isNotBlank()) { ... }
  *     }
  * }
+ *
+ * Json.decodeFromString<User>("""{"age": -100, "name": ""}""") // throws IAE from require()
```
shanshin, line 32: I think it's better to write the full name of the exception, because there is no such abbreviation anywhere else.

shanshin (approved):

PR #2833, opened 2024-10-10
Do not check kind or discriminator collisions

### LINKED ISSUE #1486: Serializer of kind ENUM cannot be serialized polymorphically with class discriminator
**Describe the bug**
java.lang.IllegalArgumentException: Serializer for ChangeAlarmTypeRequest of kind ENUM cannot be serialized polymorphically with class discriminator

All subclasses of MethodExecuteCode  have this problem

**To Reproduce**
```
polymorphic(MethodExecuteCode::class) {
    subclass(MethodExecuteStateTypeCode::class)
    subclass(ChangeAlarmTypeRequest::class)
    subclass(IncidentMethodCode::class)
}


interface MethodExecuteCode

@Serializable
enum class MethodExecuteStateTypeCode: MethodExecuteCode {

    @SerialName("create") CREATE,
    @SerialName("enable") ENABLE,
    @SerialName("disable") DISABLE,
    @SerialName("delete") DELETE,

}

@Serializable
enum class ChangeAlarmTypeRequest : MethodExecuteCode {
    @SerialName("alarm") START,
    @SerialName("cancel") STOP,
    @SerialName("update") UPDATE
}

@Serializable
enum class IncidentMethodCode: MethodExecuteCode {

    @SerialName("accept") ACCEPT,
    @SerialName("reject") REJECT,

}

```
**Environment**
 - Kotlin version: 1.5.0
 - Library version: 1.2.1
 - Kotlin platforms: JVM Android
 - Gradle version: 7.0.0
 - Other relevant context: Java 8

### LINKED ISSUE #2753: Don't check kind if needn't type discriminator for enum and primitive
Output the enum and primitive correctly.
If we add a PrimitiveKind descriptor to a subclass of polymorphic serializer. There is error #1486
If we add a wrapper for the primitive(https://github.com/Kotlin/kotlinx.serialization/issues/1252#issuecomment-2257655863). The result will be wrapped even if the `classDiscriminatorMode = ClassDiscriminatorMode.NONE` like below.
```json
{
  "data": {
    "value": "FOO"
  }
}
```

Related: #1252 #1486

https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/internal/Polymorphic.kt#L41
Avoid failing if needn't discriminator for `checkKind`. Add a condition in `checkKind` or `encodePolymorphically`
```kotlin
val kind = actual.descriptor.kind
if(needDiscriminator || (kind !is SerialKind.ENUM && kind !is PrimitiveKind)) checkKind(kind )
```

sandwwraith:
for subclasses' polymorphic serializers if Json.classDiscriminatorMode is set to NONE.

This should simplify the code in situations where JSON is expected to be only sent and not parsed.

Fixes #2753
Fixes #1486

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonClassDiscriminatorModeTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonClassDiscriminatorModeTest.kt
@@ -99,7 +99,7 @@ class ClassDiscriminatorModeNoneTest :
                 subclass(Modify::class)
             }
         }
-        val j = Json { serializersModule = module; classDiscriminatorMode = ClassDiscriminatorMode.NONE }
+        val j = Json(default) { serializersModule = module; classDiscriminatorMode = ClassDiscriminatorMode.NONE }
         parametrizedTest { mode ->
             assertEquals("""{"cmd":"CREATE"}""", j.encodeToString(Command(Modify.CREATE), mode))
         }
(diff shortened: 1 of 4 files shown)


shanshin (commented):
I think somewhere in the documentation you should specify a new behavior so that it does not come as a surprise to the user.

PR #2846, opened 2024-11-01
Reformat "libs.versions.toml" and "settings.gradle.kts"

ShreckYe:
Hello, I was experimenting with a new module I added and noticed that some project configuration files are not properly formatted. When I committed the code, IntelliJ IDEA tried to format it for me. I guess this can be fixed for contributors like me so it doesn't cause merge conflicts when people accidentally format code in different branches to be merged.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -13,7 +13,7 @@ guava = "31.1-jre"
 guava24 = "24.1.1-jre"
 jackson = "2.13.3"
 okio = "3.9.0"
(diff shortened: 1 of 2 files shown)


sandwwraith:
Tip: you can use 'Only changes uncommitted to VCS' in reformat settings to avoid changes in unrelated lines.
image

Thanks for the correction though!

sandwwraith (approved):

ShreckYe:
Thanks. I did not know this feature.

PR #2852, opened 2024-11-08
Correctly parse invalid numbers in JsonLiteral.long and other extensions

### LINKED ISSUE #2849: Missing json validation on parsing with intermediate JsonObject representation
**Describe the bug**

Long values are not validated to be actual longs when parsed from JsonObject

**To Reproduce**
```

@Serializable
data class A(val a: Long)

fun main() {
    val str = """{"a":"12:34:45"}"""
    println(Json.decodeFromJsonElement<A>(Json.decodeFromString<JsonObject>(str)))
    println(Json.decodeFromString<A>(str))
}
```

**Expected behavior**

Both crash with SerializationException. 
Actually first prints `A(a=12)`, while second one crashes. Btw, the error message in the second one is really misleading. 

**Environment**
 - Kotlin version: 2.0.20
 - Library version: 1.7.3
 - Kotlin platforms: JVM

sandwwraith:
Content must be consumed fully, with no leftovers after number. Also simplify try/catching logic — JsonLiteral.long should throw NumberFormatException, while decoding from JsonElement should throw JsonDecodingException

Fixes #2849

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt
@@ -4,6 +4,7 @@
 
 package kotlinx.serialization.json.serializers
 
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.*
 import kotlinx.serialization.test.*
 import kotlin.test.*
@@ -200,5 +201,18 @@ class JsonPrimitiveSerializerTest : JsonTestBase() {
         expectedActualULongs.forEach { (expected, actual) ->
             assertUnsignedNumberEncoding(expected, actual, JsonPrimitive(actual))
         }
+    }
+
+    @Serializable
+    class OuterLong(val a: Long)
+
+    @Test
+    fun testRejectingIncorrectNumbers() = parametrizedTest { mode ->
+        checkSerializationException({
+            default.decodeFromString(OuterLong.serializer(), """{"a":"12:34:45"}""", mode)
+        }, {
+            if (mode == JsonTestingMode.TREE) assertContains(it, "Failed to parse literal '\"12:34:45\"' as a long value at element: \$.a")
+            else assertContains(it, "Unexpected JSON token at offset 5: Expected closing quotation mark at path: \$.a")
+        })
     }
 }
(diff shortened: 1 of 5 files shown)


formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt
```diff
@@ -201,4 +202,17 @@ class JsonPrimitiveSerializerTest : JsonTestBase() {
...
+
+    @Serializable
+    class OuterLong(val a: Long)
+
+    @Test
+    fun testRejectingIncorrectNumbers() = parametrizedTest { mode ->
+        checkSerializationException({
+            default.decodeFromString(OuterLong.serializer(), """{"a":"12:34:45"}""", mode)
+        }, {
+            if (mode == JsonTestingMode.TREE) assertContains(it, "Failed to parse literal '\"12:34:45\"' as a long value at element: \$.a")
```
shanshin, line 214: Should we also check exception type?

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -223,12 +223,16 @@ internal abstract class AbstractJsonLexer {
...
+        expectedToken: Byte,
+        wasConsumed: Boolean = true,
+        message: (expected: String, source: String) -> String = { expected, source -> "Expected $expected, but had '$source' instead" }
+    ): Nothing {
         // Slow path, never called in normal code, can avoid optimizing it
         val expected = tokenDescription(expectedToken)
         val position = if (wasConsumed) currentPosition - 1 else currentPosition
         val s = if (currentPosition == source.length || position < 0) "EOF" else source[position].toString()
-        fail("Expected $expected, but had '$s' instead", position)
+        fail(message(expected, s), position)
```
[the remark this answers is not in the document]
sandwwraith: Yes, because generally, we do not know where input ends (e.g. in the case of `{"a": 12312:xdwadaw213123123413.....}`, incorrect part may potentially be as long as whole source itself

formats/json-tests/commonTest/src/kotlinx/serialization/json/serializers/JsonPrimitiveSerializerTest.kt
sandwwraith: `checkSerializationException` makes sure that only `SerializationException`s are accepted

shanshin (approved):

PR #2853, opened 2024-11-08
Add inline reified version of encodeToString as a Json member to to streamline experience for newcomers.

### LINKED ISSUE #2850: Consider moving reified overload of `encodeTo` and `decodeFrom*` into popular formats
Otherwise, the tooling starts crumbling and suggesting wrong overloads with lack of discoverability.

Also, things like `Json.encodeToString(foo)` and `Json.encodeToString(FooSerializer, foo)` live in different packages, providing weird autocompletion side-effects (when tab- based completion accidentally import the wrong method)

sandwwraith:
decodeFromString was already there for @FormatLanguage reasons.

Fixes #2850


hfhbd:
This still means every 3rd party author needs to copy these inline functions too, isn't it?

sandwwraith:
@hfhbd No, why? Original inline extension functions are not going anywhere.

hfhbd:
Yes that's true but the user still needs to import the extension functions from kotlinx.serialization while using FooFormat in `com.foo` because it is an extension function and not part of the SerialFormat as a member (not requiring an import).

sandwwraith:
Yes, just as any other extension function? I do not see a problem here

hfhbd:
It is not a problem for the json format, but a general feedback I got from users of my format that they forgot to import the extension method so instead fixing it in Json only (by copying the methods), a general solution to migrate the extension functions to members would improve all 3rd party formats too. Then no other format authors needs to copy the extension functions to have them as a member.

sandwwraith:
>  a general solution to migrate the extension functions to members would improve all 3rd party formats, too

That is true, but unfortunately, interfaces can't contain final inline functions.

So if your users complain about the same problem a lot, the only workaround here is to copy these inline functions, as in this PR

PR #2860, opened 2024-11-18
Support preserving unknown fields in ProtoBuf format

### LINKED ISSUE #2655: Supporting protobuf preserving unknown fields
I would like to support preserving unknown fields for ProtoBuf, here is an issue for discussion beforehand.

## Motivation

Basically it is required in [protobuf spec](https://protobuf.dev/programming-guides/proto3/#unknowns), so it needs to be implemented to make `kotlinx.serialization` a valid protobuf parser.

## Basic design

A new helper class will be introduced

```kotlin
interface UnknownFieldsHolder: Map<ProtoId, ProtoField>
```

Users can add to class property according to their needs:

```kotlin
@Serializable
data class ForwardCompatibleData(
    @ProtoNumber(1) val name: String,
    // No @ProtoNumber needed
    // Maybe introducing a new annotation?
    val unknownFields: UnknownFieldsHolder,
)
```

### Behaviours

* The ProtoBuf format should keeps all undeclare fields in the `UnknownFieldsHolder` when deserializing;
* All data in the `UnknownFieldsHolder` should be written back to wire as is during serializing;
* If new fields are added in `oneof` field, the new value will be stored in `UnknownFieldsHolder`, leaving the @ProtoOneOf property to be null;
* Other formats should not deseralize or serialize this property.

### Enumeration

Enums in protobuf can also have `unknown values` in old version of code, but it is much more easier to store it in sealed interface / objects styles:

```kotlin
sealed interface IPhoneType {
  val value: Int
}
data object WorkPhone: IPhoneType {
  override val value: Int
    get() = 1
}
data object HomePhone: IPhoneType {
  override val value: Int
    get() = 2
}
data class UnknownPhoneType(override val value: Int): IPhoneType
```

### Other formats

Do other formats need similar functionality?

xiaozhikang0916:
Introducing new annotation `@ProtoUnknownFields` and built-in type `ProtoMessage`, to store any fields with proto ids that are undefined in current message class.

In my understanding, both ktx-serialization and protocol buffer are designed in strong typing, all messages used in code should be well defined in kt class or proto message. Therefore I would like to keep `ProtoMessage` and `ProtoField` minimun published. Not like `JsonObject` in json format, users cannot retrieve detailed data from it.

Close #2655


xiaozhikang0916:
@sandwwraith Hi, please take a look here?

sandwwraith:
@xiaozhikang0916 Sure, when I'll have time

pdvrieze (commented):
The structure you use to represent protobuf content is somewhat complex, especially considering the protobuf wire format. You probably want to think a bit more how to represent protobuf data where the wire based information permits that. In particular you may want to distinguish primitives, "blobs", packed arrays. You may also allow users to observe a protobuf with provided metadata.

xiaozhikang0916:
@pdvrieze Actually I don't want to decode the protobuf wire data into a tree-structure type like what `JsonObject` do, neither provide retreiving method of the content.

The raw content in protobuf wire data is quite ambiguous that, you can even not possible to tell whether a `VARINT 1` is 1 or -1 or true, or tell a `LEN` field is string or sub-message. The only way to parse wire data content correctly and explictly is to parse with message definition, or with a concrete type in kotlin.

Base on it, `ProtoMessage` here is not designed for retrieving content from undefined fields, but for holding it only.

xiaozhikang0916:
Any progress about it? We have plan on upgrading Kotlin and serialization runtime version, and it would be great if we can have this functionality in near future.

docs/formats.md
```diff
@@ -642,6 +643,54 @@ is also compatible with the `message Data` given above, which means the same inp
 
 But please note that there are no exclusivity checks. This means that if an instance of `Data2` has both (or none) `homeNumber` and `workNumber` as non-null values and is serialized to protobuf, it no longer complies with the original schema. If you send such data to another parser, one of the fields may be omitted, leading to an unknown issue.
 
+### Preserving unknown fields (experimental)
+
+You may keep updating your schema by adding new fields, but you may not want to break compatibility with the old data.
+
+Kotlin Serialization `ProtoBuf` format supports preserving unknown fields, as described in the [Protocol Buffer-Unknown Fields](https://protobuf.dev/programming-guides/proto3/#unknowns).
+
+To keep the unknown fields, add a property in type `ProtoMessage` with default value `null` or `ProtoMessage.Empty`, and annotation `@ProtoUnknownFields` to your data class.
```
shanshin: Should we be more strict here, eg use only nullable (null if there are no unknown fields) or non-null property (ProtoMessage.Empty if there are no unknown fields)?

shanshin: @sandwwraith, I have a slight fear that the name `ProtoMessage` is too general - users may think that this is a universal class with a map-like structure

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -69,21 +75,34 @@ internal open class ProtobufDecoder(
...
+                ID_HOLDER_ONE_OF -> {
+                    descriptor.getElementDescriptor(i)
+                        .getAllOneOfSerializerOfField(serializersModule)
+                        .map { it.extractParameters(0).protoId }
+                        .forEach { map.putProtoId(it, i) }
+                    mapSize ++
+                }
+                ID_HOLDER_UNKNOWN_FIELDS -> {
+                    require(unknownHolderIndex == -1) {
+                        "Only one unknown fields holder is allowed in a message"
```
shanshin: Should we write descriptor name here?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -69,21 +75,34 @@ internal open class ProtobufDecoder(
...
+                    }
+                    require(descriptor.getElementDescriptor(i).nullable == ProtoMessageSerializer.descriptor.nullable) {
+                        "ProtoUnknownFields is only allowed on property with type kotlinx.serialization.protobuf.ProtoMessage, with its original serializer."
```
shanshin: And here

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -317,14 +339,16 @@ internal open class ProtobufDecoder(
                 if (index == -1) { // not found
```
shanshin: `-1` looks more and more like a magic value now, should we create a constant for it (for example, to show that `unknownHolderIndex` is initialized with `-1` for a reason)?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufReader.kt
```diff
@@ -65,6 +65,16 @@ internal class ProtobufReader(private val input: ByteArrayInput) {
         }
     }
 
+    fun readRawElement(): ByteArray {
+        return when (currentType) {
+            ProtoWireType.VARINT -> byteArrayOf(readInt(ProtoIntegerType.DEFAULT).toByte())
```
shanshin: Is there a bug here?
Int is converted to Byte, resulting in a byte array of one element. 

It's necessary to add a test for large values of each of the specified types.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
```diff
@@ -63,4 +63,14 @@ public annotation class ProtoPacked
...
+
+/**
+ * Mark a property with type [ProtoMessage] as a holder for unknown fields in protobuf message.
+ *
+ * All the contents with unregistered proto number will be stored in this field.
+ */
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY)
+@ExperimentalSerializationApi
+public annotation class ProtoUnknownFields
```
shanshin: We should add the compiler plugin check that this annotation is used only once in a class and only with a specific type of property (`ProtoMessage` now)

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoMessageSerializer.kt
```diff
@@ -0,0 +1,90 @@
...
+    override fun serialize(encoder: Encoder, value: ProtoField) {
+        if (encoder is ProtobufEncoder) {
+            encoder.encodeRawElement(value.id, value.wireType, value.data.byteArray)
+        }
+    }
+}
+
+internal class UnknownFieldsDescriptor(private val original: SerialDescriptor) : SerialDescriptor by original {
+    override val serialName: String
+        get() = "UnknownProtoFieldsHolder"
```
shanshin: Name should contains original descriptor name

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufEncoding.kt
```diff
@@ -142,10 +142,36 @@ internal open class ProtobufEncoder(
...
+        }
         serializer.descriptor == ByteArraySerializer().descriptor -> serializeByteArray(value as ByteArray)
         else -> serializer.serialize(this, value)
     }
 
+    internal fun encodeRawElement(id: Int, wireType: ProtoWireType, data: ByteArray) {
+        when(wireType) {
+            ProtoWireType.INVALID -> {}
+            ProtoWireType.VARINT -> writer.writeInt(
+                value = data.first().toInt(),
```
shanshin: Only first first byte is read as an Int, for numbers greater than 127, there will be an error.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufEncoding.kt
```diff
@@ -142,10 +142,36 @@ internal open class ProtobufEncoder(
...
+                tag = id,
+                format = ProtoIntegerType.DEFAULT
+            )
+
+            ProtoWireType.i64 -> writer.writeLong(
+                value = data.first().toLong(),
```
shanshin: Same here

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufEncoding.kt
```diff
@@ -142,10 +142,36 @@ internal open class ProtobufEncoder(
...
+                tag = id,
+                format = ProtoIntegerType.FIXED
+            )
+            ProtoWireType.SIZE_DELIMITED -> writer.writeBytes(data, id)
+            ProtoWireType.i32 -> writer.writeInt(
+                value = data.first().toInt(),
```
shanshin: And here

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoMessage.kt
```diff
@@ -0,0 +1,199 @@
...
+        result = 31 * result + wireType.hashCode()
+        result = 31 * result + data.contentHashCode()
+        return result
+    }
+}
+
+/**
+ * A data representation of a protobuf field in [ProtoField.data], without the field number and wire type.
+ */
+internal sealed interface ProtoContentHolder {
```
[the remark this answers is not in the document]
shanshin: In current implementation ProtoContentHolder is useless, `MessageContent` is not used anywhere.

It should be deleted, and `ProtoField` should only store a byte array, or indeed, there should be heirs for each type (VARINT, i64, SIZE_DELIMITED, etc), without unnecessary conversion to a byte array.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoMessage.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Number of fields holding in the message.
+     */
+    public val size: Int get() = fields.size
+
+    /**
+     * Returns a byte array representing of the message.
+     */
+    public fun asByteArray(): ByteArray =
```
shanshin: The current implementation looks like it overload the GC, but it is not used anywhere.
I would prefer to delete it in this form.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoMessage.kt
```diff
@@ -0,0 +1,199 @@
...
+    internal val wireType: ProtoWireType,
+    internal val data: ProtoContentHolder
+) {
+    companion object {
+        val Empty: ProtoField = ProtoField(0, ProtoWireType.INVALID, ProtoContentHolder.ByteArrayContent(ByteArray(0)))
+    }
+
+    fun asWireContent(): ByteArray = byteArrayOf(((id shl 3) or wireType.typeId).toByte()) + data.byteArray
+
+    val contentLength: Int
```
shanshin: The size can be calculated without creating a new byte array.

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufUnknownFieldsTest.kt
```diff
@@ -0,0 +1,362 @@
...
+
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlin.test.*
+
+class ProtobufUnknownFieldsTest {
+    @Serializable
```
shanshin, line 14: Tests are required for all supported types of unknown fields in extreme values:
VARINT, i64, SIZE_DELIMITED, i32, for values like 0/empty message, -1, MIN_VALUE, MAX_VALUE, message with length > 65kB, etc.

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufUnknownFieldsTest.kt
```diff
@@ -0,0 +1,362 @@
...
+         * 4: 42
+         * 5: {
+         *   1: {"42"}
+         *   2: 42
+         *   3: {"42"}
+         *   3: {"42"}
+         *   3: {"42"}
+         * }
+         */
+        val encoded = "082a120234321a032a2a2a202a202a202a2a120a023432102a1a0234321a0234321a023432"
```
shanshin: I would suggest building the main tests not around the hardcoded HEX values, but according to the following rule: everything that is encoded must be decoded in exactly the same form.

Steps:
- first encode what is
- then decode it to type with `ProtoUnknownFields`
- compare known fields
- encode the type with `ProtoUnknownFields` 
- this value must match the first encoded one

This is almost the same approach, but it will reduce the use of magic constants in all tests.

Of course, it is possible to check decoding from a constant HEX value, but one test will be enough for this.

docs/formats.md
xiaozhikang0916: You are right. `ProtoMessage` was introduced as general data container for ProtoBuf format for convenience of decoding/encoding and writing test code. All its read/write apis are kept internal to avoid creating universal map-like structure of proto. But it is too complex to rewrite the decoding/encoding logic with this type. I renamed this type and removed its unused fields.

xiaozhikang0916: `ProtoContentHolder` is also removed.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoTypes.kt
xiaozhikang0916: Do I need a pr in the kotlin repo?

shanshin: > Do I need a pr in the kotlin repo?

That would help a lot

xiaozhikang0916: A pr in kotlin repo: https://github.com/JetBrains/kotlin/pull/5755

shanshin:
LGTM, but I think @sandwwraith  should also take a cursory look to check the style and documentation.

sandwwraith:
Did we discuss an overall design for the feature? I have three questions:

1. If we support only `ProtoUnknownFieldHolder` type, the annotation seems unnecessary. We have full access to the descriptor everywhere, so we can simply check descriptor/serialName for equality.

2. Do we want to support only this type? It is typical to store unknown fields only in `ByteArray` — in case we are not interested in them at all, but want to preserve them when we'll serialize the message again later. Why choose a separate type? Especially given that `ProtoField` is internal, there's no way to access at least protonumbers of new fields (in case you want to, e.g., log or dump them).  `ProtoUnknownFieldHolder` probably makes more sense when ProtoField is public. But ProtoField contains a ByteArray anyway, so we're back to square one. 
3. Maybe we need functionality to parse a raw ByteArray to raw unknown fields? Then we can get in the cycle ProtoUnknownFieldHolder - ProtoField - ByteArray - inner message as ProtoUnknownFieldHolder - ProtoField - ....

xiaozhikang0916:
Hi @sandwwraith 

In my point of view, a byte array means nothing without a proto message (or class) definition, and unknown proto fields should be a format specific feature in ProtoBuf format, which should not be serialized to other format like Json. So I avoid exposing raw ByteArray field to developers, but introducing a seperate type `ProtoUnknownFieldHolder` with custom serializer to keep it internal to ProtoBuf only.

> Maybe we need functionality to parse a raw ByteArray to raw unknown fields?

In fact, I tried to write a `JsonElement` equivalent in my very first implementation, but soon I realized that it required a full rewrite of proto decoder. To keep this branch focused on one feature, I just simply wrap a ByteArray in the new `ProtoUnknownFieldHolder` type. And I am happy to see (or contribute to) the feature paring byte array to arbitrary `ProtoElement` type comes true.

To sum up, `ProtoUnknownFieldHolder` as a simple wrapper of `ByteArray` is necessary. It can be renamed to `ProtoElement` in the future if this feature is accepted, and an annotation will be necessary to mark which field should be used as the holder.

sandwwraith:
> and unknown proto fields should be a format specific feature in ProtoBuf format, which should not be serialized to other format like Json.

I imagine it is possible to transform a Kotlin's class decoded from protobuf to Json and then add something like `"unknownFields": 0xab341....`, although I do not know if there is any use case to do so. Maybe if you're writing some kind of protobuf-to-human-readable translator.

> To sum up, ProtoUnknownFieldHolder as a simple wrapper of ByteArray is necessary

In that case, is there a need for `ProtoField`? I think we can make it even more simple without it.

So, if you want to keep separate `ProtoUnknownFieldHolder`:
- Let's get rid of the annotation
- Don't try to parse inner structure to ProtoField

xiaozhikang0916:
I have tried 2 ways to improve it, but both of them have some limitations.

- A raw `ByteArray` field with annotation
  - It brings a raw `ByteArray` field to the data class, so the compiler will warn that you should  override `equals()` and `hashCode()` manually, and it brings extremely extra effort to devs.
- `ProtoUnknownFieldHolder` field without annotation
  - The proto decoder reads all infomation of elements from their annotations, and there is no class/type info in the `SerialDescriptor`, so I will need many instance comparation of `ProtoUnknownFieldHolder.serializer` here.  Please tell me if I missed anything.
  - An alternate way is injecting this annotation from compiler plugin, but it will bring more complexity to this feature.

Do you have any suggestion? @sandwwraith

sandwwraith:
> It brings a raw ByteArray field to the data class, so the compiler will warn that you should override equals() and hashCode() manually, and it brings extremely extra effort to devs.

Good point, I agree that ByteArray is not optimal in this case.

> The proto decoder reads all infomation of elements from their annotations, and there is no class/type info in the SerialDescriptor

There's `serialName,` which is usually sufficient for this kind of check. Or you can simply do `desciptor == ProtoUnknownFieldsSerializer.descriptor` as SerialDescriptors support equality, particularly for this purpose.

xiaozhikang0916:
Do we need any further more changes here? @sandwwraith

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -46,16 +46,27 @@ internal enum class ProtoWireType(val typeId: Int) {
 }
 
 internal const val ID_HOLDER_ONE_OF = -2
+internal const val ID_HOLDER_UNKNOWN_FIELDS = -3
 
+internal const val INDEX_NOT_EXISTSED = -1
```
sandwwraith: Typo, should be EXISTED

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -69,21 +75,31 @@ internal open class ProtobufDecoder(
...
+                ID_HOLDER_ONE_OF -> {
+                    descriptor.getElementDescriptor(i)
+                        .getAllOneOfSerializerOfField(serializersModule)
+                        .map { it.extractParameters(0).protoId }
+                        .forEach { map.putProtoId(it, i) }
+                    mapSize ++
+                }
+                ID_HOLDER_UNKNOWN_FIELDS -> {
+                    require(unknownHolderIndex == INDEX_NOT_EXISTSED) {
+                        "Only one unknown fields holder is allowed in a message, but get ${descriptor.getElementName(i)} and ${descriptor.getElementName(unknownHolderIndex)}"
```
sandwwraith:
```suggestion
                        "Only one unknown fields holder is allowed in a message, but two properties have ProtoUnknownFieldHolder type: ${descriptor.getElementName(i)} and ${descriptor.getElementName(unknownHolderIndex)}"
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
```diff
@@ -254,6 +270,9 @@ internal open class ProtobufDecoder(
             deserializer.descriptor == UByteArraySerializer().descriptor -> deserializeByteArray((previousValue as UByteArray?)?.asByteArray()).asUByteArray() as T
             deserializer is AbstractCollectionSerializer<*, *, *> ->
                 (deserializer as AbstractCollectionSerializer<*, T, *>).merge(this, previousValue)
+            deserializer == ProtoUnknownFieldHolderSerializer -> {
```
sandwwraith: You should compare `deserializer.descriptor == ProtoUnknownFieldHolderSerializer.descriptor` — see how it is done with `deserializeByteArray` above

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufEncoding.kt
```diff
@@ -143,11 +143,18 @@ internal open class ProtobufEncoder(
         serializer is MapLikeSerializer<*, *, *, *> -> {
             serializeMap(serializer as SerializationStrategy<T>, value)
         }
+        serializer == ProtoUnknownFieldHolderSerializer -> {
```
sandwwraith: Same here

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Reads the complete wire format bytes (tag + value) for the current field
+ * from [compositeDecoder] using its current tag.
+ */
+internal fun readRawFieldBytes(compositeDecoder: CompositeDecoder): ByteArray {
+    if (compositeDecoder is ProtobufDecoder) {
+        return readRawFieldBytes(compositeDecoder, compositeDecoder.currentTag)
+    }
+    return ByteArray(0)
```
sandwwraith: I think it's better to fail with ClassCastException to ProtobufDecoder rather than silently return wrong data

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,79 @@
...
+}
+
+internal class UnknownFieldsDescriptor(private val original: SerialDescriptor) : SerialDescriptor by original {
```
sandwwraith: I don't see the point of making a wrapper class if the only instance will be in `ProtoUnknownFieldHolderSerializer`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,79 @@
...
+package kotlinx.serialization.protobuf.internal
+
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.protobuf.*
+
+internal object ProtoUnknownFieldHolderSerializer : KSerializer<ProtoUnknownFieldHolder> {
+    override val descriptor: SerialDescriptor
```
sandwwraith: You don't really need to create new descriptor every time via `get()`. You can use a specialized [delegate descriptor constructor](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.descriptors/-serial-descriptor.html): `override val descriptor: SerialDescriptor = SerialDescriptor("kotlinx.serialization.protobuf.ProtoUnknownFieldHolder", ByteArraySerializer().descriptor)`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoUnknownFieldHolder.kt
```diff
@@ -0,0 +1,56 @@
...
+
+package kotlinx.serialization.protobuf
+
+import kotlinx.serialization.*
+import kotlinx.serialization.protobuf.internal.*
+
+/**
+ * Represents a protobuf message.
+ *
+ * Especially used as a holder of unknown proto fields in an arbitrary protobuf message.
```
sandwwraith: Since representing unknown fields is its only usage, we should rephrase this. Maybe

`Represents a thin wrapper over fields in a protobuf message. Used to store unknown proto fields.`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoUnknownFieldHolder.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+@Serializable(with = ProtoUnknownFieldHolderSerializer::class)
+public class ProtoUnknownFieldHolder internal constructor(
+    internal val fields: ByteArray
```
sandwwraith: Should we make it public? At least priniting them is probably a popular request. One may also want to try to edit them before writing back

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoUnknownFieldHolder.kt
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public fun merge(other: ProtoUnknownFieldHolder): ProtoUnknownFieldHolder {
+        return ProtoUnknownFieldHolder(fields + other.fields)
+    }
+
+    override fun hashCode(): Int {
+        return fields.contentHashCode()
+    }
+
+    override fun equals(other: Any?): Boolean {
```
sandwwraith: `toString()` is missing

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufUnknownFieldsTest.kt
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+    @Serializable
+    data class FloatData(@ProtoNumber(1) val id: Int, @ProtoNumber(2) val value: Float)
+
+    @Test
+    fun testUnknownFloat() {
+        assertUnknownFieldRoundTrip(FloatData(1, 0.0f), FloatData.serializer())
+        assertUnknownFieldRoundTrip(FloatData(1, -1.0f), FloatData.serializer())
+        assertUnknownFieldRoundTrip(FloatData(1, Float.MIN_VALUE), FloatData.serializer())
```
sandwwraith: This test fails on JS, probably due to a difference in precision. Can you investigate whether it is safe to ignore this failure?

sandwwraith (changes requested):

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufUnknownFieldsTest.kt
sandwwraith: `Expected <FloatData(id=1, value=1.4e-45)>, actual <FloatData(id=1, value=1.401298464324817e-45)>`

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
xiaozhikang0916: Changed, but can you elaborate why this pattern is recommended here?

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufUnknownFieldsTest.kt
xiaozhikang0916: I believe it is save to ignore this failure here. Json and Cbor only test float value in jvm target because of this precision issue.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt
sandwwraith: We do not require any equality contract from serializers, and keep the instantiation strategy an implementation detail. Changing `object` to `class` and vice versa may break the serializer's identity, while `SerialDescriptor.equals` is well-defined.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
xiaozhikang0916: If we check whether a deserializer is `ProtoUnknownFieldHolderSerializer` by comparing its descriptor, a wrapper is needed here to do the comparasion, otherwise, a `ByteArraySerializer` may be considered same type as `ProtoUnknownFieldHolderSerializer` in `deserializer.descriptor == ProtoUnknownFieldHolderSerializer.descriptor`.

sandwwraith: `SerialDescriptor("kotlinx.serialization.protobuf.ProtoUnknownFieldHolder", ByteArraySerializer().descriptor)` provides such wrapper

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoUnknownFieldHolder.kt
xiaozhikang0916: Sure. And I believe a list of fields data is more helpful than concating them into one byte array so that developers can know how many unknown fields is captured in this holder.

> One may also want to try to edit them before writing back

It is quite diffcult to do that in this implementation, because it is hard to know the id of these fields. To store and retrieve the id, we need a holder for arbitrary proto fields, but we just desided to remove `ProtoField` introduced in this pr.

docs/formats.md
```diff
@@ -666,6 +643,54 @@ is also compatible with the `message Data` given above, which means the same inp
 
 But please note that there are no exclusivity checks. This means that if an instance of `Data2` has both (or none) `homeNumber` and `workNumber` as non-null values and is serialized to protobuf, it no longer complies with the original schema. If you send such data to another parser, one of the fields may be omitted, leading to an unknown issue.
 
+### Preserving unknown fields (experimental)
+
+You may keep updating your schema by adding new fields, but you may not want to break compatibility with the old data.
+
+Kotlin Serialization `ProtoBuf` format supports preserving unknown fields, as described in the [Protocol Buffer-Unknown Fields](https://protobuf.dev/programming-guides/proto3/#unknowns).
+
+To keep the unknown fields, add a property in type `ProtoUnknownFieldHolder` with default value `null` or `ProtoUnknownFieldHolder.Empty` to your data class.
```
sandwwraith:
```suggestion
To keep the unknown fields, add a property of type `ProtoUnknownFieldHolder` with default value `null` or `ProtoUnknownFieldHolder.Empty` to your data class.
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/ProtoUnknownFieldHolder.kt
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Used to store unknown proto fields in a concrete class,
+ * but do not deserialize byte array to this type directly.
+ *
+ * @property fields The list of unknown fields held by this holder,
+ * each item presents for one top level field with its tagged raw byte array.
+ */
+@Serializable(with = ProtoUnknownFieldHolderSerializer::class)
+public class ProtoUnknownFieldHolder internal constructor(
+    public val fields: List<ByteArray>
```
sandwwraith: I'm still not convinced we should provide any API beyond simple `public val fields: ByteArray`. Can you provide an example where the number of such fields matters, and where having a `List` would make a big difference? IMO, it is enough to know just size and raw content for some toString/debuggability, and you can't really do anything beyond that.

If we are going to add `ProtoUnknownField` in the future, it will also be easier to hide plain byte array rather than `fields+contentSize+size` altogether.

xiaozhikang0916: Got it. Any further API for the internal fields will be suspended until a real `ProtoField` is introduced.

docs/formats.md
```diff
@@ -666,6 +643,54 @@ is also compatible with the `message Data` given above, which means the same inp
 
 But please note that there are no exclusivity checks. This means that if an instance of `Data2` has both (or none) `homeNumber` and `workNumber` as non-null values and is serialized to protobuf, it no longer complies with the original schema. If you send such data to another parser, one of the fields may be omitted, leading to an unknown issue.
 
+### Preserving unknown fields (experimental)
+
+You may keep updating your schema by adding new fields, but you may not want to break compatibility with the old data.
+
+Kotlin Serialization `ProtoBuf` format supports preserving unknown fields, as described in the [Protocol Buffer-Unknown Fields](https://protobuf.dev/programming-guides/proto3/#unknowns).
```
sandwwraith:
```suggestion
Kotlin Serialization `ProtoBuf` format supports preserving unknown fields, as described in the [Protocol Buffer - Unknown Fields](https://protobuf.dev/programming-guides/proto3/#unknowns).
```

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -133,9 +147,15 @@ internal fun extractProtoId(descriptor: SerialDescriptor, index: Int, zeroBasedD
             }
         }
     }
+    if (descriptor.getElementDescriptor(index).isUnknownFieldsDescriptor) {
```
sandwwraith, line 151: Minor: KDoc for this function has to be updated as well.

Also, it looks suspiciously similar to `fun SerialDescriptor.extractParameters`. Later investigation is needed on whether we can unify them (out of scope for this PR)

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,70 @@
...
+        ProtoWireType.INVALID -> {}
+    }
+    return output.toByteArray()
+}
+
+/**
+ * Reads the complete wire format bytes (tag + value) for the current field
+ * from [compositeDecoder] using its current tag.
+ */
+internal fun readRawFieldBytes(compositeDecoder: CompositeDecoder): ByteArray {
```
sandwwraith: You can have `compositeDecoder: ProtobufDecoder` here because of smartcast. You don't need a second `is`.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.serialization.protobuf.*
+
+internal object ProtoUnknownFieldHolderSerializer : KSerializer<ProtoUnknownFieldHolder> {
+    override val descriptor: SerialDescriptor = SerialDescriptor(
+        "kotlinx.serialization.protobuf.ProtoUnknownFieldHolder",
+        ByteArraySerializer().descriptor,
+    )
+
+    override fun deserialize(decoder: Decoder): ProtoUnknownFieldHolder {
+        if (decoder is ProtobufDecoder) {
```
sandwwraith: Follow the pattern with `.asJsonEncoder/Decoder` (https://github.com/Kotlin/kotlinx.serialization/blob/304e858ccc7066854637d86ab80056f5f2bcc094/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt#L203) — perform check here, throw IllegalStateException on failure instead of CCE.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,70 @@
...
+            return decoder.decodeStructure(descriptor) {
+                ProtoUnknownFieldHolder(readRawFieldBytes(this))
+            }
+        }
+        return ProtoUnknownFieldHolder.Empty
+    }
+
+    override fun serialize(encoder: Encoder, value: ProtoUnknownFieldHolder) {
+        if (encoder is ProtobufEncoder) {
```
sandwwraith: Same here — ISE on failed cast 

Alternatively, you can just call `encodeSerializableValue` with ByteArraySerializer if you need Json support for ProtoUnknownFieldHolder. But I'm not sure it is needed.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
```diff
@@ -0,0 +1,70 @@
...
+            encoder.writeRawBytes(value.fields)
+        }
+    }
+}
+
+/**
+ * Reads the complete wire format bytes (tag + value) for the current field
+ * from [decoder]. The [currentTag] provides the proto id for the unknown field.
+ */
+internal fun readRawFieldBytes(decoder: ProtobufDecoder, currentTag: ProtoDesc): ByteArray {
```
sandwwraith, line 54: This whole method can be moved inside `decodeRawElement(protoId: Int)` since `decoder` already knows `currentType`

sandwwraith (changes requested):

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtoUnknownFieldHolderSerializer.kt
xiaozhikang0916: `this` in the lambda of `Decoder.decodeStructure` has type in `CompositeDecoder`, which is not smartcasted to `ProtobufDecoder`. Do you suggest accessing `decoder` outside of the lambda scope here?

sandwwraith: Hm, indeed. You can access `currentTag` anyway, so you can drop `internal fun readRawFieldBytes(compositeDecoder: CompositeDecoder)` completely, leaving only `readRawFieldBytes(decoder: ProtobufDecoder, currentTag: ProtoDesc)`.

Or you can edit ProtobufDecoder to narrow the return type on overload: `override fun beginStructure(descriptor: SerialDescriptor): ProtobufDecoder`. You'll have to inline decodeStructure lambda though.

xiaozhikang0916: Yeah, `internal fun readRawFieldBytes(compositeDecoder: CompositeDecoder)` and some unnecessary type cast have been removed in my last commit.

sandwwraith (approved):
Thanks a lot for this huge missing piece in protobuf support!

xiaozhikang0916:
What should I do to progress the related PR in Kotlin https://github.com/JetBrains/kotlin/pull/5755?

sandwwraith:
@shanshin will take it from there.

PR #2865, opened 2024-11-21
Fixed keeping INSTANCE field and serializer function for serializable objects in R8 full mode

### LINKED ISSUE #2861: MinifyEnabled = true strips away Serializable objects
I am attaching a working project in which the issue can be observed. 

As per the documentation of this library , we are not to change add or modify the proguard files (if we are not using named companion objects).

That hasn't been my observation . 

I am attaching a code sample to demonstrate the issue i obtain with minifyEnabled true. The project zip file is uploaded [here](https://drive.google.com/file/d/1PrT8kHc-qmmZlXo-ByZaaGBCjAdfl2Ej/view?usp=sharing). 

Just open this project in android studio and run the android app on a device or simulator. The app would crash with the following stacktrace . 

image

Adding `-keep @kotlinx.serialization.Serializable class * {*;}`  to the proguard file would fix this situation , but this is against what the docs say

shanshin:
Details https://issuetracker.google.com/issues/379996140

Fixes #2861

### DIFF
--- a/rules/r8.pro
+++ b/rules/r8.pro
@@ -21,11 +21,5 @@
 
 -keepclassmembers @kotlinx.serialization.Serializable class ** {
     public static ** INSTANCE;
-}
-
--if @kotlinx.serialization.Serializable class ** {
-    public static ** INSTANCE;
-}
--keepclassmembers class <1> {
     kotlinx.serialization.KSerializer serializer(...);
 }
--- a/rules/r8.pro
+++ b/rules/r8.pro
@@ -10,3 +10,22 @@
 
  -if @kotlinx.serialization.Serializable class **
  -keep, allowshrinking, allowoptimization, allowobfuscation, allowaccessmodification class <1>
+
+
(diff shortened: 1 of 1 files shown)


sandwwraith:
Why it is not 

```
-if @kotlinx.serialization.Serializable class **
-keepclassmembers class <1> {
    public static <1> INSTANCE;
    kotlinx.serialization.KSerializer serializer(...);
}
```

as you advised in the original issue?

shanshin:
> as you advised in the original issue?

current rule more strict, `serializer` function will be kept only if INSTANCE field is present.
If a less strict rule is also acceptable, then it is better to use
```
-keepclassmembers @kotlinx.serialization.Serializable class ** {
    public static ** INSTANCE;
    kotlinx.serialization.KSerializer serializer(...);
}
```

shanshin:
@sandwwraith, I simplify rule, and made it less strict. Now all `INSTANCE` fields and `serializer` function will be kept in any class marked as `Serializable`

sandwwraith (approved):

PR #2871, opened 2024-12-04
README: bump version to 2.1.0

martinbonnin:


sandwwraith:
Thanks for you PR! We usually specify in the README version that was used in the latest kotlinx.serialization release, which is still 2.0.20 for now. It is updated together with next library release.

martinbonnin:
> We usually specify in the README version that was used in the latest kotlinx.serialization release, which is still 2.0.20 for now

I get it for the badge (sorry bad replace all). But usage, etc.. could encourage using Kotlin using 2.1.0? The serialization compiler plugin works fine with 2.1.0?

sandwwraith:
@martinbonnin Yes, it works fine with 2.1.0. The point is, README reflects the state of the library — if there are no changes in that regard, README is not updated.

PR #2873, opened 2024-12-05
CBOR correctly skips particular values

ardune:
The CBOR "ignoreUnknownKeys" allows for arbitrary content to be skipped.
In the case of an empty map, empty array, or tags on non-map/array values the skip would fail

Included somewhat trimmed down example test.

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
@@ -426,9 +426,11 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
             } else {
                 val header = curByte and 0b111_00000
                 val length = elementLength()
-                if (header == HEADER_ARRAY || header == HEADER_MAP) {
+                if (header == HEADER_TAG) {
+                    readNumber()
+                } else if (header == HEADER_ARRAY || header == HEADER_MAP) {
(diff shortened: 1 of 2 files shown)


sandwwraith:
Have you filed an issue somewhere already or you've decided to do a PR right away?

ardune:
I just made a PR - Apologies if I skipped a step. I would be happy to open an issue or otherwise add to the PR.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
```diff
@@ -426,9 +426,11 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
             } else {
                 val header = curByte and 0b111_00000
                 val length = elementLength()
-                if (header == HEADER_ARRAY || header == HEADER_MAP) {
+                if (header == HEADER_TAG) {
+                    readNumber()
+                } else if (header == HEADER_ARRAY || header == HEADER_MAP) {
```
sandwwraith, line 431: cc @JesusMcCloud as author of these changes. I do not see any particular problems, so I'll merge this if you don't have other comments.

JesusMcCloud: Thanks! Looks good to me!

sandwwraith (approved):
Thank you!

PR #2874, opened 2024-12-05
Introduce @JsonIgnoreUnknownKeys annotation

### LINKED ISSUE #1420: Ignore unknown keys per class / via annotation

### LINKED ISSUE #2637: Improve failure message on encountering unknown key in JSON
Right now, the failure is the following:
```
Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 20: Encountered an unknown key 'version' at path: $.name
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: {"name":"unknown", "version": 2.0}
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
```

Maybe it is worth making it more explicit and straightforward, for example:
```
Exception in thread "main" kotlinx.serialization.json.internal.JsonUnknownKeyException: Encountered an unknown key 'version' at path: $.name
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: {"name":"unknown", "version": 2.0}
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
```

Note that the actual exception can still be internal

### LINKED ISSUE #2869: Wrong path reported by json parser for unknown key
**Describe the bug**
When parsing json with unknown key, the path in the error message will only be correct if the unknown key is the first key in the object. If it is not the first key, the reported path will include the previous key at the same level as the unknown key.

**To Reproduce**
```
package ...

import kotlinx.serialization.SerializationException
import org.junit.jupiter.api.Assertions.assertEquals
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.assertThrows

class UnknownKeyParsingTest {
    @Serializable
    data class MyClass(val field: String)

    @Test
    fun testUnknownKey() {
        assertThrows<SerializationException> {
            Json.decodeFromString<MyClass>("""{ "unknownField": 123 }""")
        }.also {
            assertEquals(
                """
                    Unexpected JSON token at offset 3: Encountered an unknown key 'unknownField' at path: $
                    Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
                    JSON input: { "unknownField": 123 }
                """.trimIndent(),
                it.message
            )
        }
    }

    @Test
    fun testUnknownKeyAsSecondKey() {
        assertThrows<SerializationException> {
            Json.decodeFromString<MyClass>("""{ "field": "value", "unknownField": 123 }""")
        }.also {
            assertEquals(
                """
                    Unexpected JSON token at offset 21: Encountered an unknown key 'unknownField' at path: $.field
                    Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
                    JSON input: { "field": "value", "unknownField": 123 }
                """.trimIndent(),
                it.message
            )
        }
    }
}
```
**Expected behavior**
The expected behaviour is that the reported path is for the object that contains the unknown key, not for a sibling key.

**Environment**
 - Kotlin version: 2.0.21
 - Library version: 1.7.3
 - Kotlin platforms: JVM

sandwwraith:
for more fine-grained control over JsonBuilder.ignoreUnknownKeys setting.

Fixes #1420

Also improve error message and path handling when 'Unknown key' exception is thrown

Fixes #2869
Fixes #2637

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -13,6 +13,7 @@ In this chapter, we'll walk through features of [JSON](https://www.json.org/json
   * [Pretty printing](#pretty-printing)
   * [Lenient parsing](#lenient-parsing)
   * [Ignoring unknown keys](#ignoring-unknown-keys)
+  * [Ignoring unknown keys per class](#ignoring-unknown-keys-per-class)
   * [Alternative Json names](#alternative-json-names)
   * [Encoding defaults](#encoding-defaults)
   * [Explicit nulls](#explicit-nulls)
@@ -164,6 +165,44 @@ Project(name=kotlinx.serialization)
 
 <!--- TEST -->
 
+### Ignoring unknown keys per class
+
+Sometimes, for cleaner and safer API, it is desirable to ignore unknown properties only for specific classes.
+In that case, you can use [JsonIgnoreUnknownKeys] annotation on such classes while leaving global [ignoreUnknownKeys][JsonBuilder.ignoreUnknownKeys] setting
+turned off:
+
+```kotlin
+@Serializable
+data class Outer(val a: Int, val inner: Inner)
+
+@OptIn(ExperimentalSerializationApi::class) // JsonIgnoreUnknownKeys is an experimental annotation for now
+@Serializable
+@JsonIgnoreUnknownKeys
+data class Inner(val x: String)
+
+fun main() {
+    // 1
+    println(Json.decodeFromString<Outer>("""{"a":1,"inner":{"x":"value","unknownKey":"unknownValue"}}"""))
+    println()
+    // 2
+    println(Json.decodeFromString<Outer>("""{"a":1,"inner":{"x":"value"}, "b":2}"""))
+}
+```
+
+> You can get the full code [here](../guide/example/example-json-04.kt).
+
+Line (1) decodes successfully despite "unknownKey" in `Inner`, because annotation is present on the class. 
+However, line (2) throws `SerializationException` because there is no "b" property in `Outer`:
+
+```text
+Outer(a=1, inner=Inner(x=value))
+
+Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Encountered an unknown key 'b' at offset 31 at path: $
+Use 'ignoreUnknownKeys = true' in 'Json {}' builder or '@JsonIgnoreUnknownKeys' annotation to ignore unknown keys.
+```
+
+<!--- TEST LINES_START-->
+
 ### Alternative Json names
 
 It's not a rare case when JSON fields are renamed due to a schema version change.
@@ -184,7 +223,7 @@ fun main() {
 }
 ```
 
-> You can get the full code [here](../guide/example/example-json-04.kt).
+> You can get the full code [here](../guide/example/example-json-05.kt).
 
 As you can see, both `name` and `title` Json fields correspond to `name` property:
 
@@ -222,7 +261,7 @@ fun main() {
 }
 ```
 
-> You can get the full code [here](../guide/example/example-json-05.kt).
+> You can get the full code [here](../guide/example/example-json-06.kt).
 
 It produces the following output which encodes all the property values including the default ones:
 
@@ -261,7 +300,7 @@ fun main() {
 }
 ```
 
-> You can get the full code [here](../guide/example/example-json-06.kt).
+> You can get the full code [here](../guide/example/example-json-07.kt).
 
 As you can see, `version`, `website` and `description` fields are not present in output JSON on the first line.
 After decoding, the missing nullable property `website` without a default values has received a `null` value,
@@ -319,7 +358,7 @@ fun main() {
 }
 ```
 
-> You can get the full code [here](../guide/example/example-json-07.kt).
+> You can get the full code [here](../guide/example/example-json-08.kt).
 
 The invalid `null` value for the `language` property was coerced into the default value:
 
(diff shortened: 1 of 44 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/JsonAnnotations.kt
```diff
@@ -74,3 +74,36 @@ public annotation class JsonNames(vararg val names: String)
 @Target(AnnotationTarget.CLASS)
 @ExperimentalSerializationApi
 public annotation class JsonClassDiscriminator(val discriminator: String)
+
+
+/**
+ * Specifies whether encounters of unknown properties (i.e., properties not declared in the class) in the input JSON
+ * should be ignored instead of throwing [SerializationException].
+ *
+ * With this annotation, it is possible to allow unknown properties for specific classes, while
```
fzhinkin, line 83: Annotation's effect does not propagate to annotated class' fields:

```kotlin
@JsonIgnoreUnknownKeys @Serializable
class Outer(val inner: Inner)

@Serializable
class Inner(val x: Int)

...
// {"untitled":"unknown", "inner": {"x": 42}} // OK
// {"inner": {"x": 42, "untitled": "unknown"}} // FAIL
```

Was it a deliberate choice?
Personally, I think the current behavior is fine, but it might be worth to stress that `for specific classes` means that the annotation has a shallow effect.

guide/example/example-json-05.kt
```diff
@@ -4,16 +4,13 @@ package example.exampleJson05
 import kotlinx.serialization.*
 import kotlinx.serialization.json.*
 
-val format = Json { encodeDefaults = true }
```
fzhinkin, line 7: git's capabilities of tracking renames are quite limited and it seems like it can't handle such a chain of renames.

I see two possible solutions to keep files' git history a bit more clear:
- commit all renames separately (wow, 28 commits 😓 )
- put the new example into `example-json-32.kt` instead of 04.

WDYT?

formats/json/commonMain/src/kotlinx/serialization/json/JsonAnnotations.kt
sandwwraith: Yes, it is a deliberate choice (also we don't have technical means for this). 
I'll change 'specific' to 'annotated'.

guide/example/example-json-05.kt
sandwwraith: I'm afraid Knit functionality as a tool is quite limited in this place — it will automatically re-number all examples in a sequence (https://github.com/Kotlin/kotlinx-knit?tab=readme-ov-file#example-files)

docs/json.md
```diff
@@ -164,6 +165,44 @@ Project(name=kotlinx.serialization)
...
+    println(Json.decodeFromString<Outer>("""{"a":1,"inner":{"x":"value","unknownKey":"unknownValue"}}"""))
+    println()
+    // 2
+    println(Json.decodeFromString<Outer>("""{"a":1,"inner":{"x":"value"}, "b":2}"""))
+}
+```
+
+> You can get the full code [here](../guide/example/example-json-04.kt).
+
+Line (1) decodes successfully despite "unknownKey" in `Inner`, because annotation is present on the class.
```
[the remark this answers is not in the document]
sandwwraith: To illustrate 'shallow' effect? I think I can just move the annotation

formats/json/commonMain/src/kotlinx/serialization/json/JsonAnnotations.kt
```diff
@@ -74,3 +74,36 @@ public annotation class JsonNames(vararg val names: String)
...
 @ExperimentalSerializationApi
 public annotation class JsonClassDiscriminator(val discriminator: String)
+
+
+/**
+ * Specifies whether encounters of unknown properties (i.e., properties not declared in the class) in the input JSON
+ * should be ignored instead of throwing [SerializationException].
+ *
+ * With this annotation, it is possible to allow unknown properties for annotated classes, while
+ * general [Json.decodeFromString] would still reject them for everything else.
```
shanshin, line 84: Perhaps I'm exaggerating, but won't such a description give the user the impression that this annotation only works with `decodeFromString`, but not with `decodeFromJsonElement` or `decodeFromStream`?

shanshin (approved):

docs/json.md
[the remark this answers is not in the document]
fzhinkin: Yeap, I was talking about the 'shallow' effect.

fzhinkin (approved):

formats/json/commonMain/src/kotlinx/serialization/json/JsonAnnotations.kt
sandwwraith: Not sure about that. Maybe rephrase as `general decoding methods, such as [Json.decodeFromString] and others`?

PR #2877, opened 2024-12-10
Update to Kotlin 2.1.0 and some housekeeping:

sandwwraith:
- Update obsolete bootstrap building instructions
- Update Gradle in integration-test


integration-test/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
```
shanshin, line 3: We can use Gradle 8.10 according versions mapping https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin

shanshin (approved):

integration-test/gradle/wrapper/gradle-wrapper.properties
sandwwraith: I choose 8.7 because it is the same version used in the 'big' kotlinx.serialization project

Issue #2332, opened 2023-06-13
Poor performance in Protobuf map deserialization


ricmf:
I'm running some benchmarks serializing objects to a RocksDB and deserializing them back. Deserialization is taking much longer than serialization. I profiled the program with jvisualvm and it seems the majority of CPU is spent on hashcode calls. On line 216 the lib deserializes the map entries into a linkedhashset, which for some reason takes the most CPU time in hashcode calls. Could this be replaced by a List deserializer since on the next line it everything is put into a map? If there are repeated MapEntry elements then their keys must be equal so associateBy would do the same.

![profile](https://github.com/Kotlin/kotlinx.serialization/assets/7907754/914865ed-ca6e-40c2-ad91-0b8c3575f7a3)

https://github.com/Kotlin/kotlinx.serialization/blob/84d4af04a688286ad1b51a72130ccda691caff75/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/ProtobufDecoding.kt#L216

PR #2879, opened 2024-12-11
Speedup ProtoWriteType.from

fzhinkin:
While poking into https://github.com/Kotlin/kotlinx.serialization/issues/2332 I spotted that `ProtoWriteType.from` implementation is pretty inefficient.

There are a few ways to transform integer into a corresponding `ProtoWriteType`.
Corresponding benchmarks are here: https://github.com/fzhinkin/kx-serialization-parse-proto-type/blob/main/src/commonMain/kotlin/ParseProtoTypeBenchmark.kt

On JVM (see https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fzhinkin/c8b3335e6ba79ec06e9acffc85230e28/raw/e420db8bbb80cf03f772862eb6a58cd7509cf617/parsing-jvm.json), both 8-aligned array-based approaches works fine.
On Native (see https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fzhinkin/c8b3335e6ba79ec06e9acffc85230e28/raw/e420db8bbb80cf03f772862eb6a58cd7509cf617/parsing-macos-arm64.json), moving masking into a function works a bit faster.

Here's comparison of results proto-related benchmarks collected with existing implementation (baseline) and the implementation from this PR (optimized): https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fzhinkin/c8b3335e6ba79ec06e9acffc85230e28/raw/e420db8bbb80cf03f772862eb6a58cd7509cf617/baseline.json,https://gist.githubusercontent.com/fzhinkin/c8b3335e6ba79ec06e9acffc85230e28/raw/e420db8bbb80cf03f772862eb6a58cd7509cf617/optimized.json

Suspiciously looking `ProtoBaseline.toBytesExplicit` shows bimodal results, thus the results.
For parsing related benchmarks, the proposed change slightly improves performance.

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
@@ -26,11 +26,13 @@ internal enum class ProtoWireType(val typeId: Int) {
             ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
         }
 
-        fun from(typeId: Int): ProtoWireType {
-            if (typeId < 0 || typeId >= entryArray.size) {
-                return INVALID
-            }
-            return entryArray[typeId]
+        /**
+         * Extracts three least significant bits from the [value] and
+         * returns [ProtoWireType] with corresponding type id, or [ProtoWireType.INVALID]
+         * if there are no such a type.
+         */
+        fun fromLeastSignificantBits(value: Int): ProtoWireType {
+            return entryArray[value and 0b111]
         }
     }
 
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
@@ -22,8 +22,15 @@ internal enum class ProtoWireType(val typeId: Int) {
     ;
 
     companion object {
+        private val entryArray = Array(8) { typeId ->
+            ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
+        }
+
         fun from(typeId: Int): ProtoWireType {
-            return ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
+            if (typeId < 0 || typeId >= entryArray.size) {
+                return INVALID
+            }
+            return entryArray[typeId]
         }
     }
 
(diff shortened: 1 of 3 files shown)


fzhinkin:
Additionally, checked how replacing the enum with raw int values will affect the performance - the difference is negligibly small and does not justify such a change.

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -22,8 +22,15 @@ internal enum class ProtoWireType(val typeId: Int) {
     ;
 
     companion object {
+        private val entryArray = Array(8) { typeId ->
```
fzhinkin, line 25: My bad, the size is chosen based on the fact the only call site explicitly zero all the bits except three least significant.

Perhaps, as there's only one call site, explicitly move this operation inside `from` and rename `from` to `fromLowerBits`. WDYT?

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -22,8 +22,15 @@ internal enum class ProtoWireType(val typeId: Int) {
...
+            ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
+        }
+
         fun from(typeId: Int): ProtoWireType {
-            return ProtoWireType.entries.find { it.typeId == typeId } ?: INVALID
+            if (typeId < 0 || typeId >= entryArray.size) {
+                return INVALID
+            }
```
pdvrieze, line 32: From an optimization perspective it may be valid to not check for `typeId<0` (and maybe have a larger array of `INVALID`s) to avoid the if check (instead relying on the array check that does this too).

fzhinkin: The array bounds check will "give" us `ArrayIndexOutOfBounds`, not the `ProtoWireType.INVALID` :)

Anyway, the implementation was implicitly based on the idea that the only call site extracts 3 least-significant bits from an int before passing it down this methods, so when this method is inlined the compiler can get rid of the `if`.

sandwwraith (approved):
Great find!

sandwwraith:
Although this change was added in #2768 in 2024, before that, there were direct Int constants. So it is not the possible cause of slowness reported in #2332

fzhinkin:
> So it is not the possible cause of slowness reported in https://github.com/Kotlin/kotlinx.serialization/issues/2332

It's definitely not, but it improves a performance a bit without large-scale updates. You know, ten bucks is ten bucks. ;)

PR #2904, opened 2025-01-20
Drop 'withJava()' calls in build configuration

Tapchicoma:


settings.gradle.kts
```diff
@@ -63,6 +63,8 @@ project(":kotlinx-serialization-json-tests").projectDir = file("./formats/json-t
 
 include(":kotlinx-serialization-protobuf")
 project(":kotlinx-serialization-protobuf").projectDir = file("./formats/protobuf")
+include(":kotlinx-serialization-protobuf:test-protobuf")
```
sandwwraith, line 66: Given that it is a subproject of `serialization-protobuf`, name `test-protobuf` seems like a repetition. `test-model` or `test-data` seem to fit here more

sandwwraith (changes requested):

settings.gradle.kts
shanshin: I don't quite agree, because in this case the project name will be `test-model`, what will be perceived ambiguously.
Very often, project names are used (in reports, logs, etc), rather than their full paths. 

But it seems to me that it is very insignificant.

Tapchicoma: How about `test-protobuf-model`?

sandwwraith: `proto-test-model` then

integration-test/build.gradle.kts
```diff
@@ -46,9 +46,7 @@ kotlin {
     wasmWasi {
         nodejs()
     }
-    jvm {
-        withJava()
-    }
+    jvm()
```
sandwwraith, line 49: integration-test uses `withJava` to verify that Kapt works correctly with kotlinx-serialization. Can you recommend how to check it instead?

settings.gradle.kts
Tapchicoma: Updated project name

integration-test/build.gradle.kts
Tapchicoma: Restored this call to `withJava()` as without it Kapt stops working until Kotlin 2.1.20-Beta2 release. And this brings up the non-smooth user migration path - so I decided to lift error to warning for `withJava()` method usage.

build.gradle.kts
```diff
@@ -166,7 +166,12 @@ gradle.taskGraph.whenReady {
 
 // == projects lists and flags ==
 // getters are required because of variable lazy initialization in Gradle
-val unpublishedProjects get() = setOf("benchmark", "guide", "kotlinx-serialization-json-tests")
+val unpublishedProjects get() = setOf(
+    "benchmark",
+    "guide",
+    "kotlinx-serialization-json-tests",
+    "test-protobuf",
```
shanshin, line 173: `proto-test-model` should be mentioned here

Tapchicoma: Fixed

sandwwraith (approved):

Tapchicoma:
@tbogdanova could you also pick these changes in the `kotlin-community/dev` branch?

Tapchicoma:
Force pushed as GitHub does not support `--autosquash` feature on "Rebase and merge" :cry:

Tapchicoma:
> There is 'squash and merge' button tho?

Space when merging branch containing `fixup!` commit adds `--autosquash` option which merge `fixup!` commits into the relevant commits. Github does not do it - see [this discussion](https://github.com/orgs/community/discussions/40804).

PR #2912, opened 2025-01-24
Remove java plugin usage

Tapchicoma:


buildSrc/src/main/kotlin/source-sets-conventions.gradle.kts
```diff
@@ -28,6 +28,7 @@ kotlin {
     explicitApi()
 
     jvm {
+        withJava()
```
shanshin, line 31: Are we returning `withJava()`?

Tapchicoma: Yes, it will be a warning usage of this

shanshin (approved):

PR #2925, opened 2025-02-10
Remove deprecated APIs usage and enable warning-as-errors for build scripts

Tapchicoma:


buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
```diff
@@ -45,6 +45,7 @@ kotlin {
         watchosDeviceArm64()
 
         // Deprecated, but not removed
+        @Suppress("DEPRECATION")
```
Tapchicoma, line 48: I am not sure this target support is required and maybe we just could remove it?

sandwwraith: Yes, it is required for one of our customers

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -120,28 +128,28 @@ object Java9Modularity {
...
+                it.scriptSources
+            })
             source(sourceFile)
             destinationDirectory.set(temporaryDir)
-            multiPlatformEnabled.set(compileTask.multiPlatformEnabled)
+            multiPlatformEnabled.set(compileTask.get().multiPlatformEnabled)
             compilerOptions {
                 jvmTarget.set(JvmTarget.JVM_9)
-                // To support LV override when set in aggregate builds
-                languageVersion.set(compileTask.compilerOptions.languageVersion)
```
sandwwraith, line 140: Why did you remove LV + optIns copying? IIRC lack of proper LV caused some problems in aggregate/user projects builds.

sandwwraith:
Also, PTAL on Kotlin master build failure. It seems we can't turn -Werror on builds while `withJava` is deprecated

buildSrc/src/main/kotlin/Java9Modularity.kt
Tapchicoma: On line 139 in the new code the whole `compilerOptions` object from relevant compile task are used as a base for this task. This should also include proper LV + optIns.

sandwwraith:
You have to remove `-Werror` for now, wait for serialization to update to 2.1.20, remove `withJava`, and then this PR can be merged

Tapchicoma:
I have to rebase to fix newly introduced warnings in build scripts (now first commit) plus suppress deprecation warning for `withJava()` DSL introduced in Kotlin 2.1.20. This suppression should be removed once project will be updated to Kotlin 2.1.20 by default.

sandwwraith (approved):

Issue #2938, opened 2025-02-27
Sealed value class is serialized/deserialized incorrectly with default json polymorphism style


Stream29:
**Describe the bug**
Value class that contains a primitive value cannot be serialized polymorphically with 'type' parameter. 
The kotlinx.serialization-json library does not mention it and produces incorrect behaviour.

**To Reproduce**

```
@Serializable
sealed interface SealedInterface

@Serializable
@JvmInline
value class SealedValue(val value: Int) : SealedInterface

@Serializable
enum class SealedEnum : SealedInterface {
    A, B
}
```

```
        val json = Json {
            prettyPrint = true
            isLenient = true
            ignoreUnknownKeys = true
        }
        val valueClassInstance = SealedValue(1)
        val valueJson = json.encodeToString<SealedInterface>(valueClassInstance)
        println(valueJson)
        val deserializedValue = json.decodeFromString<SealedInterface>(valueJson)
```

This code outputs:
```
1
```

and then throw exception:
```
kotlinx.serialization.json.internal.JsonDecodingException: Expected JsonObject, but had JsonLiteral as the serialized body of SealedInterface at element: $
JSON input: 1
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:400)
	at kotlinx.serialization.json.Json.decodeFromString(Json.kt:149)
```

**Expected behavior**

When facing the sealed value class that contains primitive type, a exception should be thrown directly, like the exception on sealed enum:
```
java.lang.IllegalStateException: Enums cannot be serialized polymorphically with 'type' parameter. You can use 'JsonBuilder.useArrayPolymorphism' instead
	at kotlinx.serialization.json.internal.PolymorphicKt.checkKind(Polymorphic.kt:71)
	at kotlinx.serialization.json.internal.StreamingJsonEncoder.encodeSerializableValue(StreamingJsonEncoder.kt:253)
	at kotlinx.serialization.json.internal.JsonStreamsKt.encodeByWriter(JsonStreams.kt:99)
	at kotlinx.serialization.json.Json.encodeToString(Json.kt:125)
```

**Environment**
 - Kotlin version: [2.1.0]
[...]

PR #2940, opened 2025-02-27
Fix issue #2938. Add check for sealed value class when serializing.

Stream29:
Add check for invalid type when serializing.
Add unit test for that.


sandwwraith:
Thanks for you PR! Unfortunately we won't merge it, see my reply in the issue: https://github.com/Kotlin/kotlinx.serialization/issues/2938#issuecomment-2722151926

PR #2942, opened 2025-03-03
Supply proper equals(), hashCode(), and toString() for SerialDescriptor() wrapper

### LINKED ISSUE #2941: WrappedSerialDescriptor.equals doesn't work
**Describe the bug**

Hi, I hit this while creating a generic custom serializer with our format-specific type support. My descriptor is like this

```
override val descriptor: SerialDescriptor = SerialDescriptor("android.util.SparseArray", surrogateSerializer.descriptor)
```

And I can't compare it in `encodeSerializableValue` as the comparison seems to be instance based.

**To Reproduce**

```
val a = SerialDescriptor("android.util.SparseArray", surrogateSerializer.descriptor)
val b = SerialDescriptor("android.util.SparseArray", surrogateSerializer.descriptor)
assertThat(a).isEqualTo(b)
```

**Expected behavior**

The test should pass, but I got:

```
java.lang.AssertionError: expected:<kotlinx.serialization.descriptors.WrappedSerialDescriptor@a47353e> but was:<kotlinx.serialization.descriptors.WrappedSerialDescriptor@e7ebf9f>
at org.junit.Assert.fail(Assert.java:89)
```

**Environment**
 - Kotlin version: 2.0.20
 - Library version: 1.8.0
 - Kotlin platforms: Android
 - Gradle version: 8.9

sandwwraith:
Fixes #2941

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -121,7 +121,25 @@ public fun SerialDescriptor(serialName: String, original: SerialDescriptor): Ser
     return WrappedSerialDescriptor(serialName, original)
 }
 
-internal class WrappedSerialDescriptor(override val serialName: String, original: SerialDescriptor) : SerialDescriptor by original
+internal class WrappedSerialDescriptor(override val serialName: String, private val original: SerialDescriptor) :
+    SerialDescriptor by original {
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is WrappedSerialDescriptor) return false
+
+        return serialName == other.serialName && original == other.original
+    }
+
+    override fun hashCode(): Int {
+        var result = serialName.hashCode()
+        result = 31 * result + original.hashCode()
+        return result
+    }
+
+    override fun toString(): String = toStringImpl()
+
+}
 
 /**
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
```diff
@@ -121,7 +121,25 @@ public fun SerialDescriptor(serialName: String, original: SerialDescriptor): Ser
...
 
-internal class WrappedSerialDescriptor(override val serialName: String, original: SerialDescriptor) : SerialDescriptor by original
+internal class WrappedSerialDescriptor(override val serialName: String, private val original: SerialDescriptor) :
+    SerialDescriptor by original {
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is WrappedSerialDescriptor) return false
+
+        return serialName == other.serialName && original == other.original
```
shanshin, line 131: Is it ok if we check `serialName` inside `original`?

shanshin (approved):

core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
sandwwraith: Interesting point. Ideally, we should not check it. However, given that `equalsImpl` requires lambda to check type parameters, and we do not have any way to get them here, it seems that we can't avoid this check. 

At least, this should solve the case when the descriptor is identical, but object is created twice.

PR #2964, opened 2025-03-31
Update forward compatibility notes

### LINKED ISSUE #2924: Using Kotlinx Serialization 1.8.0 with Kotlin 1.9.23
Hello!

I am trying to use Kotlinx Serialization on a project that uses Kotlin 1.9.23. I've noticed that the most recent version, 1.8.0, seems to have a compile dependency on Kotlin 2.1.0: https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-serialization-core-jvm/1.8.0

Does this mean that I am unable to use it on versions previous to 2.1.0? Any help, advice, or knowledge, would be much appreciated.

Thank you!

EDIT: To clarify, when trying to use it in my 1.9.23 project, I get multiple build errors as the 2.1.0 library is clashing with everything else.

sandwwraith:
Fixes #2924

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -88,5 +88,10 @@ This would be indicated by a compiler error specific to a particular feature.
 * New core library versions may or may not require Kotlin compiler plugin update,
 depending on a particular release.
 We'll try to avoid these situations; however, in case of some unexpected issues, it may be necessary.
-So it is possible to have a situation where upgrading serialization runtime from `1.x` to `1.y` requires an update of Kotlin version from `1.4.0` to `1.4.x`.
+So it is possible to have a situation where upgrading serialization runtime from `1.x` to `1.y` requires an update of Kotlin version from `1.4.0` to `1.x.y`.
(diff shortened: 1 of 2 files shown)


docs/compatibility.md
```diff
@@ -88,5 +88,10 @@ This would be indicated by a compiler error specific to a particular feature.
 * New core library versions may or may not require Kotlin compiler plugin update,
 depending on a particular release.
 We'll try to avoid these situations; however, in case of some unexpected issues, it may be necessary.
-So it is possible to have a situation where upgrading serialization runtime from `1.x` to `1.y` requires an update of Kotlin version from `1.4.0` to `1.4.x`.
+So it is possible to have a situation where upgrading serialization runtime from `1.x` to `1.y` requires an update of Kotlin version from `1.4.0` to `1.x.y`.
```
shanshin, line 91: Can I use other variables? 
With this writing, it seems like you need to use compiler `1.7.8` to switch from `1.7` to `1.8`.

shanshin (approved):

PR #2978, opened 2025-04-10
update: first update for serialization documentation restructuring

daniCsorbaJB:
This PR is for the first batch of documentation updates for [KT-66330](https://youtrack.jetbrains.com/issue/KT-66330/Docs-Integrate-Serialization-docs-from-GitHub-to-kotlinlang.org)

Additional PRs will be opened on this one.


docs/images/serialization.svg
pdvrieze, line 1: Calling the middle step Primitives is misleading (as this is not Kotlin/Java primitive types). Instead the middle steps are actually serialization library calls (with a given set, including primitive types, but also supporting structures and "special" types).

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+network or stored in a database or a file. Deserialization is the opposite process of converting external data back into a runtime object.
+Together, they are essential to most applications that exchange data with third parties.
+
+Some data serialization formats, such as [JSON](https://www.json.org/json-en.html) and [Protocol Buffers](https://protobuf.dev/) are particularly common.
+Being language-neutral and platform-neutral, these formats enable data exchange between systems written in any modern language.
+
+To convert an object tree to a string or to a sequence of bytes, it must go through two mutually intertwined processes: 
+
+1. Serialization: Objects are transformed into a sequence of their primitive values.
+This universal process varies depending on the object and is managed by a serializer.
```
pdvrieze, line 13: This is misleading. Serialization translates objects to a sequence of encoder calls. There is no intermediate format, and such a format would not actually work (many formats allow for "special casing").

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+2. Encoding: The primitive sequence is converted into the desired output format, controlled by an encoder.
+
+![Serialization flow](serialization.svg){width=700}
+
+The reverse process involves parsing the input format, decoding the primitive values, and then deserializing the resulting
+stream into objects.
```
pdvrieze, line 19: The reverse process involves the decoder invoking the input format to guide it in parsing the input (there is no intermediate format and that cannot exist - note that formats may not allow parsing without the knowledge of specific data structures (generally non-self-describing binary formats)).

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you're new to serialization in Kotlin, we recommend starting with the [Get Started with Serialization](serialization-get-started.md) guide.
+This section provides a step-by-step guide to help you set up and use Kotlin serialization in your projects.
+By following these steps, you can quickly get up to speed with the basics before diving into more complex topics.
+
+## Kotlin serialization libraries
+
+The `kotlinx.serialization` library offers support for all platforms, including JVM, JavaScript, Native.
```
pdvrieze, line 27: It may be worthwhile to clarify more that the serialization library provides the format agnostic infrastructure. Then the `-json` and `-cbor` libraries are format implementations (they are fundamentally different and are independent even though first-party).

pdvrieze (commented):
You want to be accurate in how serialization works. The "intermediate data" representation is misleading, an intermediate format is fundamentally impossible (even without supporting format specific escape hatches such as `JsonPolymorphicSerializer`)

docs/images/serialization.svg
daniCsorbaJB: Hey @pdvrieze - thank you for your comments 💯 - we are still in the review process for these updates (just a draft PR at this point), but we will take a look at your feedback during the review

docs/serialization.tree
```diff
@@ -0,0 +1,24 @@
+<?xml version='1.0' encoding='utf-8'?>
+<!DOCTYPE instance-profile SYSTEM "https://resources.jetbrains.com/writerside/1.0/product-profile.dtd">
+<instance-profile id="serialization" name="Serialization" start-page="serialization.md">
+    <snippet id="serialization">
+        <toc-element topic="serialization.md" toc-title="Introduction"/>
+        <toc-element toc-title="Get started">
+            <toc-element topic="serialization-get-started.md"/>
```
sandwwraith: Am I right that these files do not exist for now (in this PR, at least?)

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
+[//]: # (title: Serialization)
```
sandwwraith: I'm finding it a bit complicated when both old and new markdown files are located in the same folder. Let's do the following:

- Leave old .md in `docs` folder. The folder should be as-is because people are using it right now, and unrelated files will confuse them.
- The new files should be in the new folder. Let's name it `docs-website`, for example (or any other name you see fit).
- Changes should be self-contained, with no errors. (Except maybe references to the files that are parts of subsequent PRs, such as `For more information, see [Serialize classes](serialization-customization-options.md).`
- Then we can iterate on each `toc-element` that can be one or a couple of .md files

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+**Serialization** is the process of converting data used by an application to a format that can be transferred over a
+network or stored in a database or a file. Deserialization is the opposite process of converting external data back into a runtime object.
+Together, they are essential to most applications that exchange data with third parties.
+
+Some data serialization formats, such as [JSON](https://www.json.org/json-en.html) and [Protocol Buffers](https://protobuf.dev/) are particularly common.
+Being language-neutral and platform-neutral, these formats enable data exchange between systems written in any modern language.
+
+To convert an object tree to a string or to a sequence of bytes, it must go through two mutually intertwined processes: 
+
+1. Serialization: Objects are transformed into a sequence of their primitive values.
```
sandwwraith: I believe this particular explanation about primitives stream is too low-level for overview page. It confuses and distracts from the main purpose, so I suggest putting it aside until the paragraph about custom serializers.

To be honest, I think it is a bit outdated and does not fit anywhere except plugin's KEEP (https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0149-serialization.md#core-api-overview-and-mental-model)

docs/topics/serialization.md
sandwwraith: Well, if we laid out every call to `encodeInt/encodeString/encode...` with its parameter, we will get a sequence of primitives (in memory, not on a wire)

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+There are also community-maintained libraries that support more serialization formats, such as [YAML](https://yaml.org/) or [Apache Avro](https://avro.apache.org/).
+
+You can find out more about experimental serialization formats in [Alternative and custom serialization formats](alternative-serialization-formats.md).
+
+## Supported serialization types
+
+Kotlin serialization supports a variety of built-in types, including all primitive types and composite types from the Kotlin standard library like the `List` type.
+For more information, see [Serialize built-in types](serialization-serialize-builtin-types.md).
+
+> Not all types from the Kotlin standard library are serializable. In particular, [ranges](ranges.md) and the [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) class are not serializable at the moment.
```
sandwwraith: ranges.md does not exist (even in full version of the PR)

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
+[//]: # (title: Serialization)
+
```
sandwwraith: I do not quite like this page. Correct me if I'm wrong — it is supposed to replace the current one on https://kotlinlang.org/docs/serialization.html , right?

In that case, it feels lacking in comparison to both the current page and GitHub's readme. I find it very valuable that both of these contain setup instructions on the go, so if you are already familiar with the library and need to remind yourself about dependency coordinates, you just need to copy it without clicking around. It also helps people who rush to code and want to experiment with everything themselves.

In other words, this page in PR lacks basic setup instructions and simple examples. I suggest taking them from the current https://kotlinlang.org/docs/serialization.html page. Other setup cases (e.g., Maven) can be put in a separate get-started.md or setup.md

So the structure on this page could look like this:

1. Introduction (Serialization is the process of converting data used by an application to a format that can be transferred over a network or stored in a database or a file......)
2. Kotlin serialization libraries (from this doc)
3. Example: JSON serialization with Gradle setup (from current website)
4. Supported formats
5. What's next and references

docs/topics/serialization.md
```diff
@@ -0,0 +1,78 @@
...
+All libraries except JSON serialization (`kotlinx-serialization-json`) are [experimental](components-stability.md), which means their API can be changed without notice.
+For more details about JSON serialization, see [JSON serialization overview](configure-json-serialization.md).
+
+There are also community-maintained libraries that support more serialization formats, such as [YAML](https://yaml.org/) or [Apache Avro](https://avro.apache.org/).
+
+You can find out more about experimental serialization formats in [Alternative and custom serialization formats](alternative-serialization-formats.md).
+
+## Supported serialization types
+
+Kotlin serialization supports a variety of built-in types, including all primitive types and composite types from the Kotlin standard library like the `List` type.
```
sandwwraith: As far as I see, this and What's next section mainly serve as the index for further reading, just like Table of contents in the current guide. 
I like how you expanded each reference with explanations instead of simply giving the link. However, it feels disconnected now, given that it consists of two paragraphs now. By-passers will likely not understand that they should read this topic to understand what sections of the documentation are there.
I.e. I would expect that `What's next` contains all follow-up links and will probably skip `Supported serialization types` section. I suggest removing this section and merging it with What's next:


```text
## What's next
* Learn the basics of Kotlin serialization in the [Get started with serialization guide](serialization-get-started.md).
* Learn how built-in types, collections, and Serializable classes are processed by kotlinx.serialization (link to builtin types)
* Learn Json customization options [JSON serialization overview](configure-json-serialization.md).
* Learn how to write and customize your serializers (custom serializers link)
* ... etc
* ```

File: docs/topics/serialization.md [hunk 10]
```diff
@@ -0,0 +1,78 @@
...
+
+> Not all types from the Kotlin standard library are serializable. In particular, [ranges](ranges.md) and the [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) class are not serializable at the moment.
+> Support for their serialization may be added in the future.
+>
+{style="note"}
+
+Additionally, classes annotated with `@Serializable` are fully supported for serialization, enabling the conversion of class instances to and from formats like JSON.
+For more information, see [Serialize classes](serialization-customization-options.md).
+
+## What's next
```
sandwwraith: I know that this is common approach in all our docs, but I don't quite like the 'What's next' name of the section. It feels logical when you read page top-to-bottom, but if you're using it just as a reference/starter it feels out of place. I would rename it to something 'What this guide can also offer you' or 'Next sections of the guide' or just 'Table of contents in this guide'. 

But maybe you know better, since this name was picked a long ago.

docs/topics/serialization.md
pdvrieze: The main challenge is that readers may assume there is such intermediate "file"/memory object. Maybe describe it as:
 - The serializer translates the object into a sequence of generic calls to the format(encoder) for the structure and individual elements of the object. It knows about individual objects, but not about formats
 - The encoder then uses this to write the data in a manner appropriate to the format. It takes the generic structure and element information, and does not know about objects.
 - There are special cases where serializers know about formats (by testing the type) that can support unusual cases. (For example JsonTransformingSerializer)
 - Similarly a format may know about certain serializers to handle them in a more appropriate way. E.g. unsigned types, ByteArrays, the default list serializers.

docs/topics/serialization.md
pdvrieze: Maybe an interesting observation to add somewhere is that while encoding can generally simply skip elements either in the serializer or in the format, this is not true for decoders. Decoders must read the byte/text streams and consume them (even if it is just to explicitly ignore the parsing results).

pdvrieze (commented):
I added some further comments. It may be that some of them are better used in other pages (not in this request)

docs/serialization.tree
daniCsorbaJB: you are correct 👍 We discussed we will be using this PR as the basis to merge the other PRs into as we go, so it will eventually contain all of these.

docs/topics/serialization.md
daniCsorbaJB: That's how the linking is set up in our doc-base. If we integrate these serialization docs in kotlinlang.org it will point to this page with this link (https://kotlinlang.org/docs/ranges.html)

docs/topics/serialization.md
daniCsorbaJB: Hmm 🤔 
If users eventually read this on kotlinlang.org, I think the consistency could help them feel a sense of familiarity with the documentation. Since users use What's next everywhere else, this could help reinforce that further. That said, I'm not against renaming this section, although I would refrain from using words like "guide." For the user seeing this page on kotlinlang.org, they might not necessarily think of this as a guide, more like an entry point to everything related to serialization. 

For example, we have a tutorial in multiplatform docs about [Create a multiplatform app using Ktor and SQLDelight](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-ktor-sqldelight.html) for some of our users, they would meet serialization here for the first time. Up until this point, they've seen that every time on the bottom of a page, there are next steps after reading said page.  

> but if you're using it just as a reference/starter it feels out of place. 

That's a very good point! 👍  I think we could at least add some additional points here, even if they might be redundant for someone who is reading from top to bottom. We could even divide it further based on experience, using some intro statements before each list.

docs/topics/serialization.md
daniCsorbaJB: > it is supposed to replace the current one on https://kotlinlang.org/docs/serialization.html , right?

You are correct 👍 

It would serve as an entry point and a quick overview. So it would change in scope. The current [serialization page](https://kotlinlang.org/docs/serialization.html) is that way right now because the entire documentation wasn't integrated into the website. 

It would serve as something similar to these entry points: [Kotlin for data analysis](https://kotlinlang.org/docs/data-analysis-overview.html) or [coroutines](https://kotlinlang.org/docs/coroutines-overview.html).

sandwwraith: Aaaand coroutines guide does not mention (at least on overview and get started) that you should have some library added to Gradle. I think that may be confusing for some users. I would suggest that serialization page, even as overview, should contain basic setup instructions

docs/topics/serialization.md
sandwwraith: Btw, coroutines page does exactly this: https://kotlinlang.org/docs/coroutines-guide.html
Although this 'guide' page is kinda not the root one: we have https://kotlinlang.org/docs/coroutines-overview.html as a starter.

docs/topics/serialization.md
daniCsorbaJB: Very interesting point 🤔 
I'm not against changing "primitives" to something like a sequence of calls to try to move away from hinting at an intermediate file in the process:

> To convert an object tree to a string or to a sequence of bytes, it must go through two mutually intertwined processes: 
> 
> 1. Serialization: Objects are described as a sequence of structural and value-based calls.
> A serializer performs this description by analyzing each object's contents.
> 2. Encoding: An encoder uses the sequence of calls to produce the final output in the target format.
> 
> Serialization flow{width=700}
> 
> The reverse process involves parsing the input format, decoding it as a sequence of structural and value-based calls,
> and then deserializing the resulting stream into objects.

But perhaps this is again too low-level, as @sandwwraith said, and we should introduce it on a more advanced page. 🤔 

For the rest of the points:
I think the main issue is that if we go into so much implementation detail, we lose conceptual clarity and make the entry point too steep for newcomers.

docs/topics/serialization.md
daniCsorbaJB: I don't think we should keep the old version. 🤔 
It should be replaced the same way it was with coroutines: https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/basics.md

This way, we avoid the overhead of updating or maintaining the old files. The new files are meant to fully replace them. (but we should only do that when we are ready with the last PR)

sandwwraith: It will be replaced once we'll finish with the review. In a meanwhile, you need a place to merge subsequent PRs to. I think it should be a separate working directory.

docs/topics/serialization.md
pdvrieze: > But perhaps this is again too low-level, as @sandwwraith said, and we should introduce it on a more advanced page. 🤔
> 
> For the rest of the points: I think the main issue is that if we go into so much implementation detail, we lose conceptual clarity and make the entry point too steep for newcomers.

I think you are right about a more advanced page. The biggest challenge I see is where users want to do something more advanced and then get confused (see many of the Q&A type questions). That is why I wrote it in terms of what each bit "knows" about (in the normal case).

Perhaps an image (a quick draw) helps more:
kotlinx_serialization

daniCsorbaJB: Thank you very much! 💯 — I saved this and return to it for a more advanced page! 👍

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,72 @@
+[//]: # (title: Serialization)
```
sandwwraith, line 1: `docs-website/images/serialization.svg` is no longer needed, so you can put it aside

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,72 @@
...
+>
+{style="note"}
+
+Additionally, classes annotated with `@Serializable` are fully supported for serialization, enabling the conversion of class instances to and from formats like JSON.
+For more information, see [Serialize classes](serialization-customization-options.md).
+
+## What's next
+
+* Learn the basics of Kotlin serialization in the [Get started with serialization tutorial](serialization-get-started.md).
+* See how the `kotlinx.serialization` library processes [primitives, collections, and other built-in types](serialization-builtin-types.md)
```
sandwwraith: At line 56, you have `serialization-serialize-builtin-types.md` as the link. What's the correct file name?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,72 @@
...
+There are also community-maintained libraries that support more serialization formats, such as [YAML](https://yaml.org/) or [Apache Avro](https://avro.apache.org/).
+
+You can find out more about experimental serialization formats in [Alternative and custom serialization formats](alternative-serialization-formats.md).
+
+## Supported serialization types
+
+Kotlin serialization supports a variety of built-in types, including all primitive types and composite types from the Kotlin standard library like the `List` type.
+For more information, see [Serialize built-in types](serialization-serialize-builtin-types.md).
+
+> Not all types from the Kotlin standard library are serializable. In particular, [ranges](ranges.md) and the [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) class are not serializable at the moment.
```
sandwwraith: I find it a little weirdly specific that we mention ranges and Regex on the overview page. Maybe move the note to builtin-types.md doc?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,72 @@
...
+
+Additionally, classes annotated with `@Serializable` are fully supported for serialization, enabling the conversion of class instances to and from formats like JSON.
+For more information, see [Serialize classes](serialization-customization-options.md).
+
+## What's next
+
+* Learn the basics of Kotlin serialization in the [Get started with serialization tutorial](serialization-get-started.md).
+* See how the `kotlinx.serialization` library processes [primitives, collections, and other built-in types](serialization-builtin-types.md)
+* To explore more complex JSON serialization scenarios, see [JSON serialization overview](configure-json-serialization.md).
+* Dive into the [Serialize classes](serialization-customization-options.md) section to learn how to serialize classes and how to modify the default behavior of the `@Serializable` annotation.
```
sandwwraith: Comparing to the existing guide, section with polymorphism is missing from the ToC. Did you move it as a sub-section to some of the documents? I think it is important enough to have its own place in 'What's next' section.

docs-website/topics/serialization.md
daniCsorbaJB: it should be: serialization-serialize-builtin-types.md
Really nice catch, thank you 💯

docs-website/topics/serialization.md
daniCsorbaJB: It's already there as well — so it's safe to remove 👍

docs-website/topics/serialization.md
daniCsorbaJB: Sure 👍 I added the following:

* See how to serialize different types through a shared base type in [Serialize polymorphic classes](serialization-polymorphism.md).

I tried to keep it as newbie friendly as possible - what do you think? 🤔

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
+[//]: # (title: Serialization)
+
+**Serialization** is the process of converting data used by an application to a format that can be transferred over a
+network or stored in a database or a file. Deserialization is the opposite process of converting external data back into a runtime object.
```
sarahhaggarty:
```suggestion
network, or stored in a database or a file. Deserialization is the opposite process of converting external data back into a runtime object.
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+Together, they are essential to most applications that exchange data with third parties.
+
+Some data serialization formats, such as [JSON](https://www.json.org/json-en.html) and [Protocol Buffers](https://protobuf.dev/) are particularly common.
```
sarahhaggarty:
```suggestion
Some data serialization formats, such as [JSON](https://www.json.org/json-en.html) and [Protocol Buffers](https://protobuf.dev/), are particularly common.
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+Being language-neutral and platform-neutral, these formats enable data exchange between systems written in any modern language.
```
sarahhaggarty:
```suggestion
These formats are language- and platform-neutral, so you can use them to exchange data between systems written in any modern language.
```
I think this reads more clearly if we talk about formats first. WDYT?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+Kotlin provides this functionality through the [`kotlinx.serialization` libraries](#kotlin-serialization-libraries),
+which support multiple platforms and data formats.
+
+If you're new to serialization in Kotlin, we recommend starting with the [Get Started with Serialization](serialization-get-started.md) tutorial.
+It walks you through adding the Kotlin serialization library to your project and shows you how to serialize and deserialize your first class.
```
sarahhaggarty, line 13: The text refers to `kotlinx.serialization` libraries as well as the Kotlin serialization library. Can you choose one to use consistently?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+
+<a href="serialization-get-started.md"><img src="get-started-serialization.svg" width="700" alt="Get started with serialization" style="block"/></a>
+
+## Kotlin serialization libraries
+
+The `kotlinx.serialization` library offers support for all platforms, including JVM, JavaScript, Native.
```
sarahhaggarty:
```suggestion
The `kotlinx.serialization` libraries offer support for all platforms, including the JVM, JavaScript, and Native platforms.
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+It works with various serialization formats, such as JSON, CBOR, and Protocol buffers. For the complete list of supported serialization,
```
sarahhaggarty:
```suggestion
The libraries work with various serialization formats, such as JSON, CBOR, and Protocol buffers. For the complete list of supported serialization,
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+see the [supported formats](#supported-serialization-formats) section.
```
sarahhaggarty:
```suggestion
see [Supported serialization formats](#supported-serialization-formats).
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+
+All Kotlin serialization libraries are part of the `org.jetbrains.kotlinx:` group, with names
+starting with `kotlinx-serialization-` and suffixes that reflect the serialization format.
+For example:
+
+* `org.jetbrains.kotlinx:kotlinx-serialization-json` provides JSON serialization for Kotlin projects.
```
sarahhaggarty:
```suggestion
* `org.jetbrains.kotlinx:kotlinx-serialization-json` provides JSON serialization.
```
To make it parallel in structure with the next bullet.

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+* `org.jetbrains.kotlinx:kotlinx-serialization-cbor` provides CBOR serialization.
+
+Platform-specific dependencies are automatically managed, so you don't need to add them manually.
```
sarahhaggarty: Can you explain what is managing the dependencies? The libraries?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+Use the same dependencies for JVM, JavaScript, Native, and multiplatform projects.
```
sarahhaggarty: Is this sentence about not platform-specific depenencies? It's a bit confusing because you say above that it's automatically managed and then here that you should use some dependencies 🤔

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+
+The `kotlinx.serialization` libraries follow their own versioning structure, independent of Kotlin.
```
sarahhaggarty: Can you explain the versioning structure here?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Supported serialization formats
+
+`kotlinx.serialization` includes libraries for various serialization formats:
+
+* [JSON](https://www.json.org/json-en.html): [`kotlinx-serialization-json`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#json)
+* [Protocol buffers](https://protobuf.dev/): [`kotlinx-serialization-protobuf`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#protobuf)
+* [CBOR](https://cbor.io/): [`kotlinx-serialization-cbor`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#cbor)
+* [Properties](https://en.wikipedia.org/wiki/.properties): [`kotlinx-serialization-properties`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#properties)
+* [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md): [`kotlinx-serialization-hocon`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#hocon) (only on JVM)
```
sarahhaggarty: What do you think about making this a table?

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+
+All libraries except JSON serialization (`kotlinx-serialization-json`) are [experimental](components-stability.md), which means their API can be changed without notice.
```
sarahhaggarty:
```suggestion
All libraries, except for JSON serialization (`kotlinx-serialization-json`), are [Experimental](components-stability.md). Their APIs might change at any time.
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+Kotlin serialization supports a variety of built-in types, including all primitive types and most composite types from the Kotlin standard library like the `List` type.
+For more information, see [Serialize built-in types](serialization-serialize-builtin-types.md).
+
+Additionally, classes annotated with `@Serializable` are fully supported for serialization, enabling the conversion of class instances to and from formats like JSON.
+For more information, see [Serialize classes](serialization-customization-options.md).
+
+## What's next
+
+* Learn the basics of Kotlin serialization in the [Get started with serialization tutorial](serialization-get-started.md).
+* See how the `kotlinx.serialization` library processes [primitives, collections, and other built-in types](serialization-serialize-builtin-types.md)
```
sarahhaggarty:
```suggestion
* See how the `kotlinx.serialization` libraries process [primitives, collections, and other built-in types](serialization-serialize-builtin-types.md)
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+* To explore more complex JSON serialization scenarios, see [JSON serialization overview](configure-json-serialization.md).
```
sarahhaggarty:
```suggestion
* Explore more complex JSON serialization scenarios in the [JSON serialization overview](configure-json-serialization.md).
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+* Dive into the [Serialize classes](serialization-customization-options.md) section to learn how to serialize classes and how to modify the default behavior of the `@Serializable` annotation.
```
sarahhaggarty:
```suggestion
* Dive into [Serialize classes](serialization-customization-options.md) to learn how to serialize classes and modify the default behavior of the `@Serializable` annotation.
```

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,68 @@
...
+All libraries except JSON serialization (`kotlinx-serialization-json`) are [experimental](components-stability.md), which means their API can be changed without notice.
+For more details about JSON serialization, see [JSON serialization overview](configure-json-serialization.md).
+
+There are also community-maintained libraries that support more serialization formats, such as [YAML](https://yaml.org/) or [Apache Avro](https://avro.apache.org/).
+
+You can find out more about experimental serialization formats in [Alternative and custom serialization formats](alternative-serialization-formats.md).
+
+## Supported serialization types
+
+Kotlin serialization supports a variety of built-in types, including all primitive types and most composite types from the Kotlin standard library like the `List` type.
```
sarahhaggarty, line 57: Again, swapping between "Kotlin serialization" and `kotlinx.serialization`. Please make the starting sentences of "Supported serialization formats" and "Supported serialization types" consistent :)

docs-website/topics/serialization.md
daniCsorbaJB: I like it 👍

docs-website/topics/serialization.md
daniCsorbaJB: As discussed via chat these are different: I clarified it in the beginning of the doc.

docs-website/topics/serialization.md
daniCsorbaJB: Good point I moved things around a little and added a link to the get started where we show how to add this 

> Kotlin serialization offers support for all platforms, including JVM, JavaScript, and Native.
You can use the same [dependency declaration](serialization-get-started.md#add-plugins-and-dependencies-for-kotlin-serialization) regardless of the target platform.

docs-website/topics/serialization.md
daniCsorbaJB: Hmm 🤔 it seems to follow the same sermver structure - so I corrected it to:

The `kotlinx.serialization` libraries follow their own versioning, independent of Kotlin.
You can find the latest release versions on [GitHub](https://github.com/Kotlin/kotlinx.serialization/releases).

docs-website/topics/serialization.md
daniCsorbaJB: Hmm 🤔 I like the idea 👍 Converted it to a table

docs-website/topics/serialization.md
```diff
@@ -0,0 +1,70 @@
...
+| [JSON](https://www.json.org/json-en.html)            | [`kotlinx-serialization-json`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#json)             | all supported platforms | Stable       |
+| [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) | [`kotlinx-serialization-hocon`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#hocon)           | JVM only                | Experimental |
+| [Protocol Buffers](https://protobuf.dev/)            | [`kotlinx-serialization-protobuf`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#protobuf)     | all supported platforms | Experimental |
+| [CBOR](https://cbor.io/)                             | [`kotlinx-serialization-cbor`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#cbor)             | all supported platforms | Experimental |
+| [Properties](https://en.wikipedia.org/wiki/.properties) | [`kotlinx-serialization-properties`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#properties) | all supported platforms | Experimental |
```
sarahhaggarty:
```suggestion
| [JSON](https://www.json.org/json-en.html)            | [`kotlinx-serialization-json`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#json)             | All supported platforms | Stable       |
| [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) | [`kotlinx-serialization-hocon`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#hocon)           | JVM only                | Experimental |
| [Protocol Buffers](https://protobuf.dev/)            | [`kotlinx-serialization-protobuf`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#protobuf)     | All supported platforms | Experimental |
| [CBOR](https://cbor.io/)                             | [`kotlinx-serialization-cbor`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#cbor)             | All supported platforms | Experimental |
| [Properties](https://en.wikipedia.org/wiki/.properties) | [`kotlinx-serialization-properties`](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/README.md#properties) | All supported platforms | Experimental |
```

sarahhaggarty (approved):
Awesome! Just one last suggestion. LGTM! 🚀

PR #2979, opened 2025-04-10
Add an adapted version of the EfficientBinaryFormat to the default tests.

pdvrieze:
This patch set creates a simple implementation of the EfficientBinaryFormat. It is included in the existing json tests (special casing json specific serializers). This should ensure that serialization works with less forgiving formats that require in order serialization and must write defaults.


sandwwraith:
I still do not think that we will support such formats in general, given their inability to handle `@EncodeDefault(NEVER)`. Also, JSON tests largely expect encoders to be `JsonEncoder`, and supporting this would be quite inconvenient.

PR #2983, opened 2025-04-14
Fixed ProGuard rules for obfuscation to work correctly

### LINKED ISSUE #2976: Missing R8 rule for library obfuscation
**Describe the bug**

class (it's not the exact code but similar example):
```kotlin

sealed interface Navigation

@Serializable
@Parcelize
data class Details(val data: Data): Navigation

@Serializable
@Parcelize
sealed class Data: Parcelable

@Serializable
@Parcelize
data class Data1(val id: Int): Data()

@Serializable
@Parcelize
data class Data2(val name: String): Data()

```

Setup:
* library aar obfuscated using R8 containing `@Serializable` annotated classes for androix.navigation compose type-safe navigation
* application which has dependency to that library and is also obfuscated using R8

In runtime when I open the screen when navigation classes are used it crashes with `Missing serializer` exception

**To Reproduce**

Issue appears only if library and app are obfuscated (independently).
If library is obfuscated and app not then everything works.
If library is not obfuscated and app is then it also works.

**Expected behavior**

Not crash

**Environment**
 - Kotlin version: 2.1.20
 - Library version: 1.8.1
 - Kotlin platforms: Android
 - Gradle version: 8.13
 - androidx.navigation: 2.8.9
 - Android Gradle Plugin: 8.9.0

To fix this I added a rule to library obfuscation (not consumer proguard rules so it's not applied in the app obfuscation phase):

```
-if @kotlinx.serialization.Serializable class **
-keep class <1>$Companion { *; }
```

shanshin:
The current rules only apply if the companion class is called `Companion`. If obfuscation is triggered, the rule does not work and the static `Companion` field is not kept, which breaks the lookup of a companion.

A rule has also been added that preserves the named companion class name and field, as they must match to correctly search for the named companion in runtime.

Fixes #2976

### DIFF
--- a/rules/common.pro
+++ b/rules/common.pro
@@ -2,7 +2,15 @@
 # This avoids serializer lookup through `getDeclaredClasses` as done for named companion objects.
 -if @kotlinx.serialization.Serializable class **
 -keepclassmembers class <1> {
-    static <1>$Companion Companion;
+    static <1>$* Companion;
+}
+
+# Keep names for named companion object from obfuscation
+# Names of a class and of a field are important in lookup of named companion in runtime
+-keepnames @kotlinx.serialization.internal.NamedCompanion class *
+-if @kotlinx.serialization.internal.NamedCompanion class *
+-keepclassmembernames class * {
+    static <1> *;
 }
 
 # Keep `serializer()` on companion objects (both default and named) of serializable classes.


sandwwraith (approved):

PattaFeuFeu:
Does the change implemented in this PR affect the [Android section of the README](https://github.com/Kotlin/kotlinx.serialization#android) where it currently says the following?

<details><summary>Android section of README</summary>
<p>
### Android

By default, proguard rules are supplied with the library.
[These rules](rules/common.pro) keep serializers for _all_ serializable classes that are retained after shrinking,
so you don't need additional setup.

**However, these rules do not affect serializable classes if they have named companion objects.**

If you want to serialize classes with named companion objects, you need to add and edit rules below to your `proguard-rules.pro` configuration. 

Note that the rules for R8 differ depending on the [compatibility mode](https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md) used.

<details>
<summary>Example of named companion rules for ProGuard and R8 compatibility mode</summary>

```proguard
# Serializer for classes with named companion objects are retrieved using `getDeclaredClasses`.
# If you have any, replace classes with those containing named companion objects.
-keepattributes InnerClasses # Needed for `getDeclaredClasses`.

-if @kotlinx.serialization.Serializable class
com.example.myapplication.HasNamedCompanion, # <-- List serializable classes with named companions.
com.example.myapplication.HasNamedCompanion2
{
    static **$* *;
}
-keepnames class <1>$$serializer { # -keepnames suffices; class is kept when serializer() is kept.
    static <1>$$serializer INSTANCE;
}
```
</details>


<details>
<summary>Example of named companion rules for R8 full mode</summary>

```proguard
# Serializer for classes with named companion objects are retrieved using `getDeclaredClasses`.
# If you have any, replace classes with those containing named companion objects.
-keepattributes InnerClasses # Needed for `getDeclaredClasses`.

-if @kotlinx.serialization.Serializable class
com.example.myapplication.HasNamedCompanion, # <-- List serializable classes with named companions.
com.example.myapplication.HasNamedCompanion2
{
    static **$* *;
}
-keepnames class <1>$$serializer { # -keepnames suffices; class is kept when serializer() is kept.
    static <1>$$serializer INSTANCE;
}

# Keep both serializer and serializable classes to save the attribute InnerClasses
-keepclasseswithmembers, allowshrinking, allowobfuscation, allowaccessmodification class
com.example.myapplication.HasNamedCompanion, # <-- List serializable classes with named companions.
com.example.myapplication.HasNamedCompanion2
{
    *;
}
```
</details>
</p>
</details>

PR #2984, opened 2025-04-14
Fix buffer resizing overflow and introduce safe power-of-two capacity growth

bernardladenthin:
- Previously, calculating next capacity for large values like 1_073_741_824 caused overflow and returned Integer.MIN_VALUE.
- Rewrote `ensureCapacity` to use `Long` for arithmetic to guard against exceeding Int.MAX_VALUE.
- Introduced `nextPowerOfTwoCapacity(minCapacity: Int)` to safely calculate the next power-of-two ≥ minCapacity.
- Capped capacity at Int.MAX_VALUE to prevent buffer allocation errors.
- Added comprehensive unit tests covering negative, small, exact, and large input values to validate correct capacity growth behavior.

Note: You can compare the fixed behavior with the old logic using:
```kotlin
fun nextPowerOfTwoCapacity(minCapacity: Int): Int {
    return minCapacity.takeHighestOneBit() shl 1
}
```

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
@@ -43,18 +43,47 @@ internal class ByteArrayOutput {
     private var position: Int = 0
 
     private fun ensureCapacity(elementsToAppend: Int) {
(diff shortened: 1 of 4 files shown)


bernardladenthin:
wrong branch...

PR #2985, opened 2025-04-14
Fix buffer resizing overflow and introduce safe power-of-two capacity growth

bernardladenthin:
- Previously, calculating next capacity for large values like 1_073_741_824 caused overflow and returned Integer.MIN_VALUE.
- Rewrote `ensureCapacity` to use `Long` for arithmetic to guard against exceeding Int.MAX_VALUE.
- Introduced `nextPowerOfTwoCapacity(minCapacity: Int)` to safely calculate the next power-of-two ≥ minCapacity.
- Capped capacity at Int.MAX_VALUE to prevent buffer allocation errors.
- Added comprehensive unit tests covering negative, small, exact, and large input values to validate correct capacity growth behavior.

Note: You can compare the fixed behavior with the old logic using:
```kotlin
fun nextPowerOfTwoCapacity(minCapacity: Int): Int {
    return minCapacity.takeHighestOneBit() shl 1
}
```

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
@@ -62,29 +62,6 @@ internal class ByteArrayOutput {
         array.copyInto(newArray, startIndex = 0, endIndex = this.position)
         return newArray
     }
-    
-    /**
-     * Calculates the next power-of-two capacity based on the required minimum size.
-     *
-     * This function ensures the returned value is at least as large as `minCapacity`,
-     * and is always a power of two, unless `minCapacity` is less than or equal to zero,
-     * in which case it returns 0. If the calculated power of two exceeds `Integer.MAX_VALUE`,
-     * it returns `Integer.MAX_VALUE`.
-     *
-     * It's useful for resizing arrays with exponential growth.
-     *
-     * @param minCapacity The minimum required capacity.
-     * @return A capacity value that is a power of two and ≥ minCapacity, or 0 if `minCapacity` is ≤ 0.
-     */
-    fun nextPowerOfTwoCapacity(minCapacity: Int): Int {
-        if (minCapacity <= 0) return 0
-
-        val highestOneBit = minCapacity.takeHighestOneBit()
-        val maxHighestOneBit = Integer.MAX_VALUE.takeHighestOneBit()
-
-        // Check if shifting would exceed the maximum allowed value
-        return if (highestOneBit < maxHighestOneBit) highestOneBit shl 1 else Integer.MAX_VALUE
-    }
 
     fun copyFrom(src: ByteArrayOutput) {
         write(src.array, count = src.position)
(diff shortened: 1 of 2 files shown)


bernardladenthin:
The method under test (`nextPowerOfTwoCapacity`) must be called from `ByteArrayOutput`, I guess. Anyway, I’d also like to test at least `ensureCapacity`. I didn’t find any existing tests for it, did I miss them? Also, is there a CI setup to execute all tests automatically?

pdvrieze:
As you are considering the case of very large buffers, you may want to consider that it is not actually possible to create an array of size 2^31 (MaxInt + 1), but only of size 2^31-1 (MaxInt). This maximum size is however not a power of 2. Meaning that you almost half the maximum supported size to 2^30 (or approx 1bln values instead of 2bln).

Ps. yes there is a CI setup (using Jetbrains private infrastructure)

bernardladenthin:
Right, my initial goal was to address the existing issue with the `NegativeArraySizeException` and the incorrect size calculation. This affects anyone working with more than `1_073_741_824` bytes, meaning the buffer is currently limited in an unfortunate way to roughly 1 GB. Additionally, the exception points to an error within the user's own implementation, which can be quite confusing.

In the short term, it might be helpful to at least fix the issue up to `Int.MAX_VALUE`. In my example, this then leads to a clear `IllegalArgumentException("Required capacity exceeds maximum array size (Int.MAX_VALUE).")`, which at least clearly indicates the actual problem. That seems like a better outcome than the current behavior.

Of course, a proper refactoring would still be ideal to fully support large data sets.

bernardladenthin:
Hi, is there any update on this pull request? At the moment, it makes CBOR unusable, and the main goal here is to restore functionality first. A refactor could still follow afterwards with enough time. Thanks a lot!

sandwwraith:
@fzhinkin can you please take a look?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
```diff
@@ -43,12 +43,18 @@ internal class ByteArrayOutput {
...
+
+        val requiredCapacity = requiredCapacityLong.toInt()
+        if (requiredCapacity <= array.size) {
             return
         }
-        val newArray = ByteArray((position + elementsToAppend).takeHighestOneBit() shl 1)
-        array.copyInto(newArray)
-        array = newArray
+
+        val newCapacity = nextPowerOfTwoCapacity(requiredCapacity)
```
fzhinkin, line 56:
```suggestion
        val newCapacity = (requiredCapacityLong.takeHighestOneBit() shl 1)
            .coerceAtMost(Int.MAX_VALUE.toLong()).toInt()
```

It seems like the only reason to have a dedicated function is to test it, but since the function is trivial, I would rather embed it here, and simplify this expression (note that we don't have to cast `requiredCapacityLong` to `int` anymore).

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
```diff
@@ -43,12 +43,18 @@ internal class ByteArrayOutput {
     private var position: Int = 0
 
     private fun ensureCapacity(elementsToAppend: Int) {
-        if (position + elementsToAppend <= array.size) {
+        val requiredCapacityLong = position.toLong() + elementsToAppend.toLong()
+        if (requiredCapacityLong > Int.MAX_VALUE) {
+            throw IllegalArgumentException("Required capacity exceeds maximum array size (Int.MAX_VALUE).")
```
fzhinkin:
```suggestion
            throw SerializationException("Capacity required to hold a value serialized to CBOR ($requiredCapacityLong) exceeds maximum supported byte array length (${Int.MAX_VALUE}).")
```

It's better to:
- show the exact value that triggered an error (it might be worth to include the exact `elementsToAppend` value to the error message too, as it might help to debug an issue, when someone will have to)
- be more precise about what's happening, as it may help users facing the issue to deal with it
- throw `SerializationException` to depict that it's a problem with serializing some particular object, and it's not some internal assertion being failed.

fzhinkin (changes requested):
@bernardladenthin, thank you for opening the PR, and sorry for the delay with review.

I suggest simplifying the change and tuning the exception message, but otherwise it looks good.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
bernardladenthin: I would not recommend this change as proposed, because it again **removes test coverage** for this logic. I understand the motivation for inlining, but this leaves a non-trivial piece of functionality without dedicated tests. **This lack of testing is precisely why the current serialization is faulty**: there are no tests, and this change would discard them again. A better long-term solution would avoid imposing such limits altogether, although that is outside the scope of my change.

fzhinkin: The non-trivial piece of functionality is the `ByteArrayOutput`'s expansion, and it remains untested.

fzhinkin:
@bernardladenthin, are you willing to continue with this PR?

bernardladenthin:
Yes, please let's find a solution that works for now. For the future, I'd like to provide a patch using my library https://github.com/bernardladenthin/streambuffer that may resolve this kind of issue completely.

bernardladenthin:
> Yes, please let's find a solution that works for now. For the future, I'd like to provide a patch using my library https://github.com/bernardladenthin/streambuffer that may resolve this kind of issue completely.

I'm trying to implement this in the next days. Please give me some time.

PR #2987, opened 2025-04-15
Suppress `REDUNDANT_ELSE_IN_WHEN` in `ProtobufTaggedEncoder.encodeNull()`

lunakoly:
[KT-76635](https://youtrack.jetbrains.com/issue/KT-76635) makes this `when` exhaustive, and without the suppress our aggregate build for the corresponding branch fails.


sandwwraith:
And why exactly we need to replace dev repo? Current `dev` depends on stable 2.1.20

lunakoly:
Sorry, rebased the branch onto `kotlin-community/dev`, but forgot that it should be pushed separately. I've changed the base to `kotlin-community/dev` and I'll push a separate branch for `dev` with the last commit only

lunakoly:
See: https://github.com/Kotlin/kotlinx.serialization/pull/2987

sandwwraith:
PRs to `kotlin-community/dev` should be reviewed by @woainikk

woainikk (approved):

PR #2991, opened 2025-04-17
Replace some object comparisons with type checks

lunakoly:
A second cherry-pick to `dev` for [pull/2987](https://github.com/Kotlin/kotlinx.serialization/pull/2987).

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
@@ -20,7 +20,7 @@ public fun <T> readJson(json: Json, element: JsonElement, deserializer: Deserial
     val input = when (element) {
         is JsonObject -> JsonTreeDecoder(json, element)
         is JsonArray -> JsonTreeListDecoder(json, element)
-        is JsonLiteral, JsonNull -> JsonPrimitiveDecoder(json, element as JsonPrimitive)
+        is JsonLiteral, is JsonNull -> JsonPrimitiveDecoder(json, element)
     }
     return input.decodeSerializableValue(deserializer)
 }
--- a/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
+++ b/formats/json/jsMain/src/kotlinx/serialization/json/internal/DynamicEncoders.kt
@@ -235,10 +235,10 @@ private class DynamicObjectEncoder(
 
     @OptIn(ExperimentalSerializationApi::class)
     fun selectMode(desc: SerialDescriptor) = when (desc.kind) {
-        StructureKind.CLASS, StructureKind.OBJECT, SerialKind.CONTEXTUAL -> WriteMode.OBJ
-        StructureKind.LIST, is PolymorphicKind -> WriteMode.LIST
-        StructureKind.MAP -> WriteMode.MAP
-        is PrimitiveKind, SerialKind.ENUM -> {
+        is StructureKind.CLASS, is StructureKind.OBJECT, is SerialKind.CONTEXTUAL -> WriteMode.OBJ
+        is StructureKind.LIST, is PolymorphicKind -> WriteMode.LIST
+        is StructureKind.MAP -> WriteMode.MAP
+        is PrimitiveKind, is SerialKind.ENUM -> {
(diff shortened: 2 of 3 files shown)


sandwwraith:
I'm not sure I like the new style. Can you explain the benefits to me, please?

lunakoly:
It prevents from accidental comparisons resulting in `false` as the default implementation of `equals()` in non-`data` objects may return `false` when comparing one instance of `object MyObject` with another. Instead of doing `is MyObject`, it's also possible to turn the existing object into a `data object`, but I was hesitant to do so in `kotlinx.serialization` (at least as part of my current work on data-flow based exhaustiveness.

If you wish to discuss it, I propose to continue in the related Slack thread. I'll send you a link

JakeWharton:
> It protects from accidental comparisons resulting in `false` as the default implementation of `equals()` in non-`data` objects may return `false` when comparing one instance of `object MyObject` with another.

Where are you seeing this happen? With multiple class loaders? Java deserialization? Something else?

lunakoly:
The default example I keep in my head is, indeed, Java deserialization ([KT-40218](https://youtrack.jetbrains.com/issue/KT-40218)). While I am not aware if this specific example is relevant for `kotlinx.serialization`, I take the fact it [was considered](https://github.com/Kotlin/KEEP/blob/master/proposals/data-objects.md#deserialization-specific-design) during the design of `data object` as a sign they should be preferred in general when defining `object`s within `sealed` hierarchies, and if marking some objects `data` is not possible, they should be checked via `is MyObject`.

It is worth mentioning that, we are not yet sure if we want the compiler to emit a warning on attempts to compare non-`data` objects by `==`. There are scenarios when turning the existing `object`s into `data` requires additional boilerplate.

sandwwraith:
I have hard time imagining someone would willingly deserialize Kotlin's `object` with Java. In that case, it is reasonable to have a `data object` with all precautions. For all other cases, I doubt `is` check is more preferable code style. 

After all, `object` was chosen precisely because I can do `when(..) MyObj -> ` or `x === MyObj`.

pdvrieze:
> > It protects from accidental comparisons resulting in `false` as the default implementation of `equals()` in non-`data` objects may return `false` when comparing one instance of `object MyObject` with another.
> 
> Where are you seeing this happen? With multiple class loaders? Java deserialization? Something else?

Looking at the bytecode generated for Kotlin objects there are some small holes that allow multiple initialisation of objects (mainly involving java serialization - ironically, and sun.misc.Unsafe). However, if it is really needed, the solution would be to tighten the implementation of object's instantiation. Note that initialisation of the static INSTANCE field is [synchronized by the JVM](https://docs.oracle.com/javase/specs/jvms/se24/html/jvms-5.html#jvms-5.5) as long as the class initializer doesn't do crazy things (serialization doesn't).

Classloaders is a different case. Types are fully identified by FQN+"loading classloader" and that independence is by design (an `is MyObject` check will also be false in this case as the types are unrelated when loaded by different classloaders).

pdvrieze:
> The default example I keep in my head is, indeed, Java deserialization ([KT-40218](https://youtrack.jetbrains.com/issue/KT-40218)). While I am not aware if this specific example is relevant for `kotlinx.serialization`, I take the fact it [was considered](https://github.com/Kotlin/KEEP/blob/master/proposals/data-objects.md#deserialization-specific-design) during the design of `data object` as a sign they should be preferred in general when defining `object`s within `sealed` hierarchies, and if marking some objects `data` is not possible, they should be checked via `is MyObject`.

If you think it worthwhile, go ahead and update the code generation for object to add `readObject` and `writeObject` methods to prevent java serialization from creating a mess (or maybe for those who actually use java and kotlin serialization they could write `readObject`/`writeObject` themselves).

> 
> It is worth mentioning that, we are not yet sure if we want the compiler to emit a warning on attempts to compare non-`data` objects by `==`. There are scenarios when turning the existing `object`s into `data` requires additional boilerplate.

lunakoly:
@pdvrieze Thank you for your analysis, it is a good point. However, we are not considering updating code generation of `object`s (the [old rationale](https://youtrack.jetbrains.com/issue/KT-40218/Singleton-object-when-serialized-and-deserialized-is-not-equal-to-itself#focus=Comments-27-7267277.0-0) by Roman still holds).

I am closing this merge request since its change is not required.

PR #2996, opened 2025-04-23
Kotlin User Projects: add kotlin_additional_cli_options and logging

woainikk:
(https://youtrack.jetbrains.com/issue/KT-75078)](https://youtrack.jetbrains.com/issue/KT-75078)

### DIFF
--- a/buildSrc/src/main/kotlin/global-compiler-options.gradle.kts
+++ b/buildSrc/src/main/kotlin/global-compiler-options.gradle.kts
@@ -3,6 +3,11 @@
  */
 
 import org.jetbrains.kotlin.gradle.tasks.*
+import kotlin.collections.joinToString
+
+val kotlin_additional_cli_options = providers.gradleProperty("kotlin_additional_cli_options")
+    .map { it.split(" ") }
+    .orNull
 
 val globalCompilerArgs
     get() = listOf(
@@ -18,6 +23,10 @@ tasks.withType(KotlinCompilationTask::class).configureEach {
         // Unconditional compiler options
         freeCompilerArgs.addAll(globalCompilerArgs)
 
+        if (kotlin_additional_cli_options != null) {
+            freeCompilerArgs.addAll(kotlin_additional_cli_options)
+        }
+
         val isMainTaskName = name.startsWith("compileKotlin")
         if (isMainTaskName) {
             val werrorEnabled = when (kotlin_Werror_override?.lowercase()) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/global-compiler-options.gradle.kts
```diff
@@ -28,7 +37,7 @@ tasks.withType(KotlinCompilationTask::class).configureEach {
             }
 
             allWarningsAsErrors = werrorEnabled
-
+            logger.info("allWarningsAsErrors=$werrorEnabled")
```
sandwwraith, line 40: Is this logging really needed?

buildSrc/src/main/kotlin/global-compiler-options.gradle.kts
```diff
@@ -46,3 +55,9 @@ tasks.withType<Kotlin2JsCompile>().configureEach {
 tasks.withType<KotlinNativeCompile>().configureEach {
     compilerOptions { freeCompilerArgs.add("-Xpartial-linkage-loglevel=ERROR") }
 }
+
+tasks.withType<KotlinCompilationTask<*>>().configureEach {
+    doFirst {
```
sandwwraith, line 60: I suggest printing only the additional args if they were provided: `if (kotlin_additional_cli_options?.isNotEmpty()) logger.info(...)`

woainikk: I think it's useful to have all options in log. now it look like this: 
Added Kotlin compiler flags: -Xexplicit-api=strict, -Xsuppress-version-warnings, -Xexpect-actual-classes, -P, plugin:org.jetbrains.kotlinx.serialization:disableIntrinsic=false, -Xreport-all-warnings, -Xrender-internal-diagnostic-names, -nowarn

buildSrc/src/main/kotlin/global-compiler-options.gradle.kts
woainikk: yes, that will allow us to make sure we use correct settings

sandwwraith: I mean, that's not exactly what I would want to see in the build log

image

sandwwraith: Maybe move it to `Added Kotlin compiler flags` section then?

sandwwraith:
```
tasks.withType<KotlinCompilationTask<*>>().configureEach {
    doFirst {
      logger.info("Added Kotlin compiler flags: ${compilerOptions.freeCompilerArgs.get().joinToString(", ")}")
      logger.info("Werror state: ${compilerOptions.allWarningsAsErrors.get()}")
   }
}

woainikk: fair, fixed

PR #2998, opened 2025-04-28
Mention type unsoundness of `serializer()` in the documentation

### LINKED ISSUE #2948: Documentation for `serializer(KType)` doesn't mention that it's unsound
```kotlin
import kotlin.reflect.*

fun main() {
    println(Json.encodeToString(serializer(typeOf<Double>()), "string"))
}
```

compiles just fine, but throws

```
Exception in thread "main" java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number (java.lang.String and java.lang.Number are in module java.base of loader 'bootstrap')
	at kotlinx.serialization.internal.DoubleSerializer.serialize(Primitives.kt:113)
	at kotlinx.serialization.json.internal.StreamingJsonEncoder.encodeSerializableValue(StreamingJsonEncoder.kt:259)
	at kotlinx.serialization.json.internal.JsonStreamsKt.encodeByWriter(JsonStreams.kt:99)
	at kotlinx.serialization.json.Json.encodeToString(Json.kt:125)
```

The documentation doesn't mention the intended usage pattern (which is probably to do an unchecked cast of the result into a `KSerializer` of the required type?) and doesn't explain that the returned value is prone to crashes (or surprising results!) otherwise.

Some fun combinations:
* `Json.encodeToString(serializer(typeOf<Int>()), Long.MAX_VALUE) == "-1"`
* `Json.encodeToString(serializer(typeOf<Int>()), 3.1415) == "3"`

dkhalanskyjb:
Fixes #2948

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Serializers.kt
+++ b/core/commonMain/src/kotlinx/serialization/Serializers.kt
@@ -58,13 +58,13 @@ public inline fun <reified T> SerializersModule.serializer(): KSerializer<T> {
  * Creates a serializer for the given [type].
  * [type] argument is usually obtained with [typeOf] method.
  *
- * This overload works with full type information, including type arguments and nullability,
- * and is a recommended way to retrieve a serializer.
- * For example, `serializer<typeOf<List<String?>>>()` returns [KSerializer] that is able
- * to serialize and deserialize list of nullable strings — i.e. `ListSerializer(String.serializer().nullable)`.
- *
  * Variance of [type]'s type arguments is not used by the serialization and is not taken into account.
  * Star projections in [type]'s arguments are prohibited.
+ *
+ * **Pitfall**: the returned serializer may return incorrect results or throw a [ClassCastException] if it receives
+ * a value that's not a valid instance of the [KType], even though the type allows passing such a value.
+ * Consider using the `serializer()` overload accepting a type argument (for example, `serializer<List<String>>()`),
+ * which returns the serializer with the correct type.
  *
  * @throws SerializationException if serializer cannot be created (provided [type] or its type argument is not serializable).
  * @throws IllegalArgumentException if any of [type]'s arguments contains star projection
@@ -80,10 +80,15 @@ public fun serializer(type: KType): KSerializer<Any?> = EmptySerializersModule()
  * The nullability of returned serializer is specified using the [isNullable].
  *
  * Note that it is impossible to create an array serializer with this method,
- * as array serializer needs additional information: type token for an element type.
+ * as an array serializer needs additional information: type token for an element type.
  * To create array serializer, use overload with [KType] or [ArraySerializer] directly.
  *
  * Caching on JVM platform is disabled for this function, so it may work slower than an overload with [KType].
+ *
+ * **Pitfall**: the returned serializer may return incorrect results or throw a [ClassCastException] if it receives
+ * a value that's not a valid instance of the [KType], even though the type allows passing such a value.
+ * Consider using the `serializer()` overload accepting a type argument (for example, `serializer<List<String>>()`),
+ * which returns the serializer with the correct type.
  *
  * @throws SerializationException if serializer cannot be created (provided [kClass] or its type argument is not serializable)
  * @throws SerializationException if [kClass] is a `kotlin.Array`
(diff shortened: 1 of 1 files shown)


core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -58,14 +58,14 @@ public inline fun <reified T> SerializersModule.serializer(): KSerializer<T> {
  * Creates a serializer for the given [type].
  * [type] argument is usually obtained with [typeOf] method.
  *
- * This overload works with full type information, including type arguments and nullability,
- * and is a recommended way to retrieve a serializer.
- * For example, `serializer<typeOf<List<String?>>>()` returns [KSerializer] that is able
```
dkhalanskyjb, line 63: I don't know what this dependent-type-style syntax means and was assuming that this was a copy-paste artifact of some sort.

sandwwraith: I think I mistyped the function call: it should be `serializer(typeOf<List<String?>>())`. Can you please bring this paragraph back in the correct form?

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -60,12 +60,17 @@ public inline fun <reified T> SerializersModule.serializer(): KSerializer<T> {
  *
  * This overload works with full type information, including type arguments and nullability,
  * and is a recommended way to retrieve a serializer.
- * For example, `serializer<typeOf<List<String?>>>()` returns [KSerializer] that is able
- * to serialize and deserialize list of nullable strings — i.e. `ListSerializer(String.serializer().nullable)`.
+ * For example, `serializerOrNull(typeOf<List<String?>>)` returns [KSerializer] that is able
```
sandwwraith:
```suggestion
 * For example, `serializer(typeOf<List<String?>>())` returns [KSerializer] that is able
```

This is documentation for regular `serializer()` function, and typeOf needs invocation

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -80,11 +85,16 @@ public fun serializer(type: KType): KSerializer<Any?> = EmptySerializersModule()
...
  *
  * Note that it is impossible to create an array serializer with this method,
- * as array serializer needs additional information: type token for an element type.
+ * as an array serializer needs additional information: type token for an element type.
  * To create array serializer, use overload with [KType] or [ArraySerializer] directly.
  *
  * Caching on JVM platform is disabled for this function, so it may work slower than an overload with [KType].
  *
+ * **Pitfall**: the returned serializer may return incorrect results or throw a [ClassCastException] if it receives
+ * a value that's not a valid instance of the [KType], even though the type allows passing such a value.
```
sandwwraith:
```suggestion
 * a value that's not a valid instance of the [KClass], even though the type allows passing such a value.
```

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -103,11 +113,14 @@ public fun serializer(
  * This overload works with full type information, including type arguments and nullability,
  * and is a recommended way to retrieve a serializer.
  * For example, `serializerOrNull<typeOf<List<String?>>>()` returns [KSerializer] that is able
```
sandwwraith: This line also needs change to `serializerOrNull(typeOf<List<String?>>())` while we're here

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -120,12 +133,18 @@ public fun serializerOrNull(type: KType): KSerializer<Any?>? = EmptySerializersM
  *
  * This overload works with full type information, including type arguments and nullability,
  * and is a recommended way to retrieve a serializer.
- * For example, `serializer<typeOf<List<String?>>>()` returns [KSerializer] that is able
- * to serialize and deserialize list of nullable strings — i.e. `ListSerializer(String.serializer().nullable)`.
+ * For example, `serializer(typeOf<List<String?>>)` returns [KSerializer] that is able
```
sandwwraith:
```suggestion
 * For example, `serializer(typeOf<List<String?>>())` returns [KSerializer] that is able
```

core/commonMain/src/kotlinx/serialization/Serializers.kt
```diff
@@ -169,11 +194,14 @@ public fun SerializersModule.serializer(
  * This overload works with full type information, including type arguments and nullability,
  * and is a recommended way to retrieve a serializer.
  * For example, `serializerOrNull<typeOf<List<String?>>>()` returns [KSerializer] that is able
```
sandwwraith: And here, if you do not mind

sandwwraith (approved):

PR #3005, opened 2025-05-16
Migrate to DGPv2

whyoleg:
Migrate to new Dokka Gradle Plugin ([migration guide](https://kotlinlang.org/docs/dokka-migration.html))

Because of the Gradle task name change, the kotlinlang deploy configuration should also be updated.
Related: https://github.com/JetBrains/kotlin-web-site/pull/XXXX (Should be merged at the same time)


formats/json-io/build.gradle.kts
```diff
@@ -33,12 +33,8 @@ kotlin {
...
-tasks.named<DokkaTaskPartial>("dokkaHtmlPartial") {
-    dokkaSourceSets {
-        configureEach {
-            externalDocumentationLink {
-                url.set(URI("https://kotlin.github.io/kotlinx-io/").toURL())
-            }
-        }
+dokka.dokkaSourceSets.configureEach {
+    externalDocumentationLinks.register("kotlinx-io") {
+        url("https://kotlinlang.org/api/kotlinx-io/")
```
whyoleg, line 38: JIC: old URL is not accessible anymore

formats/json/build.gradle.kts
```diff
@@ -52,11 +52,4 @@ kotlin {
     }
 }
 
-// This task should be disabled because of no need to build and publish intermediate JsWasm sourceset
-tasks.whenTaskAdded {
```
whyoleg, line 56: This workaround is not needed, locally build and publish tasks works find without it, but it caused some issues locally at my side.
I can revert it, if you know the reason why it's needed

buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
```diff
@@ -18,62 +11,23 @@ dependencies {
...
-                reportUndocumented.set(false)
-                skipDeprecated.set(true)
-            }
+    dokkaSourceSets.configureEach {
+        includes.from(rootDir.resolve("dokka/moduledoc.md").path)
+        reportUndocumented = true
 
-            sourceLink {
-                localDirectory.set(rootDir)
+        perPackageOption {
```
whyoleg, line 23: Previously, there were multiple `perPackageOption` - most of them are obsolete or separate configurations for `internal` package. So, obsolete were removed and `internal` are handled by one regex now

formats/json/build.gradle.kts
sandwwraith: cc @igoriakovlev maybe this workaround is not needed anymore?

buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
sandwwraith: Are `@Deprecated` declarations skipped now by default?

formats/json-io/build.gradle.kts
sandwwraith: Local build gives me `Failed to download package-list from https://kotlinlang.org/api/kotlinx-io/package-list, this might suggest that remote resource is not available, module is empty or dokka output got corrupted`, and `Source/Sink` declarations in the doc are not linked, despite URL being available

whyoleg: Hm, that looks strange, local build works fine on my side.
I can make it use the local package list (downloaded), as it was before for okio. The only case when the package list will become outdated is when a package is added or removed, so probably it should not be a big deal here to just use the local package list.
WDYT?

buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
whyoleg: Nope, only `HIDDEN` ones. Looks like I missed the flag during refactoring
Is it desirable behavior to skip deprecated with `WARNING` and `ERROR` declarations in all modules?

formats/json/build.gradle.kts
whyoleg: JIC, I've now additionally checked, and `./gradlew compileJsWasmMainKotlinMetadata` executed successfully, so we don't need this hack anymore.
I believe initially, this source set contained some declarations that were available in JS and wasm-js targets, but without a shared declaration, as there is no shared source set in the stdlib for JS and wasm-js.
But I don't see anything like this now.

buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
whyoleg: I've brought back `skipDeprecated = true`

formats/json-io/build.gradle.kts
whyoleg: I've added local package lists for both kotlinx-io and okio, so that the build will be less dependent on network failures.

JIC: Dokka by default still relies on downloading the package list of the stdlib.

buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
```diff
@@ -18,62 +11,24 @@ dependencies {
...
+    dokkaSourceSets.configureEach {
+        includes.from(rootDir.resolve("dokka/moduledoc.md").path)
+        reportUndocumented = true
+        skipDeprecated = true
 
-            sourceLink {
-                localDirectory.set(rootDir)
+        perPackageOption {
+            matchingRegex = ".*\\.internal(\\..*)?"
+            suppress = true
```
sandwwraith, line 26: reportUndocumented = false ?

whyoleg: There is no need, as suppressed packages are skipped during analysis, and so no warnings will be present for them

sandwwraith (approved):

PR #3006, opened 2025-05-19
Make throwXXX functions return Nothing instead of Unit

Martmists-GH:
This allows using them in e.g. when statements, but also might clarify code behavior to the compiler.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/PluginExceptions.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/PluginExceptions.kt
@@ -7,7 +7,7 @@ import kotlinx.serialization.descriptors.SerialDescriptor
 
 @OptIn(ExperimentalSerializationApi::class)
 @InternalSerializationApi
-public fun throwMissingFieldException(seen: Int, goldenMask: Int, descriptor: SerialDescriptor) {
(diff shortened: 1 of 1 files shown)


sandwwraith:
These functions are marked with `@InternalSerializationApi`, have a little bit weird contract with bit masks, and are intended to be called by the plugin-generated code. 

Do you actually use them in your code?

Martmists-GH:
I'd attempted to use them in custom serializers as I hadn't found the proper way to error during (de-)serialization.

sandwwraith:
Why didn't you use `throw MissingFieldException(missingFields, descriptor.serialName)` directly?

sandwwraith:
I've taken a look around, and there seem to be a lot of downsides due to the signature change:
- kotlinx-serialization compiler plugin for Wasm has to be adapted to the new signature because Wasm virtual machine is quite strict in terms of types and return values (see failed build log)
- Public API JVM signatures will also change, which may lead to problems with linkage and binary compatibility
- In these cases, old signatures and old declarations have to be left in the source code for compatibility reasons.

Given that this API is internal and is not called explicitly, the benefits of having `Nothing` do not outweigh the downsides at all to make this change. Users should create and throw `MissingFieldException` directly.

pdvrieze:
Maybe adding some documentation on why they exist (and are public) together with guidance to just throw the exception directly would be useful.

PR #3019, opened 2025-06-10
Add JsonObject field reading syntax sugar

huicunjun:
Add Enhanced JsonObject deserialization and field reading syntax sugar


formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt
```diff
@@ -347,3 +353,25 @@ internal fun unexpectedJson(key: String, expected: String): Nothing =
...
+/**
+ * Convenience methods to get typed elements from [JsonObject]
+ */
+public inline fun <reified T> JsonElement.bean(): T = Json.decodeFromString<T>(this.toString())
+public inline fun <reified T> JsonObject.getBeanOrNull(key: String): T? = this[key]?.bean<T>()
+public fun JsonObject.getJsonObject(key: String): JsonObject? = this[key]?.jsonObject
+public fun JsonObject.getJsonArray(key: String): JsonArray? = this[key]?.jsonArray
+public fun JsonObject.getJsonPrimitive(key: String): JsonPrimitive? = this[key]?.jsonPrimitive
+public fun JsonObject.getJsonNull(key: String): JsonNull? = this[key]?.jsonNull
+public fun JsonObject.getIntOrNull(key: String): Int? = this[key]?.jsonPrimitive?.intOrNull
```
pdvrieze: This (and similar) functions should include a check that it is not a string `isString` should return `false`

huicunjun: > This (and similar) functions should include a check that it is not a string `isString` should return `false`

If I don't check isString , it seems to have no consequences, I guess it should be the converted type?
image

huicunjun:
@pdvrieze  Can you review it

pdvrieze:
@huicunjun It looks good to me (not an official maintainer so can't do anything beyond that)

PR #3023, opened 2025-06-20
Add SavedState to the format doc

kuanyingchou:
Hi, we added KotlinX Serialization support [in May](https://developer.android.com/jetpack/androidx/releases/savedstate#version_13_2). It would be nice if it can be included in the format doc.


sandwwraith:
That's cool to hear, congratulations!

sandwwraith (approved):

PR #3027, opened 2025-06-25
Rename KClass<*>.isInterface helper in the JS part

broadwaylamb:
We are going to introduce a new KClass member in
the standard library with the same name marked
with an opt-in annotation, so in the usages of this
extension property would stop compiling because
the member is always favored and there is no opt-in.

See KT-78581


core/jsMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -73,8 +73,10 @@ internal actual fun isReferenceArray(rootClass: KClass<Any>): Boolean = rootClas
  * WARNING: may be broken in arbitrary time in the future without notice
  *
  * Should be eventually replaced with compiler intrinsics
+ *
+ * TODO: Remove this when KT-78581 lands into a release
```
sandwwraith: Do you have a release version in mind?

sandwwraith:
PRs should be based against `dev` branch

core/jsMain/src/kotlinx/serialization/internal/Platform.kt
broadwaylamb: 2.2.20 hopefully

sandwwraith:
```suggestion
 * TODO: Remove this when KT-78581 lands into a 2.2.20 release
```

sandwwraith (approved):

PR #3029, opened 2025-06-25
Update JsonPath tests:

sandwwraith:
- Uncomment test that was fixed by some other PR
- Remove jvmOnly clause, because tests are working on other platforms too
- Add test with useArrayPolymorphism=false


shanshin (approved):

sandwwraith:
Apparently `testHugeNestingToCheckResize` fails on mingw without any info, investigating...

PR #3031, opened 2025-06-26
Update to Kotlin 2.2

sandwwraith:
(Api dump updated because of KT-73954)


hfhbd:
Could you also unmute this test to use the generated serializer: https://github.com/Kotlin/kotlinx.serialization/blob/c86e7903f92c98f2842ca413bb5690c32b428158/formats/json-tests/commonTest/src/kotlinx/serialization/InstantSerializationTest.kt#L86

PR #3032, opened 2025-06-27
Prepare 1.9.0 release

sandwwraith:


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+1.9.0 / 2025-06-27
+==================
+
+This release updates Kotlin version to 2.2.0, includes several bugfixes and provides serializers for kotlin.time.Instant.
+
+## Add kotlin.time.Instant serializers
+
+Instant class was moved from kotlinx-datetime library to Kotlin standard library.
```
hfhbd, line 8: Do you also need a specific serialization plugin version to support the new runtime classes for generated serializers?

sandwwraith: They were added into the plugin as default in 2.1.20 IIRC
For all previous versions, you can use `@Serializable(InstantSerializer::class)` I think

hfhbd: Hm, https://youtrack.jetbrains.com/issue/KT-75759/Add-the-serializer-for-kotlin.time.Instant-to-the-list-of-standard-serializers is marked for 2.2.0. IMHO you should also add the needed compiler plugin version in the release notes.

sandwwraith: Hm, indeed. Thanks for pointing that out

PR #3035, opened 2025-07-02
🍒 Rename KClass<*>.isInterface helper in the JS part

broadwaylamb:
This is a cherry-pick of #3027 into the `kotlin-community/k1/dev` branch.


sandwwraith:
@woainikk is `k1/dev` branch still actual? I though we dropped its support

woainikk:
K1 projects are still running until we deprecate K1 compiler, so yes, we support the branch  as regression check, but do not rebase it

PR #3038, opened 2025-07-07
Support a new publication scheme

fzhinkin:


buildSrc/src/main/kotlin/publishing-conventions.gradle.kts
```diff
@@ -232,12 +232,14 @@ fun MavenPublication.signPublicationIfKeyPresent() {
     }
 }
 
-fun RepositoryHandler.addSonatypeRepository() {
+// Artifacts are published to an intermediate repo (libs.repo.url) first,
+// and then deployed to the central portal.
+fun RepositoryHandler.addPublishingRepository() {
     maven {
-        url = mavenRepositoryUri()
```
shanshin, line 237: Is `mavenRepositoryUri()` unused now? If so, it's better to delete it

shanshin (approved):

PR #3041, opened 2025-07-11
Revised the ProGuard rules and added tests on R8

### LINKED ISSUE #3033: New ProGuard implementation in #2983 causes R8 warning
**Describe the bug**
Since `kotlinx-serialization-json` `v1.9.0`, I get an R8 warning when building a release version of my Android app using Gradle.

I get the following warning:

`The type "<1>$*" is used in a field rule. The characters in this type are legal for the JVM, but unlikely to originate from a source language. Maybe this is not the rule you are looking for.`

When checking the configuration.txt file in the mapping folder (when compiling using Android Studio), I see exactly one occurrence of the `<1>$*` string. It occurs only in the `r8\kotlinx-serialization-common.pro` file.

It looks like this is introduced since #2983 (specifically, in this commit: https://github.com/Kotlin/kotlinx.serialization/pull/2983/commits/666595327ace9a880b6c256f6aa10b9b0a6884c2)

Is this a reason to rollback to the previous version? Or is there something I should change in my code to prevent this?

**Environment**
 - Kotlin version: [2.0.21]
 - Library version: [1.9.0]
 - Kotlin platforms: [JVM 21.0.6]
 - Gradle version: [8.13]
 - IDE version [Android Studio 2025.1.1]
 - Targeting Android SDK 36, compiling using SDK 36, minimum SDK 31

shanshin:
- The warning message about unusual symbols in field type is eliminated.
- Added correct rules for named companions
- Changed code for the locating name companion by annotation. Since the list of nested classes is not saved by R8, static fields are analyzed
- added tests on the rules using R8 in full mode and ProGuard compatibility mode

Fixes #3033


core/commonTest/src/kotlinx/serialization/SerializersLookupNamedCompanionTest.kt
```diff
@@ -63,31 +63,22 @@ class SerializersLookupNamedCompanionTest {
...
-            )
-        }
-
-        shouldFail<SerializationException>(beforeKotlin = "1.9.20", onJs = false, onNative = false, onWasm = false) {
-            assertEquals(
-                SealedInterface.serializer().descriptor.toString(),
-                serializer(typeOf<SealedInterface>()).descriptor.toString()
-            )
-        }
+        assertSame<KSerializer<*>>(PlainSerializer, serializer(typeOf<PlainWithCustom>()))
```
sandwwraith: Is this change related? I suggest extracting it to separate commit

Also change base branch to `dev` btw

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+dependencies {
+    "commonImplementation"(project(":kotlinx-serialization-core"))
+    "commonImplementation"("org.jetbrains.kotlin:kotlin-test")
+    "commonImplementation"("org.jetbrains.kotlin:kotlin-test-junit")
+    "commonImplementation"(libs.junit.junit4)
+    "commonImplementation"(kotlin("test-junit"))
+}
+
+tasks.compileTestKotlin {
+    configureCompilation(true)
```
sandwwraith: I think it is better to call function with explicit parameter name: `configureCompilation(r8FullMode = true)`

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+import kotlin.reflect.KClass
+
+fun getR8Checker(): R8Checker {
+    val usageFile = File(System.getProperty("r8.output.usage"))
+    val mapFile = File(System.getProperty("r8.output.map"))
+
+    return R8CheckerImpl(parseR8Output(mapFile, usageFile))
+}
+
+interface R8Checker {
```
sandwwraith: I don't think we should do interface/Impl split in a 250-lines-long utils file for a small test

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -77,12 +77,19 @@ private fun <T: Any> Class<T>.findInNamedCompanion(vararg args: KSerializer<Any?
     }
 }
 
-private fun <T: Any> Class<T>.findNamedCompanionByAnnotation(): Any? {
-    val companionClass = declaredClasses.firstOrNull { clazz ->
-        clazz.getAnnotation(NamedCompanion::class.java) != null
+private fun <T : Any> Class<T>.findNamedCompanionByAnnotation(): Any? {
+    // search static field with type marked by kotlinx.serialization.internal.NamedCompanion - it's the companion
+    val field = declaredFields.firstOrNull { field ->
+        Modifier.isStatic(field.modifiers) && field.type.getAnnotation(NamedCompanion::class.java) != null
```
sandwwraith, line 84: Mention in the comment that `declaredClasses` unavailable after R8

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+kotlin {
+    // use toolchain from settings
+    jvmToolchain(jdkToolchainVersion)
+}
+
+sourceSets {
+    // create the source set for storing sources and using dependency configuration
+    val common by creating
```
sandwwraith: `common` name can be misleading because of the strong associations with MPP. Also, `test/testCompatible` names themselves are quite vague. WDYT of renaming everything to:

- '(test)Shared'
- 'testR8Full'
- 'testPGCompatible'

Or something similar.

Also, you can just reuse regular `main` sourceset?

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+
+    val test by getting {
+        kotlin.srcDirs(common.kotlin.srcDirs)
```
sandwwraith: Also, seeing `src` next to `common.pro` seems confusing. I would prefer one of the two:

1. Leave .pro files in `rules` directory, move all test logic to something called `rules-integration-tests`
2. Rename `src` dir to something like `integrationTestSrc`

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+val extractBaseJarTask = tasks.register<Task>("extractBaseJar") {
+    inputs.property("jdkVersion", jdkToolchainVersion)
+    outputs.file(baseJar)
+
+    doLast {
+        val javaLauncher = javaToolchains.launcherFor {
+            languageVersion.set(JavaLanguageVersion.of(jdkToolchainVersion))
+        }.get()
+        val javaHomeDir = javaLauncher.metadata.installationPath.asFile
+        val baseJmod = javaHomeDir.resolve("jmods").resolve("java.base.jmod")
```
sandwwraith: Shouldn't we check that file exists and throw an error otherwise?
E.g. "cannot find file with base java module, make sure that specified jdk_toolchain_version is 9 or higher"

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+
+        val extractDir = temporaryDir.resolve("java-base")
+
+        extractDir.deleteRecursively()
+        extractDir.mkdirs()
+        // unpack jmod file
+        val jdkBinDir = javaHomeDir.resolve("bin")
+
+        val jmodFile = if (System.getProperty("os.name").startsWith("Windows")) {
```
sandwwraith: There's saying that `spi.ToolProvider` (https://docs.oracle.com/javase/10/docs/api/java/util/spi/ToolProvider.html) can get jmod programmatically, without hassle of manipulating different system binaries. May help here.

Also, apparently jmod is just a ZIP file, maybe we can repackage it as is with ZIP api (or even just change the extension)

rules/build.gradle.kts
```diff
@@ -0,0 +1,261 @@
...
+ */
+fun KotlinCompile.configureCompilation(r8FullMode: Boolean) {
+    // R8 output files
+    val mode = if (r8FullMode) "full" else "compatible"
+    val mapFile = layout.buildDirectory.file("r8/$mode/mapping.txt")
+    val usageFile = layout.buildDirectory.file("r8/$mode/usage.txt")
+
+    dependsOn(extractBaseJarTask)
+
+    inputs.files(ruleFiles)
```
sandwwraith: Should we have `inputs.files(baseJar)` or is it added automatically due to `dependsOn`?

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+     * If there are several methods with name [name], [IllegalArgumentException] will be thrown.
+     * If there is no method with name [name], [IllegalArgumentException] will be thrown.
+     */
+    fun findMethod(name: String): R8Result
+
+    /**
+     * Find R8 information about a method by its name and descriptor.
+     * If there is no method with given [name] and [descriptor], [IllegalArgumentException] will be thrown.
+     */
+    fun findMethod(name: String, descriptor: String): R8Result
```
sandwwraith: I do not see usages of this function, do we need it?

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+            classMap[original] = currentClass
+            return@forEachLine
+        }
+
+        methodRegex.matchEntire(line)?.let {
+            val current = currentClass ?: throw IllegalStateException("No current class")
+
+            val (num, returnType, name, desc, _, obfuscated) = it.destructured
+            if (num.isNotEmpty()) {
+                val existed = current.methods.singleOrNull { clazz -> clazz.name == name && clazz.descriptor == desc }
```
sandwwraith: I guess it's `method`, not `clazz`

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+                if (existed == null) {
+                    current.methods += MethodEntry(name, returnType, desc, obfuscated)
+                } else {
+                    if (existed.obfuscatedName != obfuscated) {
+                        if (obfuscated == name) {
```
sandwwraith: I do not get it. Proguard can decide to un-obfuscate method? Or can two overrides with one name be unobfuscated+obfuscated?

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+                            current.methods.remove(existed)
```
sandwwraith: Also, all this remove/single on lists make traversal quadratic. I guess it doesn't matter since we have small test data, but don't you want make `methods` `MutableMap<String, MutableList<Method>>` anyway?

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+                            current.methods += existed.copy(obfuscatedName = obfuscated)
+                        }
+                    }
+                }
+            } else {
+                current.fields.put(name, FieldEntry(name, returnType, obfuscated))
```
sandwwraith: How is it different from fieldRegex?

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,249 @@
...
+                val name = "Companion"
+                current.fields.put(name, FieldEntry(name, typeName, obfuscatedName))
+            }
+        }
+    }
+
+    // process usage.txt
+    var currentUsageClass: ClassEntry? = null
+
+    usageFile.forEachLine { raw ->
```
sandwwraith: If this file contains methods that were removed due to being unused, then it better be named `unused.txt`

rules/src/common/kotlin/kotlinx/serialization/r8/R8Tests.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Test classes renames and deleted as long as methods.
+     */
+    @Test
+    fun testOptimisation() {
+        val unusedClass = checker.findClass("kotlinx.serialization.r8.UnusedClass")
+        assertTrue(unusedClass.isShrunk)
+
+        ObfuscatedClass("World").used()
```
sandwwraith: Removing this line causes test to fail, so it's technically a test data. Maybe put this call to some top-level function in Model.kt?

rules/src/common/kotlin/kotlinx/serialization/r8/R8Tests.kt
```diff
@@ -0,0 +1,119 @@
...
+     * Descriptor field should present.
+     * Using reflection here because the private field `descriptor` isn't present in the mapping file.
+     */
+    @Test
+    fun testDescriptorField() {
+        assertTrue(AccessSerializer.serializer()::class.java.declaredFields.any { it.name == "descriptor" })
+    }
+
+
+
```
sandwwraith: unnecessary blank lines

settings.gradle.kts
```diff
@@ -82,6 +82,9 @@ project(":benchmark").projectDir = file("./benchmark")
 include(":guide")
 project(":guide").projectDir = file("./guide")
 
+include(":proguard-rules")
+project(":proguard-rules").projectDir = file("./rules")
```
sandwwraith: Maybe it should be named `proguard-rules-test`, because rules on their own are just a bunch of .pro files and do not need Gradle module

sandwwraith (changes requested):

core/commonTest/src/kotlinx/serialization/SerializersLookupNamedCompanionTest.kt
shanshin: Not fully related but I think we should add these changes somewhere

rules/build.gradle.kts
shanshin: This approach it is not quite applicable, because it is necessary to distinguish between the JVM running Gradle and the JVM specified in toolchain.
`ToolProvider. findFirst` will use Gradle JVM and it's often JVM 1.8.

> Also, apparently jmod is just a ZIP file, maybe we can repackage it as is with ZIP api (or even just change the extension)

jmod is not zip file, it's internal JVM format

rules/build.gradle.kts
shanshin: I think it's better to add this dependency

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
shanshin: No

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
shanshin: It's even worse :)
R8/ProGuard can decide to inline some part of the method to another (e.g. place field initialization into `<init>` or `<clinit>`).
In such cases, the origin method occurs twice in the file.

Example from `mapping.txt`:
```
kotlinx.serialization.r8.R8Tests -> kotlinx.serialization.r8.R8Tests:
# {"id":"sourceFile","fileName":"R8Tests.kt"}
    ...
    5:5:kotlinx.serialization.r8.R8Checker getChecker():13:13 -> <init>
    ...
    1:1:kotlinx.serialization.r8.R8Checker getChecker():13:13 -> getChecker
    ...
```

rules/build.gradle.kts
sandwwraith: Tbh I was sure that Gradle is also running on toolchain-provisioned JVM. But yea, now it seems unlikely

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
shanshin: I have the same question :)
But it's common name for this files in Android.

rules/src/common/kotlin/kotlinx/serialization/r8/R8Tests.kt
shanshin: If we move this call to `Model.kt` it will be shrinked as only `*Tests` classes are alwayes kept. So I move it to separate method in `R8Tests`.

rules/src/common/kotlin/kotlinx/serialization/r8/DeletionChecks.kt
shanshin: Unused branch

rules/build.gradle.kts
shanshin: Move tests and Gradle files to subdirectory

rules/rules-integration-tests/src/shared/kotlinx/serialization/r8/DeletionChecks.kt
```diff
@@ -0,0 +1,213 @@
...
+
+            val (num, returnType, name, desc, _, obfuscated) = it.destructured
+            if (num.isNotEmpty()) {
+                val signature = name + desc
+                val existed = current.methods[signature]
+                if (existed == null) {
+                    current.methods[signature] = MethodEntry(name, returnType, desc, obfuscated)
+                } else {
+                    if (existed.obfuscatedName != obfuscated) {
+                        if (obfuscated == name) {
```
sandwwraith, line 129: Please add comment explaining the logic of "If method saved its original name in one of the obfuscations, we use original, otherwise we use either of obfuscated"

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -77,12 +77,20 @@ private fun <T: Any> Class<T>.findInNamedCompanion(vararg args: KSerializer<Any?
     }
 }
 
-private fun <T: Any> Class<T>.findNamedCompanionByAnnotation(): Any? {
-    val companionClass = declaredClasses.firstOrNull { clazz ->
-        clazz.getAnnotation(NamedCompanion::class.java) != null
+private fun <T : Any> Class<T>.findNamedCompanionByAnnotation(): Any? {
+    // search static field with type marked by kotlinx.serialization.internal.NamedCompanion - it's the companion
+    // declaredClasses after R8 even if `-keepattributes InnerClasses, EnclosingMethod` is specified, so we use declaredFields
```
sandwwraith:
```suggestion
    // declaredClasses are erased after R8 even if `-keepattributes InnerClasses, EnclosingMethod` is specified, so we use declaredFields
```

rules/rules-integration-tests/build.gradle.kts
```diff
@@ -0,0 +1,271 @@
...
+
+dependencies {
+    sharedImplementation(project(":kotlinx-serialization-core"))
+    sharedImplementation("org.jetbrains.kotlin:kotlin-test")
+    sharedImplementation("org.jetbrains.kotlin:kotlin-test-junit")
+    sharedImplementation(libs.junit.junit4)
+    sharedImplementation(kotlin("test-junit"))
+}
+
+// extend commonImplementation by all test compilation tasks
```
sandwwraith:
```suggestion
// extend sharedImplementation by all test compilation tasks
```

PR #3042, opened 2025-07-11
Suppress the deprecation warning on a reference to Kotlin 2.0 language version

lunakoly:
See [KT-76136](https://youtrack.jetbrains.com/issue/KT-76136)


Tapchicoma (approved):
Not sure if we are still assembling this project with LV <2.0

lunakoly:
The issue is not with <2.0 but =2.0

PR #3059, opened 2025-08-11
Documentation rewrite - Get started page

daniCsorbaJB:
This is the second part of the Kotlin Serialization rewrite. 

Related YouTract ticket is: [KT-80054](https://youtrack.jetbrains.com/issue/KT-80054) [Docs][Libraries] Get started with Kotlin Serialization)

### DIFF
--- a/docs-website/topics/serialization-get-started.md
+++ b/docs-website/topics/serialization-get-started.md
@@ -276,6 +276,6 @@ Congratulations! You have successfully serialized an object to JSON and deserial
 
 ## What's next
 
-* Learn how to serialize built-in types like numbers and strings, in [Serialize built-in types](serialization-serialize-builtin-types.md).
+* Learn how to serialize basic types such as primitives and strings, as well as certain standard library classes, in [Serialize built-in types](serialization-serialize-builtin-types.md).
 * Discover how to customize class serialization and adjust the default behavior of the `@Serializable` annotation in [Serialize classes](serialization-customization-options.md).
 * Dive deeper into handling JSON data and configuring JSON serialization in the [JSON serialization overview](configure-json-serialization.md).


docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Maven
+
+Add the serialization plugin and library to your `pom.xml` file:
+
+1. Specify the Kotlin and serialization version in the `<properties>` section:
+
+    ```xml
+    <properties>
+    <kotlin.version>%kotlinVersion%</kotlin.version>
```
sandwwraith, line 46: Indentation is a little off — versions should be one tab more to the right

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+
+[Serialization](serialization.md) converts objects into a format that can be stored or transmitted and later reconstructed.
+
+This tutorial shows you how to add the necessary plugins and dependencies for Kotlin serialization and how to serialize and deserialize objects in JSON format.
+
+## Add plugins and dependencies for Kotlin serialization
+
+To include the `kotlinx.serialization` library in your project, add the corresponding plugin and dependency configuration based on your build tool.
+
+> To set up the Kotlin compiler plugin for Bazel, see the example provided in the [rules_kotlin repository](https://github.com/bazelbuild/rules_kotlin/tree/master/examples/plugin/src/serialization).
```
sandwwraith: I think this note should be moved after Gradle/Maven sections

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+   <dependencies>
+       <dependency>
+           <groupId>org.jetbrains.kotlinx</groupId>
+           <artifactId>kotlinx-serialization-json</artifactId>
+           <version>${serialization.version}</version>
+       </dependency>
+   </dependencies>
+   ```
+
+### Add the Kotlin serialization library to a multiplatform project
```
sandwwraith: We also mention Proguard rules (specific for Android) in the README.md. Do you plan to add them here?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+commonMain {
+   dependencies {
+      implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:%serializationVersion%")
+   }
+}
+```
+
+## Serialize objects to JSON
+
+Serialization is the process of converting an object into a format that can be easily stored or transmitted, such as JSON.
```
sandwwraith: This sentence repeats the one the whole document is started with

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+In Kotlin, you can serialize objects to JSON using the `kotlinx.serialization` library.
+
+To make a class serializable, you need to mark it with the [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation.
+This annotation instructs the compiler to generate the code required for serializing and deserializing instances of the class.
+For more information, see [The @Serialization annotation](serialization-customization-options.md#the-serializable-annotation).
+
+Let's look at an example:
+
+1. Import the necessary serialization libraries:
```
sandwwraith: "Import libraries" sounds not very technically correct. Library is something you can depend on, while you import _declarations_ from the said library

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+    ```
+
+2. Make a class serializable by annotating it with `@Serializable`:
+
+    ```kotlin
+    @Serializable
+    data class Data(val a: Int, val b: String)
+    ```
+
+   > The `@Serializable` annotation enables default serialization of all properties in the primary constructor.
```
sandwwraith: We serialize all properties with backing fields, not only the primary constructor ones.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+   > You can customize serialization behavior with custom constructors, optional properties, and more.
```
sandwwraith: What 'custom constructors' you mean here? I don't quite understand.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+   > 
+   > For more information, see [Serialize classes](serialization-customization-options.md).
+   >
+   {style="note"}
+
+3. Use the [`Json.encodeToString()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json/encode-to-string.html) function to serialize an instance of this class:
+
+    ```kotlin
+    // Imports the necessary libraries for serialization and JSON handling
```
sandwwraith: Same comment about declarations here

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+        val obj = Json.decodeFromString<Data>("""{"a":42, "b": "str"}""")
+        println(obj)
+        // Data(a=42, b=str)
+    }
+    ```
+   {kotlin-runnable="true"}
+
+Congratulations! You have successfully serialized an object to JSON and deserialized it back into an object in Kotlin!
+
+## What's next
```
sandwwraith, line 277: Should we add whole chapters to the TOC again? Or maybe add a separate page after all with the text like "See all chapters in this guide here — link-to-toc.md"

docs-website/topics/serialization-get-started.md
daniCsorbaJB: hmm 🤔 my issue with that is that way it might seem like it's related to Maven. Since it's just a note without specific instructions I left it in the generic part of the doc.

docs-website/topics/serialization-get-started.md
daniCsorbaJB: 🤔 
Since this isn't an introductory page, we should only list logical next steps here rather than broad exploration. (The ToC will be on the left side of the screen so if someone wants to wander, they can do so) 

My thoughts were that Built-in types are the simplest next topic, making them a natural continuation point.
We’ve also mentioned the` @Serializable` annotation, so linking to a page that explains it in detail might also be a sensible continuation.
Lastly, because the example on this page uses JSON, the JSON serialization overview might also be an appropriate next step.

docs-website/topics/serialization-get-started.md
daniCsorbaJB: good question 🤔 
I think we could add it to the Create custom serializer page (where we talk about `serializer()` and companion objects) Although, we could place a link to that part of the page here. 🤔 
What do you think?

docs-website/topics/serialization-get-started.md
daniCsorbaJB: fair point - seems pretty redundant - removed it 👍

docs-website/topics/serialization-get-started.md
daniCsorbaJB: hmm 🤔 I agree. The original intention was to fit info about non-default constructor for deserialization, but it reads really weird here.

What do you think about this version: 

> You can customize serialization behavior with property-level annotations, optional properties, and more. 

(like `@SerialName` `@Transient` etc)

docs-website/topics/serialization-get-started.md
sandwwraith: It's weird to start the dependencies section with Bazel. The first buildsystem mentioned should be our default — Gradle.

sandwwraith: You can use `### Other build systems` section or something like that to mention Bazel

sandwwraith: We can also mention there for example our 'raw' compiler plugin artifact to be used with `-Xplugin` CLI arg. I can add it there later

docs-website/topics/serialization-get-started.md
sandwwraith: I think it's a bad idea. It is not always related to custom serializer, regular `@Serializable` classes can have named companions as well. I would expect separate page for Proguard then (linked here)

docs-website/topics/serialization-get-started.md
sandwwraith: Also we can promote Amper here, adding instructions with it as well

pdvrieze: If there is a lot of stuff on different build systems it even makes sense to have a separate page for it (linked here). It would also help those who want to find instructions for setting it up in different build systems as they often would not expect it in a generic getting started page.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+           <groupId>org.jetbrains.kotlinx</groupId>
+           <artifactId>kotlinx-serialization-json</artifactId>
+           <version>${serialization.version}</version>
+       </dependency>
+   </dependencies>
+   ```
+
+### Add the Kotlin serialization library to a multiplatform project
+
+To use Kotlin serialization in multiplatform projects, add the JSON serialization library dependency to your common source set:
```
pdvrieze: Maybe change it to "To use Kotlin serialization for JSON in multiplatform projects..." (and maybe: "This will automatically include the core serialization library.") Currently it implies that JSON is needed if a different format is used.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
+[//]: # (title: Get started with Kotlin serialization)
+
+[Serialization](serialization.md) converts objects into a format that can be stored or transmitted and later reconstructed.
+
+This tutorial shows you how to add the necessary plugins and dependencies for Kotlin serialization and how to serialize and deserialize objects in JSON format.
+
```
pdvrieze: Maybe explicitly state that the library supports multiple formats, but this tutorial focuses on JSON.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,232 @@
...
+      implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:%serializationVersion%")
+   }
+}
+```
+
+## Serialize objects to JSON
+
+In Kotlin, you can serialize objects to JSON using the `kotlinx.serialization` library.
+
+To make a class serializable, you need to mark it with the [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation.
```
pdvrieze, line 177: In between these lines, you may also want to remind readers that the plugin must be applied (annotation+plugin) → serializer. People don't read carefully, so leaving it out in this context is likely going to lead to some (too many) people not applying the plugin and wondering why it doesn't work.

daniCsorbaJB:
> I've realized we have a tip
> 
> > We also provide kotlinx-serialization-core artifact that contains all serialization API but does not have a bundled serialization format with it
> 
> in our current README.md, which is missing from this document. Perhaps it is better to place it in the `serialization.md/Kotlin serialization libraries` or `Supported serialization formats` section in the form of
> 
> > Pro tip: if you do not want a dependency on a particular format (e.g., you are writing your own serialization format), you can use `kotlinx-serialization-core` library.

I really like the second option — let's go with that (let's put it in the supported serialization formats page)

docs-website/topics/serialization-get-started.md
daniCsorbaJB: Thank you for your thoughts. I see what you mean, but that would be redundant to follow up the previous section repeating what we just said. I'd put serialization as an intermediate area potentially coming after the basics (where newer users could have gotten used to applying plugins to their projects) so I don't think we have to resort to repeating ourselves here.

docs-website/topics/serialization-get-started.md
daniCsorbaJB: good idea — I'll add it! 👍

docs-website/topics/serialization-get-started.md
pdvrieze: It is easily added if a lot of users miss it, and then get stuck (but the need to apply the plugin should be prominent in the document - wherever it is).

docs-website/topics/serialization-get-started.md
daniCsorbaJB: Get started page should be enough — Let's merge Gradle + Maven into Tabs (very compact takes up not too much space)

docs-website/topics/serialization-get-started.md
daniCsorbaJB: Fair point — thank you for the context — I added it as a collapsible section then! 👍

docs-website/topics/serialization-get-started.md
```diff
@@ -121,6 +106,69 @@ commonMain {
 }
 ```
 
+This dependency automatically includes the core serialization library as well.
+
+### Configure R8 for Kotlin serialization in Android projects {initial-collapse-state="collapsed" collapsible="true"}
+
+The Kotlin serialization library includes default [ProGuard rules](https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro) that keep serializers for all serializable classes retained after shrinking.
```
sandwwraith: Why did you remove `so you don't need additional setup.` line? It is very important. Please bring it back, e.g. `This means that you do not need to do anything. However, these rules don't apply to classes with named companion objects.`

sandwwraith (changes requested):

sandwwraith:
> I really like the second option — let's go with that (let's put it in the supported serialization formats page)

So, do you plan to address it?

docs-website/topics/serialization-get-started.md
daniCsorbaJB: Hmm 🤔 to me it felt a bit redundant — it is already implied that if there are default rules for all serializable classes that you don't need to do anything, but a little clarification can be good 👍 
How about?

> The Kotlin serialization library includes default [ProGuard rules](https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro), so you don't need additional setup to keep serializers for all serializable classes retained after shrinking.

sandwwraith: Sounds good.

It's not redundant because the next paragraph starts with "this is how you can provide rules for companions", so clear emphasis should be put on "you don't need to do anything UNLESS you have named companion objects"

docs-website/topics/serialization-get-started.md
daniCsorbaJB: 🙏 It will be 👍 — on kotlinlang.org, users will be able to see five big sections on the right of their screen when they open this page. (the most prominent, because of the additional subsection,s will be the one having this info)

We haven't faced this issue with other features, so I'm confident this will work (if we ever run into an issue where this might become a problem, we'll address it right away 👌  )

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
+[//]: # (title: Get started with Kotlin serialization)
+
+[Serialization](serialization.md) converts objects into a format that can be stored or transmitted and later reconstructed.
```
sarahhaggarty:
```suggestion
[Serialization](serialization.md) converts objects into a format you can store or transmit and later reconstruct.
```
What do you think about making this more active?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+
+Kotlin serialization supports multiple formats. 
+This tutorial shows you how to add the necessary plugins and dependencies for Kotlin serialization and how to serialize and deserialize objects in JSON format.
```
sarahhaggarty:
```suggestion
This tutorial shows you how to add the necessary plugins and dependencies for Kotlin serialization, and how to serialize and deserialize objects in JSON format.
```

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Add plugins and dependencies for Kotlin serialization to your project
```
sarahhaggarty:
```suggestion
## Add plugins and dependencies
```
I think we can shorten this since it's clear we're talking about the Kotlin serialization library.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+
+</tab>
+</tabs>
+
+> To set up the Kotlin compiler plugin for Bazel, follow the example from the [rules_kotlin repository](https://github.com/bazelbuild/rules_kotlin/tree/master/examples/plugin/src/serialization).
+> Bazel isn't officially supported by the Kotlin team, and this repository is maintained independently.
+>
+{style="tip"}
+
+### Add the Kotlin serialization library to a multiplatform project
```
sarahhaggarty:
```suggestion
### Add the library to a multiplatform project
```

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+   }
+}
+```
+
+This dependency automatically includes the core serialization library as well.
+
+### Configure R8 for Kotlin serialization in Android projects {initial-collapse-state="collapsed" collapsible="true"}
+
+The Kotlin serialization library includes default [ProGuard rules](https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro), so you don't need additional setup to keep serializers for all serializable classes after shrinking.
+These rules don't apply to classes with named companion objects.
```
sarahhaggarty:
```suggestion
However, these rules don't apply to classes with named companion objects.
```

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+# If you have any such classes, replace the examples below with your own
+-keepattributes InnerClasses # Required for getDeclaredClasses
+
+-if @kotlinx.serialization.Serializable class
+com.example.myapplication.HasNamedCompanion, # <-- List serializable classes with named companions
+com.example.myapplication.HasNamedCompanion2
+{
+    static **$* *;
+}
+-keepnames class <1>$$serializer { # Using -keepnames is enough; class is kept when serializer() is kept
```
sarahhaggarty: I'm not sure what the part after the semi-colon means?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+      implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:%serializationVersion%")
+   }
+}
+```
+
+This dependency automatically includes the core serialization library as well.
+
+### Configure R8 for Kotlin serialization in Android projects {initial-collapse-state="collapsed" collapsible="true"}
+
+The Kotlin serialization library includes default [ProGuard rules](https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro), so you don't need additional setup to keep serializers for all serializable classes after shrinking.
```
sarahhaggarty: Would it be possible to briefly explain what R8 is for the readers who aren't Android developers? Maybe linking to https://developer.android.com/topic/performance/app-optimization/enable-app-optimization?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+> 
+{style="tip"}
+
+## Serialize objects to JSON
+
+In Kotlin, you can serialize objects to JSON using the `kotlinx.serialization` library.
+
+To make a class serializable, you need to mark it with the [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation.
+This annotation instructs the compiler to generate the code required for serializing and deserializing instances of the class.
+For more information, see [The @Serialization annotation](serialization-customization-options.md#the-serializable-annotation).
```
sarahhaggarty:
```suggestion
For more information, see [The `@Serialization` annotation](serialization-customization-options.md#the-serializable-annotation).
```

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+    ```kotlin
+    import kotlinx.serialization.*
+    import kotlinx.serialization.json.*
+    ```
+
+2. Make a class serializable by annotating it with `@Serializable`:
+
+    ```kotlin
+    @Serializable
+    data class Data(val a: Int, val b: String)
```
sarahhaggarty: Could you improve the example by making it more real world like? For example, the data class could be called `Book` and have `title` which is a `String` and `yearPublished` as an `Int`.

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+    fun main() {
+        // Deserializes a JSON string into an instance of the Data class
+        val obj = Json.decodeFromString<Data>("""{"a":42, "b": "str"}""")
+        println(obj)
+        // Data(a=42, b=str)
+    }
+    ```
+   {kotlin-runnable="true"}
+
+Congratulations! You have successfully serialized an object to JSON and deserialized it back into an object in Kotlin!
```
sarahhaggarty:
```suggestion
Congratulations! You have successfully serialized an object to JSON and deserialized it back into an object in Kotlin.
```
Nitpick. Just to make it sound not so excitable :)

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+
+## What's next
+
+* Learn how to serialize standard types, including built-in types like numbers and strings, in [Serialize built-in types](serialization-serialize-builtin-types.md).
```
sarahhaggarty: By "standard" do you mean basic types?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+* Discover how to customize class serialization and adjust the default behavior of the `@Serializable` annotation in the [Serialize classes](serialization-customization-options.md) section.
```
sarahhaggarty: Will it be a section in this page? Or is it a page itself?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+    data class Data(val a: Int, val b: String)
+
+    fun main() {
+        // Serializes an instance of the Data class into a JSON string
+        val json = Json.encodeToString(Data(42, "str"))
+        println(json)
+        // {"a":42,"b":"str"}
+    }
+    ```
+   {kotlin-runnable="true"}
```
sarahhaggarty: Can you give it an ID so we can see how many people run it in GA?

docs-website/topics/serialization-get-started.md
```diff
@@ -0,0 +1,281 @@
...
+    data class Data(val a: Int, val b: String)
+
+    fun main() {
+        // Deserializes a JSON string into an instance of the Data class
+        val obj = Json.decodeFromString<Data>("""{"a":42, "b": "str"}""")
+        println(obj)
+        // Data(a=42, b=str)
+    }
+    ```
+   {kotlin-runnable="true"}
```
sarahhaggarty: It would also be great to have an ID here.

docs-website/topics/serialization-get-started.md
sandwwraith: `@Serializable` btw

docs-website/topics/serialization-get-started.md
daniCsorbaJB: I like it 👍

docs-website/topics/serialization-get-started.md
daniCsorbaJB: If I understand correctly, it means you only need to preserve the class name for the `serializer()` call, so you don’t need a stronger keep rule than -keepnames?

We could perhaps write it as : 

> Using -keepnames is enough for the serializer() call to reference the class correctly

@sandwwraith does this sound correct?

docs-website/topics/serialization-get-started.md
daniCsorbaJB: sure — we could add the link to the shrinking word 👍

docs-website/topics/serialization-get-started.md
daniCsorbaJB: hmm 🤔 yea - but I think we can remove it as a whole to make it more concise: 

> * Learn how to serialize built-in types like numbers and strings, in [Serialize built-in types](serialization-serialize-builtin-types.md).

docs-website/topics/serialization-get-started.md
daniCsorbaJB: no the page itself so I rewrite it like: 

* Discover how to customize class serialization and adjust the default behavior of the `@Serializable` annotation in [Serialize classes](serialization-customization-options.md).

docs-website/topics/serialization-get-started.md
daniCsorbaJB: added: id="serialize-get-started"

docs-website/topics/serialization-get-started.md
daniCsorbaJB: added one here as well — good idea! 🙏

docs-website/topics/serialization-get-started.md
daniCsorbaJB: sure thing 👍  — rewrote both examples with those examples.

docs-website/topics/serialization-get-started.md
sandwwraith: LGTM

docs-website/topics/serialization-get-started.md
sandwwraith: Collections are not built-in or basic types, they belong to Kotlin standard library. Hence the name "standard types".

sarahhaggarty: Perhaps "standard library types" would work? Just "standard types" isn't a term we've been using in our docs so far and the relationship to the standard library isn't obvious.

daniCsorbaJB: we could be explicit: 
> Learn how to serialize built-in types, including basic types such as primitives and strings, as well as certain standard library classes, in [Serialize built-in types](serialization-serialize-builtin-types.md).

After discussing further with @sarahhaggarty I suggest let's go with the more concise one I suggested above:

> Learn how to serialize built-in types like numbers and strings, in [Serialize built-in types](https://github.com/Kotlin/kotlinx.serialization/pull/serialization-serialize-builtin-types.md).

sandwwraith: What about "Learn how to serialize basic types such as primitives and strings, as well as certain standard library classes, in [Serialize built-in types](https://github.com/Kotlin/kotlinx.serialization/pull/serialization-serialize-builtin-types.md)."?

daniCsorbaJB: that also sounds good 👍

PR #3060, opened 2025-08-11
[Wasm] fix tests for className wasm (KT-69621)

alex28sh:
In this issue [KT-69621](https://youtrack.jetbrains.com/issue/KT-69621) we changed wasm Throwable to use qualified name of class rather than simple name when throwing exceptions. So, I fixed serialization tests appropriately


alex28sh:
I suspect, right now tests should fail, since changes from https://github.com/JetBrains/kotlin/tree/rr/shefer/KClass.qualifierName-by-default have not been merged yet. So, I need this fix to merge them

sandwwraith:
You may also need to cherry-pick changes to `kotlin-community/dev`, since it is the branch used in Aggregate. cc @woainikk

Issue #2696, opened 2024-05-31
Strings are converted to Int in non-lenient mode


spand:
**Describe the bug**
It seems that strings will be automatically converted to `Int` by default in Json. That is highly surprising and especially so when lenient is set to false.

**To Reproduce**
```
    @Serializable
    data class Asd(
        val itsAnInt: Int,
    )

    @Test
    fun asd() {
        Json.decodeFromString<Asd>("""{"itsAnInt":"1"}""")
    }
```

**Expected behavior**
I expected it to throw an exception

**Environment**
 - Kotlin version: 1.9.10
 - Library version: 1.6.0
 - Kotlin platforms: JVM
 - Gradle version: 7.5.1

sandwwraith:
Can you explain why it is problematic for strings to be parsed as integers? From my experience, people find it more convenient when they don't have to think whether a particular value was quoted, since the serialization side is not always in their control. kotlinx.serialization itself always produces unquoted integers on `encodeToString` though.
Also it simplifies working with maps (see #2438)

JakeWharton:
While this is for 32-bit integers, I've also seen integers larger than 53 bits quoted so as to not lose precision with JS clients/servers/proxies. Having them transparently unwrapped when a proper 64-bit integer type is available is convenient.

spand:
Concretely we had a bug where a developer had used `Int` in the type definition where it should have been a String since the possible values were more than just quoted numbers. You could put that on the developer but at least if this feature had not existed the developer would have been forced to handle the "not a number" case explicitly.

In the abstract it is quite odd that such behavior is standard because
- It forces users to support behavior they might not have expected or wouldnt want to (Hyrum's Law). Concretely if we would need another tool to support the same json then we probably couldnt or it would be annoying like it would have been in kotlinx.serialization without this feature. All because we are forced to support this feature for every number.
- Principle of Least Surprise. No one in our team expected such a feature to exist. I really find it odd to have such a feature enabled by default. Especially in contrast to `ignoreUnknownKeys` that we enable so many places to be forward compatible (which still has the correct default imo).
- There is something off about not being able generate the same json as whats encoded. I assume people quote numbers because they cannot fit into a JS number despite json having no such restriction. So why cannot we generate the same ? A `@JsonQuoted` opt-in seemed more appropriate ?

I believe I have made my point here so I will stop unless you want more.

[...]

Issue #1042, opened 2020-09-06
Serializer coerce string to int 


herrjemand:
**Describe the bug**

When requiring type of Int in a serializable class, type coerce string to int without throwing an error.

**To Reproduce**
Attach a code snippet or test data if possible.

``` kotlin
import kotlinx.serialization.*
import kotlinx.serialization.json.*

@Serializable 
class Person(val age: Int)

fun main() {
    val obj = Json.decodeFromString<Person>("{\"age\": \"22\"}")
    println(obj) // Person(age=22)
}
```

**Expected behavior**

EXCEPTION

**Environment**
 - Kotlin version: [e.g. 1.4.0-RC]
 - Library version: [e.g. 1.0.0-RC]
 - Kotlin platforms: Multiplatform
 - Gradle version: [e.g. 6.5.1]

herrjemand:
Same thing with ints to strings

elizarov:
Please, check out `isLenient` setting for JSON. https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#lenient-parsing

Does it do what you need?

herrjemand:
@elizarov isLenient makes parser more flexible. For my project I need strict parsing. 

I have tested with isLinient = false, and it produces the same result. *(

elizarov:
Thanks! Now I see what you want. Why do you want it, though? What's wrong with accepting `"22"` for an integer in your case?

herrjemand:
@elizarov FIDO protocols require strict JSON parsing to get certification.

elizarov:
What else? Where can we read more?

herrjemand:
What @elizarov what do you mean?

elizarov:
Can you, please, provide a link to such an exemple standard to make sure we can study it and see what else we might be missing for compliance.

qwwdfsad:
@herrjemand could you please also elaborate on what you have been using prior to kotlinx-serialization? I've checked Moshi, GSON and Jackson, none of them supports this functionality by default.

Jackson will provide `CoercionConfig` in the next release that supports a lot of various coercion strategies, but for now it's unavailable

herrjemand:
@qwwdfsad

As a start https://fidoalliance.org/specs/fido-uaf-v1.1-id-20170202/fido-uaf-protocol-v1.1-id-20170202.html

gerob311:
[...]

PR #3063, opened 2025-08-15
RFC: Allow disabling coercion between JSON primitives

pschichtel:
Currently, and independent of the `isLenient` option, the JSON decoders will coerce Strings to other primitiv types, allowing e.g. `"true"` to be decoded as `Boolean`. We are currently transitioning from a very strict decoder setup based on Jackson over to kotlinx serialization for some of the nicer kotlin-specific functionality and the lack of reflection.

This PR is currently meant as an RFC to gauge if this is something that would be accepted upstream. As such there are some caveats:

* There are no additional tests, however the existing tests show that the default behavior is unchanged.
* No documentation has been written for the new functionality.
* There are some TODO's which will require closer inspection for a final implementation and some dedicated tests.
* The new option should probably be marked as `@ExperimentalSerializationApi`, but it is not currently.

I would obviously add those things, if there is a chance this would be pulled.

The implementation is rather simple: Introduce a boolean option and in and around the lexer prevent the use of quoted strings, where they wouldn't be necessary unless coercion is wanted. In some places the necessary functionality was already in place as separate methods (decodeBoolean vs decodeBooleanLenient), in others I was able to add minimal changes, usually as part of existing checks around string quotes. I tried to keep the interfaces stable.

Related issues:

* https://github.com/Kotlin/kotlinx.serialization/issues/1042
* https://github.com/Kotlin/kotlinx.serialization/issues/2696

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
@@ -438,14 +438,18 @@ internal abstract class AbstractJsonLexer {
     }
 
     // Allows consuming unquoted string
-    fun consumeStringLenient(): String {
+    fun consumeStringLenient(): String =
+        consumeOther(allowQuoted = true)
+
+    fun consumeOther(allowQuoted: Boolean): String {
+        // TODO this string peeking stuff might break things...
         if (peekedString != null) {
             return takePeeked()
         }
         var current = skipWhitespaces()
         if (current >= source.length || current == -1) fail("EOF", current)
         val token = charToTokenClass(source[current])
-        if (token == TC_STRING) {
+        if (allowQuoted && token == TC_STRING) {
             return consumeString()
         }
 
@@ -587,7 +591,10 @@ internal abstract class AbstractJsonLexer {
         throw JsonDecodingException(position, message + " at path: " + path.getPath() + hintMessage, source)
     }
 
-    fun consumeNumericLiteral(): Long {
+    fun consumeNumericLiteral(): Long =
+        consumeNumericLiteral(coercePrimitives = true)
+
+    fun consumeNumericLiteral(coercePrimitives: Boolean): Long {
         /*
          * This is an optimized (~40% for numbers) version of consumeString().toLong()
          * that doesn't allocate and also doesn't support any radix but 10
@@ -595,7 +602,7 @@ internal abstract class AbstractJsonLexer {
         var current = skipWhitespaces()
         current = prefetchOrEof(current)
         if (current >= source.length || current == -1) fail("EOF")
-        val hasQuotation = if (source[current] == STRING) {
+        val hasQuotation = if (coercePrimitives && source[current] == STRING) {
             // Check it again
             // not sure if should call ensureHaveChars() because threshold is far greater than chars count in MAX_LONG
             if (++current == source.length) fail("EOF")
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -658,6 +658,8 @@ public class JsonBuilder internal constructor(json: Json) {
      */
     public var useArrayPolymorphism: Boolean = json.configuration.useArrayPolymorphism
 
+    public var allowPrimitiveCoercion: Boolean = json.configuration.allowPrimitiveCoercion
+
     /**
      * Module with contextual and polymorphic serializers to be used in the resulting [Json] instance.
      *
@@ -695,7 +697,8 @@ public class JsonBuilder internal constructor(json: Json) {
             allowStructuredMapKeys, prettyPrint, explicitNulls, prettyPrintIndent,
             coerceInputValues, useArrayPolymorphism,
             classDiscriminator, allowSpecialFloatingPointValues, useAlternativeNames,
-            namingStrategy, decodeEnumsCaseInsensitive, allowTrailingComma, allowComments, classDiscriminatorMode
+            namingStrategy, decodeEnumsCaseInsensitive, allowTrailingComma, allowComments, classDiscriminatorMode,
+            allowPrimitiveCoercion
         )
     }
 }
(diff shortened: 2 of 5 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -438,14 +438,18 @@ internal abstract class AbstractJsonLexer {
     }
 
     // Allows consuming unquoted string
-    fun consumeStringLenient(): String {
+    fun consumeStringLenient(): String =
+        consumeOther(allowQuoted = true)
+
+    fun consumeOther(allowQuoted: Boolean): String {
+        // TODO this string peeking stuff might break things...
```
pschichtel, line 445: Looking at the code around `peekedString` I'm pretty sure this is not actually a problem, because this branch would only ever be hit when attempting to read a string. Right?

pdvrieze:
Looks like a good idea to me

pschichtel:
I just rebased this onto latest dev. @sandwwraith @Tapchicoma Is there any chance to get feedback on the "does this have a chance of getting pulled?" question? I've been using this in production in a cloud service, so it's seen quite a bit of traffic without any issues so far.

pschichtel:
I've rebased this again. @sandwwraith @fzhinkin Have I completely missed some process here? _Any_ feedback would be appreciated.

sandwwraith:
Hi! Sorry, it took me a long time to check out your proposal.

In general, I like the idea of adding such a setting. It would indeed help with stricter requirements and migrating between frameworks, such as Jackson.

However, there are several key points that have to be addressed: 

- Conversion from `JsonElement`. You can take a look at `AbstractJsonTreeDecoder` to see how it's done. I think it would be straightforward to add it there, since we have `JsonPrimitive.isString` property.
- What to do with things like `Map<Int, Int>` or `Map<String, Boolean>`. (see also #2749 for reference). In Json, keys can only be strings. So now both `{"42": "42"}` and `{"42": 42}` could be mapped to the `Map<Int, Int>`. I'm not sure whether this flag should prohibit the former. And it probably shouldn't be blocking key parsing at all. What do you think?
- Since changes involve a Json parser, I would say this is very desirable not only to write tests, but also benchmarks for the new feature. You can take `TwitterFeedBenchmark` as an example.

What do you think of this? If you want to continue working, I'd happily review your contribution.

pdvrieze:
The choice of how to handle/serialize maps with non-string keys is ultimately a choice for the format to make (as there is no "native" mapping). The current behaviour to use quoted strings is probably the most elegant (although a strict parameter could prohibit it altogether).

Fundamentally I think that the behaviour of map serialization with non-string keys is a separate one (where configuration may specify an alternative approach as for complex keys). The case for non-string primitive keys is that (to support them at all) the format maps them to strings and serializes/deserializes them as strings.

For map values, they should follow the coercion rule (`Map<Int, Int>` would only allow `{"42": 42 }`

pschichtel:
@sandwwraith thanks for your feedback. Yes I'm still interested in working on this.

> Conversion from JsonElement. You can take a look at AbstractJsonTreeDecoder to see how it's done. I think it would be straightforward to add it there, since we have JsonPrimitive.isString property.

What issue are you referring to? Decoding from JsonElement already works with my changes, doesn't it?

> to do with things like `Map<Int, Int>` or `Map<String, Boolean>`. (see also [Quoted booleans are being allowed even with `isLenient = false` #2749](https://github.com/Kotlin/kotlinx.serialization/issues/2749) for reference). In Json, keys can only be strings. So now both `{"42": "42"}` and `{"42": 42}` could be mapped to the `Map<Int, Int>`. I'm not sure whether this flag should prohibit the former. And it probably shouldn't be blocking key parsing at all. What do you think?

May intention wasn't to touch anything regarding that behavior in this PR. as @pdvrieze suggests I think this is a separate topic for a separate PR. My intention here was merely to make primitive decoding, where it happens, more strict. I think it's best to go in smaller steps here.

> changes involve a Json parser, I would say this is very desirable not only to write tests, but also benchmarks for the new feature. You can take `TwitterFeedBenchmark` as an example.

True! I'll have a look at that.

I have a few lose ends to tie up before I can work on this again, but I definitely will. We currently use this via a fork in production, so there is a strong interest in getting this upstream :)

sandwwraith:
> What issue are you referring to? Decoding from JsonElement already works with my changes, doesn't it?

You're right, I overlooked your changes there.

> May intention wasn't to touch anything regarding that behavior in this PR.

Well, since code path for `Map` is the same, you're touching it in any case. In particular, this code throws a decoding exception now:

```kotlin
@Test
fun testStrictParsing() {
    val j = Json {
        allowPrimitiveCoercion = false
    }
    val input = """{"42": 42}"""
    println(j.decodeFromString<Map<Int, Int>>( input))
}
```

I do not mind this while the flag is experimental, but we would need to make a decision whether this behavior is desirable or if we should allow such deserialization. But I agree, it is possible to do it later.

> I have a few loose ends to tie up before I can work on this again, but I definitely will. 

Glad to hear that!

Gama11:
> Well, since code path for Map is the same, you're touching it in any case. In particular, this code throws a decoding exception now

We ran into one such case in our codebase (actually it was a value class that wraps an int) with this branch and solved it  by implementing a dedicated "map key serializer":

```kotlin
@Serializable
@JvmInline
value class IntWrapper(val value: Int)

private object IntWrapperKeySerializer : KSerializer<IntWrapper> {
    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("IntWrapperMapKey", PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: IntWrapper) = encoder.encodeString(value.value.toString())
    override fun deserialize(decoder: Decoder) = IntWrapper(decoder.decodeString().toInt())
}

typealias IntWrapperStringMap = Map<
    @Serializable(with = IntWrapperKeySerializer::class)
    IntWrapper,
    String,
    >

@Serializable
data class Foo(
    val map: IntWrapperStringMap
)
```

Coming from Jackson, this didn't seem too surprising, [since you also need dedicated map key (de)serializers there](https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/deser/KeyDeserializers.html).

pschichtel:
@sandwwraith Ahh now I understand what you meant. Honestly I think it's the only logical consequence of enabling this flag, that keys must be decoded as strings, if necessary by using custom deserializer. With the great support for value classes this almost feels like a non-issue honestly. We should probably flesh out the documentation on the option to explicitly state this property, but I would argue that people looking for such a strict configuration are expecting this anyway.

Issue #2966, opened 2025-04-01
Q: How to not write a CBOR tag when value is null?


travishaagen:
For CBOR serialization, I assign a tag as so,

```
@Serializable
data class Something(
    @ValueTags(MY_TAG) @Contextual val myObject: MyObject?
)
```

and I have a serializer like

```
object MyObject AsByteArraySerializer : KSerializer<MyObject> {
    override val descriptor: SerialDescriptor
        get() = SerialDescriptor("MyObject", ByteArraySerializer().descriptor)

    override fun deserialize(decoder: Decoder): MyObject = MyObject(decoder.decodeSerializableValue(ByteArraySerializer()))

    override fun serialize(
        encoder: Encoder,
        value: MyObject,
    ) = encoder.encodeSerializableValue(ByteArraySerializer(), value.toBytes())
}
```

It works fine for non-null values, but when null, I don't want the tag written to the byte stream. I need to do this to be in compliance with another project's CBOR protocol expectations.

For example, right now, when the value is null, I'm writing a tag,

```
      70726576                          # "prev"
   D8 2A                                # tag(42)
      F6                                # primitive(22)
```

but I need it not to write the tag

```
      70726576                          # "prev"
   F6                                   # primitive(22)
```

what can I do to control this?

pdvrieze:
@travishaagen Writing of null values is managed by the container of the value (`Something`), not by the serializer of the value. If you specify `null` as the default that will likely work. There may be some other configuration options for cbor to use.

travishaagen:
@pdvrieze thanks for the reply, but setting the default to `null` didn't change the outcome.

@ardune  @JesusMcCloud @sandwwraith 

^ see that you folks have discussed CBOR and `null` handling in the past (e.g., https://github.com/Kotlin/kotlinx.serialization/pull/2952, https://github.com/Kotlin/kotlinx.serialization/issues/2848 )

do you have any insights on making CBOR tags optional with this library?

ardune:
@travishaagen
I'll preface my response with: the following is a bad approach and smarter people should give you a better one.

I personally use a custom build that exposes functions I need.
Specific example:
https://github.com/ardune/kotlinx.serialization/commit/50c9a263f87e9a506da9df3a27102b667c480ec5

Look at "TaggedByteStringSerializer" - it directly encodes and checks for tags

I did this approach as I wanted to be strict about tags in some cases and more lenient in others.

[...]

PR #3074, opened 2025-09-06
CBOR untaggedNullValueTags configuration option (#2966)

travishaagen:
Introduces a new CBOR configuration option called `untaggedNullValueTags`. When enabled, CBOR value-tags will not be serialized for `null` values, and deserialization will not fail validation. This is allowed by the CBOR spec as mentioned in this [issue](https://github.com/Kotlin/kotlinx.serialization/issues/2966#issuecomment-2800034902).

When `untaggedNullValueTags=true`, the encoding for the provided unit test is [https://cbor.me/?bytes=bf6161a0d8226162f6d8386163f66164f6ff](https://cbor.me/?bytes=bf6161a0d8226162f6d8386163f66164f6ff) where the tags shown below are key-tags,

```
BF          # map(*)
   61       # text(1)
      61    # "a"
   A0       # map(0)
   D8 22    # tag(34)
      61    # text(1)
         62 # "b"
   F6       # primitive(22)
   D8 38    # tag(56)
      61    # text(1)
         63 # "c"
   F6       # primitive(22)
   61       # text(1)
      64    # "d"
   F6       # primitive(22)
   FF       # primitive(*)
```

For comparison, when `untaggedNullValueTags=false`, the encoding is [https://cbor.me/?bytes=bf6161cca0d8226162f6d8386163d84ef66164d85accf6ff](https://cbor.me/?bytes=bf6161cca0d8226162f6d8386163d84ef66164d85accf6ff)

```
BF          # map(*)
   61       # text(1)
      61    # "a"
   CC       # tag(12)
      A0    # map(0)
   D8 22    # tag(34)
      61    # text(1)
         62 # "b"
   F6       # primitive(22)
   D8 38    # tag(56)
      61    # text(1)
         63 # "c"
   D8 4E    # tag(78)
      F6    # primitive(22)
   61       # text(1)
      64    # "d"
   D8 5A    # tag(90)
      CC    # tag(12)
         F6 # primitive(22)
   FF       # primitive(*)
```

**Note** however, that Kotlinx Serialization cannot currently deserialize the above anyhow.

This change is needed for a real world use-case, where "untagged `null` values" are a requirement for value-tags. Thank you!


travishaagen:
@fzhinkin Hello! Thanks for taking a look. How can I help move this forward?

fzhinkin:
The change makes sense.

I would only consider a different name for configuration option: all binary flags use a verb in their name, and I don't see why should we make an exception here. Perhaps, `skipValueTagsForNulls`?

Note that configuration options tend to get a companion-annotation to allow a more fine-grained control over behavior. And while an annotation was not proposed here yet (and I would definitely like to postpone its introduction until it is really necessary), it's worth thinking about how the annotation would look like / what name it would have, to use a coherent option name here. One option here would be just adding an additional boolean parameter to `@ValueTags`, and then `encodeValueTags` could be eventually replaced with something like `valueTagsEncoding: EncodingStrategy = Always|Never|SkipForNulls`.

Issue #3045, opened 2025-07-14
Encoding value classes fails in Protocol Buffers


ScottPeterJohnson:
**Describe the bug**
Attempting to encode value classes with default generated serializers fails using Protocol Buffers.

**To Reproduce**
`
fun main(){
    ProtoBuf {}.encodeToByteArray(Example(1))
}
@Serializable
@JvmInline
value class Example(val raw : Int)
`

**Expected behavior**
Encodes correctly as the base type like e.g. json does.
Currently gives:
Exception in thread "main" kotlinx.serialization.SerializationException: No tag in stack for requested element
	at kotlinx.serialization.protobuf.internal.ProtobufTaggedBase.popTag(ProtobufTaggedBase.kt:55)
	at kotlinx.serialization.protobuf.internal.ProtobufTaggedEncoder.encodeInline(ProtobufTaggedEncoder.kt:178)

Possibly related:
https://github.com/Kotlin/kotlinx.serialization/issues/1774

**Workaround:**
Write a custom serializer:
`fun main(){
    ProtoBuf {}.encodeToByteArray(Example(1))
}

@JvmInline
@Serializable(with = ExampleSerializer::class)
value class Example(val raw : Int)
object ExampleSerializer : KSerializer<Example> {
    override val descriptor: SerialDescriptor
        get() = Int.serializer().descriptor
    override fun serialize(
        encoder: Encoder,
        value: Example
    ) {
        encoder.encodeInt(value.raw)
    }

    override fun deserialize(decoder: Decoder): Example {
        return Example(decoder.decodeInt())
    }
}`

**Environment**
 - Kotlin version: 2.2.0
 - Library version: 1.9.0
 - Kotlin platforms: JVM
 - Gradle version: 8

fzhinkin:
I reproduced the issue. Our protobuf encoder is not capable of serializing inline value classes as top-level objects, only as other objects fields. Which is definitely a bug.

fzhinkin:
Seems like it won't be a simple fix as I initially thought:
- we need to figure out how to deal with `Proto<Something>`-annotated fields of inline value classes
- the same as above, but for the case of nested inline values classes  
- potentially, there might be the same issues for inline values classes that are fields of regular classes, we have to figure out something there too
- Kotlin might start supporting multi-field value classes at some point, it's worth considering them when addressing single-field value classes issue

PR #3079, opened 2025-09-12
Fix ProtoBuf packing for kotlin unsigned types

KosmX:
ProtoBuf's `ProtoPacked` annotation only worked for the original signed primitives and ByteArray, and did not work for unsigned variants.

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
@@ -68,11 +68,11 @@ internal val ProtoDesc.integerType: ProtoIntegerType
 }
 
 internal val SerialDescriptor.isPackable: Boolean
-    get() = when (kind) {
-        PrimitiveKind.STRING,
-        !is PrimitiveKind -> false
-        else -> true
-    } || isInline && elementsCount == 1 && getElementDescriptor(0).isPackable
+    get() = when {
+        isInline -> elementsCount == 1 && getElementDescriptor(0).isPackable
+        kind is PrimitiveKind -> kind != PrimitiveKind.STRING
+        else -> false
+    }
 
 internal val ProtoDesc.isPacked: Boolean
     get() = (this and PACKEDMASK) != 0L
--- a/formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/PackedArraySerializerTest.kt
+++ b/formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/PackedArraySerializerTest.kt
@@ -205,6 +205,7 @@ class PackedArraySerializerTest {
         assertEquals(expectedByteCarrier, decodedBytes)
         assertEquals(expectedIntCarrier, decodedInts)
     }
+
     @Test
     fun testEncodePackedUnsigned() {
         val byteCarrier = PackedUByteCarrier(ubyteArrayOf(1.toUByte(), 2.toUByte(), 128.toUByte()))


fzhinkin:
@KosmX, thanks for opening the PR!

Related issue: https://github.com/Kotlin/kotlinx.serialization/issues/3045

fzhinkin:
Before moving further with this PR (for the part where inline value classes support is fixed) we need to address the questions from https://github.com/Kotlin/kotlinx.serialization/issues/3045#issuecomment-3211256099

KosmX:
Okay, I'm not a Kotlin maintainer or anything, but here are my thoughts:
I'll start with the last, that's probably the easiest or most important
> Kotlin might start supporting multi-field value classes at some point, it's worth considering them when addressing single-field value classes issue

As long as value classes use their default serializer, I think, they should be inlined completely:  
For example I have a `value class Complex(val real: Double, val imaginary: Double)`, I want it to be extremely compact in a huge array.
And that's the same for serialization, especially for binary formats like protobuf.  

And, if for some reason, I do not want to serialize the value class this way, I can still go the long way, and write a custom serializer, or maybe we can add an annotation for serializable value classes (like DontInlineSerialize)

If we want to match standards/specifications exactly, don't bend protobuf's rules, maybe do the inlining the opposite way: inline only for unsigned primitives, and for explicit SerializeInline annotation.

Currently we try to inline things in all formats.

> potentially, there might be the same issues for inline values classes that are fields of regular classes, we have to figure out something there too

The way I edited the `isPackable` getter, it is recursive on nested value classes. I should write a test-case for that.

> we need to figure out how to deal with Proto<Something>-annotated fields of inline value classes

If the value class is inlined, I would look for these annotations recursively, from the deepest. 

On the implementation question, I have some ideas, but nothing precise-enough, and I think it would be better to have the complete specification first, maybe involve more people, ask what their thoughts are.

sandwwraith:
Hi @KosmX! I think we can merge this without solving all the issues mentioned. Do you want to continue working with this PR? If so, I'll submit my portion of comments

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
```diff
@@ -68,12 +68,11 @@ internal val ProtoDesc.integerType: ProtoIntegerType
 }
 
 internal val SerialDescriptor.isPackable: Boolean
-    @OptIn(kotlinx.serialization.ExperimentalSerializationApi::class)
     get() = when (kind) {
         PrimitiveKind.STRING,
         !is PrimitiveKind -> false
         else -> true
-    }
+    } || isInline && elementsCount == 1 && getElementDescriptor(0).isPackable
```
sandwwraith: IMO it would be nicer to rewrite it as one `when`:
```kotlin
when {
 isInline -> elementsCount == 1 && getElementDescriptor(0).isPackable
 kind is PrimitiveKind -> kind != PrimitiveKind.STRING
 else -> false
}
```

Also, we have a similar function to 'unwrap' inline class descriptors in the Json module: https://github.com/Kotlin/kotlinx.serialization/blob/a33ef021ae479949a7663cd82cfbc32026a2c9de/formats/json/commonMain/src/kotlinx/serialization/json/internal/WriteMode.kt#L48

I wonder whether it makes sense to copy it here as well. Doing so, we will also have support for unwrapping `SerialKind.CONTEXTUAL` descriptors. Do you think it is needed?

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/PackedArraySerializerTest.kt
```diff
@@ -164,4 +191,33 @@ class PackedArraySerializerTest {
...
+
+        val byteHex = """0a03010280"""
+        val intHex = """0a050102038001"""
+
+        val decodedBytes = ProtoBuf.decodeFromHexString<PackedUByteCarrier>(byteHex)
+        val decodedInts = ProtoBuf.decodeFromHexString<PackedUintCarrier>(intHex)
+
+        assertEquals(expectedByteCarrier, decodedBytes)
+        assertEquals(expectedIntCarrier, decodedInts)
+    }
```
sandwwraith, line 207: Empty line is missing before next test start

formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/internal/Helpers.kt
KosmX: I don't want to have extra features in this PR, it is just a fix.

PR #3081, opened 2025-09-18
Migrate Shadow's coordinate

Goooler:
I didn't update Shadow to v9 as it requires min Gradle 8.11

https://github.com/GradleUp/shadow/releases/tag/8.3.0

### DIFF
--- a/benchmark/build.gradle.kts
+++ b/benchmark/build.gradle.kts
@@ -29,6 +29,10 @@ tasks.processJmhResources {
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
 
+tasks.shadowJar {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+}
+
 tasks.jmhJar {
     archiveBaseName.set("benchmarks")
     archiveVersion.set("")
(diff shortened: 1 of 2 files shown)


benchmark/build.gradle.kts
```diff
@@ -29,6 +29,10 @@ tasks.processJmhResources {
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
 
+tasks.shadowJar {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+}
```
Goooler, line 34: I suggested this 

https://github.com/GradleUp/shadow/releases/tag/9.0.1

shanshin:
> I didn't update Shadow to v9 as it requires min Gradle 8.11

@sandwwraith, I can upgrade the Gradle version to the latest 8.x or even 9.x (if it doesn't cause build breakage)

benchmark/build.gradle.kts
sandwwraith: Changelog says: "Default duplicatesStrategy back to EXCLUDE. This strategy is consistent with 8.x series behavior, which is more compatible for most users upgrading."
So the default strategy is still EXCLUDE, you just want to write it explicitly?

Goooler: Yeah.

sandwwraith:
Big thanks!

PR #3082, opened 2025-09-18
Doc rewrite update 3 - Serialize built in types

daniCsorbaJB:
This is the third part of the Kotlin Serialization rewrite.

Related YouTract ticket is: [KT-80887](https://youtrack.jetbrains.com/issue/KT-80887) [Docs] Create a serialize built-in classes guide for kotlinx.serialization

### DIFF
--- a/docs-website/topics/serialization-serialize-builtin-types.md
+++ b/docs-website/topics/serialization-serialize-builtin-types.md
@@ -1,0 +1,378 @@
+[//]: # (title: Serialize built-in types)
+
+The Kotlin serialization library supports various built-in types, including basic types such as primitives and strings, composite types, and several standard library classes.
+The following sections describe these types in detail and provide examples of how to serialize them.
+
+## Basic types
+
+Kotlin serialization provides built-in serializers for types that are represented as a single value in serialized data.
+This includes, primitives, strings, and enums.
+
+For example, here’s how you can serialize a `Long` type:
+
+```kotlin
+// Imports the necessary library declarations
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Data(val signature: Long)
+
+fun main() {
+    val data = Data(0x1CAFE2FEED0BABE0)
+    println(Json.encodeToString(data))
+    // {"signature":2067120338512882656}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Numbers
+
+You can serialize all Kotlin number types, including integers and floating-point numbers, using their natural JSON representations:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import kotlin.math.PI
+
+//sampleStart
+@Serializable
+class Data(
+    val answer: Int,
+    val pi: Double
+)                     
+
+fun main() {
+    val data = Data(42, PI)
+    println(Json.encodeToString(data))
+    // {"answer":42,"pi":3.141592653589793}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Long numbers as strings
+
+When you serialize Kotlin `Long` values to JSON, JavaScript's native number type can't represent the full range of a Kotlin `Long` type,
+leading to precision loss.
+
+Kotlin/JS handles these large `Long` numbers correctly, but JavaScript's native methods don't.
+A common workaround is to represent long numbers with full precision using the JSON string type.
+Kotlin Serialization supports this approach with [`LongAsStringSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.builtins/-long-as-string-serializer/).
+To apply it to a `Long` property, use the `@Serializable` annotation:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Data(
+    @Serializable(with=LongAsStringSerializer::class)
+    val signature: Long
+)
+
+fun main() {
+    val data = Data(0x1CAFE2FEED0BABE0)
+    println(Json.encodeToString(data))
+    // {"signature":"2067120338512882656"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> You can also specify serializers like `LongAsStringSerializer` for all properties in a file.
+> For more information, see [Specify serializers for a file](third-party-classes.md#specify-serializers-for-a-file).
+>
+{style="tip"}
+
+### Enum classes
+
+All `enum` classes are serializable by default without the `@Serializable` annotation.
+When serialized in JSON, an `enum` is encoded as a string:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+// The @Serializable annotation isn't required for enum classes
+enum class Status { SUPPORTED }
+
+@Serializable
+class Project(val name: String, val status: Status) 
+
+fun main() {
+    val data = Project("kotlinx.serialization", Status.SUPPORTED)
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","status":"SUPPORTED"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> On Kotlin/JS and Kotlin/Native, you must use the `@Serializable` annotation for an `enum` class to use as a root object,
+> such as in `encodeToString<Status>(Status.SUPPORTED)`.
+>
+{style="note"}
+
+#### Customize serial names of enum entries
+
+> For more information on customizing serial names, see [Customize serial names](serialization-customization-options.md#customize-serial-names).
+>
+{style="tip"}
+
+To customize the serial names of enum entries, use the `@SerialName` annotation and mark the enum class with `@Serializable`:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+// Requires the @Serializable annotation because of @SerialName
+@Serializable
+enum class Status { @SerialName("maintained") SUPPORTED }
+
+@Serializable
+class Project(val name: String, val status: Status) 
+
+fun main() {
+    val data = Project("kotlinx.serialization", Status.SUPPORTED)
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","status":"maintained"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## Composite types
+
+Kotlin Serialization supports several composite types from the standard library, but some classes,
+such as ranges and the [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) class aren't supported yet.
+
+### Pair and triple
+
+You can serialize the [`Pair`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/) and [`Triple`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-triple/) classes from the Kotlin standard library:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Project(val name: String)
+
+fun main() {
+    val pair = 1 to Project("kotlinx.serialization")
+    println(Json.encodeToString(pair))
+    // {"first":1,"second":{"name":"kotlinx.serialization"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Collections
+
+Kotlin Serialization supports collection types such as [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), [`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/), and [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/).
+Lists and sets are serialized as JSON arrays, and maps are represented as JSON objects.
+
+#### Serialize lists
+
+Kotlin Serialization serializes [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/) types as JSON arrays.
+Here’s an example with a list of classes:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Project(val name: String)
+
+fun main() {
+    val list = listOf(
+        Project("kotlinx.serialization"),
+        Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(list))
+    // [{"name":"kotlinx.serialization"},{"name":"kotlinx.coroutines"}]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+#### Serialize sets
+
+[`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/) types are serialized as JSON arrays, just like [`List` types](#serialize-lists):
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Project(val name: String)
+
+fun main() {
+    val set = setOf(
+        Project("kotlinx.serialization"),
+        Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(set))
+    // [{"name":"kotlinx.serialization"},{"name":"kotlinx.coroutines"}]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+#### Serialize maps
+
+Kotlin serialization supports [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/) types with primitive or enum keys:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Project(val name: String)
+
+fun main() {
+    // Creates a map with Int keys
+    val map = mapOf(
+        1 to Project("kotlinx.serialization"),
+        2 to Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(map))
+    // {"1":{"name":"kotlinx.serialization"},"2":{"name":"kotlinx.coroutines"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+
+> JSON doesn't natively support complex or composite keys.
+> To encode structured objects as map keys, see [Encode structured map keys](serialization-json-configuration.md#encode-structured-map-keys).
+>
+{style="note"}
+
+#### Deserialization behavior of collections
+
+Kotlin uses the declared type to deserialize JSON.
+For example, a `List` preserves duplicates, while a `Set` enforces uniqueness:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+data class Data(
+    val a: List<Int>,
+    val b: Set<Int>
+)
+
+fun main() {
+    val data = Json.decodeFromString<Data>("""
+        {
+            "a": [42, 42],
+            "b": [42, 42]
+        }
+    """)
+    // Duplicates are removed from data.b because Set enforces unique elements
+    println(data)
+    // Data(a=[42, 42], b=[42])
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> For more information about collections in Kotlin, see [Collections overview](collections-overview.md).
+>
+{style="tip"}
+
+## Unit and singleton objects
+
+The Kotlin [`Unit`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/) type and other singleton objects are serializable.
+A [singleton](object-declarations.md) is a class with only one instance, where the state is defined by the object itself rather than by external properties.
+In JSON, singleton objects are serialized as empty structures:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+object SerializationVersion {
+    val libraryVersion: String = "1.0.0"
+}
+
+fun main() {
+    println(Json.encodeToString(SerializationVersion))
+    // {}
+    println(Json.encodeToString(Unit))
+    // {}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> You can use serialized singleton objects in [closed polymorphic hierarchies](serialization-polymorphism.md#serialize-objects-in-sealed-hierarchies)
+> to represent cases without additional fields.
+>
+{style="tip"}
+
+## Duration
+
+Kotlin's [`Duration`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/) class is serialized to a JSON string using the ISO-8601-2 format:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import kotlin.time.*
+
+//sampleStart
+fun main() {
+    val duration = 1000.toDuration(DurationUnit.SECONDS)
+    println(Json.encodeToString(duration))
+    // "PT16M40S"
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## Nothing
+
+The [`Nothing`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-nothing.html) type is serializable by default.
+It has no instances, so encoding or decoding it throws an exception.
+Use `Nothing` when a type is syntactically required, but not involved in serialization, like in [polymorphic classes with generic base types](serialization-polymorphism.md#serialize-polymorphic-types-with-generic-base-types):
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+sealed class ParametrizedParent<out R> {
+    @Serializable
+    data class ChildWithoutParameter(val value: Int) : ParametrizedParent<Nothing>()
+}
+
+fun main() {
+    println(Json.encodeToString(ParametrizedParent.ChildWithoutParameter(42)))
+    // {"value":42}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## What's next
+
+* Dive into the [Serialize classes](serialization-customization-options.md) section to learn how to serialize classes and how to modify the default behavior of the `@Serializable` annotation.
+* To explore more complex JSON serialization scenarios, see [JSON serialization overview](configure-json-serialization.md).
+* Learn more about polymorphism and serializing different types through a shared base in [Serialize polymorphic classes](serialization-polymorphism.md).


docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+    // {"answer":42,"pi":3.141592653589793}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Long numbers as strings
+
+When you serialize Kotlin `Long` values to JSON, JavaScript's native number type can't represent the full range of a Kotlin `Long` type,
+leading to precision loss.
```
pdvrieze: Is it worthwhile to note that this is a limitation of JS, not Json that is effectively arbitrary precision? It is somewhat implied by the explanation below, but can be confusing in the distinction.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+    // {"first":1,"second":{"name":"kotlinx.serialization"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Collections
+
+Kotlin Serialization supports collection types such as [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), [`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/), and [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/).
+Lists and sets are serialized as JSON arrays, and maps are represented as JSON objects.
```
pdvrieze: Maybe explicitly mention that this is specific to the Json format. What about:
The way lists and sets are serialized is format specific, but generally mapped to common structures. For example, in Json lists and sets are serialized as ...

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+
+> JSON doesn't natively support complex or composite keys.
+> To encode structured objects as map keys, see [Encode structured map keys](serialization-json-configuration.md#encode-structured-map-keys).
+>
+{style="note"}
```
pdvrieze, line 298: Maybe mention that other formats may support this in different ways.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+{kotlin-runnable="true"}
+
+> You can use serialized singleton objects in [closed polymorphic hierarchies](serialization-polymorphism.md#serialize-objects-in-sealed-hierarchies)
+> to represent cases without additional fields.
+>
+{style="tip"}
+
+## Duration
+
+Kotlin's [`Duration`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/) class is serialized to a JSON string using the ISO-8601-2 format:
```
pdvrieze: IIRC, this is not json specific.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
+[//]: # (title: Serialize built-in types)
+
+The Kotlin serialization library supports various built-in types, including basic types such as primitives and strings, composite types, and several standard library classes.
```
sandwwraith: You've mentioned 'composite types' without explaining what they are. I suggest just leaving 'basic types such as primitives and strings, as well as certain standard library classes'

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+The following sections describe these types in detail and provide examples of how to serialize them.
+
+## Basic types
+
+Kotlin serialization provides built-in serializers for types that are represented as a single value in serialized data.
+This includes, primitives, strings, and enums.
```
sandwwraith: Unnecessary `,` after 'includes' ?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Data(
+    @Serializable(with=LongAsStringSerializer::class)
```
sandwwraith: I'd prefer removing explicit `with`: `@Serializable(LongAsStringSerializer::class)`

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> You can also specify serializers like `LongAsStringSerializer` for all properties in a file.
+> For more information, see [Specify serializers for a file](third-party-classes.md#specify-serializers-for-a-file).
+>
+{style="tip"}
+
```
sandwwraith, line 115: I think it might worth it to add section aboun unsigned numbers here as well, given they're stable in stdlib and IR compiler has been default for years (https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/value-classes.md#unsigned-types-support-json-only). Just don't forget mentioning 
> Unsigned types are output as unsigned only in JSON format. Other formats such as ProtoBuf and CBOR use built-in serializers that use an underlying signed representation for unsigned types.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+fun main() {
+    val data = Project("kotlinx.serialization", Status.SUPPORTED)
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","status":"maintained"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## Composite types
```
sandwwraith: I would say we don't need this 'composite types' notion until maybe the chapter about custom serializers. So you can replace this with 'Standard library types'

Alternatively, you should start with the definition of composite types: "Every type which is composed out of the basic types dicussed above, is called composite".
Besides, one can always write a serializer that represents any complex type as string. So I'm in favor of removing 'composite types' notion for now.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+
+Kotlin Serialization supports several composite types from the standard library, but some classes,
+such as ranges and the [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) class aren't supported yet.
```
sandwwraith: I wouldn't say "yet" because we do not have any plans to add the support for them. Maybe saying "but not all, such as ranges and Regex" is sufficient.

docs-website/topics/serialization-serialize-builtin-types.md
sandwwraith: +1

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+    // {"1":{"name":"kotlinx.serialization"},"2":{"name":"kotlinx.coroutines"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+
+> JSON doesn't natively support complex or composite keys.
+> To encode structured objects as map keys, see [Encode structured map keys](serialization-json-configuration.md#encode-structured-map-keys).
```
sandwwraith, line 296: You can also add 
> Other formats may represent Maps with object keys in a more natural ways.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+    println(Json.encodeToString(pair))
+    // {"first":1,"second":{"name":"kotlinx.serialization"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Collections
+
+Kotlin Serialization supports collection types such as [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), [`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/), and [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/).
```
sandwwraith: I would rephrase this slightly. The thing is, we also support concrete implementations (e.g. `ArrayList` and `LinkedHashSet`), so our users should not have impression that only basic intrefaces are supported. Mutable counterparts should be mentioned as well.

And btw, regular `Array`s and primitive `Int/Long/...Array` too.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+fun main() {
+    val set = setOf(
+        Project("kotlinx.serialization"),
+        Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(set))
+    // [{"name":"kotlinx.serialization"},{"name":"kotlinx.coroutines"}]
+}
+//sampleEnd
+```
```
sandwwraith, line 257: Since Sets have their own section, maybe it is worth mentioning that

> Duplicate entries in Sets do not cause exceptions on deserialization by default and behavior on duplicates is implementation-defined.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+
+> JSON doesn't natively support complex or composite keys.
+> To encode structured objects as map keys, see [Encode structured map keys](serialization-json-configuration.md#encode-structured-map-keys).
+>
+{style="note"}
+
+#### Deserialization behavior of collections
+
+Kotlin uses the declared type to deserialize JSON.
```
sandwwraith, line 302: I think original 
> During deserialization, the type of the resulting object is determined by the static type that was specified in the source code—either as the type of the property or as the type parameter of the decoding function.

explains what happens better. It also answers the question "What happens if I specify ArrayList or LinkedList"

sandwwraith: Although you can mention this in the beginning of the section, add the suggested tip to the Set section and drop this section completely.

docs-website/topics/serialization-serialize-builtin-types.md
sandwwraith: Yes, maybe just say "string" instead of "JSON string"?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,378 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> You can use serialized singleton objects in [closed polymorphic hierarchies](serialization-polymorphism.md#serialize-objects-in-sealed-hierarchies)
+> to represent cases without additional fields.
+>
+{style="tip"}
+
+## Duration
```
sandwwraith: We also have two new serializers for `Instant` btw (https://github.com/Kotlin/kotlinx.serialization/pull/2945). They probably should be mentioned here (and requirement for Kotlin 2.2)

sandwwraith (changes requested):
see comments

docs-website/topics/serialization-serialize-builtin-types.md
pdvrieze: Stating that unsigned is only unsigned in Json is inaccurate, other formats support unsigned directly (and at full range) - maybe something that the protobuf serializer can be extended to do.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: Sounds good to me as well — let's drop the term then 👍

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: Good point - it is a bit confusing - I restructured this part a bit: 

### `Long` numbers as strings

You can represent `Long` numbers as strings in JSON.
This is useful in JavaScript environments, where JavaScript's `Number` type can't precisely represent all Kotlin `Long` values, which may lead to precision loss.

Use [`LongAsStringSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.builtins/-long-as-string-serializer/) with the `@Serializable` annotation to encode `Long` values as strings in JSON:

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: very nice point thank you!  - added this clarification. (I omitted the "for example" part - and just wrote "In JSON, lists and sets are serialized as....

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: the focus is on JSON here — unless we specify how it differs I think it might be just stating the obvious here. (to me it would read like: In JSON it works like this. In other formats it might work differently.) But if we have a very critical difference here, then I think we could list it.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: thank you — dropped it 👍

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: Hmm 🤔 this one feels a bit fluffy to me, could you please clarify?  what do we mean naturally? 
I think we could mention something specific like CBOR? 🤔

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: nice one 💯 I added it to this section, so now it should be 

### Duration and Instant

(added `kotlin-min-compiler-version="2.2"` to the example as well)

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: Good point 👍 

I added it like so: 
> Kotlin Serialization supports collection types, including both immutable and mutable variants of [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), [`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/), and [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/).
It also supports their concrete implementations such as [`ArrayList`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-array-list/) and [`LinkedHashSet`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-linked-hash-set/), as well as generic and primitive array types.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: How about? 

> By default, you can deserialize sets with duplicate entries. The behavior for handling duplicates depends on the specific `Set` implementation.
> 
{style="tip"}

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: 👍 I added a section — please let me know what you think

sandwwraith: @pdvrieze IIRC protobuf encodes `UInt` directly as `Int`, so it is signed in the output

docs-website/topics/serialization-serialize-builtin-types.md
sandwwraith: Kotlin's Map representation is dependent on the format. If a format supports Maps with non-primitive keys, then we can output them as such. CBOR indeed supports maps with object keys and we can use it as the example.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,431 @@
...
+    println(Json.encodeToString(pair))
+    // {"first":1,"second":{"name":"kotlinx.serialization"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Collections
+
+Kotlin Serialization supports collection types, including both immutable and mutable variants of [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), [`Set`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/), and [`Map`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/).
```
sandwwraith: `List` and so are not immutable, they're *read-only*, it is a different term with a different meaning. We use it across our documentation (https://kotlinlang.org/docs/collections-overview.html#collection-types), please pay attention, as it may confuse users.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,431 @@
...
+        Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(set))
+    // [{"name":"kotlinx.serialization"},{"name":"kotlinx.coroutines"}]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> By default, you can deserialize sets with duplicate entries. The behavior for handling duplicates depends on the specific `Set` implementation.
```
sandwwraith: I wouldn't say it depends on `Set` implementation, but rather on Set deserializer and the rest of kotlinx.serialization framework. I believe saying 'implementation-defined' is enough, it should be common and familiar term.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,431 @@
...
+fun main() {
+    val counted = 239.toUByte()
+    println(Json.encodeToString(Counter(counted, "tries")))
+    // {"counted":239,"description":"tries"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> Unsigned numbers are currently only supported in the JSON format.
```
sandwwraith: 'Supported' is not the best choice of words here IMO. They're working in other formats as well, just have different default representation. I would say
"Unsigned numbers have special handling to be represented this way in the JSON format. Other formats may or may not have this special representation. For example, ProtoBuf and CBOR serialize these types using their signed counterparts."

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,431 @@
...
+```
+{kotlin-runnable="true"}
+
+> For more information about collections in Kotlin, see [Collections overview](collections-overview.md).
+>
+{style="tip"}
+
+### Unit and singleton objects
+
+The Kotlin [`Unit`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/) type and other singleton objects are serializable.
```
sandwwraith: Kotlin or Kotlin's ?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,431 @@
...
+fun main() {
+    val duration = 1000.toDuration(DurationUnit.SECONDS)
+    println(Json.encodeToString(duration))
+    // "PT16M40S"
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+Similarly, Kotlin's [`Instant`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-instant/) type
```
sandwwraith: I explicitly mentioned that here should be Kotlin 2.2 requirement, as we do not version our docs and someone may open this section with the old Kotlin and get frustration because nothing is working.

docs-website/topics/serialization-serialize-builtin-types.md
pdvrieze: @sandwwraith Indeed, the current protobuf implementation encodes as Int/signed, but could be updated to encode as unsigned (behind some flag for backwards compatibility). The comment was more on the phrasing somehow making Json out as special. Your suggestion in the earlier comment addresses that.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: we can go with Kotlin's sure 👍

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: you are right — Corrected 👍

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: 🤔 
What do you think about this version? (to make it a bit more concise) 

Although JSON preserves the full range of unsigned numbers, other serialization formats may handle them differently. For example, ProtoBuf and CBOR serialize these types using their signed counterparts.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: perfect — I added it thank you

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -288,7 +288,9 @@ fun main() {
 ```
 {kotlin-runnable="true"}
 
+Map serialization depends on the format.
 In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+Other formats, such as CBOR, support maps with non-primitive or object keys and preserve them as such.
```
sandwwraith: Aren't `non-primitive` and `object` the same? There's no need for `or object`.

docs-website/topics/serialization.md
```diff
@@ -19,6 +19,11 @@ It walks you through adding the Kotlin serialization library to your project and
 Kotlin serialization offers support for all platforms, including JVM, JavaScript, and Native.
 You can use the same [dependency declaration](serialization-get-started.md#add-plugins-and-dependencies-for-kotlin-serialization) regardless of the target platform.
 
+> When you’re not using a specific format library, for example, when you're writing your own serialization format,
```
sandwwraith: I think it's better to move this note slightly down, after line 36 or 39. The `org.jetbrains.kotlinx:kotlinx-serialization-xxx` notion is introduced only on line 35.

pdvrieze: Also, custom formats are an advanced topic that would really better go in its own document.

daniCsorbaJB: This is just the overview, but it has a section in a more advanced page 👍

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
+[//]: # (title: Serialize built-in types)
+
+The Kotlin serialization library supports various built-in types, including basic types such as primitives and strings, as well as certain standard library classes.
```
sarahhaggarty:
```suggestion
The Kotlin serialization library supports a variety of built-in types, including basic types such as primitives and strings, as well as certain standard library classes.
```
This sounds more natural to me but it's up to you :)

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+The following sections describe these types in detail and provide examples of how to serialize them.
```
sarahhaggarty:
```suggestion
The following sections describe these types in detail and show how to serialize them.
```

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+
+## Basic types
+
+Kotlin serialization provides built-in serializers for types that are represented as a single value in serialized data.
+This includes primitives, strings, and enums.
+
+For example, here’s how you can serialize a `Long` type:
```
sarahhaggarty:
```suggestion
For example, here's how you can serialize a `Long` type:
```

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+class Data(val signature: Long)
+
+fun main() {
+    val data = Data(0x1CAFE2FEED0BABE0)
+    println(Json.encodeToString(data))
+    // {"signature":2067120338512882656}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
sarahhaggarty: Can you add an ID?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+    // {"answer":42,"pi":3.141592653589793}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Unsigned numbers
+
+Kotlin serialization supports Kotlin's [unsigned integer types](unsigned-integer-types.md) like `UByte` and `UInt`.
+In JSON, these values are serialized as regular JSON numbers and preserve their full unsigned range:
```
sarahhaggarty, line 59: What do you think about "retain" instead of "preserve"?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+fun main() {
+    val data = Project("kotlinx.serialization", Status.SUPPORTED)
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","status":"SUPPORTED"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> On Kotlin/JS and Kotlin/Native, you must use the `@Serializable` annotation for an `enum` class to use as a root object,
```
sarahhaggarty: Do you mean "In Kotlin/JS and Kotlin/Native applications"? As it is, it's not clear what's being referred to 🤔

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+> such as in `encodeToString<Status>(Status.SUPPORTED)`.
+>
+{style="note"}
+
+#### Customize serial names of enum entries
+
+> For more information on customizing serial names, see [Customize serial names](serialization-customization-options.md#customize-serial-names).
```
sarahhaggarty: I think it makes sense for the content of this tip to be moved to the bottom of the section and then it doesn't need to be  a tip anymore. What do you think?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+In JSON, lists and sets are serialized as JSON arrays, and maps are represented as JSON objects.
+
+Kotlin uses the declared type to deserialize JSON.
+During deserialization, the type of the resulting object is determined by the static type specified in the source code.
+This type can be either the type of the property or the type parameter of the decoding function.
+
+#### Serialize lists
+
+Kotlin serialization serializes [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/) types as JSON arrays.
+Here’s an example with a list of classes:
```
sarahhaggarty:
```suggestion
Here's an example with a list of classes:
```

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+        Project("kotlinx.coroutines")    
+    )
+    println(Json.encodeToString(set))
+    // [{"name":"kotlinx.serialization"},{"name":"kotlinx.coroutines"}]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> By default, you can deserialize sets with duplicate entries. The behavior for handling duplicates is implementation-defined.
```
sarahhaggarty: By "implementation-defined" do you mean that developers need to write some code for that?

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+In JSON, maps are represented as objects. Since JSON object keys are always strings, keys are encoded as strings even if they are numbers in Kotlin.
+Other formats, such as CBOR, support maps with non-primitive keys and preserve them as such.
+
+> JSON doesn't natively support complex or composite keys.
+> To encode structured objects as map keys, see [Encode structured map keys](serialization-json-configuration.md#encode-structured-map-keys).
+>
+{style="note"}
+
+#### Deserialization behavior of collections
+
```
sarahhaggarty, line 301: Missing an intro here :)

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+)
+
+fun main() {
+    val data = Json.decodeFromString<Data>("""
+        {
+            "a": [42, 42],
+            "b": [42, 42]
+        }
+    """)
+    // Duplicates are removed from data.b because Set enforces unique elements
```
sarahhaggarty: "The Set type"? Or "a Set"? It feels like something is missing from this sentence.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+fun main() {
+    val duration = 1000.toDuration(DurationUnit.SECONDS)
+    println(Json.encodeToString(duration))
+    // "PT16M40S"
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+Starting with Kotlin 2.2.0, you can serialize Kotlin's [`Instant`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-instant/) type
```
sarahhaggarty, line 385: It's best not to mention the Kotlin version this is supported in as our docs are supposed to reflect the latest stable release.

docs-website/topics/serialization-serialize-builtin-types.md
```diff
@@ -0,0 +1,432 @@
...
+    println(Json.encodeToString(ParametrizedParent.ChildWithoutParameter(42)))
+    // {"value":42}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## What's next
+
+* Dive into the [Serialize classes](serialization-customization-options.md) section to learn how to serialize classes and how to modify the default behavior of the `@Serializable` annotation.
```
sarahhaggarty: Similar comment to the get started PR about whether this is an actual page or a section within the page.

docs-website/topics/serialization.md
```diff
@@ -33,6 +33,11 @@ For example:
 The `kotlinx.serialization` libraries follow their own versioning, independent of Kotlin.
 You can find the latest release versions on [GitHub](https://github.com/Kotlin/kotlinx.serialization/releases).
 
+> When you’re not using a specific format library, for example, when you're writing your own serialization format,
```
sarahhaggarty:
```suggestion
> When you're not using a specific format library, for example, when you're writing your own serialization format,
```

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: sounds better to me as well 👍 let's go with it it

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: added: id="serialize-long-class"

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: I think I prefer preserve here slightly. 
Retain to me is more like is more retain a state/ continue to exist. s

“preserve” feels closer to what's happening here maintain integrity / fidelity across a transformation.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: hmm 🤔 more like on those targets 🤔 so what do you think about:
When targeting Kotlin/JS or Kotlin/Native, you must use...

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: yea — let's go with the Set type

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: As mentioned, this is an exception and it's only documented here as of now — so for now let's keep it in, but we'll monitor it.

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: good point! 👍 removed it

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: 😱 re-added it — nice catch

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: nice, I like it 👍  — moved it

docs-website/topics/serialization-serialize-builtin-types.md
daniCsorbaJB: yes — it could be controlled but very complex - so removed it as it might just confuse people this early

docs-website/topics/serialization-serialize-builtin-types.md
sarahhaggarty: Works for me!

docs-website/topics/serialization-serialize-builtin-types.md
sandwwraith: No, it means that it is controlled by the library and we do not provide any guarantees. Without this clarification, the sentence "By default, you can deserialize sets with duplicate entries." doesn't answer what exactly happens when duplicate is encountered.

PR #3083, opened 2025-09-18
[DO NOT MERGE] Upgraded Gradle version to 9.0.0

shanshin:


shanshin:
New limitation: minimal JVM version is 17

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -138,10 +138,11 @@ object Java9Modularity {
             source(compileTask.map { it.sources })
             source(compileTask.map { it.javaSources })
             // part of work-around for https://youtrack.jetbrains.com/issue/KT-60541
-            source(compileTask.map {
-                @Suppress("INVISIBLE_MEMBER")
-                it.scriptSources
-            })
+// TODO fix the migration issue
+//            source(compileTask.map {
+//                @Suppress("INVISIBLE_MEMBER")
```
shanshin, line 143: @Tapchicoma, could you please help with fixing this problem?

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -178,11 +179,12 @@ object Java9Modularity {
             @Suppress("DEPRECATION")
             if (taskKotlinLanguageVersion.get() < KotlinVersion.KOTLIN_2_0) {
                 // part of work-around for https://youtrack.jetbrains.com/issue/KT-60541
-                @Suppress("INVISIBLE_MEMBER")
-                commonSourceSet.from(compileTask.map {
-                    @Suppress("INVISIBLE_MEMBER")
-                    it.commonSourceSet
-                })
+// TODO fix the migration issue
+//                @Suppress("INVISIBLE_MEMBER")
```
shanshin, line 183: Same here

bom/build.gradle.kts
```diff
@@ -48,7 +48,8 @@ publishing {
 }
 
 fun DefaultMavenPublication.unsetModuleDescriptorGenerator() {
-    @Suppress("NULL_FOR_NONNULL_TYPE")
-    val generator: TaskProvider<Task?> = null
-    setModuleDescriptorGenerator(generator)
+// TODO fix the migration issue
+//    @Suppress("NULL_FOR_NONNULL_TYPE")
```
shanshin, line 52: And here

sandwwraith:
@Tapchicoma PTAL

gradle/libs.versions.toml
```diff
@@ -1,11 +1,11 @@
 [versions]
 kotlin = "2.2.0"
```
Tapchicoma, line 2: I would suggest first updating this repo to Kotlin 2.2.21, as it contains an important fix for Gradle 9.0.0 support

shanshin: Ok.
However, the problems with Java modularity are not due to Gradle compatibility - they are due to using the outdated KGP API and I'm not aware how to correct them correctly.

And one problem is that `@Suppress("NULL_FOR_NONNULL_TYPE")` does not work in K2 - perhaps there are also ready-made solutions here on how to do this in a different way.

PR #3090, opened 2025-10-15
Fix type in BIGNUM_NEGATIVE tag name

fzhinkin:


formats/cbor/commonMain/src/kotlinx/serialization/cbor/Tags.kt
```diff
@@ -1,7 +1,7 @@
 package kotlinx.serialization.cbor
 
-import kotlinx.serialization.cbor.internal.SuppressAnimalSniffer
 import kotlinx.serialization.*
+import kotlinx.serialization.cbor.internal.*
```
pdvrieze, line 5: This change doesn't seem to be needed, but rather due to different configuration of the import organizing.

sandwwraith: I prefer having star imports though, so let's keep it even it's not directly related

PR #3091, opened 2025-10-17
Deprecation of apple x64 target

mMaxy:
Kotlin is deprecating x64 family of apple targets, due to Apple also stopping the support for them.

See KT-78660 for more details.


sandwwraith:
PR should be open against `dev`

PR #3097, opened 2025-10-23
feat: adding the Serialize classes page

daniCsorbaJB:
This is the fourth part of the Kotlin Serialization rewrite task.

Related YouTract ticket is here: [KT-81889](https://youtrack.jetbrains.com/issue/KT-81889) [Docs] Create a Serialize classes page for kotlinx.serialization

### DIFF
--- a/docs-website/topics/serialization-customization-options.md
+++ b/docs-website/topics/serialization-customization-options.md
@@ -1,6 +1,6 @@
 [//]: # (title: Serialize classes)
 
-The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with backing fields.
+The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with [backing fields](properties.md#backing-fields).
 You can customize this behavior to fit your specific needs.
 This page covers various serialization techniques, showing you how to specify which properties are serialized, and how the serialization process is managed.
 
@@ -11,7 +11,7 @@ Before starting, make sure you've imported the [necessary library dependencies](
 The `@Serializable` annotation enables the automatic serialization of class properties,
 allowing a class to be converted to and from formats such as JSON.
 
-In Kotlin, only properties with [backing fields](properties.md#backing-fields) are serialized:
+In Kotlin, only properties with backing fields are serialized:
 
 ```kotlin
 // Imports declarations from the serialization library
@@ -415,7 +415,7 @@ fun main() {
 
 A class annotated with `@Serializable` must declare all parameters in its primary constructor as properties.
 
-If you need to perform additional initialization logic before assigning values to properties, use a secondary constructor.
+If you need to perform additional initialization logic before assigning values to properties, use a [secondary constructor](classes.md#secondary-constructors).
 The primary constructor can remain `private` and handle property initialization.
 
 Here's an example where the secondary constructor parses a string into two values, which are then passed to the primary constructor for serialization:


docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+class Project(
+    // Declares a property with a backing field - serialized
+    var name: String
+) {
+    // Declares a property with a backing field - serialized
```
sandwwraith: Is "Declares ..." our standard for technical writing? I do not remember seeing it. IMO, it is quite redundant, as we know that this line is a declaration, so we get smth like "Declaration declares a property with a backing field"

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+    // Prints only the name and the stars properties
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","stars":9000}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+Kotlin Serialization natively supports nullable properties.
+Like [other _default values_](#set-default-values-for-optional-properties), `null` values aren't encoded in the JSON output:
```
sandwwraith: I get the following tooltip on hover: 

Image

IMO it's out of place and confusing. Also, link points to "Set default values for optional properties﻿" which doesn't mention dropping of default values. Maybe it should point to "Manage the serialization of default properties with @EncodedDefault﻿" instead?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true" validate="false"}
+
+> If you need to handle `null` values from third-party JSON, you can [coerce them to a default value](serialization-json-configuration.md#coerce-input-values).
+>
+{style="tip"}
+
+Similarly, when the input contains an [optional property](#set-default-values-for-optional-properties) (a property with a default value), its initializer isn't called.
```
sandwwraith: I do not find similarities with previous example tbh. Maybe this section better be placed next to optional properties definition?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+
+```kotlin
+// Imports declarations from the serialization and JSON handling libraries
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+// The Box<T> class can be used with built-in types like Int
+// Alternatively, with user-defined types like Project
```
sandwwraith: "or with @Serializable user-defined types like Project"

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+Kotlin Serialization provides several ways to modify how classes are serialized.
+This section covers techniques for customizing property names, managing default values, and more.
+
+### Customize serial names
+
+By default, property names in the serialization output, such as JSON, match their names in the source code.
+
+You can customize these names, called _serial names_,
+with the [`@SerialName`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serial-name/) annotation.
+Use it to make a property name shorter or more descriptive in the serialized output:
```
sandwwraith: I do not think we should advise users making their names shorter

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+    println(Json.encodeToString(Project("kotlin/kotlinx.serialization")))
+    // {"owner":"kotlin","name":"kotlinx.serialization"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Validate data in the primary constructor
+
+You can validate constructor parameters before storing it in properties to ensure that a class is serializable and invalid data is rejected during deserialization.
```
sandwwraith: "A deserialization process works like a regular constructor in Kotlin and **calls all init blocks**" — is an important detail, you didn't have to remove it

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+        {"name":""}
+    """)
+    println(data)
+    // Exception in thread "main" java.lang.IllegalArgumentException: name cannot be empty
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Set default values for optional properties
```
sandwwraith: Summing up all commentaries above, I think this section should be put much earlier — before dropping `null`s and mentioning that initializers are not always called for optionals. I think it even worth its own sub-section in the navigation pane on the right. E.g.:
``` 
- Optional properties
  - setting default value
  - default values are dropped by default (example with null)
  - initializer is not always called
  - controlling behavior with @EncodeDefault
  - making properties @Required
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" validate="false"}
+
+> To avoid exceptions from unknown keys in JSON, including those marked with the `@Transient` annotation, enable the [`ignoreUnknownKeys`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-builder/ignore-unknown-keys.html) configuration property.
+> For more information, see the [Ignore unknown keys](serialization-json-configuration.md#ignore-unknown-keys) section.
+>
+{style="tip"}
+
+### Manage the serialization of default properties with `@EncodedDefault`
```
sandwwraith: typo: EncodeD

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
+[//]: # (title: Serialize classes)
+
+The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation in Kotlin enables the serialization of all properties in classes defined by the primary constructor.
```
daniCsorbaJB: nice, good point - changed it to properties with backing fields here as well 👍

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: 🤔 It's used, but I see where you are coming form, so I agree.

What do you think about these options?
My immediate thought is to use "Defines" instead (to retain the third-person structure of the comment — which is the preferred style) 

> // Defines a property with a backing field - serialized

Alternatively, I don't think it's an issue to deviate from the preferred style here, so we could make it even shorter for this one:
> // Property with a backing field - serialized.

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,505 @@
...
+
+Before starting, make sure you have the [necessary library dependencies imported](serialization-get-started.md).
+
+## The @Serializable annotation
+
+The `@Serializable` annotation enables the automatic serialization of class properties,
+allowing a class to be converted to and from formats such as JSON.
+
+In Kotlin, only properties with backing fields are serialized.
+Properties defined with [accessors](properties.md) or delegated properties without backing fields are excluded from serialization:
```
daniCsorbaJB: 🤔 Yeah, good suggestion. I think we can get away with it with the link + the example following it 👍

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I fixed the tooltip to include the rest of the information. 

image

> Also, link points to "Set default values for optional properties﻿" which doesn't mention dropping of default values. Maybe it should point to "Manage the serialization of default properties with @EncodedDefault﻿" instead?

An interesting thought 🤔  — my intention here is to point the user to a place where they can read about the concept of default values.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: fixed — thank you 💯

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I really like the idea 👍  I restructured it

I'm a bit unsure about `@Required` and `@Transient`. They mention default values, but I think those might be a better fit in ## Customize class serialization. 

But moving everything else about Optional properties solves a lot of issues 👍

docs-website/topics/serialization-customization-options.md
pdvrieze: In addition you may want to write it in such a way that it doesn't imply that there is no way to serialize "other" properties, while this can be done with custom serializers (not that you should really go into those here). Maybe just replace "excluded from serialization:" with something like "excluded from plugin managed serialization" (managed seems better than generated to me here)

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: hmm 🤔 I was/am concerned it makes the section a bit heavier on implementation detail that way. (especially the part about works like a regular constructor)

What do you think about these versions? 🤔 

Version 1:
> During deserialization, Kotlin calls the class's primary constructor and runs its initializer blocks,
just like when you create an instance.
This allows you to validate constructor parameters and reject invalid data during deserialization.
Here's an example:

Version 2:
> To ensure that a class remains serializable and that invalid data is rejected during deserialization, add validation checks in an `init` block. During deserialization, Kotlin calls the class’s primary constructor and runs all initializer blocks.
Here's an example:

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: But with the new structure I think this is resolved! 👍

docs-website/topics/serialization-customization-options.md
pdvrieze: I think both of them are somewhat confusing. I think it is relevant here to distinguish between primary constructor classes and those without them. In the first case, the primary constructor (and init blocks) will be called. In the second case, the init blocks will still be called (but a serialization specific constructor is generated that, and none of the existing constructors are called - initialisation inside non-primary constructors will not be invoked).

sandwwraith: > Kotlin calls the class's primary constructor

We don't call the primary constructor 😅 . So this version is more correct:

> After deserialization, kotlinx.serialization plugin runs class' initializer blocks,
just like when you create an instance.
This allows you to validate constructor parameters and reject invalid data during deserialization.
Here's an example:

daniCsorbaJB: I think we could clarify further — what do you think about this? 🤔 

> During deserialization, Kotlin runs all initializer blocks, and calls the class's primary constructor if it has one.

Shift the focus a little, (initializer blocks are called) and further reinforces that it calls only the primary constructor.

docs-website/topics/serialization-customization-options.md
sandwwraith: Transient indeed fits into `## Customize class serialization.`. `@Required` directly affects properties with optional values (and doesn't make sense without one), so I think it's still better to place it here.

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,510 @@
...
+// Imports declarations from the serialization and JSON handling libraries
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+@Serializable
+// The Box<T> class can be used with built-in types like Int
+// or with @Serializable user-defined types like Project
+class Box<T>(val contents: T)
+@Serializable
```
sandwwraith, line 133: missing blank line in between Box and Project?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,510 @@
...
+```
+{kotlin-runnable="true"}
+
+When you serialize a generic class like `Box<T>`, the JSON output depends on the actual type you specify for `T` at compile time.
+If that type isn't serializable, you get a compile-time error.
+
+## Optional properties
+
+A property is optional if it isn't required in the serialized input or output.
+Properties with _default values_ are optional during serialization.
```
sandwwraith:
```suggestion
Properties with _default values_ are optional during deserialization, and can be skipped during serialization, if format is configured to do so. JSON is configured to skip encoding of default values out of the box.
```

IMO this just repeats the sentence above, better to delete that one and leave the suggested version

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,510 @@
...
+    // Project(name=kotlinx.serialization, language=Kotlin)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Serialize nullable properties
+
+Kotlin Serialization natively supports nullable properties.
+Like other default values, `null` values aren't encoded in the JSON output:
```
sandwwraith: I have a slight concern that user may easily overlook "Like other default values" and think that only nulls are not encoded in the output. In the original version, we used example with `val language: String = "Kotlin"` to demonstrate that it is not encoded. Maybe leave it here as well? Especially given that identical class is used as "Set default values for optional properties" example. Or you can combine both these examples into one, e.g.:

```kotlin
@Serializable
data class Project(val name: String, val language: String = "Kotlin") // Demonstartes non-nullable optional prop

@Serializable
data class User(val name: String, val alias: String? = null) // Demonstrates nullable with null default
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,510 @@
...
+        {"name":"kotlinx.serialization","language":null}
+    """)
+    println(data)
+    // JsonDecodingException
+}
+//sampleEnd
+```
+{kotlin-runnable="true" validate="false"}
+
+> If you need to handle `null` values from third-party JSON, you can [coerce them to a default value](serialization-json-configuration.md#coerce-input-values).
```
sandwwraith, line 238: Maybe `explicitNulls` flag is also worth mentioning here

docs-website/topics/serialization-customization-options.md
sandwwraith: Or you can just bring link to "Manage the serialization of default properties with `@EncodeDefault`" section back, I think link formatting will make this part easier to notice.

sandwwraith (changes requested):

docs-website/topics/serialization-customization-options.md
sandwwraith: > calls the class's primary constructor if it has one.

We still don't do it 😭 .
kotlinx.serialziation plugin generates a special [deserialization constructor](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0149-serialization.md#layout-of-synthetic-classes-and-methods-for-structures), which is called instead of the primary one. This constructor makes sure that all the optional properties are set and all initialization blocks are called. 

You can include info about the deserialization constructor if you want, though, but maybe it's low-level enough to omit it.

pdvrieze: And of course the primary constructor does not have a body anyway (conceptually - technically the initialisation blocks would be), so there is no constructor code to "not" invoke.

daniCsorbaJB: > We still don't do it 😭 .

Ah sorry the reply didn't show up for me when I was already typing a response to @pdvrieze 😓 
(Messages are 1 minute apart 😅 ) Thanks for the extra information! 👍 
I agree it feels a bit too low level, so let's go with this one then: 

> After deserialization, the `kotlinx.serialization` plugin runs the class's initializer blocks,
just like when you create an instance.
This allows you to validate constructor parameters and reject invalid data during deserialization.
Here's an example:

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: good point — I added an extra "for example", before the JSON sentence but let's go with that.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: Both are good ideas. I prefer linking here. It makes the example a bit more concise, so let's go with that 👍

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: sounds like a good idea — added it to the tip 👍

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,515 @@
...
+To always serialize a property, regardless of its value or format settings, use the [`@EncodeDefault`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/) annotation.
+You can also change this behavior by setting the [`EncodeDefault.Mode`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/-mode/) parameter.
+
+Let's look at an example, where the `language` property is always included in the serialized output,
+while the `projects` property is excluded when it's an empty list:
+
+```kotlin
+// Imports declarations from the serialization and JSON handling libraries
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
```
sandwwraith, line 318: Also, it is better to add `import kotlinx.serialization.EncodeDefault.Mode.NEVER` here and use `@EncodeDefault(NEVER)`

daniCsorbaJB: nice 👍 
I added it — in addition, I also added some clarification for the opt-in as well, like in the JSON docs. 👍 
Please let me know what you think 🙏

docs-website/topics/serialization-customization-options.md
pdvrieze: You probably want to make sure readers are not confused whether or not nulls will be (by default) omitted/optional on reading even if not explicitly given as default.

docs-website/topics/serialization-customization-options.md
sandwwraith: I think opt-in won't be necessary: #3106

docs-website/topics/serialization-customization-options.md
sandwwraith: I've skimmed through our docs and I like the approach from 'Classes' page much more (https://kotlinlang.org/docs/classes.html#initializer-blocks). 
For _declarations_, it just describes what it is: "Class with a primary constructor that initializes name and age"
For _statements/expressions_ (i.e., lines of code that do something inside the function) it indeed uses verbs: "Creates an instance of the Person class". 

So in this case, when describing properties, I think we should stick to "Property with a backing field - serialized."

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,524 @@
...
+fun main() {
+    val data = Project("kotlinx.serialization")
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> Learn more about configuring this JSON behavior in the [Encode default values](serialization-json-configuration.md#encode-default-values) section.
```
sandwwraith: I would rephrase this as "You can change JSON setting to affect all properties with default values, learn here how" to highlight the fact that you don't have to put `@EncodeDefault` everywhere if you don't want to

sandwwraith (changes requested):
I do not know whether we'll finish whole review before mid-December (I expect next release of serialization there), but we can leave opt-in note for now and delete it later.

Besides that, please fix "Defines..." wording and my other comment and we're good to go

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: oh nice!  — I'll remove it then (I don't think it's realistic that we'll be publishing before mid-December)

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: Good idea — much cleaner that way. 

So based on that what do you think about: 

> You can configure a `Json` instance to [encode default values](serialization-json-configuration.md#encode-default-values) for all properties by default. 
>
{style="tip"}

sandwwraith (approved):

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
+[//]: # (title: Serialize classes)
+
+The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with backing fields.
+You can further customize this behavior to fit your specific needs.
```
sarahhaggarty:
```suggestion
You can customize this behavior to fit your specific needs.
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
+[//]: # (title: Serialize classes)
+
+The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with backing fields.
```
sarahhaggarty:
```suggestion
By default, the [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation serializes all class properties that have backing fields.
```
What do you think of this version?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+You can further customize this behavior to fit your specific needs.
+This page covers various serialization techniques, showing you how to specify which properties are serialized and how the serialization process is managed.
```
sarahhaggarty:
```suggestion
This page covers various serialization techniques, showing you how to specify which properties are serialized, and how the serialization process is managed.
```
Is the process managed by the library?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+    val data = Project("kotlinx.serialization").apply { stars = 9000 }
+    // Prints only the name and the stars properties
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","stars":9000}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-annotation-example"}
+
+### Serialization of class references
```
sarahhaggarty: What do you think about "Serialize class references" as a header? So that the header style matches the others on the page?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-class-references"}
+
+> To reference non-serializable classes, annotate the corresponding properties with [`@Transient`](#exclude-properties-with-the-transient-annotation), or
+> provide a [custom serializer](create-custom-serializers.md) for them.
+>
+{style="tip"}
+
+### Serialization of repeated object references
```
sarahhaggarty, line 83: What do you think about "Serialize repeated object references" as a header? So that the header style matches the others on the page?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+{kotlin-runnable="true" id="serialize-repeated-references"}
+
+> If you attempt to serialize a circular structure, it results in stack overflow.
+> To exclude references from serialization, use [the @Transient annotation](#exclude-properties-with-the-transient-annotation).
+>
+{style="tip"}
+
+### Generic class serialization
+
+Generic classes in Kotlin support type-polymorphism, which is enforced by Kotlin Serialization at
```
sarahhaggarty:
```suggestion
Generic classes in Kotlin support type-polymorphism, which is enforced by Kotlin serialization at
```
Not sure this should be capitalized?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+compile-time. For example, consider a generic serializable class `Box<T>`:
+
+```kotlin
```
sarahhaggarty, line 122: What do you think about this as an alternative? It is the same example with just slightly different names:
```

// A generic wrapper that can hold various types of data
@Serializable
class Payload<T>(val value: T)

@Serializable
data class Repository(val name: String, val language: String)

@Serializable
class BackupData(
    val issueCount: Payload<Int>,
    val mainRepo: Payload<Repository>
)

fun main() {
    val backup = BackupData(
        issueCount = Payload(42),
        mainRepo = Payload(Repository("kotlinx.serialization", "Kotlin"))
    )
    println(Json.encodeToString(backup))
    // {"issueCount":{"value":42},"mainRepo":{"value":{"name":"kotlinx.serialization","language":"Kotlin"}}}
}
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","stars":9000}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-annotation-example"}
+
+### Serialization of class references
+
+Serializable classes can contain properties that reference other classes.
```
sarahhaggarty: Do you define somewhere what it actually means for a class to be serializable? 🤔

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-generic-classes-example"}
+
+When you serialize a generic class like `Box<T>`, the JSON output depends on the actual type you specify for `T` at compile time.
+If that type isn't serializable, you get a compile-time error.
+
+## Optional properties
+
+Properties with _default values_ are optional during deserialization, and can be skipped during serialization,
```
sarahhaggarty:
```suggestion
Properties with _default values_ are optional during deserialization and can be skipped during serialization,
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+if the format is configured to do so.
+For example, JSON is configured to skip encoding of default values out of the box.
```
sarahhaggarty:
```suggestion
For example, JSON skips encoding default values out of the box.
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-repeated-references"}
+
+> If you attempt to serialize a circular structure, it results in stack overflow.
+> To exclude references from serialization, use [the @Transient annotation](#exclude-properties-with-the-transient-annotation).
+>
+{style="tip"}
+
+### Generic class serialization
```
sarahhaggarty: What do you think about "Serialize generic classes" as an alternative header? So that the header style matches the others on the page?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+
+> If you need to handle `null` values from third-party JSON, you can [coerce them to a default value](serialization-json-configuration.md#coerce-input-values).
+> 
+> You can also [omit explicit `null` values](serialization-json-configuration#omit-explicit-nulls) from the encoded JSON with the `explicitNulls` property.
+>
+{style="tip"}
+
+### Initializers in optional properties
+
+When the input contains an optional property, its initializer isn't called.
```
sarahhaggarty: By "input", do you mean the class header? Or constructor?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="default-properties-example"}
+
+> You can configure a `Json` instance to [encode default values](serialization-json-configuration.md#encode-default-values) for all properties by default.
+>
+{style="tip"}
+
+To always serialize a property, regardless of its value or format settings, use the [`@EncodeDefault`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/) annotation.
+You can also change this behavior by setting the [`EncodeDefault.Mode`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/-mode/) parameter.
```
sarahhaggarty:
```suggestion
Alternatively, you can set the [`EncodeDefault.Mode`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/-mode/) parameter.
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+
+Let's look at an example, where the `language` property is always included in the serialized output,
```
sarahhaggarty:
```suggestion
Let's look at an example where the `language` property is always included in the serialized output,
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+    // language is always serialized
+    println(Json.encodeToString(userA))
+    // {"name":"Alice","projects":[{"name":"kotlinx.serialization","language":"Kotlin"}]}
+
+    // Excludes projects because it's an empty list
+    // and EncodeDefault.Mode is set to NEVER, so it's not serialized
+    println(Json.encodeToString(userB))
+    // {"name":"Bob"}
+}
+//sampleEnd
```
sarahhaggarty, line 350: What do you think about changing `UserA` and `UserB` to `adminUser` and `guestUser`? Otherwise the example is very nice 😄

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+```
+{kotlin-runnable="true" id="serialize-encode-default"}
+
+### Make properties required with the `@Required` annotation
+
+You can annotate a property with [`@Required`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-required/) to make it required in the input.
```
sarahhaggarty:
```suggestion
Annotate a property with [`@Required`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-required/) to make it required in the input.
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+    println(data)
+    // MissingFieldException
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="required-annotation-example"}
+
+## Customize class serialization
+
+Kotlin Serialization provides several ways to modify how classes are serialized.
```
sarahhaggarty: What do you think about "control" instead of "modify"?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+This section covers techniques for customizing property names, controlling property inclusion, and more.
+
+### Customize serial names
+
+By default, property names in the serialization output, such as JSON, match their names in the source code.
+
+You can customize these names, called _serial names_,
+with the [`@SerialName`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serial-name/) annotation.
+Use it to make a property name more descriptive in the serialized output:
```
sarahhaggarty, line 392: Here you say that you should use it to make the property  more descriptive. But in the example you abbreviate it. So maybe we should change the wording or the example?

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="serial-names"}
+
+### Define constructor properties for serialization
+
+A class annotated with `@Serializable` must declare all parameters in its primary constructor as properties.
+
+If you need to perform additional initialization logic before assigning values to properties, use a secondary constructor.
+The primary constructor can remain private and handle property initialization.
```
sarahhaggarty:
```suggestion
The primary constructor can remain `private` and handle property initialization.
```

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+    println(Json.encodeToString(Project("kotlin/kotlinx.serialization")))
+    // {"owner":"kotlin","name":"kotlinx.serialization"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="serialize-constructor-properties"}
+
+### Validate data in the primary constructor
+
+After deserialization, the `kotlinx.serialization` plugin runs the class's initializer blocks,
```
sarahhaggarty: Could you add a link here to the Kotlin section about initializer blocks?  https://kotlinlang.org/docs/classes.html#initializer-blocks

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,516 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true" id="validate-constructor-example"}
+
+### Exclude properties with the `@Transient` annotation
+
+You can exclude a property from serialization with the [`@Transient`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-transient/) annotation.
+Transient properties must have a default value.
+
+If you explicitly specify a value for a transient property in the input, even if it matches the default value, a `JsonDecodingException` is thrown:
```
sarahhaggarty, line 486: Again, I'm not sure what "input" is referring to exactly here 🤔

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: technically neither — like in the examples above "input" refers to the serialization input (the JSON input in this example)

{"name":"kotlinx.serialization","language":"Java"}

We could further clarify it by adding _serialization_ before input. (but that might make it a bit redundant in other examples 🤔 )

docs-website/topics/serialization-customization-options.md
pdvrieze: I'm not sure it is good to imply that the annotation serializes. While it is "simpler" it is also misleading, potentially leading to misunderstandings. What about stating it indirectly:
"By default, the [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation causes all class properties that have backing fields to be serialized."

docs-website/topics/serialization-customization-options.md
pdvrieze: The overall process starts with the format (which could be Json or another library provided format), and then parts of it are determined by the actual serializers (either plugin generated or custom). The core library does not actually do much in terms of the process (except define a way in which it should happen). In other words, there is no serialization "loop" that formats plug in to, but rather the format manages/controls the process as restricted/defined by the core library.

docs-website/topics/serialization-customization-options.md
pdvrieze: This header (and the ones below) provide specifics as to how specific kinds of elements (attributes) are handled, but this is not an active user thing. As such the verb "serialize" is misleading.

docs-website/topics/serialization-customization-options.md
pdvrieze: Unlike the above ones which all cover elements, this header covers types. A subtle, but meaningful distinction.

docs-website/topics/serialization-customization-options.md
pdvrieze: It would be worthwhile making clear that this behaviour is configurable. I'm not sure that "out of the box" makes this clear (or would be clear to those whose English is not that good). Maybe:
"For example, in its default configuration JSON skips encoding default values."

docs-website/topics/serialization-customization-options.md
pdvrieze: This could indeed be rephrased. Looking at the example code it clear what happens, but as written it is somewhat confusing. Partially that is because it is not clear what "its" refers to (actually the property, not the input - input has no initializer)

docs-website/topics/serialization-customization-options.md
pdvrieze: As the section leaves out custom serializers (as a reference to more details for this "advanced" option), it is more a matter of adjustment/modification (I would prefer adjustment, but it might be a bit of a "hard" word).

docs-website/topics/serialization-customization-options.md
pdvrieze: And in reality the key function is compatibility with other serialization systems/existing formats. I think the wording needs to be adjusted. Maybe too advanced would be that it may also be used to ensure a property's serialized name is stable under refactoring/renaming of the underlying kotlin property.

docs-website/topics/serialization-customization-options.md
pdvrieze: The input here is the Json formatted data that is being deserialized (converted into Kotlin datastructures). What is actually going on here is that the serializer does not record transient properties in its descriptor, so when it encounters the property name it is treated as any other unknown attribute, which by default is to throw an exception (there is an option to ignore unknown fields).

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I also think I prefer the original phrasing here — it captures what happens a bit better (as @pdvrieze says, it doesn't perform serialization on its own - it essentially generates the default [serializer](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/) under the hood, which enables the serialization of class properties with backing fields )

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: 🤔 
I'd say these two state it, or should we go more explicit — essentially, a class is serializable if it is annotated with `@Serializable` (a KSerializer exists for the class)

> The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with backing fields.


> The `@Serializable` annotation enables the automatic serialization of class properties,
allowing a class to be converted to and from formats such as JSON.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: sure, I like it 👍 — I changed the first comment but otherwise I added it - thank you!

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I agree with both of you here 👍 

Let's go for a mix: 

> For example, the default JSON configuration skips encoding default values.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: 🤔 I think we could go for something like this to clarify:

> When the input includes a value for an optional property, the property's initializer isn't called.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I think it should be ok to change to Serialize generic classes 👍 
I wanted to distinguish the behavior of the library from actions the user takes with the help of titles, but I think making the style more cohesive adds more than the distinction I was going for originally.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: ah good idea 👍 💯 changed it! 😄

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I also prefer modify here slightly (also this way we can avoid repeating control from the following sentence)

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: good idea 💯

docs-website/topics/serialization-customization-options.md
sandwwraith: > and how the serialization process is managed

I think what implied here is that library manages the process, and you can manage it by using different things from library, such as `@Transient` annotation

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: I made the example match this (ensured that it's it the other way around to be more descriptive)

docs-website/topics/serialization-customization-options.md
sarahhaggarty: Thanks for the explanation! Let's stick with the original then.

docs-website/topics/serialization-customization-options.md
sarahhaggarty: Thanks for the explanation! Since there isn't a clear actor that manages the whole process, let's keep the original text.

docs-website/topics/serialization-customization-options.md
sarahhaggarty: Ah OK! Based on the first sentence, I thought that if you had a class which only had properties without backing fields, then it couldn't be serialized. I think it would help to make it more explicit. Perhaps something like "Classes annotated with `@Serializable` can contain properties that reference other classes."?

docs-website/topics/serialization-customization-options.md
sarahhaggarty: Previously, when referring to "input" it's been clear that the context is serialization by the rest of the sentence. I think it would help to be explicit and say "serialization input". For example, let's imagine someone googles and lands on this section first. I like the additional clarification of "it's" in the new sentence.

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: sure thing 👍 good idea

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,519 @@
+[//]: # (title: Serialize classes)
+
+The [`@Serializable`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serializable/) annotation enables the default serialization of all class properties with backing fields.
```
sarahhaggarty: Can you add a link to our content on backing fields? https://kotlinlang.org/docs/properties.html#backing-fields

docs-website/topics/serialization-customization-options.md
```diff
@@ -0,0 +1,519 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="serial-names"}
+
+### Define constructor properties for serialization
+
+A class annotated with `@Serializable` must declare all parameters in its primary constructor as properties.
+
+If you need to perform additional initialization logic before assigning values to properties, use a secondary constructor.
```
sarahhaggarty: Let's link to our content on secondary constructors:
https://kotlinlang.org/docs/classes.html#secondary-constructors

sarahhaggarty (approved):
It's looking great! Just two more suggestions to add links to our other content for backing fields and secondary constructors. Otherwise, LGTM! 🚀

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: oh it's added it to the second mention 😅  — moved it to first mention! 👍

docs-website/topics/serialization-customization-options.md
daniCsorbaJB: nice good idea — added it 👍

PR #3098, opened 2025-10-23
Clarify that it is preferable to have unique class' serial name

### LINKED ISSUE #3003: SerialName duplication for different polymorphic heirarchies causes collisons when `namingStrategy` is set.
**Describe the bug**

We have parallel heirarchies of polymorphic classes. We were seeing fields in one heirarchy appearing in the serialized output of the other; only when `namingStrategy` is set.

**To Reproduce**

Kotest cases:

```kotlin
@file:OptIn(ExperimentalSerializationApi::class)

import io.kotest.core.spec.style.FunSpec
import io.kotest.matchers.shouldBe
import kotlinx.serialization.ExperimentalSerializationApi
import kotlinx.serialization.SerialName
import kotlinx.serialization.json.JsonClassDiscriminator
import kotlinx.serialization.json.JsonNamingStrategy

class SerialNameCollisionTest : FunSpec({
    test("Naming strategy set - test will fail") {
        val json = Json {
            namingStrategy = JsonNamingStrategy.SnakeCase
        }

        val foo = FooBoolean(true)
        val bar = BarBoolean(true)

        json.encodeToString<Foo>(foo) shouldBe """{"type":"BOOLEAN","my_foo_value":true}"""

        // Fails at this line:
        json.encodeToString<Bar>(bar) shouldBe """{"type":"BOOLEAN","my_bar_value":true}"""
        // Expected :"{"type":"BOOLEAN","my_bar_value":true}"
        // Actual   :"{"type":"BOOLEAN","my_foo_value":true}"
    }

    test("Naming strategy not set - test will pass") {
        val json = Json { }

        val foo = FooBoolean(true)
        val bar = BarBoolean(true)

        json.encodeToString<Foo>(foo) shouldBe """{"type":"BOOLEAN","myFooValue":true}"""
        json.encodeToString<Bar>(bar) shouldBe """{"type":"BOOLEAN","myBarValue":true}"""
    }
})

@Serializable
@JsonClassDiscriminator("type")
sealed interface Foo

@Serializable
@SerialName("BOOLEAN")
data class FooBoolean(val myFooValue: Boolean) : Foo

@Serializable
@JsonClassDiscriminator("type")
sealed interface Bar

@Serializable
@SerialName("BOOLEAN")
data class BarBoolean(val myBarValue: Boolean) : Bar
```

In the failing test above, the first `encodeToString()` always provides the correct output, and the second always matches it, implying some kind of caching may be to blame.

**Expected behavior**

Serializing `BarBoolean` in the example above should never produce a field from `FooBoolean`. Failing that, the presense of `namingStrategy` should not alter this behaviour.

If neither of the above are practical, then the order in which each class is first serialized should not affect the output.

**Environment**
 - Kotlin version: 2.0.20
 - Library version: 1.8.1
 - Kotlin platforms: JVM
 - Gradle version: 8.6

sandwwraith:
Fixes #3003

### DIFF
--- a/docs/polymorphism.md
+++ b/docs/polymorphism.md
@@ -234,8 +234,9 @@ This way we can have a stable _serial name_ that is not affected by the class's 
 > In addition to that, JSON can be configured to use a different key name for the class discriminator. 
 > You can find an example in the [Class discriminator for polymorphism](json.md#class-discriminator-for-polymorphism) section.
 
-When picking a serial name for a class, avoid assigning the same name to different classes.
-Check out equality rules in documentation for [SerialDescriptor] to make sure that class descriptor will stay unique.
+> [!IMPORTANT]
+> When picking a serial name for a class, avoid assigning the same name to different classes.
+Check out equality rules in documentation for [SerialDescriptor] to make sure that the class descriptor will stay unique.
 
 ### Concrete properties in a base class
 
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptors.kt
@@ -93,6 +93,7 @@ public fun PrimitiveSerialDescriptor(serialName: String, kind: PrimitiveKind): S
 /**
  * Factory to create a new descriptor that is identical to [original] except that the name is equal to [serialName].
  * Usually used when you want to serialize a type as another type, delegating implementation of `serialize` and `deserialize`.
+ * Do not use [serialName] that is equal to the name of [original] or other serializable classes.
  *
  * Example:
  * ```


pdvrieze:
This looks, good, but it is probably also a good idea to add the same/similar warnings to the various descriptor factories in `kotlinx.serialization.descriptors.SerialDescriptors.kt`

docs/polymorphism.md
```diff
@@ -234,6 +234,9 @@ This way we can have a stable _serial name_ that is not affected by the class's
 > In addition to that, JSON can be configured to use a different key name for the class discriminator. 
 > You can find an example in the [Class discriminator for polymorphism](json.md#class-discriminator-for-polymorphism) section.
 
+When picking a serial name for a class, avoid assigning the same name to different classes.
```
fzhinkin: Should we emphasize it with some warning banner?

Like 
```Markdown
> [!WARNING]
> When picking a serial name ...
```

(or what GH allows us to use to render a message in a warning frame, `WARN`?)

pdvrieze: Given that the errors you get if you hit this bug are subtle and hard to diagnose, it is probably a good idea to warn about it.

sandwwraith: I'd say it is more of an 'Important' badge. Sounds good, but we do not have the same functionality on kotlinlang.org. @daniCsorbaJB Do you have ideas how we can adapt GH warning sections? (https://gist.github.com/cseeman/8f3bfaec084c5c4259626ddd9e516c61#other-markdown-options)  IIRC we have only "note" level now. Maybe we should add "warning/important" as well?

sandwwraith:
We have "The resulting descriptor will be uniquely identified by the given [serialName], [typeParameters] and elements structure described in [builderAction] function." comment on `buildClassSerialDescriptor` and "[serialName] must be non-blank and unique." on `PrimitiveSerialDescriptor`. So only documentation for `SerialDescriptor()` requires an update.

docs/polymorphism.md
daniCsorbaJB: @sandwwraith we have warning labels in kotlinlang. (alongside with tip + note): it shows up with a red banner. For example, [here](https://kotlinlang.org/docs/gradle-compiler-options.html#attributes-common-to-jvm-and-javascript). We exclusively use this to warn about critical things the user must absolutely be aware to proceed. 👍 


(in comparison: 

-  we use the note tag to highlight important or useful information that aids with understanding
- the tip tag to to provide advice, recommendations, references to additional reading, or the same content but on a different platform)

sandwwraith: Then I think it should be a warning label after migration

Issue #2885, opened 2024-12-16
JSON serialization incorrect behavior


FerrumBrain:
## 0\. Setup
We created the following class hierarchy for testing serialization library:

```kotlin
Value (org.plan.research)
    ArrayValue (org.plan.research)
    BooleanArrayValue (org.plan.research)
    BooleanValue (org.plan.research)
    ByteArrayValue (org.plan.research)
    ByteValue (org.plan.research)
    CharArrayValue (org.plan.research)
    CharValue (org.plan.research)
    CompositeNullableValue (org.plan.research)
    DefaultValueAlways (org.plan.research)
    DefaultValueNever (org.plan.research)
    DoubleArrayValue (org.plan.research)
    DoubleValue (org.plan.research)
    EnumValue (org.plan.research)
    FloatArrayValue (org.plan.research)
    FloatValue (org.plan.research)
    IntArrayValue (org.plan.research)
    IntValue (org.plan.research)
    ListValue (org.plan.research)
    LongArrayValue (org.plan.research)
    LongValue (org.plan.research)
    NullValue (org.plan.research)
    ObjectValue (org.plan.research)
    ShortArrayValue (org.plan.research)
    ShortValue (org.plan.research)
    StringValue (org.plan.research)
```

`Value` hierarchy tries to use most of the available serialization API and test it on all main data types available on Kotlin/JVM.
The exact implementation details are not important in most cases. We will highlight interesting implementation details whenever necessary.

## 1\. Unhandled conflict of `classDesctiptor` config option and `JsonNames` annotation values

`Value` class has a property with custom JSON names specified:

```kotlin
@OptIn(ExperimentalSerializationApi::class)
@Serializable
sealed class Value {
    @JsonNames(
        "THIS_IS_STATUS",
        "STATUS",
        "IS_OPEN"
    )
    var status = "open"
    @Suppress("unused")
    val randomStr: String get() = status
}
```

If we specify `classDiscriminator` equal to one of the JSON names, it will create a name conflict that is not detected.
Moreover, this conflict will actually affect the serialization:

```kotlin
@Test
fun `json class descriptor name conflict`() {
    val serializer = Json {
        classDiscriminator = "THIS_IS_STATUS"
    }
    val value: Value = CompositeNullableValue(
        StringValue("foo"),
        NullValue,
        NullValue
    )
    val str = serializer.encodeToString(value)
    val decodedValue = serializer.decodeFromString<Value>(str)
[...]

Issue #1664, opened 2021-09-06
Open polymorphism, classDiscriminator, keep "type" property in the class.


hovi:
**What is your use-case and why do you need this feature?**

I have existing design. I have class that I am serializing, which contains "BlockType" property. Currently there's only one class, but I wanted to switch to open polymorphism and add subclasses, each having different properties based on this "BlockType", original class being default fallback with no additional properties.

I cannot get it working, because I get message:

`Polymorphic serializer for class spaceEngineers.model.Block has property 'BlockType' that conflicts with JSON class discriminator. You can either change class discriminator in JsonConfiguration, rename property with @SerialName annotation or fall back to array polymorphism`

If I understand correctly, framework doesn't like, that this "BlockType" property is used both as normal property in class and also as classDiscriminator. But why is that? I need to keep it while having working polymorphism and I see no reason, why it should not work.

**Describe the solution you'd like**

Either allow using classDiscriminator and a class property at the same time or add configuration, that allows that.

Dominaezzz:
You can get the discriminator from the type, no?

What if there's a mismatch between the discriminator and the type?

hovi:
It is a single JSON field, in my case called "BlockType", which is discriminator at the same time so there cannot be mismatch.

aSemy:
This is useful for me as I'm using the discriminator field to generate Typescript classes (using https://github.com/ntrrgc/ts-generator). 

I've found a couple of potential work-arounds that I'm experimenting with:

```kotlin
import kotlin.reflect.full.findAnnotation
import kotlin.reflect.full.starProjectedType
import kotlinx.serialization.SerialName
import kotlinx.serialization.Serializable
import kotlinx.serialization.json.JsonClassDiscriminator


@Serializable
@JsonClassDiscriminator("object_name")
sealed class EventData() {

  //  get the discriminator as a field (the field name, objectName, is unimportant)
  // this must be a delegated field so there's no backing field, so kxs ignores it
  val objectName: String by lazy {
    // this probably won't work for generic classes?
    kotlinx.serialization.serializer(this::class.starProjectedType).descriptor.serialName
  }

  // alternative: find the @SerialName annotation.
  // Again, the field must be delegated so there's no backing field. 
  // This requires that every subclass has @SerialName
  val objectNameByAnnotation: String by lazy {
    requireNotNull(this::class.findAnnotation<SerialName>()?.value) {
      "Couldn't find @SerialName for ${this::class}!"
    }
[...]

PR #3105, opened 2025-11-04
Class discriminator conflict check rework

sandwwraith:
Check more potential places where conflict can occur during encoding:

- from JsonNamingStrategy
- from ALL_JSON_OBJECTS
- from default polymorphic serializer

Allow regular polymorphic classes to be decoded with conflicting names because this is our behavior for sealed classes.

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonClassDiscriminatorTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonClassDiscriminatorTest.kt
@@ -8,6 +8,7 @@ import kotlinx.serialization.*
 import kotlinx.serialization.builtins.*
 import kotlinx.serialization.json.*
 import kotlinx.serialization.modules.*
+import kotlinx.serialization.test.assertFailsWithMessage
 import kotlin.test.*
 
 class JsonClassDiscriminatorTest : JsonTestBase() {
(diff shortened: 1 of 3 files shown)


sandwwraith:
Related: #1664, #2885

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -55,9 +55,15 @@ internal fun Json.deserializationNamesMap(descriptor: SerialDescriptor): Map<Str
 
 internal fun SerialDescriptor.serializationNamesIndices(json: Json, strategy: JsonNamingStrategy): Array<String> =
     json.schemaCache.getOrPut(this, JsonSerializationNamesKey) {
+        val trackingSet = mutableSetOf<String>()
         Array(elementsCount) { i ->
             val baseName = getElementName(i)
-            strategy.serialNameForJson(this, i, baseName)
+            val name = strategy.serialNameForJson(this, i, baseName)
+            if (!trackingSet.add(name)) throw JsonException(
```
fzhinkin: Should it throw `JsonEncodingException`?

fzhinkin (changes requested):

fzhinkin:
It would be nice to include a description of change into release notes; optionally, with steps required to mitigate a problem (I assume that there might be users who incorrectly serialized some objects and was not _rewarded_ with an exception before).

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
sandwwraith: yup

sandwwraith:
> It would be nice to include a description of change into release notes

I usually write a paragraph or two about major changes, and this looks like one, so I'll do it

Issue #805, opened 2020-04-16
Computed property not serialize


rj-hwang:
A data class with computed property:

```
data class Bean(val name: String) {
  val computed: String
    get() = "--$name"  
}
```

Use `kotlinx-serialization-runtime-0.20.0':

```
val result = Json(Stable).stringify(Bean.serializer(), Bean(name = "rj"))
```

the result is `{"name":"rj"}` but I want it to be `{"name":"rj","computed":"--rj"}`. How?

sandwwraith:
It is not supported yet because it is quite a rare use case

grangej:
I need this feature also. 

Example use case: my image urls are computed and not stored in the database.

fluidsonic:
Same for me.

The server adds computed properties to certain data classes to have their logic server-side instead of client-side.

sandwwraith:
Also #1578

slava-shor-emporus:
Here is our use case. We have various projects written in multiple languages and legacy projects we cannot replace. Consider we have simple data class like:

slava-shor-emporus:
Here is our use case.
We have various projects written in multiple languages and legacy projects we cannot replace. Consider we have simple data classes like:
```kotlin
data class Person(val firstName: String, val lastName: String) {
  val fullName get() = "$firstName $lastName"
}
```
And a legacy system expects a JSON payload with all 3 properties: `firstName`, `lastName`, and `fullName` because it will persist in some DB as is.
There is no sense in deserializing the getter but, still useful to serialize its value for legacy systems.
In some other serialization libraries serializing getters and ignoring them on serialization is a normal practice.
Is there any workaround for `kotlinx.serialization`?

slava-shor-emporus:
> I need this feature also.
> 
> Example use case: my image urls are computed and not stored in the database.

We have both! 😁 Both are computed and stored in the DB as well.

felipeleon73:
I solved it this way:
```
@Serializable
data class Person(
    val firstName: String, 
    val lastName: String,  
    @SerialName("fullName ") 
    var _fullName = "") {
       val fullName get() = _fullName
       init {
           val fullName get() = "$firstName $lastName"
       }
}
```

Do you see any contraindications?

pdvrieze:
@felipeleon73 I would go with a serialization delegate (a private nested data class) that is used with a custom serializer to serialize instances (you can write the actual serialization code by hand, your choice).

iseki0:
[...]

PR #3106, opened 2025-11-05
Stabilize @EncodeDefault annotation and its modes.

sandwwraith:
It is widely popular and de-facto is a stable API.

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/Annotations.kt
+++ b/core/commonMain/src/kotlinx/serialization/Annotations.kt
@@ -199,12 +199,10 @@ public annotation class Transient
  */
 @MustBeDocumented
 @Target(AnnotationTarget.PROPERTY)
-@ExperimentalSerializationApi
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/Annotations.kt
```diff
@@ -199,12 +199,10 @@ public annotation class Transient
  */
```
fzhinkin, line 199: Should we explicitly mention what is considered a default value and what's not (related to https://github.com/Kotlin/kotlinx.serialization/issues/805)?

sandwwraith: We have explanation of what is serialized and what not in our docs here: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#backing-fields-are-serialized

fzhinkin (approved):

PR #3109, opened 2025-11-10
Generate unique Automatic-Module-Name for metadata JARs

### LINKED ISSUE #2901: Ambiguous JPMS module reference in maven projects
In maven projects, published "umbrella" artifacts (like "kotlinx-serialization-json") works as a gateway to corresponding `-jvm` artifacts. The former do not have a JPMS module bundled inside, so the module name is inferred from the artifact name. As a result, there are two artifacts providing the same module:
- `kotlinx-serialization-json` provides `kotlinx.serialization.json` as it was inferred from the module name.
- `kotlinx-serialization-json-jvm` provides `kotlinx.serialization.json`, well, because it actually provides it.

**To Reproduce**

Here's a reproducer: https://github.com/fzhinkin/kotlinx-serialization-jpms-conflict
```
$ mvm compile
...
[WARNING] /Users/filipp.zhinkin/.m2/repository/org/jetbrains/kotlinx/kotlinx-serialization-json-jvm/1.8.0/kotlinx-serialization-json-jvm-1.8.0.jar: (-1, -1) The root is ignored because a module with the same name 'kotlinx.serialization.json' has been found earlier on the module path at: /Users/filipp.zhinkin/.m2/repository/org/jetbrains/kotlinx/kotlinx-serialization-json/1.8.0/kotlinx-serialization-json-1.8.0.jar
[ERROR] /Users/filipp.zhinkin/Development/kotlinx-serialization-jmps-conflict/src/main/kotlin/Dummy.kt: (4, 30) Unresolved reference 'json'.
[ERROR] /Users/filipp.zhinkin/Development/kotlinx-serialization-jmps-conflict/src/main/kotlin/Dummy.kt: (10, 33) Unresolved reference 'Json'.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

**Expected behavior**

"Umbrella" artifacts like `kotlinx-serialization-core` or `kotlinx-serialization-json` should provide some JPMS modules that won't clash with modules provided by actual `***-jvm` artifacts.

fzhinkin:
Fixes #2901

### DIFF
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -241,4 +241,28 @@ object Java9Modularity {
             )
         })
     }
+
+    /**
+     * For multiplatform Gradle modules, generate and set `Automatic-Module-Name` in metadata JAR's manifest.
+     *
+     * Generated automatic (JPMS) module name has the following format:
+     * `<Gradle module name>. artifact_disambiguating_module`.
+     *
+     * For multiplatform projects, a metadata artifact is the one without a platform-specific suffix,
+     * and it always depends on corresponding `-jvm` artifact (for convenience on build systems other than Gradle).
+     * For a JVM project depending on such an artifact and using JPMS, it may result in the automatic module name clash
+     * with a module provided by the `-jvm` artifact. By explicitly setting a non-clashing automatic module name in
+     * metadata JAR's manifest, we're mitigating this issue.
+     */
+    fun Project.configureMetadataJarAutomaticModuleName() {
+        val kotlin = extensions.findByType<KotlinMultiplatformExtension>() ?: return
+        val moduleName = project.name.replace('-', '.') + ".artifact_disambiguating_module"
+        tasks.withType<Jar>().named(kotlin.metadata().artifactsTaskName) {
+            manifest {
+                attributes(
+                    "Automatic-Module-Name" to moduleName,
+                )
+            }
+        }
+    }
 }
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -1,4 +1,5 @@
 import Java9Modularity.configureJava9ModuleInfo
+import Java9Modularity.configureMetadataJarAutomaticModuleName
 import org.jetbrains.kotlin.gradle.targets.js.ir.*
 
 /*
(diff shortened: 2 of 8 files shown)


buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -241,4 +241,28 @@ object Java9Modularity {
...
+     *
+     * For multiplatform projects, a metadata artifact is the one without a platform-specific suffix,
+     * and it always depends on corresponding `-jvm` artifact (for convenience on build systems other than Gradle).
+     * For a JVM project depending on such an artifact and using JPMS, it may result in the automatic module name clash
+     * with a module provided by the `-jvm` artifact. By explicitly setting a non-clashing automatic module name in
+     * metadata JAR's manifest, we're mitigating this issue.
+     */
+    fun Project.configureMetadataJarAutomaticModuleName() {
+        val kotlin = extensions.findByType<KotlinMultiplatformExtension>() ?: return
+        val moduleName = project.name.replace('-', '.') + ".artifact_disambiguating_module"
```
sandwwraith, line 259: NB: Manifest lines are supposed to be 72 chars in length 
> No line may be longer than 72 bytes (not characters), in its UTF8-encoded form. If a value would make the initial line longer than this, it should be continued on extra lines (each starting with a single SPACE).
https://docs.oracle.com/en/java/javase/25/docs/specs/jar/jar.html

So this gets automatically wrapped to next line:
```
Manifest-Version: 1.0
Automatic-Module-Name: kotlinx.serialization.core.artifact_disambiguatin
  g_module
```

While it shouldn't cause any problems, we may want to choose a shorter name to look a bit nicer. WDYT of `common_artifact` suffix?

sandwwraith (approved):

buildSrc/src/main/kotlin/Java9Modularity.kt
fzhinkin: I picked the same name we agreed to use for coroutines for consistency.

I'm not sure if we should pursue shorter name here, as nobody is expected to import that module. And I hope not so many people have to read manifest files bundled in third-party libraries (to enjoy nice name :)).

If you prefer a shorter name, than `kotlin_metadata` or just `metadata` might be a better option compared to `common_artifact`.

sandwwraith: Let's stick to kotlinx.coroutines name then

PR #3112, opened 2025-11-14
Kotlin 2.3

sandwwraith:
- Fix some test dependencies due to "Adding API dependency types to test source sets is not supported and will removed in a future version of Kotlin." warning.

- Update BCV version

- Enable Unused Return Value checker


buildSrc/src/main/kotlin/source-sets-conventions.gradle.kts
```diff
@@ -76,6 +76,8 @@ kotlin {
             freeCompilerArgs.add("-Xsuppress-version-warnings")
         }
         freeCompilerArgs.add("-Xexpect-actual-classes")
+        // for some reason, IDE does not enable feature in test source sets without this line:
```
fzhinkin, line 79: Could you please file an issue to investigate the root cause further?

formats/json-tests/commonTest/src/kotlinx/serialization/JsonPathTest.kt
```diff
@@ -126,8 +126,8 @@ class JsonPathTest : JsonTestBase() {
             outer = Sealed.Nesting(outer)
         }
         val str = json.encodeToString(Sealed.serializer(), outer)
-        // throw-away data
-        json.decodeFromString(Sealed.serializer(), str)
+        // check that data is correctly formed
```
fzhinkin, line 129: [nit] If the goal is to check correctness, should we validate that deserialized object is structurally equal to `outer`?

formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonProhibitedPolymorphicKindsTest.kt
```diff
@@ -85,6 +85,7 @@ class JsonProhibitedPolymorphicKindsTest : JsonTestBase() {
         }
     }
 
+    @IgnorableReturnValue
```
fzhinkin: Should the function has `Unit` as its return type instead?

fzhinkin (approved):
Overall, the change looks good. There are only a few minor nits from me.

buildSrc/src/main/kotlin/source-sets-conventions.gradle.kts
sandwwraith: I'm trying to minimize the example, will post once it's ready

formats/json-tests/commonTest/src/kotlinx/serialization/json/polymorphic/JsonProhibitedPolymorphicKindsTest.kt
sandwwraith: Indeed, I missed the fact its result is not used anywhere

formats/json-tests/commonTest/src/kotlinx/serialization/JsonPathTest.kt
sandwwraith: It doesn't matter much, but why not

fzhinkin (approved):

PR #3114, opened 2025-11-18
Add .serialName to MissingFieldException

### LINKED ISSUE #2958: MissingFieldException should return serialName
`MissingFieldException` currently provides the names of the `missingFields`, but it *does not* provide the name of the class/type the fields were missing from, despite being a constructor parameter.

Currently, the `serialName` of the affected type is provided to the constructor, and used to construct the error message. This should be made available as a field on the `MissingFieldException`, to enable programmatic determination of the source of the error.

This is particularly useful when deserializing into a class hierarchy, as the name of the missing field alone may be ambiguous, and exception handlers may need to know the type of the affected class to determine the best course of action.

Adding the `serialName` as a field should be a trivial addition, since it's already available in the constructor.

It might also be useful to go further, and provide a "path" to the missing fields in the JSON AST, which would provide exception handlers with even more information, which could be especially useful in error messages.

sandwwraith:
Make new constructor that requires it, and allow creation with nullable serial name only via internal mechanism.

Old constructor without serial name should be eventually hidden.

Fixes #2958


core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -88,6 +90,23 @@ public class MissingFieldException(
...
+    /**
+     * Returns a serial name of a serializable class that cannot be deserialized due to missing fields.
+     * Typically, equal to the [SerialDescriptor.serialName] of the serializable class.
+     *
+     * However, in cases the class was compiled with an old Kotlin serialization plugin,
+     * its serial name may be unavailable and this property is `null`.
+     */
+    public val serialName: String? = serialName
+
+    // Maybe we should not provide public ctor with message and cause at all?
```
sandwwraith: The more I look at this ctor, the more I think it shouldn't be here. Nobody really adds the cause to MFE (and if somebody really wants to do this, there's `initCause`), and I do not know any use case for a non-standard message. The default message makes sure there are missing fields and serialName mentioned in it.
So I'm inclined to remove it. @fzhinkin WDYT?

fzhinkin: Sure, let's remove it. If we're missing something and there'll be a significant demand in it, we can always provide such a constructor later.

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
```diff
@@ -71,15 +71,17 @@ public open class SerializationException : IllegalArgumentException {
  * [MissingFieldException] is constructed from the following properties:
  * - [missingFields] -- fields that were required for the deserialization but have not been found.
  *   They are always non-empty and their names match the corresponding names in [SerialDescriptor.elementNames]
- * - Optional `serialName` -- serial name of the enclosing class that failed to get deserialized.
+ * - `serialName` -- serial name of the enclosing class that failed to get deserialized.
```
fzhinkin:
```suggestion
 * - [serialName] -- a serial name of the enclosing class that failed to get deserialized.
```

sandwwraith: Added manually

core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt
sandwwraith: Done

Issue #3123, opened 2025-12-02
List of community-supported serialization formats is outdated


Stream29:
formats/README.md listed some libraries from the community, and lots of them are no more actively maintained or not usable:

- https://github.com/jershell/kbson
- https://github.com/charleskorn/kaml
- https://github.com/him188/yamlkt
- https://github.com/L-Briand/obor
- https://github.com/dimitark/kotlinx-serialization-ion
- https://github.com/AhmedMourad0/bundlizer
- https://github.com/xn32/json5k
- https://github.com/codanbaru/dynamap
- https://github.com/jsoizo/kotlinx-serialization-php

They may be removed from the doc.

sandwwraith:
Really sad to hear that Kaml is no longer maintained; it was quite popular in the community.
I am not aware of the specific criteria for 'not maintained' except when the repository owner explicitly states so; therefore, I believe that removing only explicitly archived repositories (#3124) is the most meaningful action.

sandwwraith:
Fixed by #3124 and #3135

PR #3124, opened 2025-12-03
Remove archived repositories from the formats list

fzhinkin:
Related issues: #3123


sandwwraith:
It should have been probably merged to `master`, but it will get there during the release

PR #3125, opened 2025-12-12
feat: adding new page for Transform JSON structure

daniCsorbaJB:
This PR adds the last JSON page to the rewritten Kotlin serialization docs

Related Youtrack ticket: [KT-81979](https://youtrack.jetbrains.com/issue/KT-81979)

### DIFF
--- a/docs-website/topics/serialization-transform-json.md
+++ b/docs-website/topics/serialization-transform-json.md
@@ -1,12 +1,12 @@
 [//]: # (title: Transform JSON structure)
 <primary-label ref="advanced"/>
 
-To control the structure and content of the JSON you generate during serialization, you can create [custom serializers](create-custom-serializers.md).
-For smaller adjustments, the [`JsonTransformingSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/) class offers a simpler way to modify JSON by working directly with the JSON element tree instead of interacting with [`Encoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-encoder/)
+To control the structure and content of the JSON generated during serialization, you can create [custom serializers](create-custom-serializers.md).
+For smaller adjustments, such as wrapping values or unwrapping arrays, the [`JsonTransformingSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/) class provides a simpler way to modify JSON by working directly with the JSON element tree instead of using the [`Encoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-encoder/)
 or [`Decoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx.serialization-core/kotlinx.serialization.encoding/-decoder/) manually.
 
-> These sections build on concepts detailed in [Create custom serializers](create-custom-serializers.md).
-> If you're not familiar with custom serializers, we recommend reviewing that page first.
+> These sections are based on concepts explained in [Create custom serializers](create-custom-serializers.md).
+> If you're not familiar with custom serializers, we recommend reading that page first.
 >
 {style="note"}
 
@@ -17,12 +17,12 @@ In addition to transforming JSON structures, you can also use `JsonContentPolymo
 
 ## Modify JSON structure
 
-You can make structural adjustments to JSON by transforming the JSON element tree.
+You can adjust the JSON structure by transforming the JSON element tree.
 The following examples demonstrate common use cases, such as wrapping or unwrapping arrays and omitting specific properties.
 
 ### Wrap a single object in an array during deserialization
 
-Some APIs return a single JSON object when there is one item and an array when there are multiple.
+Some APIs return a single JSON object for one item and a JSON array for multiple items.
 To deserialize both cases into a [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/):
 
 1. Create a subclass of `JsonTransformingSerializer` and specify a serializer in its constructor.
@@ -114,14 +114,12 @@ fun main() {
 
 ### Omit specific properties during serialization
 
-You can use `JsonTransformingSerializer` when you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value.
-
-To do so:
+If you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value, use `JsonTransformingSerializer`:
 
 1. Create a subclass of `JsonTransformingSerializer` and specify a serializer in its constructor.
 2. Override the `transformSerialize()` function.
 
-Here's an example:
+Here's an example where the `Project` class has a `language` property, which is omitted when its value is `"Kotlin"`:
 
 ```kotlin
 // Imports declarations from the serialization library
@@ -157,8 +155,6 @@ fun main() {
 ```
 {kotlin-runnable="true"}
 
-In this example, the `Project` class has a `language` property, which is omitted when its value is `"Kotlin"`.
-
 > When serializing an object directly, you need to explicitly pass the custom serializer to the `encodeToString()`
 > function to ensure that the custom serialization logic is applied. For more information, see the [Pass serializers manually](third-party-classes.md#pass-serializers-manually) section.
 >
@@ -168,10 +164,10 @@ In this example, the `Project` class has a `language` property, which is omitted
 
 In [polymorphic serialization](serialization-polymorphism.md), JSON often contains a dedicated _class discriminator_ property that identifies the concrete subtype during deserialization.
 
-When no class discriminator is present in the JSON input, you can use [`JsonContentPolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-content-polymorphic-serializer/) to infer the type from the structure of the JSON.
-This serializer allows you to override the `selectDeserializer()` function to choose the correct deserializer based on the JSON content.
+If the JSON input doesn't contain a class discriminator, you can use [`JsonContentPolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-content-polymorphic-serializer/) to infer the type from the structure of the JSON.
+This serializer lets you override the `selectDeserializer()` function to choose the correct deserializer based on the JSON content.
 
-Let's look at an example where all values share a common base type with a `name` property:
+Here's an example where all values share a common base type with a `name` property:
 
 ```kotlin
 @Serializable
@@ -186,12 +182,12 @@ data class BasicProject(override val name: String): Project()
 data class OwnedProject(override val name: String, val owner: String) : Project()
 ```
 
-> In this example, the serializer decides which subtype to use based on the JSON content, so you don't need a `sealed` class for the class hierarchy.
+> In this example, the serializer chooses which subtype to use based on the JSON content, so you don't need a `sealed` class for the class hierarchy.
 >
 {style="note"}
 
-To distinguish between `BasicProject` and `OwnedProject`, override the `selectDeserializer()` function to select a serializer
-based on the presence of the `owner` key in the JSON object:
+To distinguish between `BasicProject` and `OwnedProject`, override the `selectDeserializer()` function.
+You can use this function to check whether the JSON object contains the `owner` key, and return the corresponding serializer:
 
 ```kotlin
 // Creates a custom serializer that selects deserializer based on the presence of "owner"
@@ -205,7 +201,7 @@ object ProjectSerializer : JsonContentPolymorphicSerializer<Project>(Project::cl
 ```
 
 When you use this serializer to serialize data, Kotlin serialization uses the serializer of the value's actual runtime type.
-This could be the one [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:
+This could be the serializer [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:
 
 ```kotlin
 // Imports declarations from the serialization library
@@ -258,7 +254,7 @@ fun main() {
 
 You can add custom behavior to the default serializer that Kotlin serialization generates, by using the default serializer as a delegate.
 
-To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically created `generatedSerializer()` as the base serializer in your custom `JsonTransformingSerializer`.
+To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically generated `generatedSerializer()` as the base serializer in your custom `JsonTransformingSerializer`.
 
 Here's an example that updates the JSON structure during deserialization by combining multiple input properties into a single `name` property expected by the target class:
 
@@ -334,7 +330,8 @@ Using these APIs, you can implement two-stage conversions:
 * Decode the input into a `JsonElement` first and then convert that element into a Kotlin value.
 * Convert a Kotlin value into a `JsonElement` first and then encode that element with the encoder.
 
-Here's an example that shows how to implement a custom `KSerializer` to fully control how a type is encoded and decoded in JSON:
+Let's look at an example of a custom `KSerializer` that fully controls how values of a `Response` type are encoded and decoded in JSON.
+This serializer encodes an `Ok` response directly as a JSON value and an `Error` response as a JSON object that contains the error message:
 
 ```kotlin
 // Imports declarations from the serialization library
@@ -416,8 +413,6 @@ fun main() {
 ```
 {kotlin-runnable="true"}
 
-In this example, the `Response` class has a custom serializer that handles `Ok` values directly as JSON values, but handles `Error` values as JSON objects containing the error message.
-
 ### Preserve unknown JSON attributes
 
 A common use case for a custom JSON-specific serializer is preserving JSON properties from the input that your serializable class doesn't define.
@@ -475,7 +470,6 @@ fun main() {
 {kotlin-runnable="true"}
 
 In this example, the preserved JSON properties remain at the same level within the input JSON object as the properties defined in the serializable class.
-
 
 ## What's next
 


docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
+[//]: # (title: Transform JSON structure)
+<primary-label ref="advanced"/>
+
+> This section builds on concepts from [Serialize polymorphic classes](serialization-polymorphism.md) and [Create custom serializers](create-custom-serializers.md).
```
sandwwraith: 1. Polymorphism required only for one section of the document, let's remove it from general requirements.
2. Let's not start the page with the note, move it one paragraph down

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+> If you're not familiar with these concepts, we recommend reviewing them first.
+>
+{style="note"}
+
+To control the structure and content of the JSON you generate during serialization, you can create [custom serializers](create-custom-serializers.md).
+For smaller adjustments, the [`JsonTransformingSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/) class offers a simpler way to modify JSON by working directly with the JSON element tree instead of interacting with [`Encoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-encoder/)
+or [`Decoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx.serialization-core/kotlinx.serialization.encoding/-decoder/) manually.
+
+The `JsonTransformingSerializer` implements the [`KSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/) interface, which
```
sandwwraith: > implements the KSerializer interface, which lets you override the transformSerialize() and transformDeserialize() functions

Such wording creates impression that `KSerializer` has transformSerialize and transformDeserialize functions, which is not true —  JsonTransformingSerializer is the interface that adds them. Please rewrite the sentence.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+lets you override the [`transformSerialize()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/transform-serialize.html) and [`transformDeserialize()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/transform-deserialize.html) functions to adjust the JSON element tree either before serialization or before deserialization.
+
+> In addition to transforming JSON structures, you can also use `JsonContentPolymorphicSerializer` to [select the appropriate polymorphic class based on the JSON content](#select-the-appropriate-polymorphic-class-based-on-the-json-content).
```
sandwwraith: IMO `style=tip` is unnecessary and adds clutter, you can just make a paragraph out of it.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+> 
+{style="tip"} 
+
+## Modify JSON structure
+
+You can make small structural adjustments to JSON by transforming the JSON element tree.
```
sandwwraith:
```suggestion
You can make structural adjustments to JSON by transforming the JSON element tree.
```

You can make any adjustments you like ;)

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+The following examples demonstrate common use cases, such as wrapping or unwrapping arrays and omitting specific properties.
+
+### Wrap a single object in an array during deserialization
+
+Some APIs return a single JSON object when there is one item and an array when there are multiple.
+If you want both cases to deserialize into a [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/), use
+`JsonTransformingSerializer` to modify the JSON element tree during deserialization.
+To implement the wrapping logic, override the `transformDeserialize()` function.
+
+> You must specify a serializer when calling the `JsonTransformingSerializer` constructor.
```
sandwwraith: Again, not sure if this should be highlighted as a note or can be just a regular paragraph. After all, _some_ serializer is a mandatory parameter in the `JsonTransformingSerializer`'s constructor.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+// Uses UserListSerializer to handle the serialization of the users property
+@Serializable
+data class Project(
+    val name: String,
+    // Specifies a custom serializer for the users property
```
sandwwraith: I think "Uses UserListSerializer to handle the serialization of the users property" or "Specifies UserListSerializer to handle the serialization of the users property" fits here more. Comment for the whole class is probably unnecessary.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    @Serializable(with = UserListSerializer::class)
+    val users: List<User>
+)
+
+@Serializable
+data class User(val name: String)
+
+// Specifies the default List serializer
```
sandwwraith: We specify default list serializer only in ctor super call, I think this comment would be misleading. WDYT of "Creates transforming serializer that delegates to default List serializer" or "Creates serializer that transforms results of default List serializer"?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+object UserListSerializer : JsonTransformingSerializer<List<User>>(ListSerializer(User.serializer())) {
+    // If not an array, wrap the element as a single-item array
+    override fun transformDeserialize(element: JsonElement): JsonElement =
+        if (element !is JsonArray) JsonArray(listOf(element)) else element
+}
+
+fun main() {
+    // Deserializes a single JSON object wrapped as an array
```
sandwwraith:
```suggestion
    // Deserializes a single JSON object and wraps it into an array
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+data class Project(
+  val name: String,
+  @Serializable(with = UserListSerializer::class)
+  val users: List<User>
+)
+
+@Serializable
+data class User(val name: String)
+
+// Unwraps single-element lists into a single object during serialization
```
sandwwraith: Let's be consistent with previous example: "Creates serializer that transforms results of default List serializer" or something similar. We have `// Unwraps single-element lists into a single JSON object` near `return element.singleOrNull() ?: element`, I believe that is enough?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","users":{"name":"kotlin"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Omit specific properties during serialization
+
+You can omit properties from the JSON output. This can be useful when they have default values, match specific values, or when the values are missing.
```
sandwwraith: I do not understand `This can be useful when they have default values, match specific values, or when the values are missing.` sentence at all. When property has a default value, it is omitted automatically by framework. Original wording was "Imagine that you cannot specify a default value for the language property in the Project data model for some reason", indicating that property _does not_ have default Kotlin value. "Match specific values" also unclear to me.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+    // Applies the custom serializer to omit the language property 
+    println(Json.encodeToString(ProjectSerializer, data))
+    // {"name":"kotlinx.serialization"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the `Project` class has a `language` property, which is omitted when its value is `"Kotlin"`.
```
sandwwraith: I think it worth moving this paragraph before the example, otherwise there's too much code without explanation of the high-level goal.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+The custom serializer is passed to `encodeToString()` to ensure the custom transformation is applied.
+
+> When serializing an object directly, you need to explicitly pass the custom serializer to the `encodeToString()`
+> function to ensure that the custom serialization logic is applied. For more information, see the [Pass serializers manually](third-party-classes.md#pass-serializers-manually) section.
+>
+{style="note"}
+
+## Select the appropriate polymorphic class based on the JSON content
+
+In [polymorphic serialization](serialization-polymorphism.md), JSON often includes a dedicated _class discriminator_ property that identifies the concrete subtype during deserialization.
```
sandwwraith:
```suggestion
In [polymorphic serialization](serialization-polymorphism.md), JSON often contains a dedicated _class discriminator_ property that identifies the concrete subtype during deserialization.
```

IMO "includes" is a more "action" verb — _someone_ (author, encoder, etc) can include something, but JSON (as a format, not our library) cannot perform any action by itself.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+When no class discriminator is present in the JSON input, you can use [`JsonContentPolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-content-polymorphic-serializer/) to infer the type from the structure of the JSON.
+This serializer allows you to override the `selectDeserializer()` function to choose the correct deserializer based on the JSON content.
+
+> When you use this serializer, the appropriate deserializer is chosen at runtime.
```
sandwwraith: I'm not sure what you mean by this tip. You can return any serializer you want from `selectDeserializer`, it can be registered or not registered, default or non-default.

sandwwraith: I think you got confused with 
> When you use this serializer to serialize data, either [registered](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#registered-subclasses) or the default serializer is selected for the actual type at runtime:

note later. What it means if you use this serializer to **encode** the data, it will try to pick the appropriate one for you, as `selectDeserializer` affects only **decoding**. Maybe you can rewrite the tip in this direction?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+@Serializable
+abstract class Project {
+    abstract val name: String
+}
+
+@Serializable
+data class BasicProject(override val name: String): Project()
+
+@Serializable
+data class OwnedProject(override val name: String, val owner: String) : Project()
```
sandwwraith, line 222: And here I start liking old Knit docs more, as they allow snippets to be broken down by more explanations, instead of one long example.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    println(Json.decodeFromString(ListSerializer(ProjectSerializer), string))
+    // [OwnedProject(name=kotlinx.serialization, owner=kotlin), BasicProject(name=example)]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the serializer decides which subtype to use based on the JSON content, so you don't need a `sealed` class for the class hierarchy.
+
+## Add custom behavior to the default serializer
```
sandwwraith, line 252: Are you going to put "Experimental" badge on the section like in other PRs?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+You can add custom behavior to the default serializer that Kotlin serialization generates.
+
+To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically created `generatedSerializer()` as the base serializer in your custom implementation:
```
sandwwraith: Maybe we should recommend this page: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/compatibility.md#experimental-api instead of the default one. It provides more information about our binary compatibility policy for experimental apis. Our "Experimental" is more like "Alpha/Beta" in Kotlin terms.

sandwwraith:
```suggestion
To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically created `generatedSerializer()` as the base serializer in your custom transformer:
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the serializer decides which subtype to use based on the JSON content, so you don't need a `sealed` class for the class hierarchy.
+
+## Add custom behavior to the default serializer
+
+You can add custom behavior to the default serializer that Kotlin serialization generates.
```
sandwwraith:
```suggestion
You can add custom behavior to the default serializer that Kotlin serialization generates, by using the default serializer as a delegate.
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    val project = Json.decodeFromString<Project>("""{"type":"basic","basic-name":"example"}""")
+
+    println(project)
+    // BasicProject(name=example)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the custom serializer updates the JSON structure during deserialization by renaming a field so it matches the `name` property defined in the base class.
```
sandwwraith: Again, explanation "what's happening" is better be placed before example.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+<!-- Would another example like this work? to focus more on the custom behavior not the polymorphism:
```
sandwwraith: This one would work as well, if that's what you wanted to ask

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically created `generatedSerializer()` as the base serializer in your custom implementation:
+
+```kotlin
+// Imports declarations from the serialization library
+import kotlinx.serialization.*
+import kotlinx.serialization.builtins.*
+import kotlinx.serialization.json.*
+
+//sampleStart
+// Defines a sealed hierarchy where each subtype has a name property
```
sandwwraith: Given that we indeed do not touch polymorphism much here, this comment can be dropped. Let's just say that we want everything to have a "name".

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+-->
+
+## Implement custom serialization logic in JSON
+<primary-label ref="experimental-general"/>
+
+If the transformation functions provided by `JsonTransformingSerializer` or
+`JsonContentPolymorphicSerializer` aren't enough, you can implement custom
+serialization logic by defining your own [`KSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/) class.
+
+This gives you full control over how values are serialized and deserialized by overriding the
```
sandwwraith: "This gives you full control ... by overriding.. " sounds clumsy (subjects in parts are not the same). Maybe "You can get full control ... by overriding ..."?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+`serialize()` and `deserialize()` functions directly.
+
+When you implement custom serialization logic for JSON, you can cast `Encoder` to
+`JsonEncoder` and `Decoder` to `JsonDecoder` to call JSON-specific functions such as
```
sandwwraith: Maybe also insert API links to the classes themselves? E.g. https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/

sandwwraith:
```suggestion
`JsonEncoder` and `Decoder` to `JsonDecoder` to call JSON-specific functions
```

Let's remove "such as", they're basically the only functions available. You also explain them in the next sentence.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+[`decodeJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/decode-json-element.html) and [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html).
+These functions allow you to retrieve and insert JSON elements at specific points in the serialization process.
```
sandwwraith: "at specific points" is very vague. Original wording was "current position", and that's what they do — immediately retrieve or insert JsonElements. Please stick to the original meaning.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+Both `JsonDecoder` and `JsonEncoder` expose a `json` property that gives access to the active `Json` instance, which controls how values are encoded and decoded.
+Through that instance, you can use [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html) and [`decodeFromJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-from-json-element.html) to convert
+between `JsonElement` instances and Kotlin objects.
+
+Using these APIs, you can implement two-stage conversions. For example, you can:
```
sandwwraith:
```suggestion
Using these APIs, you can implement two-stage conversions:
```

"For example" is an unnecessary over-strech, since there are hardly more use cases than these two.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+
+// Defines a sealed class for API responses
+@Serializable(with = ResponseSerializer::class)
+sealed class Response<out T> {
+    data class Ok<out T>(val data: T) : Response<T>()
+    data class Error(val message: String) : Response<Nothing>()
+}
+
+// Implements custom serialization logic for Response
+@OptIn(InternalSerializationApi::class)
```
sandwwraith: I think it's better to stick with the version that does not use InternalApi and uses `buildClassSerialDescriptor` (even though it's not completely idiomatic)

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    println(json)
+    // [{"name":"kotlinx.serialization"},{"error":"Not found"}]
+
+    println(Json.decodeFromString<List<Response<Project>>>(json))
+    // [Ok(data=Project(name=kotlinx.serialization)), Error(message=Not found)]
+}
+```
+{kotlin-runnable="true"}
+
+In this example, the `Response` class has a custom serializer that handles `Ok` values directly as JSON values, but handles `Error` values as JSON objects containing the error message.
```
sandwwraith: Again, better to explain to users what we're trying to do before dumping code on them.

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    override fun deserialize(decoder: Decoder): UnknownProject {
+        // Ensures the decoder is JSON-specific
+        val jsonInput = decoder as? JsonDecoder ?: error("Can be deserialized only by JSON")
+
+        // Reads the entire content as JSON
+        val json = jsonInput.decodeJsonElement().jsonObject
+
+        // Extracts and removes the name property
```
sandwwraith, line 450: Pay attention to comments and code: "removes the name property" is performed by `details.remove("name")` which is located in the next block. Move those two lines here.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I completely see what you mean. 🤔  I think we could break it into two to eliminate any potential confusion here:

> The `JsonTransformingSerializer` is a JSON-specific serializer that implements the `KSerializer` interface.
It provides the `transformSerialize()` and `transformDeserialize()` functions that you can override to adjust the JSON element tree before serialization or deserialization.

What do you think?

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: After reviewing it again, and thinking about it — I agree with you 👍 - let's change this to a paragraph.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: fair point 🤔 

We could instead use a steplist here (I adjusted the style a little to fit that format better - let me know if this works better): 

> Some APIs return a single JSON object when there is one item and an array when there are multiple.
> To deserialize both cases into a [`List`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/):
> 
> 1. Use `JsonTransformingSerializer` and specify a serializer in its constructor.
>    To use the standard conversion logic, pass the default serializer for the target type, such as [`ListSerializer()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.builtins/-list-serializer.html) for lists.
> 2. Override the `transformDeserialize()` function.
> 
> Here's an example:

docs-website/topics/serialization-transform-json.md
sandwwraith: Let's say `a JSON-specific abstract serializer`, to highlight that subclassing it is necessary.

docs-website/topics/serialization-transform-json.md
sandwwraith: I'd use more specific verb than 'use'. E.g. `Extend JsonTransformingSerializer` or `Subclass JsonTransformingSerializer`

daniCsorbaJB: Both are good options — we could go for 

> 1. Create a subclass of `JsonTransformingSerializer` and specify a serializer in its constructor....

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: good point 👍 — I like the second version let's go with that.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: good point I modified them to be consistent 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: 🤔 oh I see it — I wanted to summarize loosely related things here (e.g.: match specific values = omit property if it equals/matches a value)

> cannot specify a default value

I missed this logical step, but it helped, thank you... based on that WDYT about something very straightforward like:

> You can use `JsonTransformingSerializer` when you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value.
>
>To do so:
>
>1. Create a custom serializer and override the `transformSerialize()` function. 
>2. Pass the custom serializer to `encodeToString()` to apply the custom transformation.

>Here's an example:

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I agree with your reasoning, great point. Let's go with contains 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I moved this part about passing the serializer to `encodeToString()` to apply the custom transformation.

But in my opinion, the part about what's happening in the example is fine afterward. It's easier to reference actual values after the reader's seen before.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: Interesting, thank you for raising this 🤔 — Off the top of my head, I see no reason not to change it. I'll double-check just in case, but if I see nothing I'll change it in the other docs as well; and I think that it would be a nice improvement for clarity.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: fair point — we could theoretically break it down and put the actual runnable example in just as:

```
//rest of the code before
//sampleStart
fun main() {
    val data = listOf(
        OwnedProject("kotlinx.serialization", "kotlin"),
        BasicProject("example")
    )
    val string = Json.encodeToString(ListSerializer(ProjectSerializer), data)
    println(string)
    println(Json.decodeFromString(ListSerializer(ProjectSerializer), string))
}
//sampleEnd
```

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: 🤔 I got confused by the original here a bit, but I see what you mean. Do you think something like this would work? 

> These functions allow you to retrieve and insert JSON elements for the value the serializer is currently handling.

or perhaps a slightly longer but a bit more descriptive:

> These functions allow you to retrieve and insert JSON elements at the position of the value currently handled by the serializer.

___
For reference I think with the previous version the sentence would be confusing (at least to me 🤔 )
These functions allow you to retrieve and insert JSON elements at the current position in the serialization process.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: nice — good catch - added them

docs-website/topics/serialization-transform-json.md
sandwwraith: Yes, this one is much easier to read

docs-website/topics/serialization-transform-json.md
sandwwraith: Yeah, let's try that

docs-website/topics/serialization-transform-json.md
sandwwraith: I always found structure "Let's try to do X, here you can see how" easier to grok rather than "Here's the code, read it, and only after you've tried to figure it out by yourself, we'll tell you what it's supposed to do". Perhaps it is indeed subject to taste, but I do not see "Example first, the idea after" anywhere else in the docs

docs-website/topics/serialization-transform-json.md
sandwwraith: Well, "the value the serializer is currently handling" is not 100% correct, because there's no "value" yet — only an unparsed stream of characters, and the position(string index) we are currently looking at. But you're probably right, low-level descriptions may confuse some users. So we can stick with "the value the serializer is currently handling".

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I agree with you here, we should introduce the idea first, and that introducing how the example achieves that is mostly a matter of style.

What I'm trying to avoid is putting too much text between the instructions and showing the example. Doing the opposite isn't wrong either, but I found that it's usually easier to grasp a concept if it's almost immediately followed up by an example, and then the example is summarized afterward, bridging the space between the instructions and the example. Also, it mostly depends on whether it's clear what the example is trying to showcase. 

In the docs we have both approaches, I usually aim for this:

1. Introducing the idea
2. General instructions
3. An example
4. Example summary

So in this specific section I think we now follow this logic.

1. You can use `JsonTransformingSerializer` when you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value.

2. To do so: (the two steps)

3. Example

4. In this example, the `Project` class has a `language` property, which is omitted when its value is `"Kotlin"`.

Would something like this work as well?

1. Introducing the concept
2. General instructions
3. Example summary
4. An example

Yes, then the wording would be something like: "Here's an example where the `Project` class has a `language` property, which is omitted when its value is `"Kotlin"`.

Overall, my preference here depends on how much text is there between the instructions and the example. If it's one (or two property names) I think it's fine. If it's more, I tend to favor putting it after the example. But I don't think either approach is inherently better or worse; it mostly depends on the readers. 

I think we can gather user feedback and rewrite it as needed afterward.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I agree — let's move this line above

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: nice 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: Thank you 🙏 yes definitely, how about:

> When you use this serializer to serialize data, Kotlin serialization uses the serializer of the value's actual runtime type.
> This could be one you [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I agree & like it — I changed it 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: 🤔 I prefer it as summary here, but let's keep this open and revisit perhaps during TWr review? 🤔

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: Good point — I rewrote it with `buildClassSerialDescriptor`

docs-website/topics/serialization-transform-json.md
```diff
@@ -44,33 +38,30 @@ import kotlinx.serialization.builtins.*
...
     val users: List<User>
 )
 
 @Serializable
 data class User(val name: String)
 
-// Specifies the default List serializer
+// Creates a serializer that transforms the results of the default List serializer
 object UserListSerializer : JsonTransformingSerializer<List<User>>(ListSerializer(User.serializer())) {
-    // If not an array, wrap the element as a single-item array
```
sandwwraith: I think original comment fits `if (element !is JsonArray)` more

docs-website/topics/serialization-transform-json.md
```diff
@@ -44,33 +38,30 @@ import kotlinx.serialization.builtins.*
...
 }
 
 fun main() {
-    // Deserializes a single JSON object wrapped as an array
     println(Json.decodeFromString<Project>("""
         {"name":"kotlinx.serialization","users":{"name":"kotlin"}}
     """))
     // Project(name=kotlinx.serialization, users=[User(name=kotlin)])
-
-    // Deserializes a JSON array of objects
```
sandwwraith: Why did you remove these two?

docs-website/topics/serialization-transform-json.md
sandwwraith: bump

docs-website/topics/serialization-transform-json.md
```diff
@@ -123,10 +114,14 @@ fun main() {
...
 
-You can omit properties from the JSON output. This can be useful when they have default values, match specific values, or when the values are missing.
-This helps streamline the data, reducing unnecessary information while ensuring that only relevant properties are serialized.
+You can use `JsonTransformingSerializer` when you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value.
 
-To omit properties during serialization, create a custom serializer and override the `transformSerialize()` function to remove the fields you don't want in the JSON output:
+To do so:
+
+1. Create a custom serializer and override the `transformSerialize()` function. 
+2. Pass the custom serializer to `encodeToString()` to apply the custom transformation.
```
sandwwraith: You already have "When serializing an object directly, you need to explicitly pass the custom serializer to the `encodeToString()`" tip below the example. Listing it as a point in the list may create an expression that it is the only way, while typically it is more used in `@Serializable(ProjectSerializer)` situation.

Let's rewrite this list to be consistent with a previous example:

```
1. Create a subclass of `JsonTransformingSerializer` and specify a serializer in its constructor.
2. Because we are now changing how the object is encoded, override `transformSerialize()` function.
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,481 @@
...
+    override fun selectDeserializer(element: JsonElement) = when {
+        // Selects the OwnedProject serializer if the JSON object contains an "owner" key
+        "owner" in element.jsonObject -> OwnedProject.serializer()
+        else -> BasicProject.serializer()
+    }
+}
+```
+
+When you use this serializer to serialize data, Kotlin serialization uses the serializer of the value's actual runtime type.
+This could be one you [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:
```
sandwwraith:
```suggestion
This could be the one you [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -229,79 +253,31 @@ fun main() {
...
 import kotlinx.serialization.json.*
 
 //sampleStart
+// Defines a type with a name property
 @Serializable
 data class Project(val name: String)
 
-// Adds custom logic to rename "projectName" to "name" during deserialization
+// Adds custom logic to the default serializer to rename a field during deserialization
 object ProjectSerializer : JsonTransformingSerializer<Project>(Project.serializer()) {
```
sandwwraith: Rewrite the example so it replaces default serializer, as in previous version:
```
@KeepGeneratedSerializer
@Serializable(with = BasicProjectSerializer::class)
...
Project.generatedSerializer()
...
Json.decodeFromString<Project>(...)
```

docs-website/topics/serialization-transform-json.md
sandwwraith: Are you gonna do it now or later?

docs-website/topics/serialization-transform-json.md
```diff
@@ -229,79 +253,31 @@ fun main() {
...
     override fun transformDeserialize(element: JsonElement): JsonElement {
         val jsonObject = element.jsonObject
 
-        // If "projectName" is present, map it to "name"
+        // Renames "projectName" to "name" if it exists in the JSON input
```
sandwwraith: Giving this example a second thought, it looks too much like the existing `@JsonNames` annotation, and we shouldn't recommend a custom serializer for problems that can be solved with an annotation. Maybe make the example just a little bit more complicated? E.g. by transforming two properties `firstName` and `lastName` into a single `name`.

docs-website/topics/serialization-transform-json.md
```diff
@@ -341,22 +315,22 @@ If the transformation functions provided by `JsonTransformingSerializer` or
...
 `serialize()` and `deserialize()` functions directly.
 
 When you implement custom serialization logic for JSON, you can cast `Encoder` to
-`JsonEncoder` and `Decoder` to `JsonDecoder` to call JSON-specific functions such as
+[`JsonEncoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-encoder/) and `Decoder` to [`JsonDecoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/) to call the JSON-specific functions
 [`decodeJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/decode-json-element.html) and [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html).
-These functions allow you to retrieve and insert JSON elements at specific points in the serialization process.
+These functions allow you to retrieve and insert JSON elements for the value the serializer is currently handling.
```
sandwwraith: `retrieve/insert` matches `from/into` prepositions, not `for`

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I'm waiting for an answer — let's keep this open! We can handle it in TWr review. I'll send you a message if I don't update it before merging it into the `doc-restructuring-master` branch.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I felt like these are a bit redundant together with the other comments we discussed: 🤔 

// Wraps a single JSON object into a single-element array
and
// Unwraps single-element lists into a single JSON object

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: You are right 👍 (also removed the "you" from this sentence: adding the article here would imply that the user already added it for this specific example)
> This could be the one [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: True, although "for" here isn't being used as the directional preposition counterpart to retrieve/insert.
Here, "for" just marks the target. (Like: This lets you retrieve and insert things for someone.)

That said if you think it might confuse people, we could/should break the sentence up (that way we could use the the directional prepositions): 

> These functions allow you to retrieve JSON elements from the value the serializer is currently handling, or insert JSON elements into it. 

or we could just be on the safer side and go more explicit: 

> You can use [`decodeJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/decode-json-element.html)
>  to retrieve a JSON element from the value the serializer is currently handling, and [`encodeJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html) to insert a JSON element into it.

 
 I think both could work, do you have a preference? 🤔

sandwwraith: Apologies, from/into matches decoder, not serializer. So the correct version is `These functions allow you to retrieve JSON elements from the value the decoder is currently handling, or insert JSON elements into it.`

docs-website/topics/serialization-transform-json.md
sandwwraith: Ok, the results are quite self-explanatory

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: good point 👍I agree let's make it more consistent.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: 👍 I can re-add it

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: Combining both your comments, would something like this work then? 🤔 

`
```kotlin
// Imports declarations from the serialization library

//sampleStart
@KeepGeneratedSerializer
@Serializable(with = UserNameSerializer::class)
// Defines a type with a name property
data class User(val name: String)

// Adds custom logic to the default serializer to combine input properties during deserialization
object UserNameSerializer : JsonTransformingSerializer<User>(User.generatedSerializer()) {
    override fun transformDeserialize(element: JsonElement): JsonElement {
        val jsonObject = element.jsonObject
        val first = jsonObject["firstName"]?.jsonPrimitive?.content
        val last = jsonObject["lastName"]?.jsonPrimitive?.content

        // Combines input properties into the name property
        // if the input doesn't match the expected structure
        return if (first != null && last != null) {
            JsonObject(mapOf("name" to JsonPrimitive("$first $last")))
        } else {
            jsonObject
        }
    }
}

fun main() {
    // Deserializes JSON where the name property is split across multiple input properties
   val fromExternalData = Json.decodeFromString<User>(
        """{"firstName":"John","lastName":"Smith"}"""
    )
    println(fromExternalData)
    // User(name=John Smith)

    // Deserializes JSON where the name property matches the expected structure
    val fromInternalData = Json.decodeFromString<User>(
        """{"name":"John Smith"}"""
    )
    println(fromInternalData)
    // User(name=John Smith)
}
//sampleEnd
```
{kotlin-runnable="true"}`

daniCsorbaJB:
> I see that you fixed all my comments. Great job! Are you going to add Compatibility policy/Experimental API page as we discussed in the separate PR?

Yes, let's circle back to that later! Thank you for your review! 🚀

sandwwraith:
I've filed https://youtrack.jetbrains.com/issue/KT-84383/Docs-Create-a-compatibility-policy-page-for-kotlinx.serialization so you won't forget

pdvrieze:
@sandwwraith There should probably also be some compatibility statement in relation to the supported formats (where the format configuration/serialization semantics are expected to be stable - perhaps not when configuration options are marked as experimental and don't have the default value).

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
+[//]: # (title: Transform JSON structure)
+<primary-label ref="advanced"/>
+
+To control the structure and content of the JSON you generate during serialization, you can create [custom serializers](create-custom-serializers.md).
```
koshachy:
```suggestion
To control the structure and content of the JSON generated during serialization, you can create [custom serializers](create-custom-serializers.md).
```

koshachy: I'm a bit worried that it could make the sentence hard to read... WDYT?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+For smaller adjustments, the [`JsonTransformingSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/) class offers a simpler way to modify JSON by working directly with the JSON element tree instead of interacting with [`Encoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-encoder/)
```
koshachy: offers -> provides 

More natural to me.

koshachy: interacting -> using

koshachy: Does it work?

koshachy: The only thing is not clear for me from the sentence is what is "smaller adjustments". 
Will the audience understand this?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+or [`Decoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx.serialization-core/kotlinx.serialization.encoding/-decoder/) manually.
+
+> These sections build on concepts detailed in [Create custom serializers](create-custom-serializers.md).
```
koshachy:
```suggestion
> These sections build on concepts explained in [Create custom serializers](create-custom-serializers.md).
```

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+> If you're not familiar with custom serializers, we recommend reviewing that page first.
```
koshachy:
```suggestion
> If you're not familiar with custom serializers, we recommend reading that page first.
```

docs-website/topics/serialization-transform-json.md
koshachy: Another options are:

* This documentation is based on concepts...
* This documentation uses the concepts...

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+{style="note"}
+
+The `JsonTransformingSerializer` is a JSON-specific abstract serializer that implements the [`KSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/) interface.
+It provides the [`transformSerialize()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/transform-serialize.html) and [`transformDeserialize()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/transform-deserialize.html) functions that you can override to adjust the JSON element tree before serialization or deserialization.
+
+In addition to transforming JSON structures, you can also use `JsonContentPolymorphicSerializer` to [select the appropriate polymorphic class based on the JSON content](#select-the-appropriate-polymorphic-class-based-on-the-json-content).
+
+## Modify JSON structure
+
+You can make structural adjustments to JSON by transforming the JSON element tree.
```
koshachy: You can adjust the JSON structure?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+The following examples demonstrate common use cases, such as wrapping or unwrapping arrays and omitting specific properties.
+
+### Wrap a single object in an array during deserialization
+
+Some APIs return a single JSON object when there is one item and an array when there are multiple.
```
koshachy:
```suggestion
Some APIs return a single JSON object for one item and a JSON array for multiple items.
```

koshachy: I've tried to simplify a bit, WDYT?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+    println(Json.encodeToString(data))
+    // {"name":"kotlinx.serialization","users":{"name":"kotlin"}}
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Omit specific properties during serialization
+
+You can use `JsonTransformingSerializer` when you can't [specify a default value](serialization-customization-options.md#set-default-values-for-optional-properties), but still want to omit a property when it has a specific value.
```
koshachy: If you can't specify a default value, but still want to omit a preprety when ..., use `JsonTransformingSerializer`:

1. Create a subclass ....

koshachy: Does it work?
WDYT?

docs-website/topics/serialization-transform-json.md
koshachy: Personally, I like the version with the extended description: 

> Here's an example where the Project class has a language property, which is omitted when its value is "Kotlin":

But not sure that how it will look in the docs (and will it look better?)

koshachy: Especially, when I see that you use the same approach in the example below

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+> When serializing an object directly, you need to explicitly pass the custom serializer to the `encodeToString()`
+> function to ensure that the custom serialization logic is applied. For more information, see the [Pass serializers manually](third-party-classes.md#pass-serializers-manually) section.
+>
+{style="note"}
+
+## Select the appropriate polymorphic class based on the JSON content
+
+In [polymorphic serialization](serialization-polymorphism.md), JSON often contains a dedicated _class discriminator_ property that identifies the concrete subtype during deserialization.
+
+When no class discriminator is present in the JSON input, you can use [`JsonContentPolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-content-polymorphic-serializer/) to infer the type from the structure of the JSON.
```
koshachy: If the JSON input does not contain a class discriminator, ... from the JSON structure.

koshachy: I've tried to simplify, take a look

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+This serializer allows you to override the `selectDeserializer()` function to choose the correct deserializer based on the JSON content.
```
koshachy: Nitpicking: allows you to override -> lets you override

koshachy: Up to you tho

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+
+Let's look at an example where all values share a common base type with a `name` property:
```
koshachy: Or:

Here is a example ..

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+}
+
+@Serializable
+data class BasicProject(override val name: String): Project()
+
+@Serializable
+data class OwnedProject(override val name: String, val owner: String) : Project()
+```
+
+> In this example, the serializer decides which subtype to use based on the JSON content, so you don't need a `sealed` class for the class hierarchy.
```
koshachy: decides or chooses?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+>
+{style="note"}
+
+To distinguish between `BasicProject` and `OwnedProject`, override the `selectDeserializer()` function to select a serializer
```
koshachy: The second part is a bit hard to read... 
But I don't have a nice suggestion rn

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+    override fun selectDeserializer(element: JsonElement) = when {
+        // Selects the OwnedProject serializer if the JSON object contains an "owner" key
+        "owner" in element.jsonObject -> OwnedProject.serializer()
+        else -> BasicProject.serializer()
+    }
+}
+```
+
+When you use this serializer to serialize data, Kotlin serialization uses the serializer of the value's actual runtime type.
+This could be the one [specified in a `SerializersModule`](serialization-polymorphism.md#serialize-closed-polymorphic-classes) or the default serializer:
```
koshachy: one -> serializer

I know that there are many "serialize" words, but in this case I feel that we need it. Just not to mess up with "value's actual runtime type"

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+    // [OwnedProject(name=kotlinx.serialization, owner=kotlin), BasicProject(name=example)]
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+## Add custom behavior to the default serializer
+<primary-label ref="experimental-opt-in"/>
+
+You can add custom behavior to the default serializer that Kotlin serialization generates, by using the default serializer as a delegate.
```
koshachy, line 255: You can extend the default serializer

koshachy: Does it make sense? Or it changes the meaning?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+
+To do so, annotate a serializable class with the [Experimental](components-stability.md#stability-levels-explained) [`@KeepGeneratedSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-keep-generated-serializer/) and use the automatically created `generatedSerializer()` as the base serializer in your custom `JsonTransformingSerializer`.
```
koshachy: created or generated?
Let's try to stick to one word, so the difference will be undestandable:
create — when **the user** creates something
generate — when **the library** generates something

WDYT?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+<primary-label ref="experimental-general"/>
+
+If the transformation functions provided by `JsonTransformingSerializer` or
+`JsonContentPolymorphicSerializer` aren't enough, you can implement custom
+serialization logic by defining your own [`KSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/) class.
+
+You can get full control over how values are serialized and deserialized by overriding the
+`serialize()` and `deserialize()` functions directly.
+
+When you implement custom serialization logic for JSON, you can cast `Encoder` to
```
koshachy, line 319: This para is a bit hard to read because it has a lot of links (see the rendered version in Writerside).
They compete to each other, and takes the atterntion away.
WDYT, could we make it a bit more readable but without changing a lot?

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+Both `JsonDecoder` and `JsonEncoder` expose a `json` property that gives access to the active `Json` instance, which controls how values are encoded and decoded.
+Through that instance, you can use [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html) and [`decodeFromJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-from-json-element.html) to convert
+between [`JsonElement`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-element/) instances and Kotlin objects.
+
+Using these APIs, you can implement two-stage conversions:
+
+* Decode the input into a `JsonElement` first and then convert that element into a Kotlin value.
+* Convert a Kotlin value into a `JsonElement` first and then encode that element with the encoder.
+
+Here's an example that shows how to implement a custom `KSerializer` to fully control how a type is encoded and decoded in JSON:
```
koshachy:
```suggestion
Here's an example of how to implement a custom `KSerializer` to fully control how a type is encoded and decoded in JSON. The `Response` class has a custom serializer that handles `Ok` values directly as JSON values, but handles `Error` values as JSON objects containing the error message:
```

koshachy: Less verbose

koshachy: Moved to the intro

docs-website/topics/serialization-transform-json.md
koshachy: Since there is not much explanation, I'd like to move it before the codeblock

docs-website/topics/serialization-transform-json.md
```diff
@@ -0,0 +1,483 @@
...
+    // Deserializes JSON with properties not defined in the serializable class into UnknownProject
+    println(Json.decodeFromString(UnknownProjectSerializer, """{"type":"unknown","name":"example","maintainer":"Unknown","license":"Apache 2.0"}"""))
+    // UnknownProject(name=example, details={"type":"unknown","maintainer":"Unknown","license":"Apache 2.0"})
+
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the preserved JSON properties remain at the same level within the input JSON object as the properties defined in the serializable class.
```
koshachy, line 472: Well, I think the same rule could be applied here.

docs-website/topics/serialization-transform-json.md
koshachy: This section is hard to read because of this.
I'd like to revisit it a bit later.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: Moved it to the top 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: both changes work for me 👍

We can add some examples so it's a bit clearer (based on the ToC, people can infer this information, but adding it can clarify it further). 

> For smaller adjustments, such as wrapping values or unwrapping arrays,...

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: nice, I like "is based on concepts explained in..." 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I like it 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: it does — I like it - let's go with that 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I like it 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: both are perfect for me 😄 — If you prefer "lets you" - let's go with that 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: both work - but yes, I think chooses might fit better here as its a bit more concrete.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: hmm 🤔 do you think the issue comes from repeating selects?
WDYT about breaking the sentence in two? 

> To distinguish between `BasicProject` and `OwnedProject`, override the `selectDeserializer()` function. You can use this function to check whether the JSON object contains the `owner` key, and return the corresponding serializer:

or just a little cleaned up version:

> To distinguish between `BasicProject` and `OwnedProject`, override the `selectDeserializer()` function and check whether the JSON object contains the `owner` key to return the corresponding serializer:

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: > "value's actual runtime type"

Good point 👍 switched it

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I think it does 🤔 to me, it might read as it involves inheritance, so I prefer the original here.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: good point — I was trying to avoid repeating generated "`@KeepGeneratedSerializer` and use generated `generatedSerializer()`" but I agree it's better this way 👍

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: 🤔 I'm not sure we can get rid of the links, beacuse this is the first time users see these APIs here. But perhaps we could break the paragraph up a bit. I'm a bit on the fence, though because these ideas should belong together: 

> When you implement custom serialization logic for JSON, you can cast `Encoder` to
> [`JsonEncoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-encoder/) and `Decoder` to [`JsonDecoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/) to call the JSON-specific functions
> [`decodeJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-decoder/decode-json-element.html) and [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html).
> 
> These functions allow you to retrieve JSON elements from the value the decoder is currently handling, or insert JSON elements into it.
> 
> Both `JsonDecoder` and `JsonEncoder` expose a `json` property that gives access to the active `Json` instance, which controls how values are encoded and decoded.
> 
> Through that instance, you can use [`encodeToJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html) and [`decodeFromJsonElement()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-from-json-element.html) to convert
> between [`JsonElement`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-element/) instances and Kotlin objects.

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: It feels a bit heavier for me this way, so I'm thinking, we could rewrite it slightly. 🤔 

WDYT about using Let's look at an example start instead. Maybe I was just trained by seeing this sentence structure so much, I expect "Here's an example to be immediately followed up by an example," and the additional sentence works better for me with the Let's look at an example structure. 

So it could be something like this, WDYT:

> Let's look at an example of a custom `KSerializer` that fully controls how values of a `Response` type are encoded and decoded in JSON.
> This serializer encodes an `Ok` response directly as a JSON value and an `Error` response as a JSON object that contains the error message:

docs-website/topics/serialization-transform-json.md
daniCsorbaJB: I'm not sure it's needed here 🤔  It doesn't add new context before the example, and it would mostly repeat the intro.  The only new information is that the JSON properties remain at the same level, but I think that works better after the example.

docs-website/topics/serialization-transform-json.md
koshachy: WFM

PR #3036, opened 2025-07-05
Structured CBOR. Fixes #2975, fixes #2848, fixes #3143

JesusMcCloud:
fixes #2975, #2848

This PR introduces structured CBOR encoding and decoding

## Encoding from/to `CborElement`

Bytes can be decoded into an instance of `CborElement` with the [Cbor.decodeFromByteArray] function by either manually
specifying [CborElement.serializer()] or specifying [CborElement] as generic type parameter.  
It is also possible to encode arbitrary serializable structures to a `CborElement` through [Cbor.encodeToCborElement].

Since these operations use the same code paths as regular serialization (but with specialized serializers), the config flags
behave as expected

## Newly introduced CBOR-specific structures

* [CborPrimitive] represents primitive CBOR elements, such as string, integer, float boolean, and null.
  CBOR byte strings are also treated as primitives  
  Each primitive has a [value][CborPrimitive.value]. Depending on the concrete type of the primitive, it maps
  to corresponding Kotlin Types such as `String`, `Int`, `Double`, etc.
  Note that Cbor discriminates between positive ("unsigned") and negative ("signed") integers!  
  `CborPrimitive` is itself an umbrella type (a sealed class) for the following concrete primitives:
  * [CborNull] mapping to a Kotlin `null`
  * [CborBoolean] mapping to a Kotlin `Boolean`
  * [CborInt] which is an umbrella type (a sealed class) itself for the following concrete types
    (it is still possible to instantiate it as the `invoke` operator on its companion is overridden accordingly):
    * [CborPositiveInt] represents all `Long` numbers `≥0`
    * [CborNegativeInt] represents all `Long` numbers `<0`
  * [CborString] maps to a Kotlin `String`
  * [CborFloat] maps to Kotlin `Double`
  * [CborByteString] maps to a Kotlin `ByteArray` and is used to encode them as CBOR byte string (in contrast to a list
    of individual bytes)

* [CborList] represents a CBOR array. It is a Kotlin [List] of `CborElement` items.

* [CborMap] represents a CBOR map/object. It is a Kotlin [Map] from `CborElement` keys to `CborElement` values.
  This is typically the result of serializing an arbitrary 

# Example

```hexdump
bf                                 # map(*)
   61                              #   text(1)
      61                           #     "a"
   cc                              #   tag(12)
      1a 0fffffff                  #     unsigned(268,435,455)
   d8 22                           #   base64 encoded text, tag(34)
      61                           #     text(1)
         62                        #       "b"
[...]

PR #3126, opened 2025-12-15
CBOR: Optionally encode nullable complex props as empty map (fixes #2848)

### LINKED ISSUE #2848: CBOR serializer encodes null class instance as empty map instead of null
Traced to this line: https://github.com/Kotlin/kotlinx.serialization/blob/d4d066d72a9f92f06c640be5a36a22f75d0d7659/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoder.kt#L106

I consider this a bug. Null objects should be encoded as the simple type `null`, no matter whether they are class types or other types. 

In our use case, this is causing issues after deserialization, since the empty map is deserialized into a default instance of the type, rather than `null`.

Please consider removing this line and always encoding as `null`.

JesusMcCloud:
See discussion in #2848

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -16,6 +16,7 @@ stable, these are currently experimental features of Kotlin Serialization.
   * [Definite vs. Indefinite Length Encoding](#definite-vs-indefinite-length-encoding)
   * [Tags and Labels](#tags-and-labels)
   * [Arrays](#arrays)
+  * [Nullability of Properties](#nullability-of-properties)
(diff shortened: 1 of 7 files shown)


JesusMcCloud:
@n0900 care to test this with our stack?

sandwwraith:
Given that you said this problem can be solved with `CborElement` API, maybe it is better to avoid additional annotation?

JesusMcCloud:
Fine for me. I should be able o get back to structured CBOR by January and then this becomes obsolete

JesusMcCloud:
properly integrated into #3036

PR #3128, opened 2025-12-17
IR inliner: Enable intra-module mode for kotlinx.serialization

ddolovov:
^KT-83203


sandwwraith (changes requested):
Please avoid setting `freeCompilerArgs` in the individual files. We have `buildSrc/CompilerOptions.kt` for that.

Issue #3108, opened 2025-11-10
Polymorphic Deserialization Fails with Subclass using `JsonTransformingSerializer`


NoMathExpectation:
**Describe the bug**
In polymorphic deserialization, when the target subclass is using `JsonTransformingSerializer`, the deserialization will fail. It seems that it forgets it is deserializing an polymorphic object.
```log
kotlinx.serialization.json.internal.JsonDecodingException: Encountered an unknown key 'type' at element: $
Use 'ignoreUnknownKeys = true' in 'Json {}' builder or '@JsonIgnoreUnknownKeys' annotation to ignore unknown keys.
JSON input: {"type":"Sub"}
	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
	at kotlinx.serialization.json.internal.JsonTreeDecoder.endStructure(TreeJsonDecoder.kt:292)
	at test.DecodeTest2$Sub$$serializer.deserialize(DecodeTest2.kt:15)
	at test.DecodeTest2$Sub$$serializer.deserialize(DecodeTest2.kt:15)
	at kotlinx.serialization.json.internal.AbstractJsonTreeDecoder.decodeSerializableValue(TreeJsonDecoder.kt:351)
	at kotlinx.serialization.json.internal.TreeJsonDecoderKt.readJson(TreeJsonDecoder.kt:25)
	at kotlinx.serialization.json.Json.decodeFromJsonElement(Json.kt:170)
	at kotlinx.serialization.json.JsonTransformingSerializer.deserialize(JsonTransformingSerializer.kt:78)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:89)
	at kotlinx.serialization.json.Json.decodeFromString(Json.kt:149)
	at test.DecodeTest2.test(DecodeTest2.kt:33)
```

**To Reproduce**
```kt
class DecodeTest2 {
    @Serializable
    sealed class Base

    @OptIn(ExperimentalSerializationApi::class)
    @Serializable(Sub.Serializer::class)
    @KeepGeneratedSerializer
    @SerialName("Sub")
    class Sub : Base() {
        object Serializer : JsonTransformingSerializer<Sub>(generatedSerializer())
    }

    @Test
    fun test() {
        val base: Base = Sub()
        val jsonText = Json.encodeToString(base)
        println(jsonText)
[...]

PR #3139, opened 2026-01-14
Fix Polymorphic Deserialization Error with Subclass using JsonTransformingSerializer 

oungsi2000:
Fixes https://github.com/Kotlin/kotlinx.serialization/issues/3108

I have identified an issue where JsonTransformingSerializer fails 
during decoding when using ```JsonTreeDecoder```

After debugging, I found the cause of the issue in the endStructure method of ```JsonTreeDecoder```:

```kotlin
// JsonTreeDecoder.endStructure
for (key in value.keys) {
    if (key !in names && key != polymorphicDiscriminator) {
        throw JsonDecodingException(
            //...
        )
    }
}
```

The issue is that when a JsonTreeDecoder instance is created, 
the polymorphicDiscriminator is not explicitly set (remaining null). 
As a result, the "type" key—which holds the polymorphic information—is treated as a regular unknown key, triggering a JsonDecodingException.

I have resolved this bug by explicitly setting the discriminator in the readJson method.

However, even though all unit tests and the build have passed (including 2 new test cases), 
I am not entirely sure if this is the intended way to fix it.

Thanks


oungsi2000:
@sandwwraith Hello! I'm a first-time contributor. Could you please review this PR? Thanks

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonTransformingSerializerTest.kt
```diff
@@ -95,12 +95,19 @@ class JsonTransformingSerializerTest : JsonTestBase() {
     @Test
     fun testDocumentationSample() = parametrizedTest { streaming ->
         val correctExample = DocExample("str1")
-        assertEquals(correctExample, json.decodeFromString(DocExample.serializer(), """{"data":["str1"]}""", streaming))
-        assertEquals(correctExample, json.decodeFromString(DocExample.serializer(), """{"data":"str1"}""", streaming))
+        assertEquals(
```
sandwwraith: Please do not add irrelevant changes to the PR, especially if they're about formatting.

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -18,7 +18,7 @@ import kotlin.jvm.*
 @JsonFriendModuleApi
 public fun <T> readJson(json: Json, element: JsonElement, deserializer: DeserializationStrategy<T>): T {
     val input = when (element) {
-        is JsonObject -> JsonTreeDecoder(json, element)
+        is JsonObject -> JsonTreeDecoder(json, element, deserializer.descriptor.classDiscriminator(json))
```
sandwwraith: Sadly, this change is too broad to fix the problem. By using `classDiscriminator(json)`, you're adding a discriminator to _every_ JsonTreeDecoder, even one that was not originally polymorphic. Consider this class:
```kotlin
@Serializable(SubExample2.PolymorphicSerializer::class)
@KeepGeneratedSerializer
@SerialName("Sub")
data class SubExample2(
    val data: String
) {
    object PolymorphicSerializer : JsonTransformingSerializer<SubExample2>(generatedSerializer())
}
```
It is almost identical to yours, but it _does not have a superclass and therefore is not polymorphic_. Therefore, we expect this test to fail:

```kotlin
@Test
fun testNonPolymorphicExampleShouldNotBeParsed() {
    val input = """{"type":"Sub","data":"str1"}"""
    val j = Json { ignoreUnknownKeys = false }
    println(j.decodeFromString<SubExample2>(input)) // should throw "Encountered an unknown key 'type'"
}
```

However, with your change, this test passes.

sandwwraith (changes requested):
Thanks for your PR! As I explained in the comment, the change is too broad. The core reason for the current bug lies in the `JsonTransformingSerializer.deserialize`: by re-creating decoder in ` input.json.decodeFromJsonElement` we forgo the already existing state of the `JsonDecoder` (which contains the class discriminator we want to know). Perhaps it can be rewritten in a way that it retrieves the discriminator itself, conceptually:

```kotlin
final override fun deserialize(decoder: Decoder): T {
    val input = decoder.asJsonDecoder()
    val element = input.decodeJsonElement()
    val discriminator = when(input) {
        is StreamingJsonDecoder -> input.discriminatorHolder?.discriminatorToSkip
        is AbstractJsonTreeDecoder -> input.polymorphicDiscriminator
        else -> null
    }
    return readJson(input.json, transformDeserialize(element), tSerializer, discriminator)
//        return input.json.decodeFromJsonElement(tSerializer, transformDeserialize(element))
}
```

Note that this is a rough draft, and I won't accept it in this form: there is still a need to account for other potential subclasses, and likely a nice API should be added to `JsonDecoder` to "continue" decoding elements with the current decoder instead of creating a new one.

oungsi2000:
@sandwwraith 
Hi!
Thanks for the feedback. Before I proceed with the next commit, 
I’d like to clarify a few points and share a more structured proposal for the "nice API" you mentioned.

Regarding the JsonTreeDecoder constructor, 
I’m wondering if it’s okay to keep the form inside of ```readJson()``` I used in this PR,
(I will change discriminator passed down from StreamingJsonDecoder)
Given the many extension functions that create decoders,
I wanted to double-check if this approach aligns with the library's design.

Regarding the "Nice API" you mentioned, I’ve identified three main implementations of JsonDecoder: StreamingJsonDecoder, AbstractJsonTreeDecoder, and DynamicInput.

Since only StreamingJsonDecoder, AbstractJsonTreeDecoder need to hold discriminator state, 
I propose introducing an internal interface, PolymorphicJsonDecoder, to unify them. This allows us to retrieve the discriminator without explicit type checking in the serializer. Here is a rough draft:

```kotlin
internal interface PolymorphicJsonDecoder : JsonDecoder {
    val discriminator: String?
}

internal open class StreamingJsonDecoder(...) : PolymorphicJsonDecoder, ... {
    override val discriminator get() = discriminatorHolder?.discriminatorToSkip
}

private sealed class AbstractJsonTreeDecoder(...) : PolymorphicJsonDecoder, ... {
    override val discriminator get() = polymorphicDiscriminator
}
````
With this, the deserialize method can be simplified as follows:

```kotlin
final override fun deserialize(decoder: Decoder): T {
    //use if decoder should never be a DynamicInput.
    //otherwise use: val discriminator = (input as? PolymorphicJsonDecoder)?.discriminator
    val input = decoder.asPolymorphicJsonDecoder()
    val element = input.decodeJsonElement()
    return readJson(input.json, transformDeserialize(element), tSerializer, input.discriminator)
}
```
Does this abstraction cover the "other potential subclasses" you had in mind, or would you prefer a different approach?
Thanks

sandwwraith:
The idea with `internal interface PolymorphicJsonDecoder` is good. I also had in mind 
```kotlin
public fun <T> decodeFromJsonElementAs(element: JsonElement, serializer: DeserializationStrategy<T>): T {
        return readJson(json, element, serializer, this)
    }
```
we can add to `JsonDecoder`. But given it will be used only in JsonTransformingSerializer, I now think it is not needed as a public API, so you don't have to add it. You can also move the extraction of the discriminator to the `readJson` function (note it still should work with Dynamic decoders):

```kotlin
public fun <T> readJson(json: Json, element: JsonElement, deserializer: DeserializationStrategy<T>, previousDecoder: JsonDecoder?): T {
    val discriminator = (previousDecoder as? PolymorphicJsonDecoder)?.discriminator
    ... 
```

oungsi2000:
@sandwwraith
I've implemented the ```PolymorphicJsonDecoder``` interface and updated the logic to use ```readJson()``` instead of ```input.json.decodeFromJsonElement()```, following your suggestion to avoid exposing a new public API.

However, I noticed that ```./gradlew knit```(and the build) reports a success even when some tests fail. Is this intended behavior for this repository? It seems a bit risky as it might allow regressions to slip through.

oungsi2000:
@sandwwraith
I've implemented the PolymorphicJsonDecoder interface and updated the logic to use readJson() instead of input.json.decodeFromJsonElement(), following your suggestion to avoid exposing a new public API.

oungsi2000:
@sandwwraith
Hi there, I'm just checking in on this PR to see if there's anything else I should update or if you've had a chance to take a look. No rush, just wanted to keep it on the radar. Thanks!

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonTransformingSerializerTest.kt
```diff
@@ -123,4 +124,77 @@ class JsonTransformingSerializerTest : JsonTestBase() {
...
+        val data: String
+    ) : BaseExample() {
+        object PolymorphicSerializer : JsonTransformingSerializer<SubExample>(generatedSerializer())
+    }
+
+    @Test
+    fun testPolymorphicExampleCanBeParsed() {
+        val baseExample: BaseExample = SubExample("str1")
+        val polymorphicInput = Json.encodeToString(baseExample)
+        assertEquals(baseExample, Json.decodeFromString(polymorphicInput))
```
sandwwraith: Can you please use `assertJsonFormAndRestored` as the assertion? It tests all code paths, including streaming and tree decoders. Or you can use `parameterizedTest` lambda with `mode` parameter. You can look at the other tests in this file as an example.

formats/json/commonMain/src/kotlinx/serialization/json/internal/PolymorphicJsonDecoder.kt
```diff
@@ -0,0 +1,20 @@
...
+ * An internal extension of [JsonDecoder] that exposes the polymorphic discriminator.
+ * This interface is used during the transformation process to ensure that the
+ * class discriminator (e.g., the "type" field) is correctly propagated between
+ * decoders, especially when using [JsonTransformingSerializer].
+ */
+internal interface PolymorphicJsonDecoder : JsonDecoder {
+    /**
+     * A discriminator of the current [JsonDecoder].
+     */
+    val discriminator:String?
```
sandwwraith: formatting: space

sandwwraith (changes requested):
I'm terribly sorry this slipped away from my attention; I have only minor comments and no general objections. Can you rebase the PR on the newest `dev` and address the comments so I can merge this?

oungsi2000:
@sandwwraith 
I addressed all! please check

sandwwraith (approved):

sandwwraith:
Thanks a lot!

PR #3145, opened 2026-02-04
Expose Json exceptions structure

### LINKED ISSUE #1877: Add a less verbose error message to SerializationException
**Use case:** In my current use case, I am passing user-provided JSON input to the `Json.decodeFromString` function. In general, I expect this input to be erroneous. It could, for instance, contain a key that is not part of the actual data model. I want to make the user aware of such issues by printing a suitable error message.

**Limitation:** If I am not mistaken, the only possibility for me to detect such error conditions is by catching `SerializationException`. Unfortunately, the error message encoded into this Exception seems to be directed towards the developer and is not just objectively describing the issue. It looks as follows, for instance:
```
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 4: Encountered
an unknown key 'name'. Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: {
  "name": "xyz"
}
```

Obviously, mentioning `ignoreUnknownKeys` to the user would be highly confusing.

**Proposal:** This specific problem could be solved by adding an optional `SerializationException::shortMesssage` property and populating it with error messages that focus solely on the client-controlled portions of the serialization/deserialization process (whenever this is applicable). Do you think this would be feasible? Or am I missing something here?

Alternatively, it might make sense to create public `SerializationException` subclasses that notify the caller about one specific error condition (such as the presence of an unknown field). This would probably be the cleanest and most flexible solution, but it would also require quite a bit of additional effort.

### LINKED ISSUE #1930: Expose exception details so I can handle specific errors differently, and customise the error message

### LINKED ISSUE #2590: Sensitive Data in Exception Messages (Anonymization, GDPR, PII, …)
There are multiple places within the library where the data is included in exception messages. Not only might this data dangerous in certain situations, it might also contain sensitive information that should never land in logging systems. Many companies need to invest a lot of time into cleaning these things, but it's obviously always a lossy process.

It would be great to establish a policy in which the actual data isn't being included anymore in exception messages.

I do understand that this limits the information available for debugging, hence, an option could be offered to enable this functionality again for those users who need it. But, the default should be to assume that the data is sensitive and unsafe by default.

sandwwraith:
This commit reworks JSON exceptions in kotlinx.serialization:

- Classes JsonException, JsonDecodingException, and JsonEncodingException are now public. They have relevant public properties, such as `shortMessage` and `path`. 
Fixes #1930
Fixes #1877

- A special property `addInputsToExceptionMessages` is added to JsonConfiguration to hide user input from exception messages for security/privacy reasons.

Fixes #2590

Additionally, some exception messages were made clearer, and minor bugs (such as incorrect offsets) were fixed. This is reflected in the added tests.

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
@@ -680,17 +680,17 @@ public class JsonBuilder internal constructor(json: Json) {
      * @Serializable
      * data class User(val name: String, val age: Int)
      * 
-     * val json = Json { addInputsToExceptionMessages = false }
+     * val json = Json { exceptionsWithDebugInfo = false }
      * // Exception message will not contain the invalid input string
      * json.decodeFromString<User>("""{"name":"John","age":"invalid"}""")
      * 
-     * val debugJson = Json { addInputsToExceptionMessages = true }
+     * val debugJson = Json { exceptionsWithDebugInfo = true }
      * // Exception message will include `JSON Input: {"name":"John","age":"invalid"}` line
      * debugJson.decodeFromString<User>("""{"name":"John","age":"invalid"}""")
      * ```
      */
     @ExperimentalSerializationApi
-    public var addInputsToExceptionMessages: Boolean = json.configuration.addInputsToExceptionMessages
+    public var exceptionsWithDebugInfo: Boolean = json.configuration.exceptionsWithDebugInfo
 
     @OptIn(ExperimentalSerializationApi::class)
     internal fun build(): JsonConfiguration {
@@ -721,7 +721,7 @@ public class JsonBuilder internal constructor(json: Json) {
             coerceInputValues, useArrayPolymorphism,
             classDiscriminator, allowSpecialFloatingPointValues, useAlternativeNames,
             namingStrategy, decodeEnumsCaseInsensitive, allowTrailingComma, allowComments, classDiscriminatorMode,
-            addInputsToExceptionMessages
+            exceptionsWithDebugInfo
         )
     }
 }
--- a/formats/json/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
@@ -45,7 +45,7 @@ public sealed class JsonException(override val message: String) : SerializationE
  *   when available.
  * - [input]: the original JSON input (or its minified excerpt in the message). Large inputs
  *   are shortened with context around [offset]. Input is provided on a best-effort basis,
- *   so it may be incomplete. Input is only included when [JsonConfiguration.addInputsToExceptionMessages] is enabled.
+ *   so it may be incomplete. Input is only included when [JsonConfiguration.exceptionsWithDebugInfo] is enabled.
  * - [hint]: optional suggestions for the developer, e.g., enabling certain [Json] configuration options.
  *
  * @property shortMessage short, human-readable description of the decoding error.
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingRedactedErrorMessagesTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingRedactedErrorMessagesTest.kt
@@ -11,7 +11,7 @@ import kotlinx.serialization.test.checkDecodingException
 import kotlin.test.Test
 
 class JsonDecodingRedactedErrorMessagesTest : JsonTestBase() {
-    val redacted = Json(default) { addInputsToExceptionMessages = false }
+    val redacted = Json(default) { exceptionsWithDebugInfo = false }
 
     @Test
     fun testPrimitiveInsteadOfObjectOrList() = parametrizedTest { mode ->
--- a/formats/json/api/kotlinx-serialization-json.api
+++ b/formats/json/api/kotlinx-serialization-json.api
@@ -94,7 +94,6 @@ public final class kotlinx/serialization/json/JsonArraySerializer : kotlinx/seri
 }
 
 public final class kotlinx/serialization/json/JsonBuilder {
-	public final fun getAddInputsToExceptionMessages ()Z
 	public final fun getAllowComments ()Z
 	public final fun getAllowSpecialFloatingPointValues ()Z
 	public final fun getAllowStructuredMapKeys ()Z
@@ -104,6 +103,7 @@ public final class kotlinx/serialization/json/JsonBuilder {
 	public final fun getCoerceInputValues ()Z
 	public final fun getDecodeEnumsCaseInsensitive ()Z
 	public final fun getEncodeDefaults ()Z
+	public final fun getExceptionsWithDebugInfo ()Z
 	public final fun getExplicitNulls ()Z
 	public final fun getIgnoreUnknownKeys ()Z
 	public final fun getNamingStrategy ()Lkotlinx/serialization/json/JsonNamingStrategy;
@@ -113,7 +113,6 @@ public final class kotlinx/serialization/json/JsonBuilder {
 	public final fun getUseAlternativeNames ()Z
 	public final fun getUseArrayPolymorphism ()Z
 	public final fun isLenient ()Z
-	public final fun setAddInputsToExceptionMessages (Z)V
 	public final fun setAllowComments (Z)V
 	public final fun setAllowSpecialFloatingPointValues (Z)V
 	public final fun setAllowStructuredMapKeys (Z)V
@@ -123,6 +122,7 @@ public final class kotlinx/serialization/json/JsonBuilder {
 	public final fun setCoerceInputValues (Z)V
 	public final fun setDecodeEnumsCaseInsensitive (Z)V
 	public final fun setEncodeDefaults (Z)V
+	public final fun setExceptionsWithDebugInfo (Z)V
 	public final fun setExplicitNulls (Z)V
 	public final fun setIgnoreUnknownKeys (Z)V
 	public final fun setLenient (Z)V
@@ -145,7 +145,6 @@ public final synthetic class kotlinx/serialization/json/JsonClassDiscriminator$I
 
 public final class kotlinx/serialization/json/JsonConfiguration {
 	public fun <init> ()V
-	public final fun getAddInputsToExceptionMessages ()Z
 	public final fun getAllowComments ()Z
 	public final fun getAllowSpecialFloatingPointValues ()Z
 	public final fun getAllowStructuredMapKeys ()Z
@@ -155,6 +154,7 @@ public final class kotlinx/serialization/json/JsonConfiguration {
 	public final fun getCoerceInputValues ()Z
 	public final fun getDecodeEnumsCaseInsensitive ()Z
 	public final fun getEncodeDefaults ()Z
+	public final fun getExceptionsWithDebugInfo ()Z
 	public final fun getExplicitNulls ()Z
 	public final fun getIgnoreUnknownKeys ()Z
 	public final fun getNamingStrategy ()Lkotlinx/serialization/json/JsonNamingStrategy;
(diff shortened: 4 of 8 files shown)


formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -660,6 +660,36 @@ public class JsonBuilder internal constructor(json: Json) {
...
+     * // Exception message will not contain the invalid input string
+     * json.decodeFromString<User>("""{"name":"John","age":"invalid"}""")
+     * 
+     * val debugJson = Json { addInputsToExceptionMessages = true }
+     * // Exception message will include `JSON Input: {"name":"John","age":"invalid"}` line
+     * debugJson.decodeFromString<User>("""{"name":"John","age":"invalid"}""")
+     * ```
+     */
+    @ExperimentalSerializationApi
+    public var addInputsToExceptionMessages: Boolean = json.configuration.addInputsToExceptionMessages
```
qwwdfsad: Maybe `verboseExceptions` or `contextualExceptions` would be less directive and more self-explanatory

fzhinkin: I would expect `verboseExceptions = false` to remove hints from exceptions. And the word `contextual` has to many meaning attached to it.

I don't have a better alternative, though :/

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
```diff
@@ -660,6 +660,36 @@ public class JsonBuilder internal constructor(json: Json) {
      */
     public var serializersModule: SerializersModule = json.serializersModule
 
+    /**
+     * Specifies whether actual input data should be included in exception messages.
+     * 
+     * When `false`, exception messages will not contain sensitive input data that could be logged
```
fzhinkin, line 666: They still can contain sensitive data, if it was encoded as a map key and an exception message includes it as a path.

We definitely should elaborate why this option is needed and when to enable/disable it, and inputs leakage into logs is a nice argument in that context. But claiming that all sensitive data will be removed from an exception message might be too strict.

formats/json/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @property shortMessage short, human-readable description of the decoding error.
+ * @property offset zero-based index of the error position in the input, or `-1` if unknown.
+ * @property path JSON path to the failing element when available, or `null`.
+ * @property input original input or its excerpt.
+ * @property hint optional suggestions for the developer that can help fix or diagnose the problem.
+ */
+@ExperimentalSerializationApi
+public class JsonDecodingException @Deprecated(
+    "Use decodingExceptionOf() factory methods",
```
fzhinkin, line 59: Nobody will need to construct these exceptions outside of the `kotlinx-serialization-json`, because even if there's a custom serializer that works solely with JSON documents, it will delegate to JsonEncoder/Decoder and the actual decoding exception will come from the decoder, right?

At the same time, `to deserialize it into a target type` imply that I might want to throw it myself if the `JsonElement` could not be converted into a target type.

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
fzhinkin: If we'll take a `verbosity` route, then we can think of it not as a binary property (brief/verbose), but as different verbosity levels:
- verbose/trace - include message, path, hint and input
- debug - include message, path, hint
- info - include message only

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
```diff
@@ -0,0 +1,278 @@
...
+            default.decodeFromString(serString, input, mode)
+        }, {
+            if (mode in setOf(JsonTestingMode.STREAMING, JsonTestingMode.TREE)) {
+                message("Expected quotation mark '\"', but had '}' instead")
+                offset(13)
+                if (mode == JsonTestingMode.STREAMING) {
+                    path("$.boxed")
+                } else {
+                    // FIXME? When we are reading Json in 'raw tree' (parseToJsonElement) mode, lexer does not provide path information on syntax errors because there are no descriptors.
+                    // It's probably better to cut out path completely or try to build it without descriptors
```
fzhinkin, line 173: How difficult the path reconstruction would be?

If it's hard, I think it's better to cut out the path completely and add a hint about its unavailability. No path is better than incorrect one, if you're trying to process it somehow.

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
```diff
@@ -0,0 +1,278 @@
...
+
+        checkDecodingException(mode, {
+            default.decodeFromString(serInt, input, mode)
+        }, {
+            // For discussion:
+            // Technically, both of these messages are correct despite them being completely different.
+            // A `:` instead of `"` is a good guess, but `:`/`}` is a perfectly valid token inside Json string — for example,
+            // it can be some kind of path `{"foo:bar:baz":"my:resource:locator:{123}"}` or even URI used as a string key/value.
+            // So if the closing quote is missing, there's really no way to correctly tell where the key or value is supposed to end.
+            // Although we may try to unify these messages for consistency.
```
fzhinkin, line 146: Well, if the key contains an escaped `"`, we'll throw `Unexpected EOF` too :)

Perhaps, we should complain about "Unterminated string started at" instead?

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: I chose `inputs` because it's straightforward and highlights that we hide user's info, but not some program/source info (such as variable names).
I agree, Verbose/non-verbose may imply that we'll hide hints and paths (?) as well.

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: Or we can go and remove map keys right away. We just need to decide on replacement for them

formats/json/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
sandwwraith: > Nobody will need to construct these exceptions outside of the kotlinx-serialization-json

Yes. Deprecation is mostly a reminder to future me to use factory methods

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
sandwwraith: Shouldn't be that hard, but the path will contain input keys, not our property names. Or we can remove it here completely, too.

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
sandwwraith: Not sure if we can correctly detect that in all of the cases

sandwwraith: While we can guess offset when we have all string available in memory (https://github.com/Kotlin/kotlinx.serialization/blob/70a5f8a66762c3c5b7f0d0f36de085c53d1a3a7c/formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/StringJsonLexer.kt#L95), when we work with streams and buffers, offset can be 'lost' or modifier if buffer boundaries are shifted

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
fzhinkin: Either of `to cut out path completely or try to build it without descriptors` sounds good then.

But if ` the path will contain input keys` we might need to cut it depending on the semantics of the newly added configuration option. :)

formats/json-tests/commonTest/src/kotlinx/serialization/json/JsonDecodingErrorMessagesTest.kt
fzhinkin: But which way you're suggesting to unify message then?

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
fzhinkin: `<the path was removed, use _configuration_option_name_ to bring it back>`?

formats/json/commonMain/src/kotlinx/serialization/json/JsonExceptions.kt
fzhinkin: Ok then. I just wanted to double check that there are no scenarios where users would need to instantiate the exception.

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
fzhinkin: Maybe something like `verboseExceptionMessages: Boolean` or `exceptionsWithDebugInfo: Boolean`?

sandwwraith: I think `exceptionsWithDebugInput` or `exceptionsWithDebugInfo` is a way to go

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
sandwwraith: That would make deeply-nested map paths awfully long

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
fzhinkin: Let's use the `exceptionsWithDebugInfo` then?

formats/json/commonMain/src/kotlinx/serialization/json/Json.kt
fzhinkin: No replacement then?

sandwwraith: `<debug info disabled>` should point in the right direction to search for configuration key

sandwwraith:
@fzhinkin PTAL

fzhinkin (approved):

Issue #3144, opened 2026-01-31
Protobuf: OneOf case serialization omitted


KirillObukhov:
Serializer skips writing of selected oneof case if it contains default values (or empty).

```
@Serializable
data class Message(
    @ProtoOneOf val ofValue: OfValue = OfValue.StringValue(),
) {
    @Serializable
    sealed interface OfValue {
        @Serializable
        data class StringValue(
            @ProtoNumber(1) val value: String = "",
        ) : OfValue

        @Serializable
        data class IntValue(
            @ProtoNumber(2) val value: Int = 0,
        ) : OfValue

        @Serializable
        data class EmptyValue(
            @ProtoNumber(3) val value: EmptyBox = EmptyBox(),
        ) : OfValue
    }
}

@Serializable
class EmptyBox {
    override fun equals(other: Any?): Boolean =
        this === other || javaClass == other?.javaClass

    override fun hashCode(): Int =
        javaClass.hashCode()
}

fun main() {
    check(Message.OfValue.IntValue(42))
    check(Message.OfValue.IntValue())
    check(Message.OfValue.EmptyValue())
}

fun check(ofValue: Message.OfValue) {
    val original = Message(ofValue)
    val encoded: ByteArray =
        ProtoBuf.encodeToByteArray(original)
    val decoded: Message =
        ProtoBuf.decodeFromByteArray(encoded)
    println("= 0x${encoded.toHexString()} =")
    println(original.ofValue::class.qualifiedName)
    println(decoded.ofValue::class.qualifiedName)
}
```

Output:
```
= 0x102a =
Message.OfValue.IntValue
Message.OfValue.IntValue
= 0x =
Message.OfValue.IntValue
Message.OfValue.StringValue
= 0x =
Message.OfValue.EmptyValue
Message.OfValue.StringValue
```
so original type lost and replaced with default one.

**Environment**
 - Kotlin version: 2.3.0
 - Library version: 1.10.0
 - Kotlin platforms: JVM
 - Gradle version: 9.0

sandwwraith:
As far as I remember, @xiaozhikang0916 contributed this feature. Maybe you will be able to take a look?

xiaozhikang0916:
> As far as I remember, [@xiaozhikang0916](https://github.com/xiaozhikang0916) contributed this feature. Maybe you will be able to take a look?

Working on a fix, may open a pr in a few days.

xiaozhikang0916:
Fixes in #3146.

[...]

PR #3146, opened 2026-02-06
Fix protobuf oneof default values

xiaozhikang0916:
Override `shouldEncodeElementDefault` in `OneOfElementEncoder` to always return true. This ensures oneof fields are encoded even when they contain default values.

According to protobuf oneof semantics, the selected case must always be identifiable. So a tag with field number is always needed for one of field.

Fixes https://github.com/Kotlin/kotlinx.serialization/issues/3144

And I found that code sample in the issue shows some misusing of default values in proto messages. So I add an extra section in docs for it.


xiaozhikang0916:
Sorry, wrong target branch. I will create a new one.

PR #3149, opened 2026-02-08
Add CBOR kotlinx-io support

cedrickcooke:
Looking to follow in the footsteps of #2707 with CBOR support for kotlinx-io.

(Sorry for the noise re #3148, accidentally pointed it at the wrong branch and couldn't find the UI to change it.)

### DIFF
--- a/formats/cbor-io/commonMain/src/kotlinx/serialization/cbor/io/internal/IoCborStreams.kt
+++ b/formats/cbor-io/commonMain/src/kotlinx/serialization/cbor/io/internal/IoCborStreams.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.cbor.io.internal
+
+import kotlinx.io.*
+import kotlinx.serialization.cbor.internal.*
+
+internal class IoStreamOutput(private val sink: Sink) : Output {
+
+    override fun write(buffer: ByteArray, offset: Int, count: Int) {
+        sink.write(buffer, startIndex = offset, endIndex = offset + count)
+    }
+
+    override fun write(byteValue: Byte) {
+        sink.writeByte(byteValue)
+    }
+}
+
+internal class IoStreamInput(private val source: Source): Input {
+    override val availableBytes: Int
+        get() = source.peek().readByteArray().size
+
+    override fun read(): Int =
+        try {
+            source.readByte().toInt()
+        } catch (_: EOFException) {
+            return -1
+        }
+
+    override fun read(b: ByteArray, offset: Int, length: Int): Int =
+        source.readAtMostTo(b, startIndex = offset, endIndex = offset + length)
+
+    override fun skip(length: Int) {
+        source.skip(length.toLong())
+    }
+}
--- a/formats/cbor-io/src/commonTest/kotlin/kotlinx/serialization/cbor/io/IoTests.kt
+++ b/formats/cbor-io/src/commonTest/kotlin/kotlinx/serialization/cbor/io/IoTests.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization.cbor.io
+
+import kotlinx.io.*
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+import kotlin.test.*
+
+class IoTests {
+
+    @Serializable
+    data class Simple(val i: Int)
+
+    @Test
+    fun testEncodingDefinite() {
+        val value = Simple(42)
+        val buffer = Buffer()
+        Cbor { useDefiniteLengthEncoding = true }.encodeToSink(value, buffer)
+        assertEquals(expected = "a16169182a", actual = buffer.readByteArray().toHexString())
+    }
+
+    @Test
+    fun testEncodingIndefinite() {
+        val buffer = Buffer()
+        Cbor { useDefiniteLengthEncoding = false }.encodeToSink(Simple(42), buffer)
+        assertEquals(expected = "bf6169182aff", actual = buffer.readByteArray().toHexString())
+    }
+
+    @Test
+    fun testDecoding() {
+        val buffer = Buffer()
+        buffer.write("a16169182a".hexToByteArray())
+        val decoded = Cbor.decodeFromSource<Simple>(buffer)
+        assertEquals(expected = Simple(42), actual = decoded)
+
+        assertTrue(buffer.exhausted())
+    }
+
+    @Test
+    fun testDecodingFailsWithUnprocessedBytes() {
+        val buffer = Buffer()
+        buffer.write("bf6169182aff00".hexToByteArray())
+        assertFailsWith<SerializationException> { Cbor.decodeFromSource<Simple>(buffer) }
+    }
+}
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoder.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoder.kt
@@ -13,28 +13,18 @@ import kotlinx.serialization.encoding.*
 import kotlinx.serialization.modules.*
 import kotlin.experimental.*
 
-
-//value classes are only inlined on the JVM, so we use a typealias and extensions instead
-private typealias Stack = MutableList<CborWriter.Data>
-
-private fun Stack(initial: CborWriter.Data): Stack = mutableListOf(initial)
-private fun Stack.push(value: CborWriter.Data) = add(value)
-private fun Stack.pop() = removeLast()
-private fun Stack.peek() = last()
-
 // Writes class as map [fieldName, fieldValue]
 // Split implementation to optimize base case
-internal sealed class CborWriter(
+@CborFriendModuleApi
+public sealed class CborWriter(
     override val cbor: Cbor,
-    protected val output: ByteArrayOutput,
+    protected val output: Output,
 ) : AbstractEncoder(), CborEncoder {
-    protected var isClass = false
+    protected var isClass: Boolean = false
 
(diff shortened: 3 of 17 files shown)


formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
```diff
@@ -4,15 +4,30 @@
...
+    public fun read(): Int
+    public fun read(b: ByteArray, offset: Int, length: Int): Int
+    public fun skip(length: Int)
+}
+
+@CborFriendModuleApi
+public interface Output {
+    public fun write(buffer: ByteArray, offset: Int = 0, count: Int = buffer.size)
+    public fun write(byteValue: Byte)
+}
```
pdvrieze, line 20: These types appear to be format independent and relevant to all binary formats. It would be worthwhile to consider whether these interfaces should be put in a common context.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
```diff
@@ -4,15 +4,30 @@
 
 package kotlinx.serialization.cbor.internal
 
-internal class ByteArrayInput(private var array: ByteArray) {
+@CborFriendModuleApi
+public interface Input {
+    public val availableBytes: Int
+    /** Returns a -1 if no bytes are available. Otherwise returns a value between 0 and 255 (inclusive). */
```
pdvrieze, line 10: It is not clear what this means. What is the meaning of `availableBytes==0` and how do you deal with (pending) network reads where the bytes are not available yet (and stream size is not known).

pdvrieze (commented):
As more formats support kotlinx-io they all need some sort of text or binary input/output abstraction. It makes sense that this is more or less "internal", but such abstractions are not format specific at all. As such it makes more sense to put them into the core module (with a serialization internal api flag).

At some level it might even make sense to have these interfaces as public, but that may be better for a different discussion.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
cedrickcooke: Should I handle that as part of this PR? My original intent with this API was to be as non-invasive as possible. If so, I'm happy to make the change but I'm not sure where it should live. `formats/io`?

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
cedrickcooke: Sorry, I didn't do the best job on the docs for `Input`/`Output`, I'll go back and update them 😅.

This one comment was me attempting to describe the pre-existing behavior of `read` in `ByteArrayInput`, since it took me a minute to figure it out. I had originally thought that this should return a `Byte` (the 0-255 value returned on the happy path), but there's a failure path if there's no more data available where it returns -1 (which forces the `Int` typing).

Re. network streams, I don't think this is even handled by kotlinx-io yet? It seems like the APIs presented are all synchronous (until Kotlin/kotlinx-io#163 has an implementation, at least).

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
xiaozhikang0916: I would suggest leaving these base Input/Output interface in the `core` module, and change the parameter of `BinaryFormat` and `StringFormat` to use them.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
pdvrieze: In most cases (including Java IO streams) returning `0` means that nothing was returned, but another call may return something again (although most API's that support blocking will block instead of returning `0`). A negative (or `-1`) return value tends to mean end of file/stream.

sandwwraith:
RE: Input/Output general interfaces

When kotlinx-io is 1.0, that will no longer be needed. Its `Sink` and `Source` can be directly added to a `BinaryFormat` or `BinaryIoFormat`. Implementing encoding to/from ByteArray should be trivial via sinks, so it would be enough for the implementation to write/read Sink/Source.

formats/cbor-io/commonMain/src/kotlinx/serialization/cbor/io/internal/IoCborStreams.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    override fun write(byteValue: Byte) {
+        sink.writeByte(byteValue)
+    }
+}
+
+internal class IoStreamInput(private val source: Source): Input {
+    override val availableBytes: Int
+        get() = source.peek().readByteArray().size
```
sandwwraith, line 23: Using `peek` and reading _whole_ input to the memory is a huge potential performance problem. `exhausted()` should be used instead. Yes, it returns Boolean, so calls of `availableBytes` should also be adapted.

formats/cbor-io/commonMain/src/kotlinx/serialization/cbor/io/internal/IoCborStreams.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun read(): Int =
+        try {
```
sandwwraith, line 26: Using try-catch on the hot path is also a performance problem. `read()` seems to be called only from methods reading byte/int/long/etc. These methods should be adapted and use corresponding Sink methods instead (readByte()/readInt()/etc). It will also (probably) automatically give us IOException without the need to use `availableBytes` — see above.

formats/cbor-io/commonMain/src/kotlinx/serialization/cbor/io/internal/IoCborStreams.kt
```diff
@@ -0,0 +1,38 @@
...
+            source.readByte().toInt()
+        } catch (_: EOFException) {
+            return -1
+        }
+
+    override fun read(b: ByteArray, offset: Int, length: Int): Int =
+        source.readAtMostTo(b, startIndex = offset, endIndex = offset + length)
```
sandwwraith, line 33: Semantic is different: `readAtMostTo` can read fewer bytes than `length` in order not to block, while `Input.read` expects all the bytes. ~~In this case, a reading loop should be implemented. @fzhinkin  I do not remember why we don't have `readExactTo` in `Source`?~~

We have `Source.readTo(sink: ByteArray, startIndex: Int = 0, endIndex: Int = sink.size)`. Use it instead. Return value of `read(b, offset, len)` is not needed anyway.

formats/cbor-io/src/commonTest/kotlin/kotlinx/serialization/cbor/io/IoTests.kt
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package kotlinx.serialization.cbor.io
+
+import kotlinx.io.*
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+import kotlin.test.*
+
+class IoTests {
```
sandwwraith, line 12: Preferably, the whole test suite should be adapted to work with IO streams; see how it is done in `JsonTestBase` with `JsonTestingMode`.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoder.kt
```diff
@@ -13,28 +13,18 @@ import kotlinx.serialization.encoding.*
 import kotlinx.serialization.modules.*
 import kotlin.experimental.*
 
-
-//value classes are only inlined on the JVM, so we use a typealias and extensions instead
-private typealias Stack = MutableList<CborWriter.Data>
```
sandwwraith, line 18: This change is not related and I prefer to leave the typealias for easier reading.

sandwwraith (changes requested):
In short, the current `Input/ByteArrayInput` API does not fit the task because it was created to mimic the Java stream API, and kotlinx-io has different ideas in mind. Try to form the API based on what the actual `CborDecoder` needs.

Also, it may interfere with other work in CBOR (#3036), so maybe it's worth coordinating work in advance, or waiting for the PR instead. cc @JesusMcCloud @fzhinkin

formats/cbor-io/src/commonTest/kotlin/kotlinx/serialization/cbor/io/IoTests.kt
sandwwraith: Also, a pair of basic benchmarks (see CborBaseline.kt) would be nice.

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Streams.kt
JesusMcCloud: I think this is a separate discussion. We probably we want string and binary outputs in core, so we can then plug in whatever we want. Providing shims for ByteArrayOutput/input and Sink/Source (also for string-based IO) can then also move to core and core-io.

Just my 2 cents

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Encoder.kt
```diff
@@ -180,25 +171,28 @@ internal class IndefiniteLengthCborWriter(cbor: Cbor, output: ByteArrayOutput) :
...
+public class DefiniteLengthCborWriter(cbor: Cbor, output: Output) : CborWriter(cbor, output) {
+
+    private class Data(val bytes: ByteArrayOutput, var elementCount: Int)
 
-    private val structureStack = Stack(Data(output, -1))
-    override fun getDestination(): ByteArrayOutput =
-        structureStack.peek().bytes
+    private val structureStack = mutableListOf<Data>()
+    override fun getDestination(): Output =
+        structureStack.lastOrNull()?.bytes ?: output
```
JesusMcCloud, line 181: This will interfere with #3036, but it's minimal. I'd of course still prefer my humongous PR to be finalized at at some point without adding even more layers to what is already challenging to review ;-)

PR #3150, opened 2026-02-09
Update Dokka to 2.2.0-Beta and fix unresolved links

whyoleg:


formats/json/jsWasmMain/src/kotlinx/serialization/json/internal/createMapForCache.kt
```diff
@@ -4,9 +4,4 @@
 
 package kotlinx.serialization.json.internal
 
-/**
```
sandwwraith, line 7: Since this declaration is internal anyway, I would prefer leaving this comment since it is a hint for developer (future me). Is this a problem for Dokka?

whyoleg: Yeah, even though the declaration is `internal`, Dokka's link resolution is executed earlier in the pipeline, before `private/internal` declarations are filtered out.
And here we have a problem in the `jsWasmMain` source set: it can't resolve `ThreadLocal`. The IDE will also highlight it as `unresolved`:

Image

We could replace it with `kotlin.native.concurrent.ThreadLocal` in `jsWasmMain` KDoc - in this case, the link will be correctly resolved.
Also, in `jvmMain`, `[ThreadLocal]` is resolved to Java's [ThreadLocal](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/ThreadLocal.html), and `kotlin.native.concurrent.ThreadLocal`  is not resolved at all (both by IDE and Dokka) - so it's a bit misleading.


So, yes, I could revert these changes and fix only the link in the `jsWasmMain` source set. **WDYT?**

JIC: even if documentation is present only on `expect` declaration, the IDE will still show its documentation in platform-specific source sets on hover.

sandwwraith: > We could replace it with kotlin.native.concurrent.ThreadLocal

Isn't it enough to `import kotlin.native.concurrent.ThreadLocal`? In any case, let's specify it everywhere.

> even if documentation is present only on expect declaration, the IDE will still show its documentation in platform-specific source sets on hover.

Yes, but sometimes it is easier to navigate to declaration rather than read pop-up :)

whyoleg: > Isn't it enough to import kotlin.native.concurrent.ThreadLocal?

Yeah, that's an alternative. Previously, there were some problems with the import optimizer in the IDE, which would have removed the import if it was used only in KDoc, but it looks like it's working fine right now.

>  In any case, let's specify it everywhere.

What should we do in the `jvmMain` case, where we can't really reference `kotlin.native.concurrent.ThreadLocal` at all? :)

sandwwraith: > we can't really reference kotlin.native.concurrent.ThreadLocal at all? :)

Which is weird, given that it's declared in common source. Looks like a downside of `@OptionalExpectation`. Well, FQ name in docs it is then.

whyoleg:
@sandwwraith, so, should we do `Rebase and merge`?

PR #3167, opened 2026-03-25
CBOR: Relax value range check when decoding numbers

### LINKED ISSUE #3143: CBOR unsigned number encoding is wrong
**Describe the bug**

Direct byte encoding/decoding of Kotlin unsigned primitives is currently two’s-complement via signed primitive, which round-trips within the library, but doesn’t match the CBOR representation. CBOR has native signed and native unsigned Integer types and those should be used

*Context:*  
While working on structured CBOR and pondering extensions/shorthands for numbers other than `Long` I noticed that unsigned primitives must not take the detour through signed types, because CBOR represents signed and unsigned types with a different header/mask. This is perfectly fine for the newly introduced`CborInteger`, because it stores a numbers signum and magnitude separately.

*Problem:*  
When rebasing onto `dev` I noticed range checks got introduced to the CBOR decoder and it clicked:

Kotlin’s built-in serializers for UByte/UShort/UInt/ULong are inline wrappers over the signed primitives (Byte/Short/Int/Long) and literally call encodeByte(value.toByte()), decodeByte().toUByte(), etc. ([core/commonMain/src/kotlinx/serialization/internal/ValueClasses.kt](https://github.com/Kotlin/kotlinx.serialization/blob/605a35faa87534e24482aa00a52a2a71bebd51a3/core/commonMain/src/kotlinx/serialization/internal/ValueClasses.kt)).
* CBOR’s “unsigned integer” is major type 0 (0..2⁶⁴-1). So for example UByte(200) “should” be encoded as CBOR unsigned integer 200.
* With the range checks (pulled from dev) in [CborReader.decodeByte/decodeShort/decodeInt](https://github.com/JesusMcCloud/kotlinx.serialization/blob/82c9e7d759ae8edea1f47444e3c8db2feb2786a3/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt#L152-L157), decoding a CBOR unsigned integer 200 into a Byte [now throws ](https://github.com/JesusMcCloud/kotlinx.serialization/blob/82c9e7d759ae8edea1f47444e3c8db2feb2786a3/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt#L144)(since 200 is not within [-128..127]). That means UByte decoding via the standard serializer also fails for values > 127 even though an unsigned CBOR Int up to 255 would be within range.
* Encoding is also flawed: UByte(200).toByte() is -56, so the default encodeByte path emits a CBOR negative integer, not an unsigned 200.

**To Reproduce**
See [Testcase](https://github.com/JesusMcCloud/kotlinx.serialization/blob/4b88575c959d38fddeb9d8938eb092a79d43356d/formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborUnsignedInteroperabilityTest.kt) that always fails.

**Expected behavior**
Unsigned Kotlin number types should always be encoded as Major Type 0

**Environment**
 - Kotlin version: any
 - Library version: any
 - Kotlin platforms: all
 - Gradle version: any
 - Other relevant context: Was always there and the range checks made it even worse, which helped discover the issue

fzhinkin:
A commit e334d1c introduced a strict value range check when decoding integer values. The main intention was to prevent misinterpreting a parsed CBOR document due to integer overflow or truncation. However, the overflow check is too strict and prevent decoding unsigned Kotlin numbers from some unsigned CBOR integers. For now, we should relax the check to verify only that there will be no truncation.

Closes #3143

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
@@ -584,12 +584,10 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
         return when (majorType) {
             HEADER_BYTE_STRING, HEADER_STRING, HEADER_ARRAY
                 -> readUnsignedIntegerIgnoringMajorType { "${majorType.majorTypeName} length" }
-                .asSizedElementLength(majorType)
-
+                    .asSizedElementLength(majorType)
             HEADER_MAP
                 -> readUnsignedIntegerIgnoringMajorType { "map length" }
-                .asSizedElementLength(majorType, Int.MAX_VALUE / 2) * 2
-
+                    .asSizedElementLength(majorType, Int.MAX_VALUE / 2) * 2
             else -> when (additionalInformation) {
                 24 -> 1
                 25 -> 2
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
@@ -129,17 +129,17 @@ internal open class CborReader(override val cbor: Cbor, protected val parser: Cb
 
     override fun decodeBoolean() = parser.nextBoolean(tags)
 
-    override fun decodeByte() = parser.nextNumberWithinRange(
-        tags, Byte.MIN_VALUE.toLong(), Byte.MAX_VALUE.toLong(), "Byte"
+    override fun decodeByte() = parser.nextNumberWithMaxWidth(
+        tags, Byte.SIZE_BITS, "Byte"
     ).toByte()
-    override fun decodeShort() = parser.nextNumberWithinRange(
-        tags, Short.MIN_VALUE.toLong(), Short.MAX_VALUE.toLong(), "Short"
+    override fun decodeShort() = parser.nextNumberWithMaxWidth(
+        tags, Short.SIZE_BITS, "Short"
     ).toShort()
-    override fun decodeChar() = parser.nextNumberWithinRange(
-        tags, Char.MIN_VALUE.code.toLong(), Char.MAX_VALUE.code.toLong(), "Char"
+    override fun decodeChar() = parser.nextNumberWithMaxWidth(
+        tags, Char.SIZE_BITS, "Char"
     ).toInt().toChar()
-    override fun decodeInt() = parser.nextNumberWithinRange(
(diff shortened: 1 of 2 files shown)


formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
```diff
@@ -353,11 +353,18 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
         }
     }
 
-    internal fun nextNumberWithinRange(tags: ULongArray?, from: Long, to: Long, type: String): Long {
+    internal fun nextNumberWithMaxWidth(tags: ULongArray?, widthBits: Int, type: String): Long {
```
sandwwraith: Won't it be simpler to check if the number is in (Byte.MIN_VALUE..MAX_VALUE) or (0..UByte.MAX_VALUE)?

sandwwraith (approved):

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
fzhinkin: It would be simpler to check, but then either 4 (well, 3) numbers / 2 ranges has to be passed into the function, or corresponding parameters has to be stored elsewhere.

sandwwraith: I don't see problem in passing additional `Uxxx.MAX_VALUE` parameter

sandwwraith (approved):

formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
```diff
@@ -353,10 +358,21 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
...
+        type: String,
+    ): Long {
         val number = nextNumber(tags)
-        if (number !in from..to) {
-            throw CborDecodingException("Decoded number $number is not within the range for type $type ([$from..$to])")
+        if (number !in from..to && number !in 0..unsignedUpperBound) {
+            throw CborDecodingException(buildString {
+                append("Decoded number $number is not within the range for type $type ([$from..$to])")
+                if (unsignedUpperBound >= 0) {
+                    append(", nor it is within the range for U$type ([0..$unsignedUpperBound])")
```
averen, line 373: typo, this should read "nor is it within..."

PR #3175, opened 2026-04-07
Added templates to the issue creation wizard

ivakub:


vsartiushkin (approved):
I have reduced the number of input fields for bug report to make it more uniform with the other kotlinx libraries.

PR #3188, opened 2026-04-29
KTI-2626: Enable cache redirector for Yarn & NPM

fzhinkin:
kotlinx-serialization builds executed as part of the Kotlin aggregate build are failing intermittently due networking errors on the Yarn/Npm side. This PR sets up cache redirector.

Ideally, there would be a way to configure it for all projects maintained by the libraries team (such as coroutines, datetime, benchmarks, atomicfu, collections immutable) in a single place, but it seems to be tricky to implement. The main reason for that is reliance on APIs provided by KGP, those APIs depend on a particular KGP version and our libraries use different versions. Which may require slightly different approaches to configure redirection. So for now, it seems like the easiest fix would be making the proposed change.

^KTI-2626

### DIFF
--- a/buildSrc/src/main/kotlin/cacheRedirector.kt
+++ b/buildSrc/src/main/kotlin/cacheRedirector.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.the
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.targets.js.yarn.YarnLockMismatchReport
+
+const val DEFAULT_YARN_REGISTRY = "https://registry.yarnpkg.com"
+const val NPM_REGISTRY_CACHE = "https://cache-redirector.jetbrains.com/registry.npmjs.org"
+const val NODE_DIST_CACHE = "https://cache-redirector.jetbrains.com/nodejs.org/dist"
+const val YARN_DIST_CACHE = "https://cache-redirector.jetbrains.com/github.com/yarnpkg/yarn/releases/download"
+
+fun Project.configureJsCacheRedirector() {
+    rootProject.plugins.withType(org.jetbrains.kotlin.gradle.targets.js.yarn.YarnPlugin::class.java) {
+        rootProject.the<org.jetbrains.kotlin.gradle.targets.js.yarn.YarnRootExtension>().restoreYarnLockTaskProvider.configure {
+            doLast {
+                // yarn 1.x doesn't and won't support overriding registry used in yarn.lock, so we need to replace it manually
+                // https://github.com/yarnpkg/yarn/issues/6436#issuecomment-426728911
+                val lockFile = outputFile.get()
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/cacheRedirector.kt
```diff
@@ -0,0 +1,39 @@
...
+            doLast {
+                // yarn 1.x doesn't and won't support overriding registry used in yarn.lock, so we need to replace it manually
+                // https://github.com/yarnpkg/yarn/issues/6436#issuecomment-426728911
+                val lockFile = outputFile.get()
+                lockFile.writeText(lockFile.readText().replace(DEFAULT_YARN_REGISTRY, NPM_REGISTRY_CACHE))
+            }
+        }
+    }
+
+    plugins.withType(org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin::class) {
```
sandwwraith: This one doesn't have `rootProject`. Although you only call this method in the root buildscript, it's always called on rootProject.

buildSrc/src/main/kotlin/cacheRedirector.kt
```diff
@@ -0,0 +1,39 @@
...
+        extensions.configure(org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsEnvSpec::class.java) {
+            downloadBaseUrl.set(NODE_DIST_CACHE)
+        }
+    }
+
+    afterEvaluate {
+        rootProject.plugins.withType(org.jetbrains.kotlin.gradle.targets.js.yarn.YarnPlugin::class.java) {
+            rootProject.the<org.jetbrains.kotlin.gradle.targets.js.yarn.YarnRootEnvSpec>().downloadBaseUrl.set(YARN_DIST_CACHE)
+            rootProject.the<org.jetbrains.kotlin.gradle.targets.js.yarn.YarnRootEnvSpec>().yarnLockMismatchReport.set(YarnLockMismatchReport.WARNING)
```
sandwwraith: Are you sure that setting a warning on mismatches is safe? I don't mind running `upgradeYarnLock` once in a while

fzhinkin: Mainly the artifact of copy-pasting the script from the Kotlin repo. You're right, it's better to keep it `FAIL`.

sandwwraith (approved):

sandwwraith:
Looks like we need to re-generate the lock file because hashes are different now (why though?)

fzhinkin:
@sandwwraith, that's because urls are now pointing to cache redirector.

fzhinkin:
Discussed it elsewhere and agreed on setting up the redirector in the main branch, not only for `kotlin-community/dev`.

fzhinkin:
@sandwwraith, before rebasing it all onto `dev` I decided to explore "make js cache redirector switchable":
- turning it off and on no longer require updating the lock file (the part I like)
- lock file patching become even more complicated (the part I don't like)

WDYT? We can still start with unconditional cache redirection and see if there will be any complaints, and only then consider this approach.

sandwwraith:
I don't remember many complaints about cache-redirector in 'big' Kotlin (although it does have a switch). IMO we can start without one

fzhinkin:
@sandwwraith, okay, I dropped the flag switching the redirector and rebased the branch onto `dev`.

sandwwraith (approved):

Issue #3084, opened 2025-09-19
Proposal: Deprecate for removal the ‘External serializers’ feature with the aim of being replaced by the new 'Serializers DSL'


sandwwraith:
## Why?

The external serializers feature is incomplete and has a lot of bugs. The leading causes for this are the fundamental technical limitations of the Kotlin compiler and overall design, such as:

* The compiler does not store default values for properties anywhere in the metadata. This means it is impossible to generate the code similar to `if (prop != propDefaultValue) encoder.encodeElement(prop)` in the external serializer ([https://github.com/Kotlin/kotlinx.serialization/issues/2512](https://github.com/Kotlin/kotlinx.serialization/issues/2512))  
* The compiler does not (always) store the fact that a property has a backing field. Since we aim to serialize properties with backing fields only and treat all others as transient, it can result in inconsistent serialization or descriptor information. ([https://github.com/Kotlin/kotlinx.serialization/issues/2549](https://github.com/Kotlin/kotlinx.serialization/issues/2549))   
* Generated serializers themselves are not customizable at all. You cannot specify SerialName for a property, nor specify a serializer for a specific property (in case a serializer for its type is also external, for example: [https://github.com/Kotlin/kotlinx.serialization/issues/889](https://github.com/Kotlin/kotlinx.serialization/issues/889))   
* Java classes do not have a notion of a primary constructor. As a result, it is impossible to generate a serializer for a Java class at all, which limits the usability of this feature ([https://github.com/Kotlin/kotlinx.serialization/issues/1223](https://github.com/Kotlin/kotlinx.serialization/issues/1223))  
* Private or internal properties cannot be serialized correctly this way.

## Replacement

Replacement solution requirements:

* The user should be able to specify the serial name, serializer, and default value for the property.  
* The user should be able to specify which properties are serializable and which are not, as the compiler cannot deduce it for them.  
* The user should be able to specify which constructor to call, as we cannot deduce it for them.

We can see that the ideal solution here is just to write a custom manual serializer. However, it requires significant expertise in kotlinx.serialization, and serializers themselves are boilerplate-ish, since Encoders API is designed mainly to be called from code generated by the compiler plugin. Perhaps AI can generate something reasonable, but we shouldn’t rely entirely on it. 

A good example of avoiding writing a manual serializer is JsonTransformingSerializer, which is very popular and solves a lot of pain points. Providing a specialized solution for this particular problem also seems viable. Let’s name it Serializers DSL for now. 

[...]

Issue #1978, opened 2022-07-04
Support for additional properties


pschichtel:
It is sometimes the case that I want to parse an object with a few concrete properties and a bunch of additional properties.

OpenAPI 3's [Responses Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#responses-object) is an example for this. It has the fixed `default` default plus additional fields for various http statuses.

so this could look like:

```json
{
  "default": {...},
  "200": {...},
  "400": {...},
  "404": {...}
}
```

And it would be great to be able to parse this into something like:

```kotlin
@Serializable
data class Responses(val default: Response, statuses: Map<String, Response>)
```

Jackson has the `@JsonAnySetter` annotation which can be used to implement this.

OpenAPI/JSON Schema support specifying object schemas that have both `properties` and `additionalProperties`.

Currently I think the only solution would be a hand-written KSerializer for Responses.

pschichtel:
for reference the KSerializer I wrote for my example:

```kotlin
object ResponsesSerializer : KSerializer<Responses> {
    private val mapSerializer = MapSerializer(String.serializer(), Response.serializer())
    private const val defaultFieldName = "default"

    override val descriptor: SerialDescriptor = object : SerialDescriptor by mapSerializer.descriptor {
        override val serialName: String = "Responses"
    }

    override fun serialize(encoder: Encoder, value: Responses) {
        val pairs = listOfNotNull(
            value.default?.let { Pair(defaultFieldName, it) },
        ) + value.statuses.map { (key, value) -> Pair(key.code.toString(), value) }

        mapSerializer.serialize(encoder, pairs.toMap())
    }

    override fun deserialize(decoder: Decoder): Responses {
        val map = mapSerializer.deserialize(decoder).toMutableMap()
        val default = map.remove(defaultFieldName)
        val statuses = map.mapKeys { (key, _) -> HttpStatusCode(key.toUShort()) }
        return Responses(default, statuses)
    }
}
```

sandwwraith:
See https://github.com/Kotlin/kotlinx.serialization/issues/959#issuecomment-683881720

pschichtel:
Interesting alternative to my serializer, but it still requires me to write a custom serializer doesn't it?

sandwwraith:
Yes, it still requires one, but given that `JsonTransformingSerializer` is much easier to implement, I'm not sure that this feature needs special support

pschichtel:
[...]

Issue #1512, opened 2021-05-24
ArrayIndexOutOfBoundsException when deserializing a custom object


edenman:
**Describe the bug**
```
Index 0 out of bounds for length 0
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor.getElementDescriptor(PluginGeneratedSerialDescriptor.kt:135)
	at kotlinx.serialization.descriptors.SerialDescriptorKt$elementDescriptors$1$1.next(SerialDescriptor.kt:284)
	at kotlinx.serialization.descriptors.SerialDescriptorKt$elementDescriptors$1$1.next(SerialDescriptor.kt:279)
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptorKt.hashCodeImpl(PluginGeneratedSerialDescriptor.kt:137)
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor$_hashCode$2.invoke(PluginGeneratedSerialDescriptor.kt:44)
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor$_hashCode$2.invoke(PluginGeneratedSerialDescriptor.kt:44)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor.get_hashCode(PluginGeneratedSerialDescriptor.kt:44)
	at kotlinx.serialization.internal.PluginGeneratedSerialDescriptor.hashCode(PluginGeneratedSerialDescriptor.kt:97)
	at java.base/java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
	at kotlinx.serialization.json.internal.DescriptorSchemaCache.get(SchemaCache.kt:35)
	at kotlinx.serialization.json.internal.DescriptorSchemaCache.getOrPut(SchemaCache.kt:27)
	at kotlinx.serialization.json.internal.JsonTreeDecoder.elementName(TreeJsonDecoder.kt:221)
	at kotlinx.serialization.internal.NamedValueDecoder.getTag(Tagged.kt:330)
	at kotlinx.serialization.json.internal.JsonTreeDecoder.decodeElementIndex(TreeJsonDecoder.kt:204)
	at chat.quill.data.DefaultFoodContainerSerializer.deserialize(APIObjects.kt:40)
[...]

PR #3189, opened 2026-04-30
Add @JsonExtraKeys for capturing unknown JSON keys

inemtsev:
Adds `@JsonExtraKeys`, a property annotation that captures every key of a JSON object that does not match a declared property, and writes those keys back on encoding.

```kotlin
@Serializable
data class Project(
    val name: String,
    @JsonExtraKeys val details: JsonObject = JsonObject(emptyMap())
)

val json = Json { useExtraKeys = true }
val project = json.decodeFromString<Project>("""{"name":"example","maintainer":"Unknown","license":"Apache 2.0"}""")
// project.details == {"maintainer":"Unknown","license":"Apache 2.0"}
// json.encodeToString(project) reproduces the input
```

### Behavior

- Enabled by the new `useExtraKeys` flag on `JsonBuilder`, off by default (see performance below). Without the flag the annotated property behaves as a regular property.
- The bucket type must be `JsonObject` or `Map<String, JsonElement>`. Typed values (`Map<String, V>`) were in the first version of this PR and were removed on review: the bucket holds keys the class does not know about, so it should not presume their shape. Individual entries can be decoded with `Json.decodeFromJsonElement`.
- At most one bucket per class, and `@JsonNames` is rejected on the bucket property. Both rules are validated lazily on first use.
- Capture takes precedence over `ignoreUnknownKeys` and `@JsonIgnoreUnknownKeys`.
- Keys resolved through `@JsonNames`, a `JsonNamingStrategy`, or the class discriminator are not captured.
- On encoding, the entries are written after the regular properties under their own keys; the bucket's property name is not written. If the bucket contains a key that a declared property or the discriminator would be written under, encoding throws `JsonEncodingException` instead of emitting duplicate keys.
- Streaming and tree encoders and decoders are both supported.

### Performance

`TwitterFeedBenchmark` (macro, bucket-less classes), A/B pairs in the same JVM session:

- Flag off vs `dev`: decode 1571 ± 40 vs 1565 ± 43 ops/s, encode 3181 ± 38 vs 3202 ± 65 ops/s. Parity.
- Flag on vs flag off: decode 1440 ± 30 vs 1552 ± 15 ops/s (-7.2%), encode 2987 ± 34 vs 3241 ± 29 ops/s (-7.8%).

The flag-on cost is why the flag defaults to off. `JsonExtraKeysBenchmark` (micro) and flag-on/flag-off `TwitterFeedBenchmark` variants are committed.

### Known limitation

Descriptors produced by `@Serializer(forClass = ...)` throw from `hashCode()` (#1512), so the schema-cache lookup for the bucket index falls back to uncached computation for them. #3215 would have fixed the root cause; it was closed with a pointer to #3084, so the fallback stays.

### Docs

`docs/json.md` gets a "Preserving unknown keys (experimental)" section next to the existing hand-written serializer example for the same problem, with a knit-generated example and test.

Related issue: #1978 (asks for a typed `Map<String, Response>`; see the note on bucket types above).

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -164,3 +164,67 @@ internal inline fun Json.tryCoerceValue(
 
 internal fun SerialDescriptor.ignoreUnknownKeys(json: Json): Boolean =
     json.configuration.ignoreUnknownKeys || annotations.any { it is JsonIgnoreUnknownKeys }
+
+internal val JsonExtraKeysIndexKey = DescriptorSchemaCache.Key<Int>()
+
+/**
+ * Returns the element index of the property annotated with [JsonExtraKeys],
+ * or -1 if no such property exists in this descriptor.
+ *
+ * Validates on first computation and throws [SerializationException] if:
+ *  - more than one property is annotated with [JsonExtraKeys];
+ *  - the annotated property is not of type `Map<String, JsonElement>`.
+ *
+ * The result is memoised in the per-[Json] [DescriptorSchemaCache].
+ */
+internal fun SerialDescriptor.jsonExtraKeysIndex(json: Json): Int =
+    json.schemaCache.getOrPut(this, JsonExtraKeysIndexKey) {
+        computeJsonExtraKeysIndex()
+    }
+
+private fun SerialDescriptor.computeJsonExtraKeysIndex(): Int {
+    var foundIndex = -1
+    var duplicates: MutableList<String>? = null
+    for (i in 0 until elementsCount) {
+        if (getElementAnnotations(i).any { it is JsonExtraKeys }) {
+            if (foundIndex == -1) {
+                foundIndex = i
+            } else {
+                val list = duplicates ?: mutableListOf(getElementName(foundIndex)).also { duplicates = it }
+                list.add(getElementName(i))
+            }
+        }
+    }
+    duplicates?.let {
+        throw SerializationException(
+            "Class '$serialName' has more than one property annotated with @JsonExtraKeys: " +
+                it.joinToString(", ") { name -> "'$name'" } +
+                ". At most one such property is allowed per class."
+        )
+    }
+    if (foundIndex == -1) return -1
+    validateJsonExtraKeysProperty(foundIndex)
+    return foundIndex
+}
+
+private fun SerialDescriptor.validateJsonExtraKeysProperty(index: Int) {
+    val propertyName = getElementName(index)
+    val elementDescriptor = getElementDescriptor(index)
+    val rejection = "Property '$propertyName' of '$serialName' is annotated with @JsonExtraKeys " +
+        "but its type is not 'Map<String, JsonElement>'"
+    if (elementDescriptor.kind != StructureKind.MAP) {
+        throw SerializationException("$rejection (kind is ${elementDescriptor.kind}).")
+    }
+    val keyDescriptor = elementDescriptor.getElementDescriptor(0)
+    if (keyDescriptor.kind != PrimitiveKind.STRING) {
+        throw SerializationException(
+            "$rejection: map key type must be String but was '${keyDescriptor.serialName}'."
+        )
+    }
+    val valueDescriptor = elementDescriptor.getElementDescriptor(1)
+    if (valueDescriptor != JsonElementSerializer.descriptor) {
+        throw SerializationException(
+            "$rejection: map value type must be JsonElement but was '${valueDescriptor.serialName}'."
+        )
+    }
+}
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -189,9 +189,9 @@ internal fun SerialDescriptor.jsonExtraKeysIndex(json: Json): Int {
         // Some partially-customized descriptors (e.g. created via the @Serializer
         // companion shortcut) have a broken hashCode() that throws an
         // index-out-of-bounds exception when the schema cache hashes them.
-        // Unlike useAlternativeNames — which touches hashCode only on the
-        // unknown-key slow path and documents `useAlternativeNames = false` as
-        // the remedy — this lookup runs eagerly for every class descriptor, so
+        // useAlternativeNames touches hashCode only on the unknown-key slow
+        // path and documents `useAlternativeNames = false` as the remedy. This
+        // lookup instead runs eagerly for every class descriptor, so
         // without this fallback the mere presence of such a descriptor would
         // break on upgrade even when @JsonExtraKeys is not used at all
         // (see JsonCustomSerializersTest). Fall back to uncached computation.
@@ -267,7 +267,7 @@ internal val JsonExtraKeysCollisionNamesKey = DescriptorSchemaCache.Key<Set<Stri
  * Names that a [JsonExtraKeys] bucket is not allowed to contain when written
  * back during encoding: every JSON name that decoding would resolve to a
  * declared property (serial names, [JsonNames] aliases and naming-strategy
- * forms) — except names resolving to the bucket itself, which decoding
+ * forms), except names resolving to the bucket itself, which decoding
  * captures into the bucket and which therefore round-trip safely.
  *
  * Memoised in the per-[Json] [DescriptorSchemaCache].
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -79,6 +79,21 @@ internal fun SerialDescriptor.getJsonEncodedNames(json: Json): Set<String> {
     return if (strategy == null) jsonCachedSerialNames() else serializationNamesIndices(json, strategy).toSet()
 }
 
+// All JSON names that decode would resolve to a declared property, including @JsonNames aliases
+internal fun SerialDescriptor.getJsonDecodingNames(json: Json): Set<String> {
+    val strategy = namingStrategy(json)
+    return if (strategy == null && !json.configuration.useAlternativeNames) {
+        jsonCachedSerialNames()
+    } else {
+        val result = mutableSetOf<String>()
+        if (strategy == null) {
+            result.addAll(jsonCachedSerialNames())
+        }
+        result.addAll(json.deserializationNamesMap(this).keys)
+        result
+    }
+}
+
 internal fun SerialDescriptor.namingStrategy(json: Json) =
     if (kind == StructureKind.CLASS) json.configuration.namingStrategy else null
 
@@ -178,7 +193,13 @@ internal val JsonExtraKeysIndexKey = DescriptorSchemaCache.Key<Int>()
  * The result is memoised in the per-[Json] [DescriptorSchemaCache].
  */
 internal fun SerialDescriptor.jsonExtraKeysIndex(json: Json): Int =
-    json.schemaCache.getOrPut(this, JsonExtraKeysIndexKey) {
+    try {
+        json.schemaCache.getOrPut(this, JsonExtraKeysIndexKey) {
+            computeJsonExtraKeysIndex()
+        }
+    } catch (e: ArrayIndexOutOfBoundsException) {
+        // Defensive fallback for hand-rolled descriptors (e.g. external @Serializer)
+        // whose hashCode() crashes due to mismatched childSerializers array.
         computeJsonExtraKeysIndex()
     }
 
@@ -190,8 +211,12 @@ private fun SerialDescriptor.computeJsonExtraKeysIndex(): Int {
             if (foundIndex == -1) {
                 foundIndex = i
             } else {
-                val list = duplicates ?: mutableListOf(getElementName(foundIndex)).also { duplicates = it }
-                list.add(getElementName(i))
+                val list = duplicates
+                if (list == null) {
+                    duplicates = mutableListOf(getElementName(foundIndex), getElementName(i))
+                } else {
+                    list.add(getElementName(i))
+                }
             }
         }
     }
@@ -222,9 +247,19 @@ private fun SerialDescriptor.validateJsonExtraKeysProperty(index: Int) {
         )
     }
     val valueDescriptor = elementDescriptor.getElementDescriptor(1)
+    // Strict identity check: we deliberately reject custom JsonElement
+    // serializers because the streaming decoder synthesises a JsonObject
+    // and feeds it to the standard JsonElementSerializer (see ARCHITECTURE.md §8.3).
+    // Loosening this requires also routing the user's serializer through
+    // the synthetic decode path.
     if (valueDescriptor != JsonElementSerializer.descriptor) {
         throw SerializationException(
             "$rejection: map value type must be JsonElement but was '${valueDescriptor.serialName}'."
+        )
+    }
+    if (getElementAnnotations(index).any { it is JsonNames }) {
+        throw SerializationException(
+            "$rejection: @JsonNames is not allowed on a @JsonExtraKeys property."
         )
     }
 }
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -197,9 +197,11 @@ internal fun SerialDescriptor.jsonExtraKeysIndex(json: Json): Int =
         json.schemaCache.getOrPut(this, JsonExtraKeysIndexKey) {
             computeJsonExtraKeysIndex()
         }
-    } catch (e: ArrayIndexOutOfBoundsException) {
-        // Defensive fallback for hand-rolled descriptors (e.g. external @Serializer)
-        // whose hashCode() crashes due to mismatched childSerializers array.
+    } catch (_: ArrayIndexOutOfBoundsException) {
+        // Some partially-customized descriptors (e.g. created via the @Serializer
+        // companion shortcut) have a broken hashCode() that throws AIOOBE.
+        // The library already documents this limitation around `useAlternativeNames`
+        // (see JsonCustomSerializersTest.kt). Fall back to uncached computation.
         computeJsonExtraKeysIndex()
     }
 
@@ -249,7 +251,7 @@ private fun SerialDescriptor.validateJsonExtraKeysProperty(index: Int) {
     val valueDescriptor = elementDescriptor.getElementDescriptor(1)
     // Strict identity check: we deliberately reject custom JsonElement
     // serializers because the streaming decoder synthesises a JsonObject
-    // and feeds it to the standard JsonElementSerializer (see ARCHITECTURE.md §8.3).
+    // and feeds it to the standard JsonElementSerializer.
     // Loosening this requires also routing the user's serializer through
     // the synthetic decode path.
     if (valueDescriptor != JsonElementSerializer.descriptor) {
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
@@ -80,21 +80,6 @@ internal fun SerialDescriptor.getJsonEncodedNames(json: Json): Set<String> {
     return if (strategy == null) jsonCachedSerialNames() else serializationNamesIndices(json, strategy).toSet()
 }
 
-// All JSON names that decode would resolve to a declared property, including @JsonNames aliases
-internal fun SerialDescriptor.getJsonDecodingNames(json: Json): Set<String> {
-    val strategy = namingStrategy(json)
-    return if (strategy == null && !json.configuration.useAlternativeNames) {
-        jsonCachedSerialNames()
-    } else {
-        val result = mutableSetOf<String>()
-        if (strategy == null) {
-            result.addAll(jsonCachedSerialNames())
-        }
-        result.addAll(json.deserializationNamesMap(this).keys)
-        result
-    }
-}
-
 internal fun SerialDescriptor.namingStrategy(json: Json) =
     if (kind == StructureKind.CLASS) json.configuration.namingStrategy else null
 
@@ -189,7 +174,7 @@ internal val JsonExtraKeysIndexKey = DescriptorSchemaCache.Key<Int>()
  *
  * Validates on first computation and throws [SerializationException] if:
  *  - more than one property is annotated with [JsonExtraKeys];
- *  - the annotated property is not of type `Map<String, V>` (string-keyed map).
+ *  - the annotated property is not of type `JsonObject` / `Map<String, JsonElement>`.
  *
  * The result is memoised in the per-[Json] [DescriptorSchemaCache].
  */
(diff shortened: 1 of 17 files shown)


hrach:
Just a few days ago, I was thinking that I'd utilize this feature. 

My use case: parsing open-api specs. Additional props may be added to the object and my generator's plugin system may want to read that (those `x-` prefixed props):

```json
      "UnifiedApiResponseAdditionsReturnedTypeResponseDto" : {
        "properties" : {
          "error" : {
            "type" : "string"
          },
          "flashes" : {
            "items" : {
              "$ref" : "#/components/schemas/FlashMessage"
            },
            "type" : "array"
          },
          "result" : {
            "$ref" : "#/components/schemas/AdditionsReturnedTypeResponseDto"
          },
          "traceId" : {
            "type" : "string"
          }
        },
        "required" : [ "flashes" ],
        "type" : "object",
        "x-unified-response" : true,
        "x-unified-response-result-required" : true
      }
```

inemtsev:
> Just a few days ago, I was thinking that I'd utilize this feature.
> 
> My use case: parsing open-api specs. Additional props may be added to the object and my generator's plugin system may want to read that (those `x-` prefixed props):
> 
> ```json
>       "UnifiedApiResponseAdditionsReturnedTypeResponseDto" : {
>         "properties" : {
>           "error" : {
>             "type" : "string"
>           },
>           "flashes" : {
>             "items" : {
>               "$ref" : "#/components/schemas/FlashMessage"
>             },
>             "type" : "array"
>           },
>           "result" : {
>             "$ref" : "#/components/schemas/AdditionsReturnedTypeResponseDto"
>           },
>           "traceId" : {
>             "type" : "string"
>           }
>         },
>         "required" : [ "flashes" ],
>         "type" : "object",
>         "x-unified-response" : true,
>         "x-unified-response-result-required" : true
>       }
> ```

Yes, I think this feature would be useful in your case :)

inemtsev:
@sandwwraith @pdvrieze Please take a look

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -164,3 +179,87 @@ internal inline fun Json.tryCoerceValue(
...
+    }
+    duplicates?.let {
+        throw SerializationException(
+            "Class '$serialName' has more than one property annotated with @JsonExtraKeys: " +
+                it.joinToString(", ") { name -> "'$name'" } +
+                ". At most one such property is allowed per class."
+        )
+    }
+    if (foundIndex == -1) return -1
+    validateJsonExtraKeysProperty(foundIndex)
```
pdvrieze, line 229: Some of this code is doing more than is needed on the "happy path". I am wondering if a configuration option to disable expensive validation would be worthwhile (so it for example doesn't check for duplicates, and thus may start from the last index instead of the first).

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -164,3 +179,87 @@ internal inline fun Json.tryCoerceValue(
...
+    // Strict identity check: we deliberately reject custom JsonElement
+    // serializers because the streaming decoder synthesises a JsonObject
+    // and feeds it to the standard JsonElementSerializer (see ARCHITECTURE.md §8.3).
+    // Loosening this requires also routing the user's serializer through
+    // the synthetic decode path.
+    if (valueDescriptor != JsonElementSerializer.descriptor) {
+        throw SerializationException(
+            "$rejection: map value type must be JsonElement but was '${valueDescriptor.serialName}'."
+        )
+    }
```
pdvrieze: What about subtypes (e.g. you only want to allow primitive extra attributes)?

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
```diff
@@ -164,3 +179,87 @@ internal inline fun Json.tryCoerceValue(
...
+ */
+internal fun SerialDescriptor.jsonExtraKeysIndex(json: Json): Int =
+    try {
+        json.schemaCache.getOrPut(this, JsonExtraKeysIndexKey) {
+            computeJsonExtraKeysIndex()
+        }
+    } catch (e: ArrayIndexOutOfBoundsException) {
+        // Defensive fallback for hand-rolled descriptors (e.g. external @Serializer)
+        // whose hashCode() crashes due to mismatched childSerializers array.
+        computeJsonExtraKeysIndex()
```
pdvrieze: If a custom serializer has a crashing hashCode that should be an error, not ignored. I don't think you want to have special handling for broken custom serializers/descriptors (and note that for this to happen the user would have had to implement a custom hashCode in the first place). Separately the way descriptors are used (lots of wrapping) means that custom descriptors have very limited value so should be rare.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -46,6 +46,15 @@ internal open class StreamingJsonDecoder(
     private val configuration = json.configuration
 
     private val elementMarker: JsonElementMarker? = if (configuration.explicitNulls) null else JsonElementMarker(descriptor)
+    private val extraKeysMap = mutableMapOf<SerialDescriptor, LinkedHashMap<String, JsonElement>>()
+    private val extraKeysEmittedMap = mutableMapOf<SerialDescriptor, Boolean>()
```
pdvrieze: Why do you want to record per descriptor whether an extra key was found in it?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -46,6 +46,15 @@ internal open class StreamingJsonDecoder(
     private val configuration = json.configuration
 
     private val elementMarker: JsonElementMarker? = if (configuration.explicitNulls) null else JsonElementMarker(descriptor)
+    private val extraKeysMap = mutableMapOf<SerialDescriptor, LinkedHashMap<String, JsonElement>>()
```
pdvrieze: What about treating the extra keys as any regular collection/map element and go through the map deserializer (potentially synthesizing a single item map). This would also mean that handling of types works automatically.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -111,6 +120,16 @@ internal open class StreamingJsonDecoder(
                 discriminatorHolder
             )
             else -> if (mode == newMode && json.configuration.explicitNulls) {
+                val extraKeysIndex = try {
+                    descriptor.jsonExtraKeysIndex(json)
+                } catch (_: ArrayIndexOutOfBoundsException) {
+                    -1
+                }
```
pdvrieze: This should not throw, and if it does, it should not be silently ignored.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -111,6 +120,16 @@ internal open class StreamingJsonDecoder(
...
+                if (extraKeysIndex != -1) {
+                    val savedMap = try { extraKeysMap.remove(descriptor) } catch (_: ArrayIndexOutOfBoundsException) { null }
+                    val savedEmitted = try { extraKeysEmittedMap.remove(descriptor) } catch (_: ArrayIndexOutOfBoundsException) { null }
+                    extraKeysSavedStack.add(ExtraKeysFrame(descriptor, savedMap, savedEmitted))
+                }
```
pdvrieze, line 159: You add a lot of complexity, where using the fact that collection serializers allow for updating (yes it requires a check/cast) would make this a lot easier to handle (and controlled by the serializer of the containing object).

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -220,14 +260,15 @@ internal open class StreamingJsonDecoder(
     )
 
     private fun decodeObjectIndex(descriptor: SerialDescriptor): Int {
+        val extraKeysIndex = descriptor.jsonExtraKeysIndex(json)
```
pdvrieze: This is a potentially expensive operation. You should only retrieve that if you find an unknown key. That way code that does not have unknown keys and does not use this feature does not need to pay for it.

I see that the reason you get it here is to be able to stop recognizing the element name if annotated. Should `getJsonNameIndex` not implement that policy (and just making `index !=extraKeysIndex` always false)?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -236,20 +277,29 @@ internal open class StreamingJsonDecoder(
...
-                hasComma = handleUnknown(descriptor, key)
+                hasComma = handleUnknown(descriptor, key, extraKeysIndex)
             }
         }
         if (hasComma && !json.configuration.allowTrailingComma) lexer.invalidTrailingComma()
 
+        if (extraKeysIndex != -1 && !extraKeysEmittedMap.getOrPut(descriptor) { false }) {
+            extraKeysEmittedMap[descriptor] = true
+            return extraKeysIndex
+        }
```
pdvrieze: collection serializers allow for updating, so this can be done incrementally.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
```diff
@@ -236,20 +277,29 @@ internal open class StreamingJsonDecoder(
...
 
-    private fun handleUnknown(descriptor: SerialDescriptor, key: String): Boolean {
-        if (descriptor.ignoreUnknownKeys(json) || discriminatorHolder.trySkip(key)) {
+    private fun handleUnknown(descriptor: SerialDescriptor, key: String, extraKeysIndex: Int): Boolean {
+        if (discriminatorHolder.trySkip(key)) {
+            lexer.skipElement(configuration.isLenient)
+        } else if (extraKeysIndex != -1) {
+            val collector = extraKeysMap.getOrPut(descriptor) { LinkedHashMap() }
+            collector[key] = JsonTreeReader(configuration, lexer).read()
+        } else if (descriptor.ignoreUnknownKeys(json)) {
```
pdvrieze: You may want to think a bit more about this. It is a bit clunky (doing a ping-pong with `decodeObjectIndex`).

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt
```diff
@@ -152,6 +168,44 @@ internal class StreamingJsonEncoder(
...
     }
 
+    override fun <T> encodeSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        serializer: SerializationStrategy<T>,
+        value: T
+    ) {
+        val extraKeysIndex = descriptor.jsonExtraKeysIndex(json)
+        if (extraKeysIndex == index) {
```
pdvrieze: Because the "else" (I know you used a return) bit is much shorter (and the common option) it would make sense to flip the if condition here (so you can see the condition when looking at the else clause) - you could even use a return and not have an else clause at all.

pdvrieze: Separately, should you not check for the mode. It is only valid in object mode, not in array or other modes

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -226,6 +237,10 @@ private open class JsonTreeDecoder(
                 return index
             }
         }
+        if (extraKeysIndex != -1 && !extraKeysEmitted) {
+            extraKeysEmitted = true
+            return extraKeysIndex
+        }
```
pdvrieze: Collections are incremental, you don't need to collect the inputs to then provide the elements all at once.

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -263,7 +278,20 @@ private open class JsonTreeDecoder(
         return fallbackName ?: baseName
     }
 
-    override fun currentElement(tag: String): JsonElement = value.getValue(tag)
+    override fun currentElement(tag: String): JsonElement {
+        val descriptor = currentDescriptor
+        if (descriptor != null && extraKeysIndex != -1 && tag == descriptor.getJsonElementName(json, extraKeysIndex)) {
```
pdvrieze: should this use `getJsonElementName` or just the element name from the descriptor. The name is never written/read (unless such a key happens to be inside the map).

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt
```diff
@@ -227,6 +229,40 @@ private open class JsonTreeEncoder(
...
     }
 
+    override fun <T> encodeSerializableElement(
+        descriptor: SerialDescriptor,
+        index: Int,
+        serializer: SerializationStrategy<T>,
+        value: T
+    ) {
+        val extraKeysIndex = descriptor.jsonExtraKeysIndex(json)
+        if (extraKeysIndex == index) {
```
pdvrieze: Again, it makes sense to me to flip the if clause (so the short default comes first).

pdvrieze (commented):
I've had a look at it. The implementation is correct, but it adds a degree of complexity. There are some bits where this complexity can be reduced.

A key consideration is that the collection serializers actually support incremental parsing (updating). This would allow you to just handle/parse unknown keys incrementally (rather than having to collect them first). At that point you can just return the index of the "other attributes" element, and in the relevant (map) decode function handle parsing it.

The one gotcha is handling the class discriminator that you need to exclude from returning the index.

As optimization, the schemaCache already has code to create a map from name to index. As part of initialization it loops through all elements and annotations to determine the effective name(s) - supporting `@JsonNames`. It would certainly be more efficient to check for the `@JsonExtraKeys` element at the same time. (I know that this is a bit against the idea of SchemaCache, but the design is more extensible than it needs to be for an internal type (keep in mind multi-threading issues though - don't make the value mutable))

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
inemtsev: validateJsonExtraKeysProperty() run exactly once per descriptor per Json instance (not once per encode/decode or anything like that), so I don't think this is worth trying to optimize.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
inemtsev: Do you mean something like: Map<String, JsonPrimitive>?

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
inemtsev: You're right, this is left over code from another approach I was trying. Will fix.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
inemtsev: Actually there is a scenario that this handles even when no unknown keys are present:
```
@Serializable
data class Foo(
    val a: Int,
    @JsonExtraKeys val extras: Map<String, JsonElement> = emptyMap()
)
val input = """{"a": 1, "extras": {"x": 99}}"""
```

Regarding the second, I dont think `getJsonNameIndex` should become bucket-aware and will require it to look up jsonExtraKeysIndex(json) per-key.

I did optimize the caching, so the cost of this method is close to 0 now

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
inemtsev: Hmmm, descriptor.getElementName would stop working here if naming strategy like "JsonNamingStrategy.SnakeCase" is used.

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
inemtsev: Fair point. I'll try to refactor this for clarity.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
pdvrieze: @inemtsev Yes, that would be correct. Note that if you would delegate to the actual serializer for the property (by wrapping the unknown key in decoder/composite decoder that simulates it being inside its own map) the value type deserialization would happen automatically without additional tests (but would fail if the type does not work).

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
pdvrieze: @inemtsev The thing is that naming strategies are irrelevant for this as it is not actually name present in json anyway (although it may be that avoiding the renaming earlier on is worse than doing the renaming here)

inemtsev: Agreed. The translated name has no semantic role on the wire. The reason currentElement calls getJsonElementName is just to match the strategy-translated tag the framework already produced for the bucket

formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt
inemtsev: Done

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt
inemtsev: @pdvrieze That's a great tip actually. I'll have to change my strategy a bit though, I'll try to push in a few days.

inemtsev:
@pdvrieze @sandwwraith I've made some significant changes as per suggestions, please take a look when you have a moment

Flavien:
Thanks for this PR, IMO this is a must-have feature of any JSON library as this is the only way to ensure safe round-tripping.

inemtsev:
@sandwwraith @fzhinkin Do you think we can try to get this into the next release (maybe under experimental?) This would be very handy for some of our migrations.

sandwwraith (commented):
I've looked through the implementation and have some general comments:

1. Since this feature involves some logic on the hot parser path, it would be wise to put it under the feature flag, as it was done with `useAlternativeNames`. 
2. You need to have benchmarks with both states of this flag to verify that it doesn't slow down things a lot when enabled by default. Take a look at `TwitterFeedBenchmark` or `CoerceInputValuesBenchmark` for examples
3. If this flag would be implemented, there would be no need to care for weird try-catches around broken descriptors in `SerialDescriptor.jsonExtraKeysIndex`
4. Why `Map<String, T>` was chose instead of a simple `JsonObject`? It seems highly inlikely to me that someone would not know what keys are there, but would know their shape. Using JsonObject for the unknown keys is the most straightforward way which also would likely simplify implementation.
5. Writing extra keys back. I understand this is a nice feature to have, but I'm not sure it is mandatory. It makes a lot of complications in encoding to flatten the keys. Again, if unknown keys holder is strictly a JsonObject, we could simply dump its .toString() onto the current stream and go on our way. Please investigate this approach, otherwise I doubt the benefits of having this functionality.

inemtsev:
> I've looked through the implementation and have some general comments:
> 
> 1. Since this feature involves some logic on the hot parser path, it would be wise to put it under the feature flag, as it was done with `useAlternativeNames`.
> 2. You need to have benchmarks with both states of this flag to verify that it doesn't slow down things a lot when enabled by default. Take a look at `TwitterFeedBenchmark` or `CoerceInputValuesBenchmark` for examples
> 3. If this flag would be implemented, there would be no need to care for weird try-catches around broken descriptors in `SerialDescriptor.jsonExtraKeysIndex`
> 4. Why `Map<String, T>` was chose instead of a simple `JsonObject`? It seems highly inlikely to me that someone would not know what keys are there, but would know their shape. Using JsonObject for the unknown keys is the most straightforward way which also would likely simplify implementation.
> 5. Writing extra keys back. I understand this is a nice feature to have, but I'm not sure it is mandatory. It makes a lot of complications in encoding to flatten the keys. Again, if unknown keys holder is strictly a JsonObject, we could simply dump its .toString() onto the current stream and go on our way. Please investigate this approach, otherwise I doubt the benefits of having this functionality.

Fair points. I liked the encoding ability because if the user decided to _serialize_ the deserialized object again (for logging or something) then the "extras" are spread back into JSON. For now, I will skip encoding this and document.

pdvrieze:
> 5\. Writing extra keys back. I understand this is a nice feature to have, but I'm not sure it is mandatory. It makes a lot of complications in encoding to flatten the keys. Again, if unknown keys holder is strictly a JsonObject, we could simply dump its .toString() onto the current stream and go on our way. Please investigate this approach, otherwise I doubt the benefits of having this functionality

Comparing this with XML (where the feature has been available for quite some time now) I would say that the ability to write back the additional keys would be relevant. In many ways, the extra entries would be opaque, so `JsonObject` would be a better representation although `Map<String, JsonElement>` would be valid too.

The biggest question is what to do when the unknownKeys map is updated/manipulated such that it contains overlapping keys. This would be an invalid state of the to-be-serialized object, but there is still the decision of how to handle it:
- Make it undefined behaviour (the format ignores the possibility, so for encoding to string would just have duplicate keys, for tree serialization the last value sticks)
- Ignore all but the first values (if unknown keys are serialized first, they "win")
- First write all "normal" keys and ignore all duplicate keys in the unknown keys
- The above too, but keep last, not first
- Throw exceptions on duplicate keys (either first or collecting all)

The undefined behaviour option is cheapest (it would be roughly dumping `toString()` to the end of the object), and duplicate keys are technically allowed in Json (although it is also a security breach waiting to happen).

The other options do require either keeping track of written keys or looking up the key on the descriptor. There I would be inclined to go for the exception option as the others are surprising, especially if it ends up depending on the order of the elements in the structure (I would go to serialize the unknown keys after all "regular" keys). This can be implemented by recording the unknown keys element if it occurs (and is not empty), and on the endStructure writing the values (if needed).

Flavien:
Yes I agree the ability to write back the unknown keys is necessary for this feature to be useful. And I agree it makes more sense to store the unrecognised keys in a `JsonObject`.

inemtsev:
> Yes I agree the ability to write back the unknown keys is necessary for this feature to be useful. And I agree it makes more sense to store the unrecognised keys in a `JsonObject`.

Upon more thought, this does strike a good middle-ground. Will update PR shortly. 
@pdvrieze For when unknownKeys map is updated/manipulated, I will just throw for now. This is a very unlikely case anyway.

pdvrieze:
@inemtsev Exceptions on duplicate keys is a good solution. The main reason to avoid it is if you want to avoid the check for a duplicate, but given the potential for duplicate keys to cause security challenges (inconsistent values used with different checks/systems) it is probably worthwhile to check for duplicates (and throw).

inemtsev:
- I added a flag **useExtraKeys** and changed unknown keys to bucket into JsonObject. 
- I tried removing the try-catch once the flag existed, but 37 existing tests fail, because unlike useAlternativeNames (which only hashes descriptors on the unknown-key slow path), the bucket lookup must run eagerly for every class descriptor, so removal breaks apps with partially-customized serializers that never use @JsonExtraKeys.

#### Benchmarks
JMH, JDK 21 (arm64 macOS). Micro: `JsonExtraKeysBenchmark` (10-field object, ops/µs).
Macro: `TwitterFeedBenchmark` (ops/s). Higher is better.

#### Hot path — classes without `@JsonExtraKeys`

| Scenario | master | this PR, `useExtraKeys = false` (default) | this PR, `useExtraKeys = true` |
|---|---|---|---|
| decode 10-field object | 3.33 ± 0.12 | 3.51 ± 0.40 | 3.39 ± 0.31 |
| encode 10-field object | 10.13 ± 0.26 | 10.27 ± 0.37 | 9.84 ± 0.42 |
| decode TwitterFeed (macro) | 1562 ± 12 | **1596 ± 47 — parity** | 1449 ± 10 (−7%) |
| encode TwitterFeed (macro) | 3221 ± 86 | **3223 ± 94 — parity** | 3046 ± 113 (−5%) |

#### Feature cost — classes with a `@JsonExtraKeys` bucket (`useExtraKeys = true`)

| Scenario | Map<String, JsonElement> bucket | JsonObject bucket | Reference |
|---|---|---|---|
| decode, input has no unknown keys | 3.62 ± 0.10 | 3.61 ± 0.33 | ≥ plain decode: declaring a bucket is free on clean input |
| decode, capture 3 unknown keys | 1.76 ± 0.16 | 1.64 ± 0.12 | 2.54 ± 0.10 with `ignoreUnknownKeys` (keys dropped) |
| encode, write back 3 captured keys | 4.64 ± 0.10 | 4.67 ± 0.16 | 10.27 plain encode (10 keys vs 13 keys written) |

### Summary

Looks like with flag off the cost is basically free. But, enabling the flag costs about 5-7%. Probably keeping it false, unless needed is ideal?

inemtsev:
@sandwwraith I pushed changes listed above. 

I also discovered an optimization that would help operations like this, but I will open another PR for that:
https://github.com/Kotlin/kotlinx.serialization/pull/3215

inemtsev:
@sandwwraith @pdvrieze Two updates since the last round.

The try-catch in jsonExtraKeysIndex has to stay. #3215 would have fixed the descriptor hashCode() crash underneath it, but it was closed due to #3084, so descriptors from `@Serializer(forClass = ...)` will keep throwing when the schema cache hashes them. The catch needs to stay to handle that case now. Everything else from the review is done: useExtraKeys flag (off by default), JsonObject bucket, write-back at endStructure with an exception on collisions, and benchmarks for both flag states.

With #2860 merged, ProtoBuf preserves unknown fields and JSON doesn't. Can this go in under @ExperimentalSerializationApi in the next release so both formats have it? I'll rebase onto current dev and add a docs/json.md section mirroring the ProtoBuf one right now.

Anything else you'd like changed?

PR #3192, opened 2026-05-04
Use CompositeDecoder::decodeCollectionSize to pre-allocate collections

### LINKED ISSUE #3153: Pre-allocate Map and Collection capacities using CBOR definite-length headers
Currently, when the kotlinx-serialization-cbor library decodes a CBOR map or array, the default MapSerializer and ListSerializer instantiate empty collections (e.g., LinkedHashMap or ArrayList) using their default initial capacities (typically 16).

If a CBOR payload contains a definite-length map or array with thousands of elements, populating the collection triggers continuous, expensive memory reallocations and array-copying under the hood. On memory- and CPU-constrained environments like Android, this creates a noticeable performance bottleneck and unnecessary garbage collection overhead on hot paths.

When the CBOR decoder parses a definite-length map (Major Type 5) or array (Major Type 4), it already reads the exact size from the binary header. The library should use this parsed size to initialize the backing collections with an exact capacity.

fzhinkin:
Reimplemented `AbstractCollectionSerializer::merge` to read expected collection size first and pass it as an initial capacity to a corresponding builder. Builder were updated to take the capacity into account.

The change is quite intrusive, but for formats providing information about the collection size before reading it (it's only CBOR for now, though) it reduces the footprint and slightly improve decoding performance.

Fixes #3153


fzhinkin:
Raw benchmarking results:
- [baseline JSON](https://github.com/user-attachments/files/27416413/baseline.f5.json)
- [branch JSON](https://github.com/user-attachments/files/27416419/opt.f5.json)

Comparison: https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fzhinkin/5e63a49d551118a7eba58620b60c618c/raw/b0e9331a1fe133aa94ee8419a9a2b996eb9514d4/baseline.json,https://gist.githubusercontent.com/fzhinkin/5e63a49d551118a7eba58620b60c618c/raw/b0e9331a1fe133aa94ee8419a9a2b996eb9514d4/opt.json

core/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
fzhinkin, line 123: To discuss: we can avoid a copy here if `position == buffer.size`, but we need to explicitly invalidate the builder, JIC.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -24,11 +25,16 @@ public sealed class AbstractCollectionSerializer<Element, Collection, Builder> :
 
     @InternalSerializationApi
     public fun merge(decoder: Decoder, previous: Collection?): Collection {
-        val builder = previous?.toBuilder() ?: builder()
-        val startIndex = builder.builderSize()
         val compositeDecoder = decoder.beginStructure(descriptor)
+        val expectedSize = compositeDecoder.decodeCollectionSize(descriptor)
+        val builder = if (previous != null) {
+            previous.toBuilder().also { it.checkCapacity(expectedSize) }
```
pdvrieze: This should add the new new "expectedSize" to the current size when calling checkCapacity. The common case for this branch is for incremental parsing of list elements.

It also makes sense to pass the new needed size to `toBuilder` to avoid an extra copy in the case that a new builder is actually created (not default behaviour).

fzhinkin: Nice catch, thanks for pointing to it!

Not only the code does not work as expected, it can also fail in runtime if `checkCapacity` can't handle negative `expectedSize` (which is the case for all non-empty implementations, I guess).

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -204,7 +204,12 @@ internal class ReferenceArraySerializer<ElementKlass : Any, Element : ElementKla
     @Suppress("UNCHECKED_CAST")
     override fun ArrayList<Element>.toResult(): Array<Element> = toNativeArrayImpl<ElementKlass, Element>(kClass)
 
-    override fun Array<Element>.toBuilder(): ArrayList<Element> = ArrayList(this.asList())
+    override fun Array<Element>.toBuilder(expectedAdditionalSize: Int): ArrayList<Element> {
+        checkBuildersInitialCapacity(expectedAdditionalSize)
+        val expectedCapacity = size + expectedAdditionalSize.coerceAtLeast(0)
```
pdvrieze: It may be more optimal to coerce it to at least 1. The use case for extending lists kind of implies to add at least one new element.

fzhinkin:
Updated benchmarking results:
- [baseline.v2.json](https://github.com/user-attachments/files/27655284/baseline.v2.json)
- [opt.v2.json](https://github.com/user-attachments/files/27655285/opt.v2.json)
- [comparison](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fzhinkin/a483976ce0caafa0313e3aa2420cdbe3/raw/a84f5f870232efeffc953c42187316f2cd1d4aa6/baseline.v2.json,https://gist.githubusercontent.com/fzhinkin/a483976ce0caafa0313e3aa2420cdbe3/raw/a84f5f870232efeffc953c42187316f2cd1d4aa6/opt.v2.json)

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -189,13 +194,22 @@ internal class ReferenceArraySerializer<ElementKlass : Any, Element : ElementKla
...
     override fun ArrayList<Element>.builderSize(): Int = size
 
     @Suppress("UNCHECKED_CAST")
     override fun ArrayList<Element>.toResult(): Array<Element> = toNativeArrayImpl<ElementKlass, Element>(kClass)
 
-    override fun Array<Element>.toBuilder(): ArrayList<Element> = ArrayList(this.asList())
+    override fun Array<Element>.toBuilder(expectedAdditionalSize: Int): ArrayList<Element> {
+        checkBuildersInitialCapacity(expectedAdditionalSize)
+        val expectedCapacity = size + expectedAdditionalSize.coerceAtLeast(0)
+        return ArrayList<Element>(expectedAdditionalSize).also { it.addAll(this) }
```
sandwwraith: Should be `ArrayList<Element>(expectedCapacity)`, right?

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -8,27 +8,33 @@ package kotlinx.serialization.internal
...
     protected abstract fun Builder.checkCapacity(size: Int)
 
     abstract override fun serialize(encoder: Encoder, value: Collection)
 
     @InternalSerializationApi
     public fun merge(decoder: Decoder, previous: Collection?): Collection {
-        val builder = previous?.toBuilder() ?: builder()
-        val startIndex = builder.builderSize()
         val compositeDecoder = decoder.beginStructure(descriptor)
+        val expectedSize = compositeDecoder.decodeCollectionSize(descriptor)
```
sandwwraith, line 29: I suggest checking that it is not less than -1 right here instead of calling `checkBuildersInitialCapacity(initialCapacityHint)` in every implementation

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -213,10 +227,22 @@ internal abstract class CollectionSerializer<E, C: Collection<E>, B>(element: KS
...
+        checkBuildersInitialCapacity(initialCapacityHint)
+        return ArrayList(initialCapacityHint)
+    }
     override fun ArrayList<E>.builderSize(): Int = size
     override fun ArrayList<E>.toResult(): List<E> = this
-    override fun List<E>.toBuilder(): ArrayList<E> = this as? ArrayList<E> ?: ArrayList(this)
+    override fun List<E>.toBuilder(expectedAdditionalSize: Int): ArrayList<E> {
+        if (expectedAdditionalSize != -1) checkBuildersInitialCapacity(expectedAdditionalSize)
+        if (this is ArrayList) {
+            if (expectedAdditionalSize == -1) return this
```
sandwwraith:
```suggestion
            if (expectedAdditionalSize <= 0) return this
```

We do not need to reallocate if no elements will be added

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -8,27 +8,33 @@ package kotlinx.serialization.internal
...
     abstract override fun serialize(encoder: Encoder, value: Collection)
 
     @InternalSerializationApi
     public fun merge(decoder: Decoder, previous: Collection?): Collection {
-        val builder = previous?.toBuilder() ?: builder()
-        val startIndex = builder.builderSize()
         val compositeDecoder = decoder.beginStructure(descriptor)
+        val expectedSize = compositeDecoder.decodeCollectionSize(descriptor)
+        val builder = if (previous != null) {
+            previous.toBuilder(expectedSize)
```
sandwwraith: I think passing here `previous.toBuider(previous.collectionSize() + expectedSize.coerceAtLeast(0))` would also significantly simplify implementations

core/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
sandwwraith: builders do not live outside of `merge()` IIRC

sandwwraith (changes requested):

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
pdvrieze: Maybe the coerceAtLeast should be 1 (the idea that this would not be called if there is not at least one element added, and the exception would be much less likely than this case).

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Apparently, I was in the "let's write over 9000 LOC" mood. :/

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
sandwwraith: Incoming collection can be empty, there's no need to re-allocate existing one in this case

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Fixed

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Fixed

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Fixed

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Fixed

hfhbd:
Nice! BTW it's not only CBOR but any 3rd party format too, like my CSV format that returns a List.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -8,27 +8,36 @@ package kotlinx.serialization.internal
...
 @InternalSerializationApi
 public sealed class AbstractCollectionSerializer<Element, Collection, Builder> : KSerializer<Collection> {
     protected abstract fun Collection.collectionSize(): Int
     protected abstract fun Collection.collectionIterator(): Iterator<Element>
-    protected abstract fun builder(): Builder
+    protected abstract fun builder(initialCapacityHint: Int): Builder
     protected abstract fun Builder.builderSize(): Int
     protected abstract fun Builder.toResult(): Collection
-    protected abstract fun Collection.toBuilder(): Builder
+    protected abstract fun Collection.toBuilder(expectedSize: Int): Builder
```
sandwwraith, line 20: With the deletion of `readSize`, `Builder.checkCapacity` no longer needed and can be safely removed

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -213,10 +227,20 @@ internal abstract class CollectionSerializer<E, C: Collection<E>, B>(element: KS
...
+        if (initialCapacityHint == -1) return arrayListOf()
+        return ArrayList(initialCapacityHint)
+    }
     override fun ArrayList<E>.builderSize(): Int = size
     override fun ArrayList<E>.toResult(): List<E> = this
-    override fun List<E>.toBuilder(): ArrayList<E> = this as? ArrayList<E> ?: ArrayList(this)
+    override fun List<E>.toBuilder(expectedSize: Int): ArrayList<E> {
+        if (this is ArrayList) {
+            if (expectedSize == size) return this
+            return this.also { it.ensureCapacity(expectedSize) }
```
sandwwraith, line 233: I think it is possible to drop `if (expectedSize == size) return this` because `ensureCapacity` does nothing if expectedSize <= size

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -227,10 +251,18 @@ internal class LinkedHashSetSerializer<E>(
...
-    override fun builder(): LinkedHashSet<E> = linkedSetOf()
+    override fun builder(initialCapacityHint: Int): LinkedHashSet<E> {
+        if (initialCapacityHint == -1) return linkedSetOf()
+        return newLinkedHashSet(initialCapacityHint)
+    }
     override fun LinkedHashSet<E>.builderSize(): Int = size
     override fun LinkedHashSet<E>.toResult(): Set<E> = this
-    override fun Set<E>.toBuilder(): LinkedHashSet<E> = this as? LinkedHashSet<E> ?: LinkedHashSet(this)
+    override fun Set<E>.toBuilder(expectedSize: Int): LinkedHashSet<E> {
+        if (this is LinkedHashSet<E>) return this
```
sandwwraith, line 254:
```suggestion
        if (this is LinkedHashSet<E>) return this.also { ensureCapacity(expectedSize) }
```

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -227,10 +251,18 @@ internal class LinkedHashSetSerializer<E>(
...
+        if (expectedSize == size) return LinkedHashSet(this)
```
sandwwraith: Maybe `expectedSize <= size` just to be sure? Although it will likely never occur

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -241,10 +273,17 @@ internal class HashSetSerializer<E>(
...
-    override fun builder(): HashSet<E> = HashSet()
+    override fun builder(initialCapacityHint: Int): HashSet<E> {
+        if (initialCapacityHint == -1) return HashSet()
+        return newHashSet(initialCapacityHint)
+    }
     override fun HashSet<E>.builderSize(): Int = size
     override fun HashSet<E>.toResult(): Set<E> = this
-    override fun Set<E>.toBuilder(): HashSet<E> = this as? HashSet<E> ?: HashSet(this)
+    override fun Set<E>.toBuilder(expectedSize: Int): HashSet<E> {
+        if (this is HashSet<E>) return this
```
sandwwraith, line 275: Same here (and in other collections), ensureCapacity?

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -273,10 +320,43 @@ internal class HashMapSerializer<K, V>(
...
+        if (this is HashMap<K, V>) return this
+        if (expectedSize == size) return HashMap(this)
+        return newHashMap<K, V>(expectedSize).also { it.putAll(this) }
+    }
+
     override fun HashMap<K, V>.checkCapacity(size: Int) {}
     override fun HashMap<K, V>.insertKeyValuePair(index: Int, key: K, value: V): Unit = set(key, value)
 }
+
+private fun estimateCapacityForHashMap(requiredCapacity: Int, loadFactor: Double): Int =
```
sandwwraith:
```suggestion
private fun estimateCapacityForHashMap(requiredCapacity: Int, loadFactor: Float): Int =
```

to avoid writing `.toDouble()` everywhere :)

core/commonMain/src/kotlinx/serialization/internal/PrimitiveArraysSerializers.kt
```diff
@@ -36,17 +40,19 @@ internal object ByteArraySerializer : KSerializer<ByteArray>,
 
 @PublishedApi
 internal class ByteArrayBuilder internal constructor(
-    bufferWithData: ByteArray
+    bufferWithData: ByteArray, initialPosition: Int
 ) : PrimitiveArrayBuilder<ByteArray>() {
+    internal constructor(bufferWithData: ByteArray) : this(bufferWithData, bufferWithData.size) {
+        ensureCapacity(INITIAL_SIZE)
```
sandwwraith: This constructor is called only in case of `expectedSize == size`, right? So we do not expect any more data, or we do not know the size of the data. In that case, wouldn't it be better to delay resizing until the next element arrives, as usual?

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: This way, I'm not gonna make it to the to contributors by the number of changed lines ;)

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Only ArrayList has it

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -189,14 +193,20 @@ internal class ReferenceArraySerializer<ElementKlass : Any, Element : ElementKla
...
+    }
     override fun ArrayList<Element>.builderSize(): Int = size
 
     @Suppress("UNCHECKED_CAST")
     override fun ArrayList<Element>.toResult(): Array<Element> = toNativeArrayImpl<ElementKlass, Element>(kClass)
 
-    override fun Array<Element>.toBuilder(): ArrayList<Element> = ArrayList(this.asList())
-    override fun ArrayList<Element>.checkCapacity(size: Int): Unit = ensureCapacity(size)
+    override fun Array<Element>.toBuilder(expectedSize: Int): ArrayList<Element> {
+        val expectedCapacity = expectedSize.coerceAtLeast(0)
```
sandwwraith, line 205: looks like `coerceAtLeast` is no longer needed here

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -213,11 +223,19 @@ internal abstract class CollectionSerializer<E, C: Collection<E>, B>(element: KS
...
+    }
     override fun ArrayList<E>.builderSize(): Int = size
     override fun ArrayList<E>.toResult(): List<E> = this
-    override fun List<E>.toBuilder(): ArrayList<E> = this as? ArrayList<E> ?: ArrayList(this)
-    override fun ArrayList<E>.checkCapacity(size: Int): Unit = ensureCapacity(size)
+    override fun List<E>.toBuilder(expectedSize: Int): ArrayList<E> {
+        if (this is ArrayList) {
+            return this.also { it.ensureCapacity(expectedSize) }
+        }
+        if (expectedSize == size) return ArrayList(this)
```
sandwwraith, line 235:
```suggestion
        if (expectedSize <= size) return ArrayList(this)
```

sandwwraith (approved):

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -8,27 +8,34 @@ package kotlinx.serialization.internal
...
     @InternalSerializationApi
     public fun merge(decoder: Decoder, previous: Collection?): Collection {
-        val builder = previous?.toBuilder() ?: builder()
-        val startIndex = builder.builderSize()
         val compositeDecoder = decoder.beginStructure(descriptor)
+        val expectedSize = compositeDecoder.decodeCollectionSize(descriptor)
+        require(expectedSize >= -1) {
+            "Decoded collection size should be either non-negative, or -1 (meaning, unknown). Was: $expectedSize"
+        }
+        val builder = if (previous != null) {
```
shanshin, line 30: Not a problem but: maybe we should invert the if?
As far as I understand, in practice, `previous` is used only for protobuf. 
But judging by this code, it seems the most common case.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
```diff
@@ -273,10 +311,42 @@ internal class HashMapSerializer<K, V>(
...
+    override fun Map<K, V>.toBuilder(expectedSize: Int): HashMap<K, V> {
+        if (this is HashMap<K, V>) return this
+        if (expectedSize <= size) return HashMap(this)
+        return newHashMap<K, V>(expectedSize).also { it.putAll(this) }
+    }
+
     override fun HashMap<K, V>.insertKeyValuePair(index: Int, key: K, value: V): Unit = set(key, value)
 }
+
+private fun estimateCapacityForHashMap(requiredCapacity: Int, loadFactor: Float): Int =
```
shanshin: I’m not sure about the exact implementation of `HashMap` for all platforms, but it seems that for some of them, `loadFactor` is ignored.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
pdvrieze: XML uses the previous option too (it allows collecting all tags with same name to be collected into a list, while those tags are also randomly distributed in the document). Even outside this, it is needed:
 - Lookahead (to determine/parse the entire "list") is potentially unbounded (as all children need parsing)
 - Alternatively the format would itself need to collate list elements and then produce them all at once (but there is no good interface for that).

Key to notice is that for formats that cannot predict list size or parse the entire list in one go, the "`previous==null`" case is the least common case.

core/commonMain/src/kotlinx/serialization/internal/CollectionSerializers.kt
fzhinkin: Nice catch! Indeed, on targets other than JVM maps and sets will be 25% larger than they should be.

fzhinkin: Fixed, PTAL

core/jvmMain/src/kotlinx/serialization/internal/Platform.kt
```diff
@@ -222,3 +223,8 @@ private inline fun loadSafe(block: () -> Unit) {
     } catch (_: ClassNotFoundException) {
     }
 }
+
+internal actual fun estimateCapacityForHashMap(requiredCapacity: Int): Int =
+    ceil(requiredCapacity / DEFAULT_HASH_MAP_LOAD_FACTORY).toInt()
+
+private const val DEFAULT_HASH_MAP_LOAD_FACTORY: Float = 0.75f
```
sandwwraith: FACTOR !! :)

PR #3195, opened 2026-05-12
CBOR: Correctly decode CborLabel values

### LINKED ISSUE #3190: CBOR `@CborLabel` with negative label value decodes incorrectly
**Describe the bug**
When using negative values for `@CborLabel` when combined with `preferCborLabelsOverNames` decoding fails

**To Reproduce**
Quick Test file
``` kotlin
package kotlinx.serialization.cbor

import kotlin.test.*

class NegativeLabelsOverNamesTest {
    private val withLabel = Cbor {
        preferCborLabelsOverNames = true
        useDefiniteLengthEncoding = true
    }

    @Test
    fun withNegativeTwoLabel() {
        val target = WithNegativeTwo(
            3,
            true
        )
        // a2010321f5
        /**
         * A2    # map(2)
         *    01 # unsigned(1)
         *    03 # unsigned(3)
         *    21 # negative(1)
         *    F5 # primitive(21)
         */
        val bytes = withLabel.encodeToByteArray(target)

        // failure - kotlinx.serialization.cbor.internal.CborDecodingException: Expected an unsigned or negative integer, but found F5
        val decodeBack: WithNegativeTwo = withLabel.decodeFromByteArray(bytes)
        assertEquals(target, decodeBack)
    }

    @Test
    fun withNegativeOneLabel() {
        val target = WithNegativeOne(
            3,
            true
        )
        //a2010320f5
        /**
         * A2    # map(2)
         *    01 # unsigned(1)
         *    03 # unsigned(3)
         *    20 # negative(0)
         *    F5 # primitive(21)
         */
        val bytes = withLabel.encodeToByteArray(target)

        // failure - kotlinx.serialization.cbor.internal.CborDecodingException: CborLabel unknown: 0 for kotlinx.serialization.cbor.NegativeLabelsOverNamesTest.WithNegativeOne(id: kotlin.Int, cool: kotlin.Boolean?)
        val decodeBack: WithNegativeOne = withLabel.decodeFromByteArray(bytes)
        assertEquals(target, decodeBack)
    }

    @Serializable
    data class WithNegativeTwo(
        @CborLabel(1)
        val id: Int,
        // critical bit: negative label causes failure to decode
        @CborLabel(-2)
        val cool: Boolean? = null,
    )

    @Serializable
    data class WithNegativeOne(
        @CborLabel(1)
        val id: Int,
        // critical bit: negative label causes failure to decode
        @CborLabel(-1)
        val cool: Boolean? = null,
    )
}

```

**Expected behavior**
Tests should pass.

**Fix**
The fix I applied locally to 1.11.0 is attached as a patch (one liner):

[kotlinx.serialization-3-41-26 PM.patch](https://github.com/user-attachments/files/27260803/kotlinx.serialization-3-41-26.PM.patch)

This does change that line from:
https://github.com/Kotlin/kotlinx.serialization/commit/e334d1c3927753e579f064aa53327585d995a0c0
Which was intended to make reading strictly follow the RFC - but since it only allowed positive numbers I just changed it to the generic 'readNumber' to handle both positive and negative.

**Environment**
 - Kotlin version: 2.3.20
 - Library version: 1.11.0
 - Kotlin platforms: All
 - Gradle version: 8.7

fzhinkin:
CborLabels are regular numbers and should be decoded as such.

Closes #3190

### DIFF
--- a/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
+++ b/formats/cbor/commonMain/src/kotlinx/serialization/cbor/internal/Decoder.kt
@@ -352,7 +352,7 @@ internal class CborParser(private val input: ByteArrayInput, private val verifyO
             readByte()
             return Triple(ans, null, collectedTags)
         } else {
-            val res = readUnsignedIntegerIgnoringMajorType { majorType.majorTypeName }
+            val res = readNumber()
             readByte()
             return Triple(null, res, collectedTags)
         }
--- a/formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborLabelTest.kt
+++ b/formats/cbor/commonTest/src/kotlinx/serialization/cbor/CborLabelTest.kt
@@ -1,7 +1,7 @@
 package kotlinx.serialization.cbor
 
 import kotlinx.serialization.*
-import kotlinx.serialization.cbor.internal.CborDecodingException
+import kotlinx.serialization.cbor.internal.*
 import kotlin.test.*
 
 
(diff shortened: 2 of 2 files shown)


fzhinkin:
@ardune, thanks for spotting the issues and providing both the fix and the test. I added you as a co-author, if you don't mind.

ardune:
Bug number mentioned might be wrong?

fzhinkin:
Indeed. Thanks!

sandwwraith (approved):
Issue number in commit message is also incorrect

PR #3199, opened 2026-05-18
Align collections' SerialDescriptors with the contract

### LINKED ISSUE #3163: MapLikeDescriptor.getElementIndex throws instead of returning UNKNOWN_NAME
**Describe the bug**
The documentation states that `SerialDescriptor.getElementIndex` returns the index or `UNKNOWN_NAME` when there is no such element. But the map like descriptor will throw if you pass a non integer like descriptor. 

**To Reproduce**
```kotlin
val serializer = MapSerializer(String.serializer(), String.serializer())
val descriptor = serializer.descriptor
descriptor.getElementIndex("key") // Throws
```

**Expected behavior**
There is no documentation that this function will ever throw and it would be more suitable if it returned `UNKNOWN_NAME`.

**Environment**
 - Kotlin version: 2.3.10
 - Library version: 1.10.0
 - Kotlin platforms: JVM

sandwwraith:
getElementIndex(name) should return UNKNOWN_NAME instead of throwing, whilst getElementName(index) should throw IOOBE on invalid indices.

Fixes #3163

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptor.kt
+++ b/core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptor.kt
@@ -68,7 +68,7 @@ import kotlinx.serialization.encoding.*
  * In practice, for regular classes it is allowed to invoke `getElement*(index)` methods
  * with an index from `0` to [elementsCount] range and the element at the particular index corresponds to the
  * serializable property at the given position.
- * For collections and maps, index parameter for `getElement*(index)` methods is effectively bounded
+ * For collections and maps, index parameter for `getElement*(index)` methods should be non-negative and is effectively bounded
  * by the maximal number of collection/map elements.
  *
  * ### Thread-safety and mutability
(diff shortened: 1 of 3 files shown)


core/commonMain/src/kotlinx/serialization/descriptors/SerialDescriptor.kt
```diff
@@ -319,6 +319,8 @@ public interface SerialDescriptor {
      * userDescriptor.getElementIndex("alias") // Returns 1
      * userDescriptor.getElementIndex("lastName") // Returns CompositeDecoder.UNKNOWN_NAME = -3
      * ```
+     *
+     * @throws IllegalStateException if the current descriptor does not support children elements (e.g. is a primitive)
```
fzhinkin, line 323: Or `UnsupportedOperationException`?

sandwwraith: [Current behavior](https://github.com/Kotlin/kotlinx.serialization/blob/d317f5dcf7f3f4b92b0b9f2ea72c3b4903371306/core/commonMain/src/kotlinx/serialization/internal/Primitives.kt#L40) has been around for quite some time; I simply documented it. I do not see much benefit in changing it for a slightly better exception type.

fzhinkin (approved):

PR #3203, opened 2026-06-01
Add a settings plugin configuring cache redirector for maven repos

### LINKED ISSUE #3202: Setup cache redirector for maven repositories
Follow up for https://github.com/Kotlin/kotlinx.serialization/pull/3188: we don't configure cache redirector for maven repositories and builds are failing in CI because of that.

We have to configure it the same way it was done for coroutines or AFU.

fzhinkin:
Inspired by the implementation in the kotlinx-atomicfu.

Closes #3202

### DIFF
--- a/build-settings-logic/build.gradle.kts
+++ b/build-settings-logic/build.gradle.kts
@@ -1,0 +1,10 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+plugins {
+    `kotlin-dsl`
+}
+
+kotlin {
+    compilerOptions.allWarningsAsErrors = true
+}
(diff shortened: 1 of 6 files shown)


fzhinkin:
Tested locally by running:
```
./gradlew check --info --refresh-dependencies 2>&1 | grep 'Cached resource' | grep -v  https://cache-redirector.jetbrains.com
```

shanshin:
Should we add some opt-out property for the redirection in case if our mirrors will be unavailable or to slow for some local builds?

fzhinkin:
@shanshin, previously, when I added cache redirector for Yarn & NPM, we decided to make it unconditional for now (https://github.com/Kotlin/kotlinx.serialization/pull/3188#issuecomment-4412653859).

shanshin (approved):

PR #3206, opened 2026-06-06
Fix ProtoBuf schema generation for value classes in collections

### LINKED ISSUE #3205: ProtoBuf schema generation does not unwrap value classes inside collections
**Describe the bug**

`ProtoBufSchemaGenerator` does not unwrap a `@JvmInline value class` when it is inside a collection. A single value-class property is unwrapped to the type it wraps, but the same value class inside a `List`, an `Array`, or as a `Map` value is emitted as a separate message instead.

**To Reproduce**

```kotlin
@JvmInline
@Serializable
value class WrappedUInt(val i: UInt)

@Serializable
class Data(val items: List<WrappedUInt>)

ProtoBufSchemaGenerator.generateSchemaText(Data.serializer().descriptor)
```

This generates:

```proto
message Data {
  repeated WrappedUInt items = 1;
}

message WrappedUInt {
  required int32 i = 1;
}
```

`Array<WrappedUInt>` and `Map<String, WrappedUInt>` (the map value) behave the same way.

**Expected behavior**

The value class is unwrapped to its underlying type, the same way a single value-class field already is:

```proto
message Data {
  repeated int32 items = 1;
}
```

**Environment**
 - Kotlin version: 2.3.20
 - Library version: 1.11.0 (also reproduces on `dev`)
 - Kotlin platforms: all (the schema generator is common code)
 - Gradle version: 8.7

saar:
The schema generator only unwrapped `@JvmInline` value classes for plain message fields. Inside a `List`, `Array`, or as a `Map` value the value class was left wrapped, so the schema used the wrapper type and added an extra message for it.

Unwrapping is now shared in a small `unwrapInlineType` helper and applied in `generateListType` and `generateMapType`. Packed repeated fields and map keys are unchanged.

Added `InlineClassesInCollections` to the schema generation test, covering a value class in a `List`, an `Array`, and a `Map` value.

Fixes #3205

### DIFF
--- a/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
+++ b/formats/protobuf/commonMain/src/kotlinx/serialization/protobuf/schema/ProtoBufSchemaGenerator.kt
@@ -270,10 +270,7 @@ public object ProtoBufSchemaGenerator {
         inOneOfStruct: Boolean = false,
         indent: Int = 1,
     ): List<TypeDefinition> {
-        var unwrappedFieldDescriptor = fieldDescriptor
-        while (unwrappedFieldDescriptor.isInline) {
-            unwrappedFieldDescriptor = unwrappedFieldDescriptor.getElementDescriptor(0)
-        }
+        val unwrappedFieldDescriptor = fieldDescriptor.unwrapInlineType()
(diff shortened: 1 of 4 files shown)


sandwwraith:
Fixed by #3209

PR #3208, opened 2026-06-17
Included G Play SDK verification files for published modules

fzhinkin:


JakeWharton:
Just FYI that we (square/cashapp maintainers) are going to push back against this as a mechanism. It is wholly unrealistic to think every library will include these, so we're not going to even try. It would be nice if you all joined the pushback!

sandwwraith (approved):

sandwwraith:
@JakeWharton, do you have any idea what channel for feedback we should use? I'm all in for making this process easier.

JakeWharton:
@ZacSweers has a contact. We were not approached directly.

ZacSweers:
Sent it over and told them to reach out

PR #3209, opened 2026-06-23
Fixed of Protobuf schema generation for inlined value classes

### LINKED ISSUE #3205: ProtoBuf schema generation does not unwrap value classes inside collections

shanshin:
Fixes #3205

### DIFF
--- a/formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/schema/InlinedClassesSchemaTest.kt
+++ b/formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/schema/InlinedClassesSchemaTest.kt
@@ -9,7 +9,7 @@ import kotlin.jvm.JvmInline
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
-class InlinedClassesTest {
+class InlinedClassesSchemaTest {
     @Serializable
     class SimpleClass
 
@@ -21,6 +21,14 @@ class InlinedClassesTest {
     @Serializable
     value class WrappedUInt(val i: UInt)
 
+
+    @JvmInline
+    @Serializable
+    value class DeepWrappedUInt(val wrapped: WrappedUInt)
+
+    @Serializable
+    class DeepWrappedUIntHolder(val x: DeepWrappedUInt)
+
     @Serializable
     class SingleHolder(val uint: List<WrappedUInt>, val simple: List<WrappedSimpleClass>)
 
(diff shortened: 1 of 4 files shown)


sandwwraith:
Does this replace #3206 ?

shanshin:
> Does this replace #3206 ?

Yes. I generalized the changes and made additional improvements to support keys, values, optionality, and nullability.

formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/schema/InlinedClassesSchemaTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class InlinedClassesTest {
+    @Serializable
+    class SimpleClass
+
+    @Serializable
+    @JvmInline
+    value class WrappedSimpleClass(val s: SimpleClass)
```
sandwwraith, line 18: Do we need to test wrappers over regular Int and wrapper over `WrappedSimpleClass`?

shanshin: You mean something like `value class SuperWrapper(val x: WrappedSimpleClass)`?

shanshin: For Int it's already tested via UInt

sandwwraith: Yes. I don't remember how we serialize wrappers over wrappers

shanshin: Add test on it. All nested inline types are flattened

sandwwraith (approved):

PR #3211, opened 2026-06-29
Disable redundant targets when building as part of the aggregate build

qwwdfsad:
kotlinx-serialization full builds runs on every commit to Kotlin repository and often is on a critical path of aggregate build completion, so cutting on redundant (obsolete, deprecated and/or nearly-duplicating) targets yields both wall time and total CI consumption reduction

#KTI-3388

### DIFF
--- a/buildSrc/src/main/kotlin/Projects.kt
+++ b/buildSrc/src/main/kotlin/Projects.kt
@@ -19,3 +19,15 @@ val Project.overriddenLanguageVersion : String?
 
 val Project.teamcityInteractionEnabled : Boolean
     get() = !hasProperty("no_teamcity_interaction") && !hasProperty("build_snapshot_up")
+
+// Property so cli-compiler-options can reference it
+const val KOTLIN_ADDITIONAL_CLI_OPTIONS_PROPERTY = "kotlin_additional_cli_options"
+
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/Projects.kt
```diff
@@ -19,3 +19,15 @@ val Project.overriddenLanguageVersion : String?
 
 val Project.teamcityInteractionEnabled : Boolean
     get() = !hasProperty("no_teamcity_interaction") && !hasProperty("build_snapshot_up")
+
+// Property so cli-compiler-options can reference it
+const val KOTLIN_ADDITIONAL_CLI_OPTIONS_PROPERTY = "kotlin_additional_cli_options"
```
sandwwraith, line 24: Are they always specified though? IIRC, they're provided only if the build is launched manually to verify some hypothesis (e.g., impact of enabling new LF). In an automated aggregate, it may be empty

qwwdfsad: I'll double-check. All builds I've ran were manual (though without any extra-options) [and there are no general permissions to look at build conf], thanks

sandwwraith:
closing for now

PR #3215, opened 2026-07-04
fix: partial descriptor behavior

### LINKED ISSUE #1512: ArrayIndexOutOfBoundsException when deserializing a custom object

inemtsev:
Fixes #1512

### Problem

`PluginGeneratedSerialDescriptor` instances created without a generated serializer throw
`ArrayIndexOutOfBoundsException` from `hashCode()`, `equals()` and `toString()`. This happens
for serializers produced by the `@Serializer(forClass = ...)` companion shortcut:

```kotlin
@Serializable(Model.Companion::class)
data class Model(val a: Int, val b: String) {
    @Serializer(forClass = Model::class)
    companion object
}
```

For these, the plugin constructs the descriptor with `generatedSerializer = null`, so
`childSerializers` is empty while **elementsCount > 0** and all three identity methods walk
**getElementDescriptor(i)**, which indexes into the empty array.

The bug is invisible on ordinary encode/decode (nothing calls `hashCode`), but it breaks any
feature that uses descriptors as cache keys in `DescriptorSchemaCache`. This is exactly the
crash reported in #1512 (open since 2021). The stack trace there is this path verbatim:
`DescriptorSchemaCache.get` -> `PluginGeneratedSerialDescriptor.hashCode` -> AIOOBE. The
originally reported trigger is `useAlternativeNames = true` (the documented limitation:
`JsonCustomSerializersTest` must run with `useAlternativeNames = false`), and a later report
in that thread hits the same crash via `coerceInputValues = true`. Because the root cause is
descriptor identity itself, every schema-cache entry point is affected the same way: in #3189
it forced a try-catch around the schema-cache lookup, and any future descriptor-keyed cache
would inherit the same landmine.

### Fix

Identity of such "incomplete" descriptors is now computed from data the descriptor itself
owns: serial name, type parameters, kind, and element names, instead of the child descriptors
it cannot provide:

- Complete descriptors are untouched: same hash values, same equality semantics as before.
- Incomplete descriptors hash/compare by serialName + typeParams + kind + element names.
- Complete vs incomplete never compare equal (previously threw), which is what keeps the
equals/hashCode contract intact with two different hash formulas. The kind check also
prevents accidental equality with EnumDescriptor, which reuses the null-serializer
construction but overrides hashCode itself.
- toString() for incomplete descriptors prints element names without child type names.

The new branch executes only inside the existing lazy _hashCode initializer and in equals,
so there is no hot-path cost (verified: TwitterFeedBenchmark decode/encode at parity with
master).

| Benchmark | master (ops/s) | this PR (ops/s) | Δ |
|---|---|---|---|
| `decodeTwitter` | 1591 ± 22 | 1596 ± 38 | +0.3%, intervals overlap (parity) |
| `encodeTwitter` | 3244 ± 12 | 3200 ± 54 | -1.4%, intervals overlap (parity) |

### DIFF
--- a/core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
+++ b/core/commonMain/src/kotlinx/serialization/internal/PluginGeneratedSerialDescriptor.kt
@@ -34,13 +34,36 @@ internal open class PluginGeneratedSerialDescriptor(
     // Cache child serializers, they are not cached by the implementation for nullable types
     private val childSerializers: Array<KSerializer<*>> by lazy(LazyThreadSafetyMode.PUBLICATION) { generatedSerializer?.childSerializers() ?: EMPTY_SERIALIZER_ARRAY }
 
+    // Descriptors created without a generated serializer (e.g. via the `@Serializer(forClass = ...)`
+    // companion shortcut) have no child serializers, so getElementDescriptor is unusable and
+    // identity operations (equals/hashCode/toString) must not rely on element descriptors.
+    private val isIncomplete: Boolean
+        get() = childSerializers.size < elementsCount
+
     // Lazy because of JS specific initialization order (#789)
     internal val typeParameterDescriptors: Array<SerialDescriptor> by lazy(LazyThreadSafetyMode.PUBLICATION) {
         generatedSerializer?.typeParametersSerializers()?.map { it.descriptor }.compactArray()
     }
 
     // Can be without synchronization but Native will likely break due to freezing
-    private val _hashCode: Int by lazy(LazyThreadSafetyMode.PUBLICATION) { hashCodeImpl(typeParameterDescriptors) }
+    private val _hashCode: Int by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        if (isIncomplete) incompleteHashCode() else hashCodeImpl(typeParameterDescriptors)
+    }
+
+    // Identity based on data that is always available: serial name, type parameters,
+    // kind and element names. Kept distinct from equals of complete descriptors
+    // (mixed comparisons are never equal), so the equals/hashCode contract holds.
+    private fun incompleteHashCode(): Int {
+        var result = serialName.hashCode()
+        result = 31 * result + typeParameterDescriptors.contentHashCode()
+        result = 31 * result + kind.hashCode()
+        var namesHash = 1
+        for (i in 0 until elementsCount) {
+            namesHash = 31 * namesHash + names[i].hashCode()
+        }
+        result = 31 * result + namesHash
+        return result
+    }
 
     public fun addElement(name: String, isOptional: Boolean = false) {
         names[++added] = name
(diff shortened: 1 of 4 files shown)


inemtsev:
Hi @sandwwraith @fzhinkin, please take a look at this improvement. 

It fixes a crash path (that was commented for @JsonNames annotation in the past), but also allows me to clean up try/catch in https://github.com/Kotlin/kotlinx.serialization/pull/3189 

Would be great to apply this fix first before cleaning up the above PR.

sandwwraith:
See #3084. I think there's no need to support this case for external serializers, since it won't be fixed correctly on the descriptor side in the plugin.

PR #3218, opened 2026-07-12
Speed up TreeJsonDecoder when no alternative names exist

### LINKED ISSUE #3193: TreeJsonDecoder with `useAlternativeNames=true` is slow even if no alternative names exist
**Describe the bug**

While working on https://github.com/ohs-foundation/kotlin-fhir/pull/83 I discovered using TreeJsonDecoder (which is used when polymorphic serialization fails the hot path) the performance of kotlinx serialization is extremely poor (roughly 2x to 4x slower than `useAlternativeNames=false`)

**To Reproduce**
```kotlin
@Serializable
data class WideClass(
  val f00: String? = null,
  val f01: String? = null,
  val f02: String? = null,
  val f03: String? = null,
  val f04: String? = null,
  val f05: String? = null,
  val f06: String? = null,
  val f07: String? = null,
  val f08: String? = null,
  val f09: String? = null,
  val f10: String? = null,
  val f11: String? = null,
  val f12: String? = null,
  val f13: String? = null,
  val f14: String? = null,
  val f15: String? = null,
  val f16: String? = null,
  val f17: String? = null,
  val f18: String? = null,
  val f19: String? = null,
  val f20: String? = null,
  val f21: String? = null,
  val f22: String? = null,
  val f23: String? = null,
  val f24: String? = null,
  val f25: String? = null,
  val f26: String? = null,
  val f27: String? = null,
  val f28: String? = null,
  val f29: String? = null,
)

fun main() {
  // 5 of 30 fields populated. JSON has 5 keys; descriptor has 30 slots; 25 absent → 25 slow-path
  // invocations per decode under default `useAlternativeNames = true`.
  val sample =
    """
    {"f00":"a","f01":"b","f02":"c","f03":"d","f04":"e"}
    """
      .trimIndent()

  val withAlternates = Json { useAlternativeNames = true }
  val withoutAlternates = Json { useAlternativeNames = false }

  val element = withAlternates.parseToJsonElement(sample)
  val iterations = 5_000_000

  // Warm both Json instances + JIT.
  repeat(50_000) {
    withAlternates.decodeFromJsonElement(WideClass.serializer(), element)
    withoutAlternates.decodeFromJsonElement(WideClass.serializer(), element)
  }

  val withTime = measureTime {
    repeat(iterations) {
      withAlternates.decodeFromJsonElement(WideClass.serializer(), element)
    }
  }
  val withoutTime = measureTime {
    repeat(iterations) {
      withoutAlternates.decodeFromJsonElement(WideClass.serializer(), element)
    }
  }

  println()
  println("════════════════════════════════════════════════════════")
  println(" JsonTreeDecoder slow-path reproducer")
  println("════════════════════════════════════════════════════════")
  println("  Class:              WideClass (30 nullable String fields, 0 @JsonNames)")
  println("  JSON input:         5 keys present, 25 slots absent")
  println("  Iterations:         $iterations")
  println()
  println("  useAlternativeNames=true  (default):  ${withTime} = ${withTime.inWholeNanoseconds / iterations} ns/op")
  println("  useAlternativeNames=false (skip slow): ${withoutTime} = ${withoutTime.inWholeNanoseconds / iterations} ns/op")
  println()
  val ratio = withTime.inWholeNanoseconds.toDouble() / withoutTime.inWholeNanoseconds.toDouble()
  println("  Slowdown factor:    %.2fx".format(ratio))
  println("════════════════════════════════════════════════════════")
}
```

**Expected behavior**

Serialization should be fast :)

**Actual behavior**

Serialization is slow! :(

On my M3 Max Macbook pro `useAlternativeNames = true ` is 3x slower than `useAlternativeNames = false`. Almost the entire time is spent doing O(n*m) lookups here:

https://github.com/Kotlin/kotlinx.serialization/blob/6956af2e6073347c7832c3c5b374fa3b5a345956/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt#L254

Even though the model declares no alternative names. While it is easy to set `useAlternativeNames=false`, the default in Kotlin is true... And I think this is something that can easilly be cached.

mwaddo:
## Problem

When you decode from a `JsonElement` and `useAlternativeNames` is on (the default), it's 2-4x slower and uses about 5x more memory than when it's off, even for classes that have no `@JsonNames` on them at all.

Two things cause it, and both only matter when a class actually has alternative names:
- `endStructure` builds a new set of allowed key names every time it checks for unknown keys.
- `elementName` searches the keys to find fields that aren't present under their own name.

## Fix

Check once per object whether the class has any alternative names. If it doesn't, just do what `useAlternativeNames = false` already does, use the field's own name in `elementName`, and reuse the set of names we already have in `endStructure` instead of building a new one. Classes that do use alternative names aren't affected.

One detail I want to mention: I read the cache directly here to avoid the `getOrPut` allocation in [deserializationNamesMap()](https://github.com/Kotlin/kotlinx.serialization/blob/25c2a75/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt#L53-L54) because it was the only thing keeping memory above the `false` number even though its a small difference. Happy to move that read into `deserializationNamesMap()` itself if you'd rather me fix it at the source.

## Testing

30 field class, 5 of the 30 fields present, decoded from a `JsonElement`:

|                                       | memory per decode |
|---------------------------------------|-------------------|
| useAlternativeNames = false           | 336 bytes         |
| useAlternativeNames = true (before)   | 1872 bytes         |
| useAlternativeNames = true (with fix) | 336 bytes         |

To run it yourself:

```
./gradlew jmhJar
java -jar benchmarks.jar TreeJsonNamesBenchmark -prof gc
```

Attached are before and after logs and profiling artifacts:

[flame-alloc-after.html](https://github.com/user-attachments/files/29932585/flame-alloc-after.html)
[flame-alloc-before.html](https://github.com/user-attachments/files/29932586/flame-alloc-before.html)
[jmh-after-fix.log](https://github.com/user-attachments/files/29932587/jmh-after-fix.log)
[jmh-before-fix.log](https://github.com/user-attachments/files/29932588/jmh-before-fix.log)

Fixes #3193

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TreeJsonNamesBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TreeJsonNamesBenchmark.kt
@@ -38,7 +38,7 @@ open class TreeJsonNamesBenchmark {
         val f27: String? = null, val f28: String? = null, val f29: String? = null
     )
 
-    private val default = Json                              // useAlternativeNames = true (default)
+    private val default = Json // useAlternativeNames = true (default)
     private val noAltNames = Json { useAlternativeNames = false }
 
     private val sparse = """{"f00":"a","f07":"b","f13":"c","f21":"d","f29":"e"}"""
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TreeJsonNamesBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/TreeJsonNamesBenchmark.kt
@@ -1,0 +1,53 @@
+package kotlinx.benchmarks.json
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.jsonObject
+import org.openjdk.jmh.annotations.Benchmark
+import org.openjdk.jmh.annotations.BenchmarkMode
+import org.openjdk.jmh.annotations.Fork
+import org.openjdk.jmh.annotations.Measurement
+import org.openjdk.jmh.annotations.Mode
+import org.openjdk.jmh.annotations.OutputTimeUnit
+import org.openjdk.jmh.annotations.Scope
+import org.openjdk.jmh.annotations.State
+import org.openjdk.jmh.annotations.Warmup
+import java.util.concurrent.TimeUnit
+
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+@Fork(2)
+open class TreeJsonNamesBenchmark {
+
+    @Serializable
+    data class Wide(
+        val f00: String? = null, val f01: String? = null, val f02: String? = null,
+        val f03: String? = null, val f04: String? = null, val f05: String? = null,
+        val f06: String? = null, val f07: String? = null, val f08: String? = null,
+        val f09: String? = null, val f10: String? = null, val f11: String? = null,
+        val f12: String? = null, val f13: String? = null, val f14: String? = null,
+        val f15: String? = null, val f16: String? = null, val f17: String? = null,
+        val f18: String? = null, val f19: String? = null, val f20: String? = null,
+        val f21: String? = null, val f22: String? = null, val f23: String? = null,
+        val f24: String? = null, val f25: String? = null, val f26: String? = null,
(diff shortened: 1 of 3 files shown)


qwwdfsad (commented):
I am not sure that pretending the unreasonable performance difference is not here is a directionally correct fix.

It _might_ be the case but it should be proven we cannot do any better. 
I'll postpone the PR until we figured all the root causes, addressed them and then let's see if the difference is still significant enough to do it

mwaddo:
Yeah good point... I was glossing over the actual issue rather than fixing the root cause. Thank you for the review and feedback.

I was in the process of putting out a revision and identified the two spots you had landed on. I'll this in favor of #3220, that's the better fix. Good rewrite, I'd be happy to review if helpful and thank you for fixing it.

qwwdfsad:
No worries!

For the sake of statistics -- would it be okay for you to disclose the amount of usage (if any!) of an LLM in an attempt to fix the issue?

Issue #1990, opened 2022-07-13
Allow failing on duplicate keys


tjerkw:
# WHY

The JSON format allows for duplicate keys in objects:

```json
{'a': 2, 'a': 4}
```

but the spec says it "SHOULD" not be done.

Different parsers implement it differently.
It would by nice to configure the Json parses to fail on duplicate keys

# HOW

Allow a setting on `Json` to fail on duplicate keys:

```
val json = kotlinx.serialization.json.Json {
        ignoreDuplicateKeys = false (default should be true
    }
``

**Comment by qwwdfsad** [comment]
Unless there are compelling use cases, we would prefer not to introduce such a feature: it's way too easy to misuse it and start ignoring backend errors or API misconfiguration.

A class-specific workaround would be to write your own serializer (or use JsonTransforming one) that deliberately skips duplicates

**Comment by Lysoun** [comment]
A use case of such a feature is to check strict JSON validity. Since the spec does not recommend using duplicate keys, it seems rather natural to allow the user to forbid such a thing. Especially when trying to check the equality between two JSON, the following JSON appear to be the same using kotlinx serializers but they are not
```json

{key: 'value1', key: 'value2'} 
{key: 'value2'}
```

I don't understand how it is so easy to misuse such a feature though. The error message should be clear and give the key that was duplicated. Since it would be an option that the user has to set if they want such a strict check, well, if they don't want that check they just have to not set the option, which would anyway be the default behaviour.

Are you still firmly against it or can I make a PR for this feature? I'm pretty sure it would be very easy to implement, making a little change in ```JsonTreeReader```.

qwwdfsad:
I honestly was sure that we do fail on duplicate keys and that the request was about **allowing** duplicate keys, misreading the sentence.

The request indeed is reasonable and seems like it should be **on** by default, though it requires a proper damage assessment.

Lysoun:
@qwwdfsad Hi, I've been working in this issue but I have trouble running the tests with the JVM. I get this error:
```> Task :kotlinx-serialization-core:compileKotlinJvm FAILED
e: Module java.base cannot be found in the module graph```
Guess it must be linked to my version of the JDK. I tried using 1.8, 11 and 17 (configured in my project structure in IntelliJ and in my preferences, for kotlin compilation target) but it always fails. It seems to be a problem with the project gradle configuration but I'm very surprised that the contributors did not see that.
[...]

PR #3220, opened 2026-07-13
Rewrite TreeJsonDecoder main decoding loop to be identical to StreamingJsonDecoder

qwwdfsad:
Previously, the decoding was SerialDescriptor-driven, not JsonObject keys-driven.
It made us reimplement the logic of StreamingJsonDecoder in an inverted manner (instead of mapping actual keys to indices, we mapped indices to keys), which led to all kinds of duplicated caches and reverse quadratic lookups.

Now the logic is almost identical to the streaming decoder, the redundancy in implementation is gone, and throughput for the tree decoder is, for a baseline, twice as good, and for alternative names up to 10 times better, making alternative names to be on par with the regular mode.

Implementation-wise, it is still tagged decoder under the hood (will be fixed separately, minor inconvenience, harder to review) and contains behavioural changes, see tests for that. The behavioural changes are now more in line with streaming behaviour.

### DIFF
--- a/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamesTest.kt
+++ b/formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamesTest.kt
@@ -87,6 +87,22 @@ class JsonNamesTest : JsonTestBase() {
     }
 
     @Test
+    fun testLastPropertyWinsOnAlternativeNames() {
+        val cases = mapOf(
+            """{"data":"first","foo":"second"}""" to "second", // primary then alternative
+            """{"foo":"first","data":"second"}""" to "second", // alternative then primary
+            """{"foo":"first","_foo":"second"}""" to "second", // two alternatives
+            """{"foo":"first","foo":"second"}""" to "second", // Two primary
+        )
+        for ((input, expected) in cases) {
+            parameterizedCoercingTest { json, streaming, _ ->
+                val data = json.decodeFromString(WithNames.serializer(), input, jsonTestingMode = streaming).data
+                assertEquals(expected, data, "Failed to parse input '$input' with streaming=$streaming")
+            }
+        }
+    }
+
+    @Test
     fun testThrowsAnErrorOnDuplicateNames() {
         val serializer = CollisionWithAlternate.serializer()
         parameterizedCoercingTest { json, streaming, _ ->
@@ -99,6 +115,25 @@ class JsonNamesTest : JsonTestBase() {
                     jsonTestingMode = streaming
                 )
             }
+        }
+    }
+
+    @Test
+    fun testMissingFieldExceptionOnDuplicateAlternativeNames() {
+        val serializer = CollisionWithAlternate.serializer()
+        for (input in listOf("{}", """{"data":"a"}""")) {
+            parameterizedCoercingTest { json, streaming, msg ->
+                assertFailsWithMessage<MissingFieldException>("required", "$msg, input=$input") {
+                    json.decodeFromString(serializer, input, jsonTestingMode = streaming)
+                }
+            }
+        }
+        parameterizedCoercingTest { json, streaming, msg ->
+            assertEquals(
+                CollisionWithAlternate("a", "b"),
+                json.decodeFromString(serializer, """{"data":"a","foo":"b"}""", jsonTestingMode = streaming),
+                msg
+            )
         }
     }
 }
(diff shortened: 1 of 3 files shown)


qwwdfsad:
Benchmarks: https://gist.github.com/qwwdfsad/8523d8ff8212878b8a4b9e71c3cefe94 (decided not to commit)

Before the change:
```
Benchmark                           Mode  Cnt  Score   Error   Units
TreeBenchmark.withActualAlternate  thrpt    5  0.668 ± 0.007  ops/us
TreeBenchmark.withAllAlternates    thrpt    5  0.519 ± 0.004  ops/us
TreeBenchmark.withAlternates       thrpt    5  1.171 ± 0.032  ops/us
TreeBenchmark.withoutAlternates    thrpt    5  4.745 ± 0.034  ops/us
```

After the change:
```
Benchmark                           Mode  Cnt  Score   Error   Units
TreeBenchmark.withActualAlternate  thrpt    3  7.887 ± 0.482  ops/us
TreeBenchmark.withAllAlternates    thrpt    3  6.582 ± 0.098  ops/us
TreeBenchmark.withAlternates       thrpt    3  9.127 ± 1.023  ops/us
TreeBenchmark.withoutAlternates    thrpt    3  9.080 ± 0.694  ops/us
```

TL;DR much faster, less code, one implementation instead of two, behaviour that replicates streaming decoder

qwwdfsad:
It would be even nicer to get rid of `NamedValueDecoder` and tag system. I've duct-taped (incorrect) zero-cost tagging, the perf win is around ~20% but it effectively disables error paths -- so implementation complexity is unclear; will leave it out of this PR

formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamesTest.kt
```diff
@@ -86,6 +86,22 @@ class JsonNamesTest : JsonTestBase() {
         }
     }
 
+    @Test
+    fun testLastPropertyWinsOnAlternativeNames() {
+        val cases = mapOf(
+            """{"data":"first","foo":"second"}""" to "second", // primary then alternative
```
sandwwraith, line 92: I'm a bit concerned about this change. Currently, `@SerialName`/property name has priority over the alternate one, and it is even [documented](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-names/). I hope no one actually relies on it, though. I'd rather prefer to try to restore the original behavior, but it looks like though it is not possible in this implementation unless we fix #1990 along the way

formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonDecoder.kt
```diff
@@ -197,122 +201,98 @@ private class JsonPrimitiveDecoder(
...
-                        descriptor, index,
-                        { currentElementOrNull(name) is JsonNull },
-                        { (currentElementOrNull(name) as? JsonPrimitive)?.contentOrNull },
-                        { // an unknown enum value should be coerced to null via decodeNotNullMark if explicitNulls=false :
-                            if (setForceNull(descriptor, index)) return index
-                        }
-                    )
-                ) continue // do not read coerced value
+    // Pointer to the current entry of JsonObject that is being decoded
+    private val entries: Iterator<Map.Entry<String, JsonElement>> by lazy(LazyThreadSafetyMode.NONE) {
```
sandwwraith, line 218: Is lazy really needed here? `decodeElementIndex` is always called when class is not empty, so it looks like we can strip one level of indirection. And if `value.isEmpty()` you can simply have null iterator or something like that

sandwwraith (commented):
btw, it would be nice to apply similar fix for the `DynamicInput`. Or try to completely unify it with JsonElement decoder

formats/json-tests/commonTest/src/kotlinx/serialization/features/JsonNamesTest.kt
Snipesy: This appears to be colliding with a whole different issue. Currently on all releases with `@SerialName` the behavior of `TreeJsonDecoder` versus `StreamingJsonDecoder` diverges.

```kotlin
@Serializable
data class Config(
    @SerialName("foo") val a: String? = null, // also works without SerialName
    @JsonNames("foo") val b: String? = null,
)

fun test() {
    val input = """{"foo":"X"}"""

    println(Json.decodeFromString<Config>(input))
    println(Json.decodeFromJsonElement<Config>(Json.parseToJsonElement(input)))
}
```
```
Config(a=X, b=null) // Streaming
Config(a=X, b=X) // Tree
```

On [qwwdfsad/tree-json-opto](https://github.com/Kotlin/kotlinx.serialization/tree/qwwdfsad/tree-json-opto) it inadvertently fixes the divergence and the 2 align.

🙃

sandwwraith:
@qwwdfsad I've fixed some commits, do you want me to merge it like this or you want to continue working?

qwwdfsad:
Sorry, I completely missed the GH notification. All good

PR #3221, opened 2026-07-14
Doc restructuring update cbor protobuf

daniCsorbaJB:
This PR adds the CBOR and ProtoBuf pages

### DIFF
--- a/docs-website/topics/serialization-cbor.md
+++ b/docs-website/topics/serialization-cbor.md
@@ -1,0 +1,287 @@
+[//]: # (title: CBOR format)
+<primary-label ref="experimental-general"/>
+
+Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc7049)) is a compact binary format based on JSON.
+It supports a subset of [JSON features](configure-json-serialization.md) and produces binary output instead of text.
+
+## Add dependencies for CBOR
+
+To use CBOR in your project, add the CBOR serialization library dependency to your build file:
+
+<tabs>
+<tab id="dependency-gradle" title="Gradle">
+
+```kotlin
+// build.gradle(.kts)
+
+dependencies {
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-cbor:%serializationVersion%")
+}
+```
+
+</tab>
+
+<tab id="dependency-maven" title="Maven">
+
+```xml
+<!-- pom.xml -->
+
+<dependencies>
+    <dependency>
+        <groupId>org.jetbrains.kotlinx</groupId>
+        <artifactId>kotlinx-serialization-cbor</artifactId>
+        <version>%serializationVersion%</version>
+    </dependency>
+</dependencies>
+```
+
+</tab>
+</tabs>
+
+## Use CBOR for binary serialization
+
+The [`Cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/) class provides two main functions:
+
+* [`encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-byte-array.html) serializes objects to a binary array.
+* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a binary array.
+
+Let's look at an example where a `Project` object is serialized into a binary array and then deserialized back to its original form:
+
+```kotlin
+// Imports declarations from the serialization library
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
+    // Shows printable ASCII bytes as characters and other bytes as hex values
+    if (it in 32..127) it.toInt().toChar().toString() else
+        "{${it.toUByte().toString(16).padStart(2, '0').uppercase()}}"
+}
+
+@Serializable
+data class Project(val name: String, val language: String)
+
+@OptIn(ExperimentalSerializationApi::class)
+fun main() {
+    val data = Project("kotlinx.serialization", "Kotlin")
+    
+    // Serializes the object to a CBOR binary array
+    val bytes = Cbor.encodeToByteArray(data)
+
+    // Converts the binary array to a human-readable hex string
+    println(bytes.toAsciiHexString())
+    // {BF}dnameukotlinx.serializationhlanguagefKotlin{FF}
+    
+    // Deserializes the binary array back to an object
+    val obj = Cbor.decodeFromByteArray<Project>(bytes)
+    println(obj)
+    // Project(name=kotlinx.serialization, language=Kotlin)
+}
+```
+
+<!-- REMOVE AFTER REVIEW: no runnable examples here, because playground doesn't support these formats (yet) -->
+
+This example prints the encoded bytes in a readable mixed form. It represents printable ASCII bytes as characters and non-printable bytes as hexadecimal values.
+
+The following table shows the same output in full [CBOR hex notation](http://cbor.me/):
+
+| Hex Code                                     | CBOR Type    | Description                        |
+|----------------------------------------------|--------------|------------------------------------|
+| BF                                           | map(*)       | Start of a CBOR map                |
+| 64                                           | text(4)      | Length of the string               |
+| 6E616D65                                     | string       | The string "name"                  |
+| 75                                           | text(21)     | Length of the string               |
+| 6B6F746C696E782E73657269616C697A6174696F6E   | string       | The string "kotlinx.serialization" |
+| 68                                           | text(8)      | Length of the string               |
+| 6C616E6775616765                             | string       | The string "language"              |
+| 66                                           | text(6)      | Length of the string               |
+| 4B6F746C696E                                 | string       | The string "Kotlin"                |
+| FF                                           | primitive(*) | End of the CBOR map                |
+
+<!-- REMOVE AFTER REVIEW: moving the note from the original further down where we first talk about map formats in depth. while this is a map it doesn't have non-trivial keys, and I feel like it's better to introduce this below -->
+
+## Ignore unknown keys in CBOR
+
+CBOR is commonly used in communication with [IoT](https://en.wikipedia.org/wiki/Internet_of_things) devices where new properties may be added as part of API evolution.
+By default, unknown keys encountered during deserialization result in an error.
+
+Just like in [JSON](serialization-json-configuration.md#ignore-unknown-keys), you set the [`ignoreUnknownKeys`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/ignore-unknown-keys.html) property to `true` to ignore them during deserialization:
+
+```kotlin
+// Imports declarations from the serialization library
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+@Serializable
+data class Project(val name: String)
+
+@OptIn(ExperimentalSerializationApi::class)
+fun main() {
+
+    // Creates a Cbor instance that ignores unknown keys during deserialization
+    val format = Cbor { ignoreUnknownKeys = true }
+
+    // Decodes the CBOR input and ignores the unknown "language" key
+    val data = format.decodeFromHexString<Project>(
+        // CBOR hex notation input with an extra, unknown "language" key
+        "bf646e616d65756b6f746c696e782e73657269616c697a6174696f6e686c616e6775616765664b6f746c696eff"
+    )
+    println(data)
+    // Project(name=kotlinx.serialization)
+}
+```
+
+In this CBOR input, the following bytes represent the unknown `"language"` key:
+
+* `68`: Length of the key `"language"`
+* `6c616e6775616765`: The key `"language"`
+* `66`: Length of the value `"Kotlin"`
+* `4b6f746c696e`: The value `"Kotlin"`
+
+## Customize how CBOR encodes data
+
+According to the [RFC 8949 Major Types](https://datatracker.ietf.org/doc/html/rfc8949#section-3.1) specification, CBOR supports the following data types:
+
+* Major type 0: an unsigned integer
+* Major type 1: a negative integer
+* Major type 2: a byte string
+* Major type 3: a text string
+* Major type 4: an array of data items
+* Major type 5: a map of pairs of data items
+* Major type 6: optional semantic tagging of other major types
+* Major type 7: floating-point numbers, simple data types with no content, and the "break" stop code
+
+By default, Kotlin `ByteArray` instances are encoded as major type 4, which represents an array of data items.
+
+To encode `ByteArray` instances as major type 2, a byte string, use the [`@ByteString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-byte-string/) annotation:
+
+```kotlin
+// Imports declarations from the serialization library
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
+    if (it in 32..127) it.toInt().toChar().toString() else
+        "{${it.toUByte().toString(16).padStart(2, '0').uppercase()}}"
+}
+
+@Serializable
+data class Data(
+    // Encodes the byte array as CBOR major type 2 as a byte string
+    @ByteString
+    val type2: ByteArray,
+    // Encodes the byte array as CBOR major type 4 as an array of individual data items
+    val type4: ByteArray
+)
+
+@OptIn(ExperimentalSerializationApi::class)
+fun main() {
+    // Creates a Data object with two ByteArray fields
+    val data = Data(byteArrayOf(1, 2, 3, 4), byteArrayOf(5, 6, 7, 8))
+
+    // Serializes the Data object into a CBOR byte array
+    val bytes = Cbor.encodeToByteArray(data)
+
+    println(bytes.toAsciiHexString())
+    // {BF}etype2D{01}{02}{03}{04}etype4{9F}{05}{06}{07}{08}{FF}{FF}
+    
+    val obj = Cbor.decodeFromByteArray<Data>(bytes)
+    println(obj)
+    // Data(type2=[1, 2, 3, 4], type4=[5, 6, 7, 8])
+}
+```
+
+<!-- REMOVE AFTER REVIEW: I felt like we have shown enough times how it looks in CBOR hex notation so I removed that part to keep it a bit more concise. -->
+
+In this example, the bytes before each `ByteArray` value differ because the properties use different CBOR major types.
+
+> Instead of annotating each property with `@ByteString`, you can also encode all `ByteArray` values as major type 2 by setting the [`alwaysUseByteString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/always-use-byte-string.html) property to `true`.
+>
+{style="note"}
+
+<!-- REMOVE AFTER REVIEW: I decided to move information from the original Arrays part here, to me it fits logically together with this part -->
+
+You can also customize how CBOR encodes entire classes.
+
+By default, classes are serialized as a CBOR map, which corresponds to major type 5.
+This means that each property of the class is stored as a key-value pair.
+
+You can serialize a class as a CBOR array, major type 4, with the [`@CborArray`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-array/) annotation.
+This can be useful for encoding COSE message structures, which [RFC 9052](https://www.rfc-editor.org/rfc/rfc9052#section-2) defines as CBOR arrays.
+
+Here's an example:
+
+```kotlin
+@Serializable
+@CborArray
+data class DataClass(
+    val alg: Int,
+    val kid: String?
+)
+ 
+Cbor.encodeToByteArray(DataClass(alg = -7, kid = null))
+```
+
+With the `@CborArray` annotation, this example is encoded as a CBOR array: `0x8226f6`.
+Without it, the same class is encoded as a CBOR map: `0xa263616c6726636b6964f6`.
+
+Unlike JSON, CBOR supports maps with non-trivial keys.
+Some parsers, such as [`jackson-dataformat-cbor`](https://github.com/FasterXML/jackson-dataformats-binary), don't support this feature.
+
+> For a JSON workaround, see [Allow structured map keys](serialization-json-configuration.md#allow-structured-map-keys).
+>
+{style="tip"}
+
+By using annotations like `@ByteString` and `@CborArray`, you can customize how CBOR encodes data to better match existing specifications and, in some cases, reduce binary size.
+
+## Definite and indefinite length encoding in CBOR
+
+CBOR supports two encodings for maps and arrays: *definite length encoding* and *indefinite length encoding*.
+
+By default, Kotlin serialization uses indefinite length encoding.
+This means that the number of elements in a map or array isn't encoded explicitly, and a terminating byte is appended after the last element.
+
+Definite length encoding omits the terminating byte and encodes the number of elements at the start of the map or array.
+
+To switch between these two modes, use the [`useDefiniteLengthEncoding`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/use-definite-length-encoding.html) property.
+
+## Tags and labels in CBOR
+
+CBOR allows you to define *tags* that encode additional information for properties and values.
+You can specify these tags with the [`@KeyTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-key-tags/) and [`@ValueTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-value-tags/) annotations.
+The `encodeKeyTags`, `encodeValueTags`, `verifyKeyTags`, and `verifyValueTags` properties control
+the encoding and verification of these tags.
+
+> For more information on tagging in CBOR, see [RFC 8949 Tagging of Items](https://datatracker.ietf.org/doc/html/rfc8949#name-tagging-of-items).
+>
+{style="tip"}
+
+You can also tag classes using the [`@ObjectTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-object-tags/) annotation, which applies tags to all instances of a class.
+
+When serializing, `@ObjectTags` are encoded directly before the data of the tagged object.
+If a property has value tags and its type has object tags, the value tags are encoded before the object tags.
+The `encodeObjectTags` and `verifyObjectTags` properties control whether object tags are encoded and verified.
+If you verify only value tags and don't verify object tags, the decoder can still deserialize data with additional object tags.
+
+> For a list of well-known tags, see [`CborTag`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-tag/).
+>
+{style="tip"}
+
+CBOR supports map keys of any type.
+In COSE (CBOR Object Signing and Encryption), these keys are restricted to strings and numbers and are called *labels*.
+
+You can assign string labels with the [`@SerialName`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serial-name/) annotation and numeric labels with the [`@CborLabel`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-label/) annotation.
+The [`preferCborLabelsOverNames`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-configuration/prefer-cbor-labels-over-names.html) property allows prioritizing numeric labels over serial names when both are present.
+You can use it to keep compact labels for CBOR while still keeping readable names when serializing to JSON.
+
+Kotlin serialization also provides a predefined [`Cbor.CoseCompliant`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/-default/-cose-compliant.html) instance that follows COSE encoding requirements.
+It uses definite length encoding, encodes and verifies all tags, and prefers numeric labels over serial names.
+
+## Custom CBOR-specific serializers
+
+CBOR encoders and decoders implement the [`CborEncoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-encoder/) and [`CborDecoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-decoder/) interfaces.
+
+These interfaces extend the general `Encoder` and `Decoder` APIs, providing access to CBOR-specific configurations through the [`cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-encoder/cbor.html) property.
+Custom serializers can use this property to access the current [`Cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/) instance, produce embedded byte arrays, and read the current settings, such as `preferCborLabelsOverNames` and `useDefiniteLengthEncoding`.
+
+For more information about creating custom serializers, see [Create custom serializers](create-custom-serializers.md).
--- a/docs-website/topics/serialization-cbor.md
+++ b/docs-website/topics/serialization-cbor.md
@@ -1,8 +1,7 @@
 [//]: # (title: CBOR format)
 <primary-label ref="experimental-general"/>
 
-Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc7049)) is a compact binary format based on JSON.
-It supports a subset of [JSON features](configure-json-serialization.md) and produces binary output instead of text.
+Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc8949)) is a compact binary format that extends the JSON data model.
 
 ## Add dependencies for CBOR
 
@@ -42,13 +41,12 @@ dependencies {
 
 The [`Cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/) class provides two main functions:
 
-* [`encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-byte-array.html) serializes objects to a binary array.
-* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a binary array.
+* [`encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-byte-array.html) serializes objects to a byte array.
+* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a byte array.
 
-Let's look at an example where a `Project` object is serialized into a binary array and then deserialized back to its original form:
+Let's look at an example where a `Project` object is serialized into a byte array and then deserialized back to its original form:
 
 ```kotlin
-// Imports declarations from the serialization library
 import kotlinx.serialization.*
 import kotlinx.serialization.cbor.*
 
@@ -65,40 +63,39 @@ data class Project(val name: String, val language: String)
 fun main() {
     val data = Project("kotlinx.serialization", "Kotlin")
     
-    // Serializes the object to a CBOR binary array
+    // Serializes the object to a CBOR byte array
     val bytes = Cbor.encodeToByteArray(data)
 
-    // Converts the binary array to a human-readable hex string
+    // Converts the byte array to a human-readable hex string
     println(bytes.toAsciiHexString())
     // {BF}dnameukotlinx.serializationhlanguagefKotlin{FF}
     
-    // Deserializes the binary array back to an object
+    // Deserializes the byte array back to an object
     val obj = Cbor.decodeFromByteArray<Project>(bytes)
     println(obj)
     // Project(name=kotlinx.serialization, language=Kotlin)
 }
 ```
 
-<!-- REMOVE AFTER REVIEW: no runnable examples here, because playground doesn't support these formats (yet) -->
-
 This example prints the encoded bytes in a readable mixed form. It represents printable ASCII bytes as characters and non-printable bytes as hexadecimal values.
 
-The following table shows the same output in full [CBOR hex notation](http://cbor.me/):
+Here's the same output in full [CBOR hex notation](http://cbor.me/):
 
-| Hex Code                                     | CBOR Type    | Description                        |
-|----------------------------------------------|--------------|------------------------------------|
-| BF                                           | map(*)       | Start of a CBOR map                |
-| 64                                           | text(4)      | Length of the string               |
-| 6E616D65                                     | string       | The string "name"                  |
-| 75                                           | text(21)     | Length of the string               |
-| 6B6F746C696E782E73657269616C697A6174696F6E   | string       | The string "kotlinx.serialization" |
-| 68                                           | text(8)      | Length of the string               |
-| 6C616E6775616765                             | string       | The string "language"              |
-| 66                                           | text(6)      | Length of the string               |
-| 4B6F746C696E                                 | string       | The string "Kotlin"                |
-| FF                                           | primitive(*) | End of the CBOR map                |
 
-<!-- REMOVE AFTER REVIEW: moving the note from the original further down where we first talk about map formats in depth. while this is a map it doesn't have non-trivial keys, and I feel like it's better to introduce this below -->
+```none
+Hex code                                             | CBOR type    | Description
+-----------------------------------------------------|--------------|------------------------------------
+BF                                                   | map(*)       | Start of a CBOR map
+├── 64                                               | text(4)      | Length of the string
+│   └── 6E616D65                                     | string       | The string "name"
+├── 75                                               | text(21)     | Length of the string
+│   └── 6B6F746C696E782E73657269616C697A6174696F6E   | string       | The string "kotlinx.serialization"
+├── 68                                               | text(8)      | Length of the string
+│   └── 6C616E6775616765                             | string       | The string "language"
+├── 66                                               | text(6)      | Length of the string
+│   └── 4B6F746C696E                                 | string       | The string "Kotlin"
+└── FF                                               | primitive(*) | End of the CBOR map
+```
 
 ## Ignore unknown keys in CBOR
 
@@ -108,7 +105,6 @@ By default, unknown keys encountered during deserialization result in an error.
 Just like in [JSON](serialization-json-configuration.md#ignore-unknown-keys), you set the [`ignoreUnknownKeys`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/ignore-unknown-keys.html) property to `true` to ignore them during deserialization:
 
 ```kotlin
-// Imports declarations from the serialization library
 import kotlinx.serialization.*
 import kotlinx.serialization.cbor.*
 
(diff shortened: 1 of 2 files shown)


docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
+[//]: # (title: CBOR format)
+<primary-label ref="experimental-general"/>
+
+Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc7049)) is a compact binary format based on JSON.
```
sandwwraith, line 4: https://datatracker.ietf.org/doc/html/rfc8949 is a newer version

sandwwraith, line 4: Official docs use "The underlying data model is an extended version of the JSON data model", so let's stick to it — use 'extends' json instead of 'based on' json

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+It supports a subset of [JSON features](configure-json-serialization.md) and produces binary output instead of text.
```
sandwwraith, line 5: This is kinda confusing because you've meant what configuration flags are available in kotlinx-serialization-cbor, and you're not talking about format itself.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+</tab>
+</tabs>
+
+## Use CBOR for binary serialization
+
+The [`Cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/) class provides two main functions:
+
+* [`encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-byte-array.html) serializes objects to a binary array.
```
sandwwraith, line 45: 'binary array' term is quite ambiguous, can we use 'byte array' instead?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a binary array.
+
+Let's look at an example where a `Project` object is serialized into a binary array and then deserialized back to its original form:
+
+```kotlin
+// Imports declarations from the serialization library
```
sandwwraith, line 51: we've discussed to drop this comment, didn't we?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+    // {BF}dnameukotlinx.serializationhlanguagefKotlin{FF}
+    
+    // Deserializes the binary array back to an object
+    val obj = Cbor.decodeFromByteArray<Project>(bytes)
+    println(obj)
+    // Project(name=kotlinx.serialization, language=Kotlin)
+}
+```
+
+<!-- REMOVE AFTER REVIEW: no runnable examples here, because playground doesn't support these formats (yet) -->
```
sandwwraith, line 82: Should we file an issue to website team to add them? This should be pretty trivial

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+
+This example prints the encoded bytes in a readable mixed form. It represents printable ASCII bytes as characters and non-printable bytes as hexadecimal values.
+
+The following table shows the same output in full [CBOR hex notation](http://cbor.me/):
```
sandwwraith, line 86: I'm not sure the table is better than original notation, as tabulation there showed how data was nested. Maybe we can add corresponding tabs/spaces to the beginning of each row or show nestedness otherwise somehow?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+
+@OptIn(ExperimentalSerializationApi::class)
+fun main() {
+
+    // Creates a Cbor instance that ignores unknown keys during deserialization
+    val format = Cbor { ignoreUnknownKeys = true }
+
+    // Decodes the CBOR input and ignores the unknown "language" key
+    val data = format.decodeFromHexString<Project>(
+        // CBOR hex notation input with an extra, unknown "language" key
```
sandwwraith, line 126: this comment essentially duplicates previous, merge them: ` Decodes the CBOR input with an extra, unknown "language" key`

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+        "bf646e616d65756b6f746c696e782e73657269616c697a6174696f6e686c616e6775616765664b6f746c696eff"
+    )
+    println(data)
+    // Project(name=kotlinx.serialization)
+}
+```
+
+In this CBOR input, the following bytes represent the unknown `"language"` key:
+
+* `68`: Length of the key `"language"`
```
sandwwraith, line 136: Again, it is quite hard to pinpoint what 68 you're talking about in the long string without proper hierarchy/nestedness

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+CBOR is commonly used in communication with [IoT](https://en.wikipedia.org/wiki/Internet_of_things) devices where new properties may be added as part of API evolution.
+By default, unknown keys encountered during deserialization result in an error.
+
+Just like in [JSON](serialization-json-configuration.md#ignore-unknown-keys), you set the [`ignoreUnknownKeys`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/ignore-unknown-keys.html) property to `true` to ignore them during deserialization:
+
+```kotlin
+// Imports declarations from the serialization library
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
```
sandwwraith, line 114: We can put //sampleStart here because we've already seen all of the imports

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+fun ByteArray.toAsciiHexString() = joinToString("") {
+    if (it in 32..127) it.toInt().toChar().toString() else
+        "{${it.toUByte().toString(16).padStart(2, '0').uppercase()}}"
+}
+
```
sandwwraith, line 167: Same here, there's no need to show `toAsciiHexString` again

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+@Serializable
+data class Data(
+    // Encodes the byte array as CBOR major type 2 as a byte string
+    @ByteString
+    val type2: ByteArray,
+    // Encodes the byte array as CBOR major type 4 as an array of individual data items
```
sandwwraith, line 173:
```suggestion
    // Encodes the byte array as CBOR major type 4: an array of individual data items
```

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+    println(bytes.toAsciiHexString())
+    // {BF}etype2D{01}{02}{03}{04}etype4{9F}{05}{06}{07}{08}{FF}{FF}
+    
+    val obj = Cbor.decodeFromByteArray<Data>(bytes)
+    println(obj)
+    // Data(type2=[1, 2, 3, 4], type4=[5, 6, 7, 8])
+}
+```
+
+<!-- REMOVE AFTER REVIEW: I felt like we have shown enough times how it looks in CBOR hex notation so I removed that part to keep it a bit more concise. -->
```
sandwwraith, line 194: At least for this example, it is crucial

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+
+In this example, the bytes before each `ByteArray` value differ because the properties use different CBOR major types.
+
+> Instead of annotating each property with `@ByteString`, you can also encode all `ByteArray` values as major type 2 by setting the [`alwaysUseByteString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/always-use-byte-string.html) property to `true`.
```
sandwwraith, line 198: I think it's better to not make it a note, as it is a rather important information. Maybe even add a 4th-level header to highlight it.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+>
+{style="note"}
+
+<!-- REMOVE AFTER REVIEW: I decided to move information from the original Arrays part here, to me it fits logically together with this part -->
```
sandwwraith, line 202: No, it is a separate section, as it is not related to byte strings and annotation is applicable to classes, not properties

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+    val kid: String?
+)
+ 
+Cbor.encodeToByteArray(DataClass(alg = -7, kid = null))
+```
+
+With the `@CborArray` annotation, this example is encoded as a CBOR array: `0x8226f6`.
+Without it, the same class is encoded as a CBOR map: `0xa263616c6726636b6964f6`.
+
+Unlike JSON, CBOR supports maps with non-trivial keys.
```
sandwwraith, line 228: Why is it placed here? How is it related to arrays and classes? If you couldn't find a better place for this note, you can bring it back to its original place. Or put it near major types.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+
+> For a JSON workaround, see [Allow structured map keys](serialization-json-configuration.md#allow-structured-map-keys).
+>
+{style="tip"}
+
+By using annotations like `@ByteString` and `@CborArray`, you can customize how CBOR encodes data to better match existing specifications and, in some cases, reduce binary size.
+
+## Definite and indefinite length encoding in CBOR
+
+CBOR supports two encodings for maps and arrays: *definite length encoding* and *indefinite length encoding*.
```
sandwwraith, line 239: You can put a relevant link to the RFC here

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,287 @@
...
+You can use it to keep compact labels for CBOR while still keeping readable names when serializing to JSON.
+
+Kotlin serialization also provides a predefined [`Cbor.CoseCompliant`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/-default/-cose-compliant.html) instance that follows COSE encoding requirements.
+It uses definite length encoding, encodes and verifies all tags, and prefers numeric labels over serial names.
+
+## Custom CBOR-specific serializers
+
+CBOR encoders and decoders implement the [`CborEncoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-encoder/) and [`CborDecoder`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-decoder/) interfaces.
+
+These interfaces extend the general `Encoder` and `Decoder` APIs, providing access to CBOR-specific configurations through the [`cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-encoder/cbor.html) property.
```
sandwwraith, line 284: APIs -> interfaces

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+</tab>
+</tabs>
+
+## Use ProtoBuf for binary serialization
+
+To serialize objects with ProtoBuf, use the
+[`ProtoBuf`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-buf/) class with the [`.encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-buf/encode-to-byte-array.html) and the [`.decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-buf/decode-from-byte-array.html) functions:
+
+```kotlin
+// Imports declarations from the serialization library
```
sandwwraith, line 51: Same here

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.protobuf.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
+    if (it in 32..127) it.toInt().toChar().toString() else
+        "{${it.toUByte().toString(16).padStart(2, '0').uppercase()}}"
+}
+
```
sandwwraith, line 108: //sampleStart from here

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+@OptIn(ExperimentalSerializationApi::class)
+fun main() {
+    val data = Data(1, -2, 3) 
+    println(ProtoBuf.encodeToByteArray(data).toAsciiHexString())
+    // {08}{01}{10}{03}{1D}{03}{00}{00}{00}
+}
+```
+
+<!-- REMOVE AFTER REVIEW: Like with CBOR examples, I felt like we have shown enough times how it looks in ProtoBuf hex notation so I removed that part to keep it a bit more concise. -->
```
sandwwraith, line 206: here it is acceptable imo

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+
+## Encode empty lists in ProtoBuf
```
sandwwraith, line 208: We fixed this some time ago (https://github.com/Kotlin/kotlinx.serialization/blob/8f82c6be67a6ceb9518a0557ea40bd7d82a20ab3/formats/protobuf/commonTest/src/kotlinx/serialization/protobuf/ProtobufAbsenceTest.kt#L75), so this chapter can be dropped, I think

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+
+    println(ProtoBuf.decodeFromByteArray(DataRepeated.serializer(), packed))
+    // DataRepeated(values=[1, 2, 3])
+
+    println(ProtoBuf.decodeFromByteArray(DataPacked.serializer(), repeated))
+    // DataPacked(values=[1, 2, 3])
+}
+```
+
+<!-- REMOVE AFTER REVIEW: I wanted to add an example here as well, please let me know if this works -->
```
sandwwraith, line 310: lgtm

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+        string work_phone = 3;
+    }
+}
+```
+
+To represent this message in Kotlin:
+
+1. Define a class for the entire message.
+2. Add a property of the polymorphic type to the class. Annotate this property with [`@ProtoOneOf`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-one-of/).
+3. Create a `sealed interface` or `abstract class` to represent the fields inside the `oneof` declaration.
```
sandwwraith, line 334: Why did you reorder 2 and 3? It is more confusing now. Also, it's better to stick to to the 'oneof group' rather than 'fields inside', because fields are represented as subclasses.

Also, why did you delete examples in-text? I think they're worth keeping. Even better, you can make the same structure you used in other docs:


3. Create a `sealed interface` or `abstract class` to represent the fields inside the `oneof` declaration.
```
@Serializable sealed interface IPhoneType
```

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+// Represents the oneof group
+@Serializable sealed interface IPhoneType
+
+// Represents the home_phone field
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable @JvmInline value class HomePhone(@ProtoNumber(2) val number: String): IPhoneType
+
+// Represents the work_phone field
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable data class WorkPhone(@ProtoNumber(3) val number: String): IPhoneType
```
sandwwraith, line 365: You lost explanation comment on why one of the classes is value and another is data. Why?

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,471 @@
...
+
+<!-- REMOVE AFTER REVIEW: I'm not sure I understand the original meaning, so I slightly rewritten this one based on my understanding. "....if an instance of Data2 has both (or none).. " If I understand correctly, if both are null simply nothing is set, if one is null other is set value it works, and if both are set there is a mismatched schema (if not true we can simply add if both are null to the above explanation) --> 
+
+## Generate a ProtoBuf schema
+
+Typically, working with ProtoBuf involves using a `.proto` file and a code generator to create code for serialization and deserialization.
+However, with Kotlin serialization, you can use Kotlin classes annotated with `@Serializable` as the source for the schema, making `.proto` files optional.
+
+This approach simplifies the process when all the code involved is written in Kotlin,
+interoperability with other languages often still requires a `.proto` schema.
```
sandwwraith, line 424:
```suggestion
but interoperability with other languages often still requires a `.proto` schema.
```

sandwwraith (changes requested):

sandwwraith:
Why base for this PR is `doc-restructuring-update-10` and not `doc-restructuring-update-master`???

docs-website/topics/serialization-cbor.md
pdvrieze: Maybe phrase it as "...format extending the JSON data model". As it is the underlying data model that is extensible, not the actual wire format.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: Unfortunately, it's not possible to use sampleStart on non-runnable code examples 😢 but if it's enabled until then we can add them 👍

docs-website/topics/serialization-cbor.md
daniCsorbaJB: 🤔 I think we should go with this then (to avoid the gerund):

> Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc8949)) is a compact binary format that extends the JSON data model.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: Ofc — nice, I like it 👍

docs-website/topics/serialization-cbor.md
daniCsorbaJB: I've messaged them asking them to add it

docs-website/topics/serialization-cbor.md
pdvrieze: Also, is it really a subset or just a set of options that have the same name (and hopefully the same semantics)?

daniCsorbaJB: good point 🤔 yes, perhaps we could just delete this sentence. It feels confusing like this. — we cover `ignoreUnknownKeys` below so I don't think this is needed after all.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: 🤔 we could nest it in the list and include the full CBOR hex notation below for reference as a collapsible element.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: I moved it outside — perhaps we could add a minimal example as well? 🤔 (based on how important it is) if not that important, we can just move it out of a note.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: 🤔 yea good point — we could separate these two then: 

```
## Customize CBOR data encoding

### Encode `ByteArray` properties as byte strings
### Encode classes as CBOR arrays
```

docs-website/topics/serialization-cbor.md
daniCsorbaJB: let's put it near major types 👍

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: nice! 👍 Yes, I agree, let's remove it.

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: the order is just to represent the order of the code example. 🤔 but yes, rereading, I think we should switch it.

> make the same structure you used in other docs

Sure, I wanted to keep it more concise, but the same structure works as well 👍

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: nice catch! It's an important piece of info, so I re-added it to the step list to have this information, but in the example, it feels arbitrary to me.

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,463 @@
...
+    data class Data(
+        @ProtoNumber(1)
+        val name: String,
+    
+        @ProtoOneOf
+        val phone: IPhoneType?,
+    )
+    ```
+
+3. Create a subclass for each field in the `oneof` declaration. Each subclass can be a value class or a data class and must have a single property for that field:
```
sandwwraith: It can also be a regular class. So more 'You can use a regular class, a data class or a value class for that. Each subclass must have a single property for that field:'

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,463 @@
...
+
+This is useful if you plan to reorder properties, and it aligns with Protobuf's compatibility rules for evolving schemas.
+
+Here's an example:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.protobuf.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
```
sandwwraith: Even if we can't make //sampleStart, let's still remove `toAsciiHexString()` from all examples except the very first one. It's just clutter at this point.

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,463 @@
...
+Field #3: 1A String Length = 6, Hex = 06, UTF8 = "Kotlin"
+```
+
+> For more information about Protobuf field numbers, see the [Official Protobuf Language Guide](https://protobuf.dev/programming-guides/proto2/#assigning).
+>
+{style="tip"}
+
+## Specify integer encoding in ProtoBuf
+
+ProtoBuf encodes integer properties using varint encoding by default.
```
sandwwraith: let's put https://protobuf.dev/programming-guides/encoding/#varints to varint. Otherwise, it would sound very cryptic

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,322 @@
...
+You can customize how CBOR encodes data to better match existing specifications and, in some cases, reduce binary size.
+
+By default, Kotlin `ByteArray` values are encoded as major type 4, which represents an array of data items.
+To encode `ByteArray` properties as major type 2, a byte string, use the [`@ByteString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-byte-string/) annotation:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
```
sandwwraith: Same here: let's drop `toAsciiHexString` anyway

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,322 @@
...
+    println(obj)
+    // Data(type2=[1, 2, 3, 4], type4=[5, 6, 7, 8])
+}
+```
+
+In this example, the bytes before each `ByteArray` value differ because the properties use different CBOR major types.
+
+Here's the encoded byte array in full CBOR hex notation:
+
+```none
```
sandwwraith, line 217: you can make this one collapsible as well

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,322 @@
...
+* Major type 0: an unsigned integer
+* Major type 1: a negative integer
+* Major type 2: a byte string
+* Major type 3: a text string
+* Major type 4: an array of data items
+* Major type 5: a map of pairs of data items
+* Major type 6: optional semantic tagging of other major types
+* Major type 7: floating-point numbers, simple data types with no content, and the "break" stop code
+
+Unlike JSON, CBOR supports maps with non-trivial keys.
```
sandwwraith: I think we should expand what 'trivial' means here. It means something that cannot be represented as a string (a primitive or a enum). Can we use just 'non-primitive' here or is there something better?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,322 @@
...
+    val kid: String?
+)
+ 
+Cbor.encodeToByteArray(DataClass(alg = -7, kid = null))
+```
+
+With the `@CborArray` annotation, this example is encoded as a CBOR array: `0x8226f6`.
+Without it, the same class is encoded as a CBOR map: `0xa263616c6726636b6964f6`.
+
+## Definite and indefinite length encoding in CBOR
```
sandwwraith: IMO this should be 3rd level `###` to get under ''Customize CBOR data encoding"

sandwwraith (changes requested):

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: sure thing, agreed 👍 we can just readd them when //sampleStart works

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: good idea— added it

docs-website/topics/serialization-cbor.md
daniCsorbaJB: yes, works quite well that way I agree 👍 🎉

docs-website/topics/serialization-cbor.md
daniCsorbaJB: 🤔 I suggest we align it with the related section, WDYT:

> Unlike JSON, CBOR supports maps with structured map keys, such as instances of user-defined classes.
However, some parsers, such as [`jackson-dataformat-cbor`](https://github.com/FasterXML/jackson-dataformats-binary), don't support them.

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: Perhaps we could also combine these two sentences, WDYT: 

> For each field in the `oneof` declaration, create a subclass with a single property that corresponds to that field. Each subclass can be a regular class, data class, or value class:

sandwwraith (approved):
Great!!

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+    </dependency>
+</dependencies>
+```
+
+</tab>
+</tabs>
+
+## Use CBOR for binary serialization
+
+The [`Cbor`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/) class provides two main functions:
```
koshachy: I opened the link and there are only two functions. No other functions exist in the class.
Do we need to name them **main** functions?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+
+* [`encodeToByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/encode-to-byte-array.html) serializes objects to a byte array.
+* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a byte array.
+
+Let's look at an example where a `Project` object is serialized into a byte array and then deserialized back to its original form:
```
koshachy: Let's avoid **Let's**:

Here's an example where ...

The following example serializes a Project object into a byte array and then deserializes it back to its original form:

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+ 
+Cbor.encodeToByteArray(DataClass(alg = -7, kid = null))
+```
+
+With the `@CborArray` annotation, this example is encoded as a CBOR array: `0x8226f6`.
+Without it, the same class is encoded as a CBOR map: `0xa263616c6726636b6964f6`.
+
+### Definite and indefinite length encoding in CBOR
+
+CBOR supports [two encodings](https://datatracker.ietf.org/doc/html/rfc8949#section-3.2.2) for maps and arrays: *definite length encoding* and *indefinite length encoding*.
```
koshachy:
```suggestion
CBOR supports [two encodings](https://datatracker.ietf.org/doc/html/rfc8949#section-3.2.2) for maps and arrays: _definite length encoding_ and _indefinite length encoding_.
```

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
+[//]: # (title: CBOR format)
+<primary-label ref="experimental-general"/>
+
+Concise Binary Object Representation ([CBOR](https://datatracker.ietf.org/doc/html/rfc8949)) is a compact binary format that extends the JSON data model.
+
+## Add dependencies for CBOR
+
+To use CBOR in your project, add the CBOR serialization library dependency to your build file:
+
+<tabs>
```
koshachy: Why don't you use standard

```markdown
<tabs group="build-script">
<tab title="Kotlin" group-key="kotlin">
```

So the switcher will work across the whole site?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+* [`decodeFromByteArray()`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/decode-from-byte-array.html) deserializes objects from a byte array.
+
+Let's look at an example where a `Project` object is serialized into a byte array and then deserialized back to its original form:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+fun ByteArray.toAsciiHexString() = joinToString("") {
+    // Shows printable ASCII bytes as characters and other bytes as hex values
```
koshachy: If I understand this code correctly, it doesn't shows anything.

BTW, ASCII `127` value is non-printable

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+├── 68                                               | text(8)      | Length of the string
+│   └── 6C616E6775616765                             | string       | The string "language"
+├── 66                                               | text(6)      | Length of the string
+│   └── 4B6F746C696E                                 | string       | The string "Kotlin"
+└── FF                                               | primitive(*) | End of the CBOR map
+```
+
+## Ignore unknown keys in CBOR
+
+CBOR is commonly used in communication with [IoT](https://en.wikipedia.org/wiki/Internet_of_things) devices where new properties may be added as part of API evolution.
```
koshachy: ... Internet of Things (IoT) devices ...

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+Unlike JSON, CBOR supports maps with structured map keys, such as instances of user-defined classes.
+However, some parsers, such as [`jackson-dataformat-cbor`](https://github.com/FasterXML/jackson-dataformats-binary), don't support them.
+
+> For a JSON workaround, see [Allow structured map keys](serialization-json-configuration.md#allow-structured-map-keys).
+>
+{style="tip"}
+
+### Encode `ByteArray` properties as byte strings
+
+You can customize how CBOR encodes data to better match existing specifications and, in some cases, reduce binary size.
```
koshachy: If these two things are not heavy related, we could have two sentences then.
Smth like:

You can customize how CBOR encodes data to better match existing specifications. In some cases it can reduce binary size.

koshachy: Or...

You can customize CBOR encoding to match existing specifications. ....

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+By default, Kotlin `ByteArray` values are encoded as major type 4, which represents an array of data items.
+To encode `ByteArray` properties as major type 2, a byte string, use the [`@ByteString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-byte-string/) annotation:
+
+```kotlin
+import kotlinx.serialization.*
+import kotlinx.serialization.cbor.*
+
+@Serializable
+data class Data(
+    // Encodes the byte array as CBOR major type 2 as a byte string
```
koshachy: // Encodes the byte array as a CBOR byte string (major type 2).

WDYT? Is it easier to read?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+    alwaysUseByteString = true
+}
+```
+
+### Encode classes as CBOR arrays
+
+By default, classes are serialized as a CBOR map, which corresponds to major type 5.
+This means that each property of the class is stored as a key-value pair.
+
+You can serialize a class as a CBOR array, major type 4, with the [`@CborArray`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-array/) annotation.
```
koshachy:
```suggestion
You can serialize a class as a CBOR array (major type 4), with the [`@CborArray`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-array/) annotation.
```

koshachy: I think this information is kinda additional, so we don't need to interrupt the flow with commas.
WDYT?

docs-website/topics/serialization-cbor.md
koshachy: We don't use single asterisks.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+With the `@CborArray` annotation, this example is encoded as a CBOR array: `0x8226f6`.
+Without it, the same class is encoded as a CBOR map: `0xa263616c6726636b6964f6`.
+
+### Definite and indefinite length encoding in CBOR
+
+CBOR supports [two encodings](https://datatracker.ietf.org/doc/html/rfc8949#section-3.2.2) for maps and arrays: *definite length encoding* and *indefinite length encoding*.
+
+By default, Kotlin serialization uses indefinite length encoding.
```
koshachy: Shouldn't it be hyphenated? indefinite-length?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+This means that the number of elements in a map or array isn't encoded explicitly, and a terminating byte is appended after the last element.
```
koshachy:
```suggestion
The number of elements in a map or array isn't encoded explicitly, and a terminating byte is appended after the last element.
```

koshachy: Could be two sentences?
encoded explicitly. Instead, a terminating (break?) byte marks the end of the collection.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+
+Definite length encoding omits the terminating byte and encodes the number of elements at the start of the map or array.
```
koshachy: Definitel-length?

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+
+To switch between these two modes, use the [`useDefiniteLengthEncoding`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor-builder/use-definite-length-encoding.html) property.
+
+## Tags and labels in CBOR
+
+CBOR allows you to define *tags* that encode additional information for properties and values.
```
koshachy: https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit?tab=t.0#heading=h.b70s84q084fn

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,318 @@
...
+You can specify these tags with the [`@KeyTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-key-tags/) and [`@ValueTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-value-tags/) annotations.
```
koshachy, line 283: WDYT, is it obvious what is Key tag and what is Value? 🤔 (I think it is, but want to hear your opinion)

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,449 @@
+[//]: # (title: ProtoBuf format)
+<primary-label ref="experimental-general"/>
+
+[Protocol Buffers](https://developers.google.com/protocol-buffers), or ProtoBuf, is a language-neutral binary format that normally
```
koshachy: What if we add ProtoBuf in parentheses?

koshachy: normally == typically?

BTW, Is this an imprtant nuance to mention?

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,449 @@
...
+```
+
+## Assign field numbers for ProtoBuf serialization
+
+By default, ProtoBuf assigns field numbers automatically.
+
+To keep your schema stable over time, use the [`@ProtoNumber`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/-proto-number/) annotation to assign field numbers explicitly, without needing a separate `.proto` file.
+For example, `@ProtoNumber(1)` assigns field number 1 to a property, so the Kotlin serialization's ProtoBuf format uses that number during encoding and decoding instead of assigning one automatically.
+
+This is useful if you plan to reorder properties, and it aligns with Protobuf's compatibility rules for evolving schemas.
```
koshachy: If you plan to reorder properties, it helps to align ....

Does this work?

docs-website/topics/serialization-protobuf.md
```diff
@@ -0,0 +1,449 @@
...
+    println(ProtoBuf.decodeFromByteArray(DataPacked.serializer(), repeated))
+    // DataPacked(values=[1, 2, 3])
+}
+```
+
+## Represent `oneof` fields
+
+A [`oneof`](https://protobuf.dev/programming-guides/proto2/#oneof) field defines a group of fields where only one value can be set at a time.
+
+In Kotlin serialization, you can represent this structure with a [polymorphic](serialization-polymorphism.md) type.
```
koshachy, line 252: I think it's a question for a separate discussion.
Sometimes we say "Kotlin serialization" and sometimes "`kotlinx.serialization`".
I think these two things should be somehow aligned.

koshachy (changes requested):

docs-website/topics/serialization-cbor.md
daniCsorbaJB: Good question—yes I think we can easily drop it 👍 — there are several functions extended by it, but since those are not provided it should be good

docs-website/topics/serialization-cbor.md
daniCsorbaJB: corrected everywhere 👍

docs-website/topics/serialization-cbor.md
daniCsorbaJB: 👍 Let's rewrite it as, WDYT: 

>     // Converts bytes in the printable ASCII range to characters
>     // and other bytes to hex values
>     if (it in 32..126) it.toInt().toChar().toString() else

(we can modify the 127 to 126 then good point)

docs-website/topics/serialization-cbor.md
daniCsorbaJB: I like both. I'd modify the second sentence slightly in this case though just to clarify what "it" refers to 🤔 WDYT:

> You can customize CBOR encoding to match existing specifications. In some cases, these customizations can also reduce binary size.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: WFM! 👍

docs-website/topics/serialization-cbor.md
daniCsorbaJB: yes, works for me—I'll drop the comma before the "with" then 👍 so it can be one uninterrupted sentence

docs-website/topics/serialization-cbor.md
pdvrieze: The phrase "better match existing specifications" is somewhat vague to me. The word "specification" could be referring to CBOR, or to specific serial forms (externally defined) of the specific data to be serialized (i.e. match an external schema).

The separate part is that what you are doing is customizing the (de)serialized form instead of using the automatic default (that cannot take data semantics into account).

docs-website/topics/serialization-cbor.md
daniCsorbaJB: Good question 🤔 Yes, I think it should be.

The API and some other Kotlin serialization resources use it without hyphens, as do some other official CBOR resources in the same context, so it could be treated as a technical term. However, since the linked specification hyphenates it, I think that settles it... and it also works better as a compound modifier.

Updated all occurrences—cheers for raising it 🚀

docs-website/topics/serialization-cbor.md
daniCsorbaJB: sure, WFM! 👍 

Actually, we could even connect it more closely with the previous explanation above and expand on it:

> The number of elements in a map or array isn't encoded explicitly. Instead, a "break" stop code (`0xFF`) marks the end of a collection.

pdvrieze: Rather than "The number" use "In this encoding, the number" to start the sentence, that makes it clear that you are referring to the "indefinite length encoding".

You might also consider replacing "terminating byte" by "terminator", "terminator token", or "end of sequence marker", as the byte nature of the terminator is mostly implementation detail here.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: yes, I think it should be obvious 🤔

If the user is reading about tags and labels, they should already have the basic concepts of CBOR. If they do (or if they've read the previous section about encoding), they're already aware that, by default, class properties are encoded as key-value pairs, and the annotation names make the distinction clear.

That said, it'd be a very easy update to make it explicit, although I think it reads a bit less naturally and becomes more mechanical:

> You can tag a property's key with `@KeyTags` and its value with `@ValueTags`.

That said, I actually prefer the more readable, less mechanical approach of the original here. 🤔

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: yes, I think "typically" is better here. 
For the nuance itself, I think it helps in fact it would help to link to it imho. It's an important distinction that in Kotlin you don't need it in certain circumstances which is further expanded upon in one of the later sections:

> [Protocol Buffers](https://developers.google.com/protocol-buffers) (ProtoBuf) is a language-neutral binary format that typically
uses a separate [`.proto` file](#generate-a-protobuf-schema) to define the protocol schema.

docs-website/topics/serialization-protobuf.md
daniCsorbaJB: sure 👍 
Although I feel like if we start with "If you plan to reorder properties..." we should make the connection explciit:

> If you plan to reorder properties, assigning field numbers explicitly keeps the schema stable and aligns with Protobuf's compatibility rules for evolving schemas.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: I think this is a good clarification—thank you 🙏 

I modified as:

> You can customize the CBOR representation of your data instead of using the default encoding, for example, to match a serialized form defined by an external schema. In some cases, these customizations can also reduce binary size.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: "In this encoding, the number...." works well! 👍 

I think we should stick with the term used by the RFC— "break" stop code. That should be easy to cross-reference.

docs-website/topics/serialization-cbor.md
```diff
@@ -0,0 +1,319 @@
...
+CBOR allows you to define _tags_ that encode additional information for properties and values.
+You can specify these tags with the [`@KeyTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-key-tags/) and [`@ValueTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-value-tags/) annotations.
+The `encodeKeyTags`, `encodeValueTags`, `verifyKeyTags`, and `verifyValueTags` properties control
+the encoding and verification of these tags.
+
+> For more information on tagging in CBOR, see [RFC 8949 Tagging of Items](https://datatracker.ietf.org/doc/html/rfc8949#name-tagging-of-items).
+>
+{style="tip"}
+
+You can also tag classes using the [`@ObjectTags`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-object-tags/) annotation, which applies tags to all instances of a class.
```
koshachy, line 291: let's use another verb than "tag"

koshachy: annotate with ...

koshachy: Through the doc there are different other verbs: "use annotation", "apply annotation".

I think it is better using only one or two.

koshachy (approved):
LGTM.
Pay attention to the last comments.

docs-website/topics/serialization-cbor.md
daniCsorbaJB: Nice catch, it does feel ambiguous. 🤔 Here, "tag" refers to assigning a CBOR tag to the encoded data, while the declaration itself is annotated with `@ObjectTags`. I’ve changed the wording to "assign tags" with the `@ObjectTags` annotation to make the distinction clearer.

PR #3225, opened 2026-07-17
Improve JSON decoding performance for Okio and kotlinx-io for ASCII 

qwwdfsad:


qwwdfsad:
Performance improvements.

Before:
```
Benchmark                          Mode  Cnt     Score    Error   Units
MoshiBaseline.kotlinFromKxSource  thrpt    5   313.422 ±  8.733  ops/ms
MoshiBaseline.kotlinFromSource    thrpt    5   312.250 ±  9.838  ops/ms
MoshiBaseline.kotlinFromString    thrpt    5  1416.439 ± 29.257  ops/ms
```

After:
```
Benchmark                          Mode  Cnt     Score   Error   Units
MoshiBaseline.kotlinFromKxSource  thrpt    5  1112.266 ± 3.409  ops/ms
MoshiBaseline.kotlinFromSource    thrpt    5  1074.611 ± 6.153  ops/ms
MoshiBaseline.kotlinFromString    thrpt    5  1405.612 ± 9.831  ops/ms
```

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaselineNonAscii.kt
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.benchmarks.json.moshi
+
+import benchmarks.model.DefaultPixelEvent
+import benchmarks.model.pixelEvent
+
+open class MoshiBaselineNonAscii : MoshiBaseline() {
```
qwwdfsad, line 10: Score at this benchmark is still terrible compared to any reasonable baseline.
I have ideas what to do, but let's integrate this one first

qwwdfsad:
(force-pushed rebase onto `dev` to make https://github.com/Kotlin/kotlinx.serialization/pull/3219 pass)

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
```diff
@@ -34,7 +35,60 @@ internal class JsonToIoStreamWriter(private val sink: Sink) : InternalJsonWriter
...
+private const val HIGH_SURROGATE_HEADER = 0xd800 - (0x010000 ushr 10)
+// Value added to the low UTF-16 surrogate after masking
+private const val LOW_SURROGATE_HEADER = 0xdc00
+
+@OptIn(UnsafeIoApi::class, InternalIoApi::class)
+internal class IoSerialReader(private val source: Source) : InternalJsonReader {
+    // When the last (count'th) byte is a high surrogate, we save it here and merge with the low one on the next read()
+    private var bufferedChar: Char? = null
+
+    final override fun read(buffer: CharArray, bufferOffset: Int, count: Int): Int {
```
pdvrieze, line 49: Reading into `CharArray` expensive for a number of reason:
- you will end up storing the meaningful content either not at all (non-string content), or as utf8 (strings use this internally)
- a lot of file content is not "stored" at all: whitespace, lexical separators: ( `{`, `}`, `:`, `[`, `]`, ':`,…)
- it is valid to do byte comparison on lexical separators due to UTF8 semantics.
- the amount of underlying bytes to be read for a given amount of char elements is content dependent.

pdvrieze (commented):
I have done some of this work for XMLUtil (it has significant performance and allocation impact). The answer is that for UTF8 sources you should not work in UTF16, but for other encodings (or strings) you still want to. To provide this you want an abstraction that can do the important operations independent of the underlying format (compare with codepoint/ascii character) and allow for extracting text ranges given position markers.

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
```diff
@@ -34,7 +35,60 @@ internal class JsonToIoStreamWriter(private val sink: Sink) : InternalJsonWriter
...
+        bufferedChar?.let {
+            buffer[bufferOffset] = it
+            bufferedChar = null
+            written++
+        }
+
+        while (written < count) {
+            val remaining = count - written
+            // Nothing left, bail out
+            if (!source.request(1)) break
```
fzhinkin: You can use `source.exhausted()` instead: it does the same under the hood, but looks a bit more readable

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
```diff
@@ -34,7 +35,60 @@ internal class JsonToIoStreamWriter(private val sink: Sink) : InternalJsonWriter
...
-    override fun nextCodePoint(): Int = source.readCodePointValue()
+// Value added to the high UTF-16 surrogate after shifting
+private const val HIGH_SURROGATE_HEADER = 0xd800 - (0x010000 ushr 10)
+// Value added to the low UTF-16 surrogate after masking
+private const val LOW_SURROGATE_HEADER = 0xdc00
+
+@OptIn(UnsafeIoApi::class, InternalIoApi::class)
+internal class IoSerialReader(private val source: Source) : InternalJsonReader {
+    // When the last (count'th) byte is a high surrogate, we save it here and merge with the low one on the next read()
+    private var bufferedChar: Char? = null
```
fzhinkin: Since it's either a high surrogate, or nothing, replacing `null` value with `'\0'` can save us from a few allocations.

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/internal/OkioJsonStreams.kt
```diff
@@ -33,8 +33,63 @@ internal class JsonToOkioStreamWriter(private val sink: BufferedSink) : Internal
     }
 }
 
-internal class OkioSerialReader(private val source: BufferedSource): InternalJsonReaderCodePointImpl() {
-    override fun exhausted(): Boolean = source.exhausted()
-    override fun nextCodePoint(): Int = source.readUtf8CodePoint()
-}
+// Value added to the high UTF-16 surrogate after shifting
+private const val HIGH_SURROGATE_HEADER = 0xd800 - (0x010000 ushr 10)
```
fzhinkin, line 37: It would be nice to avoid duplicating the code responsible for surrogate pair splitting, and implement it once where the `InternalJsonReaderCodePointImpl` was.

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaselineNonAscii.kt
fzhinkin: Was it better/worse with the old implementation? Just to know what should we use as a baseline.

fzhinkin (changes requested):

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
pdvrieze: Alternatively `'\uxFFFF'` can be used as it is actually a "noncharacter" in the standard (better than 0).

benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/moshi/MoshiBaselineNonAscii.kt
qwwdfsad: It was the same.
```

Benchmark                                  Mode  Cnt     Score    Error   Units
MoshiBaselineNonAscii.kotlinFromKxSource  thrpt    5   221.977 ± 29.670  ops/ms
MoshiBaselineNonAscii.kotlinFromSource    thrpt    5   142.154 ±  4.999  ops/ms
MoshiBaselineNonAscii.kotlinFromString    thrpt    5  1500.673 ± 69.438  ops/ms
MoshiBaselineNonAscii.kotlinToOkio        thrpt    5   520.307 ±  2.308  ops/ms
MoshiBaselineNonAscii.kotlinToString      thrpt    5  3099.969 ± 71.102  ops/ms
MoshiBaselineNonAscii.moshiFromSource     thrpt    5   435.260 ± 13.457  ops/ms
MoshiBaselineNonAscii.moshiFromString     thrpt    5   290.644 ± 40.892  ops/ms
MoshiBaselineNonAscii.moshiToOkio         thrpt    5   437.311 ± 34.897  ops/ms
MoshiBaselineNonAscii.moshiToString       thrpt    5   320.161 ±  8.902  ops/ms
```

(results on top of the current branch for the rough idea)

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
qwwdfsad: Thanks! All valid and more or less aligned with my findings, but beyond the scope of this PR.
My goal is to merge the local improvement, so in the next release, there will be reasonable performance for people using streaming integrations.
Meanwhile, I'll be working on a "proper" solution (trie + better lexel abstractions, basically) that is not busy with moving bytes from one encoding to the other via a char array

formats/json-okio/commonMain/src/kotlinx/serialization/json/okio/internal/OkioJsonStreams.kt
qwwdfsad: I wanted to do so originally, but it implies `internal` + `JsonFriendModuleApi` + abi dump, and taking these are two obvious constants that I'll get rid of in the next iteration (trie), the current option seemed more meningful to me

fzhinkin (approved):

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
pdvrieze: Just trying to share what I learned doing it (mostly) for XML. It makes sense to not do it all at once.

PR #3228, opened 2026-07-20
Rename some Json implementation classes

qwwdfsad:
While working on the trie prototype (and on #3225/#2743), I've generally noticed that it is really hard to follow the current implementation and switch between various types of lexers/readers. 

Its naming is not self-consistent, is partially contradictory, and is slightly confusing. It especially stands out after some time without contributing to the project but still having an understanding of all concepts.

I've tried to keep things self-consistent and conventional, without touching any ABI. One commit -- one self-contained change

### DIFF
--- a/formats/json-io/commonMain/src/kotlinx/serialization/json/io/IoStreams.kt
+++ b/formats/json-io/commonMain/src/kotlinx/serialization/json/io/IoStreams.kt
@@ -8,7 +8,7 @@ import kotlinx.serialization.*
 import kotlinx.serialization.json.DecodeSequenceMode
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.internal.*
-import kotlinx.serialization.json.io.internal.JsonToIoStreamWriter
+import kotlinx.serialization.json.io.internal.KxIoJsonWriter
 import kotlinx.serialization.json.internal.decodeToSequenceByReader
 import kotlinx.serialization.json.io.internal.KxIoReader
 import kotlinx.io.*
@@ -25,7 +25,7 @@ public fun <T> Json.encodeToSink(
     value: T,
     sink: Sink
 ) {
-    val writer = JsonToIoStreamWriter(sink)
+    val writer = KxIoJsonWriter(sink)
     try {
         encodeByWriter(this, writer, serializer, value)
     } finally {
--- a/formats/json-io/commonMain/src/kotlinx/serialization/json/io/IoStreams.kt
+++ b/formats/json-io/commonMain/src/kotlinx/serialization/json/io/IoStreams.kt
@@ -10,7 +10,7 @@ import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.internal.*
 import kotlinx.serialization.json.io.internal.JsonToIoStreamWriter
 import kotlinx.serialization.json.internal.decodeToSequenceByReader
-import kotlinx.serialization.json.io.internal.IoSerialReader
+import kotlinx.serialization.json.io.internal.KxIoReader
 import kotlinx.io.*
 
 /**
@@ -60,7 +60,7 @@ public fun <T> Json.decodeFromSource(
     deserializer: DeserializationStrategy<T>,
     source: Source
 ): T {
-    return decodeByReader(this, deserializer, IoSerialReader(source))
+    return decodeByReader(this, deserializer, KxIoReader(source))
 }
 
 /**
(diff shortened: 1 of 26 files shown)


formats/json-io/commonMain/src/kotlinx/serialization/json/io/IoStreams.kt
```diff
@@ -25,7 +25,7 @@ public fun <T> Json.encodeToSink(
     value: T,
     sink: Sink
 ) {
-    val writer = JsonToIoStreamWriter(sink)
+    val writer = KxIoJsonWriter(sink)
```
qwwdfsad, line 28: I realize that `KxIo` and `Okio` are not technically types, but `Sink` and `BufferedSink` is going to be really confusing

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonMode.kt
```diff
@@ -11,20 +11,20 @@ import kotlinx.serialization.json.*
 import kotlinx.serialization.modules.*
 import kotlin.jvm.*
 
-internal enum class WriteMode(@JvmField val begin: Char, @JvmField val end: Char) {
+internal enum class JsonMode(@JvmField val begin: Char, @JvmField val end: Char) {
```
fzhinkin: I'm not sure if it's just "Json mode" either. `JsonLexerMode`?

But then there's `Json.selectMapMode`.

qwwdfsad: Just `LexerMode` maybe?

sandwwraith: Tbh it's more like `JsonTokens`/`LexerTokens` because it simply contains a pair of chars. But LexerMode is also fine

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
```diff
@@ -41,7 +41,7 @@ private const val HIGH_SURROGATE_HEADER = 0xd800 - (0x010000 ushr 10)
 private const val LOW_SURROGATE_HEADER = 0xdc00
 
 @OptIn(UnsafeIoApi::class, InternalIoApi::class)
-internal class IoSerialReader(private val source: Source) : InternalJsonReader {
+internal class KxIoReader(private val source: Source) : InternalJsonReader {
```
sandwwraith, line 44: IMO it should be `OkioJsonReader/KxIoJsonReader` for symmetry with `...JsonWriter`

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonMode.kt
fzhinkin: It's a mode in a sense it tells lexer when to stop (that's how I sold `LexerMode` name myself) :)

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
qwwdfsad: I wanted to do the opposite (`Json` substring is completely redundant in all of the implementation classes and is just extra-verbose).
But I cannot easily remove `Json` because we'll end up with things like `StringWriter`, ` OutputStreamWriter` that clash with JDK classes. So I've left them for writers but not readers :)

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonMode.kt
qwwdfsad: Done!

fzhinkin (approved):

sandwwraith (approved):

qwwdfsad:
I'll squash the last two commits and do a rebase merge, easier for the history

formats/json-io/commonMain/src/kotlinx/serialization/json/io/internal/IoJsonStreams.kt
pdvrieze: As there are multiple formats in the project it may be helpful to have the format name in the class. It makes working with the overall project a little easier.
In addition, while reading the changes I cannot help noticing that I am wondering whether `KxIoReader` is a format independent adaptor that allows for kotlinx.io sources to be processed by a format, or that it is specifically a `Json` parser.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt
```diff
@@ -41,8 +41,8 @@ public interface InternalJsonReader {
 public fun <T> encodeByWriter(json: Json, writer: InternalJsonWriter, serializer: SerializationStrategy<T>, value: T) {
     val encoder = StreamingJsonEncoder(
         writer, json,
-        WriteMode.OBJ,
-        arrayOfNulls(WriteMode.entries.size)
+        LexerMode.OBJ,
```
pdvrieze, line 44: I would think that both names are misleading. `WriteMode` implies that this is for writing (which doesn't work for parsing). Lexer implies it is for reading. Something neutral might be better (but I don't have a good at this point)

PR #3229, opened 2026-07-21
KT-87873: suppress warning on android native target family

mMaxy:


buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
```diff
@@ -34,19 +34,24 @@ kotlin {
     mingwX64()
     iosX64()
     watchosDeviceArm64()
+
+    // Deprecated
```
sandwwraith, line 38: It would be better to mention the ticket number and the deprecation timeline (when it will be an error and when it will be removed). And the same for Apple x64 targets.

sandwwraith (changes requested):

buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
mMaxy: Done.

sandwwraith (approved):

sandwwraith:
Will you cherry-pick to `kotlin-community/dev` yourself?

mMaxy:
> Will you cherry-pick to kotlin-community/dev yourself?

Yes, no problem :) 
https://github.com/Kotlin/kotlinx.serialization/pull/3231

PR #3230, opened 2026-07-21
Update Kotlin to 2.4

sandwwraith:


core/api/kotlinx-serialization-core.api
```diff
@@ -600,7 +600,7 @@ public final class kotlinx/serialization/internal/ArrayListSerializer : kotlinx/
 }
 
 public final class kotlinx/serialization/internal/BooleanArrayBuilder : kotlinx/serialization/internal/PrimitiveArrayBuilder {
-	public synthetic fun build$kotlinx_serialization_core ()Ljava/lang/Object;
+	public synthetic fun build$org_jetbrains_kotlinx_kotlinx_serialization_core ()Ljava/lang/Object;
```
shanshin, line 603: Interesting change in the name mangling

shanshin (approved):

core/api/kotlinx-serialization-core.api
sandwwraith: @udalov do you know why it was changed?

sandwwraith: I think it's because it is some kind of synthetic bridge, not an published api function

udalov: It's one of these:
* [KT-69701](https://youtrack.jetbrains.com/issue/KT-69701) Gradle: module name is passed inconsistently to different types of compilations
* [KT-82216](https://youtrack.jetbrains.com/issue/KT-82216) Sanitize '.kotlin_module' filename
* [KT-85412](https://youtrack.jetbrains.com/issue/KT-85412) Module name is not sanitized with older Kotlin compiler versions

PR #3234, opened 2026-08-03
IR inliner: Don't use CLI flag that enforces "intra-module" inliner mode

ddolovov:
Previously, "intra-module" inliner has been explicitly enabled in kotlinx.serialization runtime libraries when they were built with 2.3.x compiler: a5a3c97a590571de3f6692c21d00de95aa16226c.

In 2.4.x the "intra-module" inliner is enabled by default, and in 2.5.0 the full ("cross-module") inliner is enabled by default.

We need to drop all explicit usages of `-Xklib-ir-inliner` flag from build scripts to not stick with the "intra-module" mode forever.

^KT-88022

### DIFF
--- a/buildSrc/src/main/kotlin/CompilerOptions.kt
+++ b/buildSrc/src/main/kotlin/CompilerOptions.kt
@@ -14,7 +14,6 @@ val defaultCompilerArgs
         "-Xreport-all-warnings",
         "-Xrender-internal-diagnostic-names",
         "-Xreturn-value-checker=full",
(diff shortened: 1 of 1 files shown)


ddolovov:
@sandwwraith There is a failure in `:buildSrc:compileKotlin` compilation: `buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts:27:5 Unresolved reference: watchosArm32`. I am not sure it's related to my changes. Is that a problem for merging this PR?

PR #3235, opened 2026-08-11
Exceptions catching

sandwwraith:
Add exception handling to encode/decodeSerializableValue in Json codecs


formats/json/api/kotlinx-serialization-json.api
```diff
@@ -281,7 +281,7 @@ public final class kotlinx/serialization/json/JsonEncodingException : kotlinx/se
 }
 
 public abstract class kotlinx/serialization/json/JsonException : kotlinx/serialization/SerializationException {
-	public synthetic fun <init> (Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
```
fzhinkin, line 284: It seems like we need an overload for backward compatibility

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonExceptions.kt
```diff
@@ -8,13 +8,81 @@ package kotlinx.serialization.json.internal
...
+@Suppress("DEPRECATION_ERROR")
+@OptIn(ExperimentalSerializationApi::class)
+internal inline fun JsonDecoder.errorFromDeserializer(
+    cause: Throwable,
+    path: String?,
+    input: () -> CharSequence
+): JsonDecodingException {
+    val causeMessage = cause.message
+    val shortMessage =
+        "Deserializer caused " + (if (causeMessage == null) "an exception" else "'$causeMessage' exception") + " in the decoder"
```
fzhinkin: `threw` or `raised` instead of `caused`?

fzhinkin (commented):
Some minor nitpicking from me, but overall the change looks good

formats/json/api/kotlinx-serialization-json.api
sandwwraith: We don't, JsonException is a sealed class, there are no external inheritors

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonExceptions.kt
sandwwraith: I think `caused` is better, but `raised` is also possible

shanshin (approved):
/final

sandwwraith:
@fzhinkin please re-check exception message

fzhinkin (approved):

sandwwraith:
hm, there is no autosquash fixup option on github.....

PR #3236, opened 2026-08-12
Optimize ordered JSON key decoding

IRus:
## Summary

- avoid materializing JSON object keys and performing a hash lookup when string input follows descriptor order
- retain the existing lookup path for out-of-order, escaped, unknown, lenient, naming-strategy, and streamed input
- restore the outer sequential index when an object decoder is reused for a nested object
- add regression coverage and a reversed-order JMH control

## Motivation

Object decoding currently allocates a `String` for every field name and then looks it up in the descriptor name map. Profiling showed key parsing and lookup as a significant part of JSON decoding time and allocation.

The new optimistic path compares the next descriptor name directly with the source. A mismatch immediately falls back to the existing behavior, so accepted inputs and public APIs are unchanged.

## Performance

Project JMH, clean `master` versus this change:

| Benchmark | Throughput | Allocation |
|---|---:|---:|
| Five-field generic object | +50.1% | -19.9% |
| One-field integer object | +115.2% | -54.5% |
| Twitter feed | +19.3% | -52.4% |

An external `java-json-benchmark` A/B run on JDK 25 and the dedicated Claws host compared published 1.11.0 with the local candidate across 1/10/100/1000 KB payloads:

- Users deserialization: **+20.37%** geometric mean
- Clients deserialization: **+7.00%** geometric mean
- all deserialization cases: **+13.49%** geometric mean
- serialization control: **+0.52%** geometric mean, with mixed signs and overlapping confidence intervals

The external run used 15 threads, two forks, a fixed seed, and alternated baseline/candidate execution order per workload.

## Compatibility

The regular lookup remains authoritative whenever the optimistic comparison cannot be used or does not match. Tests cover hash-colliding names, different-length and reversed fields, escaped names, and recursive reuse of the same descriptor.

There are no public API changes.

## Validation

- `./gradlew :kotlinx-serialization-json-tests:jvmTest`
- `./gradlew :kotlinx-serialization-json-tests:jsNodeTest`
- `./gradlew :kotlinx-serialization-json-tests:macosArm64Test`
- `./gradlew apiCheck`
- `./gradlew :benchmark:jmhJar`
- full external Claws A/B benchmark matrix

### DIFF
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/LookupOverheadBenchmark.kt
@@ -33,6 +33,7 @@ open class LookupOverheadBenchmark {
     private val data = """{"a":""}"""
     private val doubleData = """{"a":"","b":0}"""
     private val pentaData = """{"a":"","b":0,"c":1,"d":true,"e":" "}"""
+    private val reversedPentaData = """{"e":" ","d":true,"c":1,"b":0,"a":""}"""
 
     @Serializable
     object Object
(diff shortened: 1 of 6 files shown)


qwwdfsad (commented):
That's a really neat approach, thanks!

I am not sure yet we should accept that though: we have a trie-based prefix matching in our mid-term plans, which has a similar effect (no allocations, single pass over a string), does not overoptimize over a specific (and easy to benchmark) shape and does not incur two code paths instead of one. It will be slightly less efficient though (albeit it should be ~single digit/low teens diff).

I'd suggest waiting until the prefix matching so we can make a weighted decision on the final solution

PR #3238, opened 2026-08-13
Deprecate ChunkedDecoder

sandwwraith:


shanshin:
Should we write at leas at the commit message reason why we deprecate this API and what the alternatives?

sandwwraith:
It was discussed on DM. There are no alternatives. There aren't any usages of it either, so it shouldn't cause any problems

shanshin:
So I suggest to add a comment on what you have concluded during your DM discussions.

shanshin (approved):
/final

fzhinkin (approved):
Fully agree with Sergey, it's worth explicitly mentioning the deprecation reason in the commit message.

PR #3241, opened 2026-08-17
Suppress LV 2.0 deprecation

fzhinkin:
Suppress LV 2.0's deprecation with error.

Required for https://github.com/JetBrains/kotlin/pull/7411


fzhinkin:
@bnorm, FYI

sandwwraith:
Can't we just drop this `if`? IIRC it was needed for Aggregate build with 1.9, but it has been dropped for some time?

fzhinkin:
@sandwwraith, sure, that makes sense. I removed both sections handling older Kotlin or KGP versions.

sandwwraith (approved):

PR #3245, opened 2026-08-31
Workaround watchosArm32 target removal

fzhinkin:
Added a stub function to invoke instead of the removed `watchosArm32()`.

### DIFF
--- a/buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
@@ -1,4 +1,5 @@
 import org.jetbrains.kotlin.gradle.*
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
 import org.jetbrains.kotlin.gradle.plugin.mpp.*
 
 /*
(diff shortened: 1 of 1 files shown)


fzhinkin:
Note that this time the build is failed because of the issues that should be fixed by https://github.com/Kotlin/kotlinx.serialization/pull/3244

buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
```diff
@@ -9,6 +10,12 @@ plugins {
     kotlin("multiplatform")
 }
 
+// Temporary workaround for the removed watchosArm32 target
+@Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+fun KotlinMultiplatformExtension.watchosArm32() {
```
shanshin, line 15: Perhaps we should use a similar migration in KGP, where calling this method will print a warning

shanshin (approved):

sandwwraith:
Does this mean we can fix https://youtrack.jetbrains.com/issue/KTI-3541 ?

fzhinkin:
@sandwwraith, yes, it does.

sandwwraith (approved):

buildSrc/src/main/kotlin/native-targets-conventions.gradle.kts
fzhinkin: Shared it in https://youtrack.jetbrains.com/issue/KT-87513

PR #3246, opened 2026-09-01
Add Json nesting depth calculation during parsing 

sandwwraith:
+ Set Deprecated(HIDDEN) to setters in JsonConfiguration
+ Increase stack size for mingwx64 test binaries


sandwwraith:
#3029 can be merged after this

formats/json-tests/commonTest/src/kotlinx/serialization/NestingJsonTest.kt
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.serialization
+
+import kotlinx.serialization.*
```
fzhinkin:
```suggestion
```

formats/json-tests/commonTest/src/kotlinx/serialization/NestingJsonTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.serialization.json.JsonTestBase
+import kotlinx.serialization.json.JsonTestingMode
+import kotlinx.serialization.test.checkDecodingException
+import kotlin.test.Test
+
+class NestingJsonTest: JsonTestBase() {
```
fzhinkin, line 16: Please also add tests verifying behavior one level before hitting the limit and just hitting it (I mean, 499-depth object should be parsed, 500-depth object should be rejected).

And it would be nice to check that the limit could be actually adjusted by the corresponding property.

fzhinkin (changes requested):

formats/json-tests/commonTest/src/kotlinx/serialization/NestingJsonTest.kt
sandwwraith: > And it would be nice to check that the limit could be actually adjusted by the corresponding property.

This is verified [here](https://github.com/Kotlin/kotlinx.serialization/pull/3246/changes#diff-f0fc442f1356a7c486c42cc3361e0ddc904a43f0fddf42246cf2db3aa485a7e3R50)

sandwwraith:
Apparently, #3235 added some stack variables, so now it's SoE with a 500 limit... gonna lower it more.

formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonTreeReader.kt
```diff
@@ -110,9 +119,27 @@ internal class JsonTreeReader(
...
+            TC_BEGIN_LIST -> withDepthControl { readArray() }
             else -> lexer.fail("Can't begin reading element, unexpected token")
         }
     }.invoke(Unit)
+
+    private inline fun withDepthControl(read: () -> JsonElement): JsonElement {
+        if (++stackDepth > maxNestingDepth) {
+            jsonTooNested(maxNestingDepth)
+        }
+        val result = read()
```
shanshin, line 135: Am I right that if an exception is thrown in `read()`, we’ll go back to the very beginning of the serialization and won’t access `stackDepth` again?

sandwwraith: Yes, any exception from lexer/parser should completely stop deserialization and unwind all stack

formats/json-tests/commonTest/src/kotlinx/serialization/NestingJsonTest.kt
shanshin: I also think we should check the boundary condition, like  `maxNestingDepth = 2` and JSON with depth 2 should be parsed successfully

shanshin (approved):

fzhinkin (approved):

PR #3247, opened 2026-09-01
Avoid specifying -Xjdk-release twice

fzhinkin:
With recent changes on KGP side (KT-88381) it results in a warning, which fails our builds.

### DIFF
--- a/buildSrc/src/main/kotlin/Java9Modularity.kt
+++ b/buildSrc/src/main/kotlin/Java9Modularity.kt
@@ -143,8 +143,14 @@ object Java9Modularity {
             multiPlatformEnabled.set(compileTask.get().multiPlatformEnabled)
             compilerOptions {
                 jvmTarget.set(JvmTarget.JVM_9)
-                freeCompilerArgs.addAll(
-                    listOf("-Xjdk-release=9",  "-Xsuppress-version-warnings", "-Xexpect-actual-classes")
+                freeCompilerArgs.set(
+                    compileTask.flatMap { it.compilerOptions.freeCompilerArgs }.map { args ->
+                        args.filterNot { it.startsWith("-Xjdk-release=") } + listOf(
+                            "-Xjdk-release=9",
+                            "-Xsuppress-version-warnings",
+                            "-Xexpect-actual-classes"
+                        )
+                    }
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -143,8 +143,14 @@ object Java9Modularity {
             multiPlatformEnabled.set(compileTask.get().multiPlatformEnabled)
             compilerOptions {
                 jvmTarget.set(JvmTarget.JVM_9)
-                freeCompilerArgs.addAll(
-                    listOf("-Xjdk-release=9",  "-Xsuppress-version-warnings", "-Xexpect-actual-classes")
+                freeCompilerArgs.set(
+                    compileTask.flatMap { it.compilerOptions.freeCompilerArgs }.map { args ->
```
fzhinkin, line 147: I don't really like how it looks, but hey, it makes our CI green again.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -143,8 +143,14 @@ object Java9Modularity {
...
+                        args.filterNot { it.startsWith("-Xjdk-release=") } + listOf(
+                            "-Xjdk-release=9",
+                            "-Xsuppress-version-warnings",
+                            "-Xexpect-actual-classes"
```
sandwwraith, line 151: Won't this mean we pass `-Xexpect-actual-classes` now twice? :)

KvanTTT: Maybe, but there is intentionally no CLI warning for duplicated arguments with identical values like `-Xexpect-actual-classes` (in this case it's implicit `true`), see [KT-84132](https://youtrack.jetbrains.com/projects/KT/issues/KT-84132/CLI-regression-in-deduplication-of-same-value-arguments) But the following should be reported: `-Xjdk-release=11 -Xjdk-release=17`.

sandwwraith (approved):

fzhinkin:
`@ALikhachev` suggested a neat way to configure it, I'll implement it later.

PR #3248, opened 2026-09-01
Use the hand-rolled function for checking if a character needs escapi…

qwwdfsad:
…ng when writing quoted strings

Rationale:

For large enough strings (notably: https://github.com/ej-technologies/serialization-comparison), copying and iterating over a string is fast enough, but range checks to see if we hit a slow path to escape Unicode symbols dominate the execution time.

Replacing it with a hand-rolled check produces up to 30%+ better throughput (e.g. on the benchmark from the linked repo).

**UPD**:
https://github.com/ej-technologies/serialization-comparison/pull/1 is merged.

Before this change:
```
Benchmark                             Mode  Cnt   Score   Error  Units
JsonRoundTripBenchmark.jacksonJava    avgt    5  21.126 ± 1.944  us/op
JsonRoundTripBenchmark.jacksonKotlin  avgt    5  24.521 ± 0.143  us/op
JsonRoundTripBenchmark.kotlinxStream  avgt    5  33.423 ± 0.376  us/op
JsonRoundTripBenchmark.kotlinxString  avgt    5  33.453 ± 1.762  us/op
JsonRoundTripBenchmark.moshiJava      avgt    5  32.571 ± 0.328  us/op
JsonRoundTripBenchmark.moshiKotlin    avgt    5  29.875 ± 0.156  us/op
```

After:
```
Benchmark                             Mode  Cnt   Score   Error  Units
JsonRoundTripBenchmark.jacksonJava    avgt    5  20.888 ± 0.301  us/op
JsonRoundTripBenchmark.jacksonKotlin  avgt    5  24.647 ± 0.196  us/op
JsonRoundTripBenchmark.kotlinxStream  avgt    5  24.156 ± 0.831  us/op
JsonRoundTripBenchmark.kotlinxString  avgt    5  25.656 ± 0.150  us/op
JsonRoundTripBenchmark.moshiJava      avgt    5  32.848 ± 0.180  us/op
JsonRoundTripBenchmark.moshiKotlin    avgt    5  30.142 ± 0.247  us/op
```

### DIFF
--- a/formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
+++ b/formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
@@ -30,7 +30,11 @@ public val ESCAPE_STRINGS: Array<String?> = arrayOfNulls<String>(93).apply {
     this[0x0c] = "\\f"
 }
 
-internal val ESCAPE_MARKERS: ByteArray = ByteArray(93).apply {
+internal fun isCodePointRequiringEscapeSequence(charCode: Int): Boolean =
+    charCode <= 0x1f || charCode == 0x22 || charCode == 0x5c
+
+// Note: it is 128 so Hotspot successfully eliminates all range checks
+internal val ESCAPE_MARKERS: ByteArray = ByteArray(128).apply {
     for (c in 0..0x1f) {
         this[c] = 1.toByte()
     }
--- a/formats/json/jvmMain/src/kotlinx/serialization/json/internal/StringJsonWriter.kt
+++ b/formats/json/jvmMain/src/kotlinx/serialization/json/internal/StringJsonWriter.kt
@@ -57,7 +57,7 @@ internal actual class StringJsonWriter : InternalJsonWriter {
         for (i in sz until sz + length) {
             val ch = arr[i].code
             // Do we have unescaped symbols?
-            if (ch < ESCAPE_MARKERS.size && ESCAPE_MARKERS[ch] != 0.toByte()) {
+            if (isCodePointRequiringEscapeSequence(ch)) {
                 // Go to slow path
                 return appendStringSlowPath(i - sz, i, text)
             }
--- a/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/RoundtripStringEncodingBenchmark.kt
+++ b/benchmark/src/jmh/kotlin/kotlinx/benchmarks/json/RoundtripStringEncodingBenchmark.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.benchmarks.json
+
+import kotlinx.serialization.builtins.serializer
+import kotlinx.serialization.json.Json
+import org.openjdk.jmh.annotations.Benchmark
+import org.openjdk.jmh.annotations.BenchmarkMode
+import org.openjdk.jmh.annotations.Fork
+import org.openjdk.jmh.annotations.Measurement
+import org.openjdk.jmh.annotations.Mode
+import org.openjdk.jmh.annotations.OutputTimeUnit
+import org.openjdk.jmh.annotations.Scope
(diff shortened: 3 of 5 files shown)


qwwdfsad:
Score on the https://github.com/ej-technologies/serialization-comparison/pull/1 (decided not to commit here as it's a one-shot thing): before -- ~32 us/ops, after -- ~24 us/ops.

I know you'd prefer to check `dtraceasm` and the reasoning for 128, here it is: https://bugs.openjdk.org/browse/JDK-8003585 :)

formats/json/commonMain/src/kotlinx/serialization/json/internal/StringOps.kt
```diff
@@ -30,7 +30,11 @@ public val ESCAPE_STRINGS: Array<String?> = arrayOfNulls<String>(93).apply {
     this[0x0c] = "\\f"
 }
 
-internal val ESCAPE_MARKERS: ByteArray = ByteArray(93).apply {
+internal fun isCodePointRequiringEscapeSequence(charCode: Int): Boolean =
```
fzhinkin, line 33: Please also add the test from https://github.com/Kotlin/kotlinx.serialization/blob/c09c44dc906f655004e43b113c4442baf094a986/formats/json/commonTest/src/kotlinx/serialization/json/internal/EscapeMarkersTest.kt

fzhinkin:
And as a bonus, it will be nice to add a dedicated benchmark to our suite.

qwwdfsad:
Sure, done. I am not sure we are going to use this benchmark any further tho

fzhinkin (approved):

formats/json/jvmMain/src/kotlinx/serialization/json/internal/StringJsonWriter.kt
```diff
@@ -57,7 +57,7 @@ internal actual class StringJsonWriter : InternalJsonWriter {
         for (i in sz until sz + length) {
             val ch = arr[i].code
             // Do we have unescaped symbols?
-            if (ch < ESCAPE_MARKERS.size && ESCAPE_MARKERS[ch] != 0.toByte()) {
+            if (isCodePointRequiringEscapeSequence(ch)) {
```
qwwdfsad, line 60: FTR: Making `ESCAPE_MARKERS` a power-of-two array was enough to get the same performance gain.

But avoiding touching an extra chunk of memory is generally better (the score is the same), and we are not relying on a specific JIT optimization (also: ART, K/N, etc.).

Co-authored with @fzhinkin, who was independently nerd-sniped with the same blogpost!

pdvrieze: One very small one is that you are actually not working with codepoints but utf16 characters (standard `Char`s - that could be surrogate pairs). In this case you only care about ascii characters anyway and UTF is designed smartly to support that.
As to optimization, for strings that are stored as utf8 it would be faster to compare it using byte values rather than char values (any non-ascii value has the high bit set so doesn't match and its utf8 structure can be ignored).

vlsi:
This reminds me https://github.com/netty/netty/pull/9896.
Just wondering, have you compared your approach with `ByteArray(Character.MAX_VALUE + 1)`? Looks like it might eliminate the bound checks altogether.

ingokegel:
@qwwdfsad On my Linux Zen 4 the improvement for kotlinx.serialization is even more pronounced 🎉

```
Benchmark                             Mode  Cnt   Score   Error  Units
JsonRoundTripBenchmark.jacksonJava    avgt    5  29.263 ± 4.170  us/op
JsonRoundTripBenchmark.jacksonKotlin  avgt    5  33.478 ± 3.806  us/op
JsonRoundTripBenchmark.kotlinxStream  avgt    5  27.469 ± 2.095  us/op
JsonRoundTripBenchmark.kotlinxString  avgt    5  24.743 ± 3.856  us/op
JsonRoundTripBenchmark.moshiJava      avgt    5  34.790 ± 0.877  us/op
JsonRoundTripBenchmark.moshiKotlin    avgt    5  31.695 ± 1.472  us/op
```

PR #3253, opened 2026-09-02
Upgrade to Gradle 9 and enable configuration cache

qwwdfsad:
I was tired of redefining my JAVA_HOME to older Java and waiting for 3 seconds per test instead of 0.x, that's the whole premise


shanshin:
Replacement for the old PR #3083

fzhinkin:
@qwwdfsad, builds on CI all failed with OOME. We need more heap for Gradle 9, I guess?

qwwdfsad:
@fzhinkin Gradle 9 build actually passed (it was #3247 that made it red), it's CC and the fact that we started K/N linking in parallel

qwwdfsad:
@fzhinkin all addressed, rebased on dev, ready for review

integration-test/gradle.properties
```diff
@@ -17,6 +17,8 @@ source_map_support_version = 0.5.3
 kapt.use.k2=true
 
 org.gradle.kotlin.dsl.allWarningsAsErrors=true
+org.gradle.configuration-cache=true
+org.gradle.jvmargs="-Dorg.gradle.jvmargs=-Xmx4g -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError"
```
fzhinkin: Should it be `org.gradle.jvmargs="-Xmx4g -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError"` instead (here, and in the root gradle.properties file)?

Otherwise, the change looks good

qwwdfsad: without quotes, done

fzhinkin: Now the build failed on CI w/ OOME :(

fzhinkin (approved):

PR #3263, opened 2026-09-09
Stabilize `kotlin.uuid.Uuid` serializer

whyoleg:
Inspired by #3261. `Uuid` class is stable starting from Kotlin 2.4.0.

I've also updated the documentation to more clearly state that both "hex" and "hex-and-dash" formats are supported during **deserialization**, since the serializer uses `Uuid.parse`. Feel free to suggest a better way to describe the difference between serialization and deserialization.
An alternative could be to switch to using `Uuid.parseHexDash` for deserialization, so it will work according to the original documentation, but it will be a "behavior" breaking change, so I'm not sure which one will be better here.


shanshin:
@ilya-g, can we consider UuidSerializer to be as stable as Uuid?

fzhinkin:
@shanshin, we'll discuss it the next week. But yeah, it should be stable.

pdvrieze:
@whyoleg Nothing wrong with accepting broader deserialization than serialization as long as the broader format is documented, well-designed and actually tested (aka. not an implementation detail that will bite you back in the future or with a different library). As long as tests for all formats (and erroneous versions) are present, it should be valid to support al variants.

As clarification, it would be invalid if the implementation merely ignores all dashes and allows content that is not a valid uuid serialization: `----a---f--…`

fzhinkin:
@shanshin, we can consider it stable.

With that said, we might also consider introducing some stricter serializes that serialize UUID to a particular format and accept only that format during deserialization. But is has nothing to do with ExperimentalUuidApi.

core/commonMain/src/kotlinx/serialization/builtins/BuiltinSerializers.kt
```diff
@@ -265,16 +265,16 @@ public fun Instant.Companion.serializer(): KSerializer<Instant> = InstantSeriali
...
 /**
  * Returns serializer for [Uuid].
- * Serializer operates with a standard UUID string representation, also known as "hex-and-dash" format —
- * [RFC 9562 section 4](https://www.rfc-editor.org/rfc/rfc9562.html#section-4).
+ * It is serialized as a string in the format produced by [Uuid.toString]:
+ * the standard UUID string representation, also known as "hex-and-dash" format,
+ * and described in [RFC 9562 section 4](https://www.rfc-editor.org/rfc/rfc9562.html#section-4).
  *
- * Serialization always produces lowercase string, deserialization is case-insensitive.
+ * Deserialization is case-insensitive and supports parsing both standard hex-and-dash and the hexadecimal string representations.
```
fzhinkin:
```suggestion
 * Deserialization is case-insensitive and supports parsing string representations supported by [Uuid.parse], such as hex-and-dash and the hexadecimal formats.
```

Or something like that, so we don't have to update if if `parse` will start accepting more formats.

fzhinkin (approved):

shanshin (approved):

PR #3269, opened 2026-09-09
[Json] Correctly detect overflow when parsing numeric literals

### LINKED ISSUE #3267: JsonPrimitive.longOrNull returns a wrapped value for integers above 2^64
Hey folks - hit this one on xtdb/xtdb#6038. AI summary below (which I've read through), let me know if anything's unclear 🙂 

Cheers,

James

---

**Version:** 1.10.0. The same lexer code is present on `v1.12.0-RC`.

`longOrNull` is documented to return "null if current element is not a valid representation of number", and it does so for `Long.MAX_VALUE + 1`. But an integer large enough to wrap past 2^64 comes back as a small, entirely different number instead of null.

## Reproducer

```kotlin
val p = Json.parseToJsonElement("18446744073709551623").jsonPrimitive  // 2^64 + 7
println(p.longOrNull)  // 7
println(p.intOrNull)   // 7
```

Expected: `null` from both, as for `9223372036854775808`.

Observed across the boundary:

| literal | value | `longOrNull` | `intOrNull` |
| --- | --- | --- | --- |
| `9223372036854775807` | 2^63-1 | `9223372036854775807` | `null` |
| `9223372036854775808` | 2^63 | `null` | `null` |
| `18446744073709551615` | 2^64-1 | `null` | `null` |
| `18446744073709551616` | 2^64 | `0` | `0` |
| `18446744073709551623` | 2^64+7 | `7` | `7` |
| `-18446744073709551623` | -(2^64+7) | `-7` | `-7` |
| `36893488147419103232` | 2^65 | `0` | `0` |

## Cause

`longOrNull` resolves through `parseLongImpl` to `AbstractJsonLexer.consumeNumericLiteral`, which accumulates the digits negatively and guards overflow in two places. In the digit loop:

```kotlin
accumulator = accumulator * 10 - digit
if (accumulator > 0) fail("Numeric value overflow")
```

and on the way out:

```kotlin
return when {
    isNegative -> accumulator
    accumulator != Long.MIN_VALUE -> -accumulator
    else -> fail("Numeric value overflow")
}
```

The first catches a wrap that crosses into positive, which is what rejects `18446744073709551615` (it accumulates to `1`). The second — added in #1367, for `9223372036854775808` reading as `-9223372036854775808` — catches a positive literal that accumulated to exactly `Long.MIN_VALUE`.

Neither catches a wrap that lands on any other negative value. `18446744073709551623` accumulates to `-7`, stays negative, and is negated on return to `7`.

## Scope

- `long` / `longOrNull` and `int` / `intOrNull`. The `int` accessors range-check *after* `parseLongImpl`, so a wrapped value that falls inside `Int` range passes the check.
- `StreamingJsonDecoder.decodeLong()` and `decodeInt()` call the same lexer method, so a `@Serializable` class with a `Long` field silently receives the wrapped value when decoded from JSON.

## Note

#1300 fixed the sibling case for narrow types decoded from a `JsonElement`; this is the `Long` path in the lexer itself.

fzhinkin:
Closes #3267


formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -641,8 +641,9 @@ internal abstract class AbstractJsonLexer(internal val configuration: JsonConfig
                 exponentAccumulator = exponentAccumulator * 10 + digit
                 continue
             }
-            accumulator = accumulator * 10 - digit
-            if (accumulator > 0) fail("Numeric value overflow")
+            val newAccumulator = accumulator * 10 - digit
+            if (newAccumulator > accumulator) fail("Numeric value overflow")
```
shanshin: Value `46116860184273879020` still passes this check

shanshin: In a wonderful way ` -4611686018427387902L * 10L == -9223372036854775788L ` (I love "bitwise" calculations) - by a magical coincidence, the truncated value is also negative.

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -641,8 +641,12 @@ internal abstract class AbstractJsonLexer(internal val configuration: JsonConfig
                 exponentAccumulator = exponentAccumulator * 10 + digit
                 continue
             }
+            if (accumulator < MUL_BY_10_OVERFLOW_LIMIT
```
shanshin: What about check like this,
```
if (accumulator < MUL_BY_10_OVERFLOW_LIMIT) error()
accumulator = accumulator * 10
if (accumulator < Long.MIN_VALUE + digit) error()
accumulator = accumulator - digit
```

There’s not much of a fundamental difference; it’s just that there are fewer checks.

fzhinkin: Sure! I guess, the second condition could be replaced with a check against old accumulator value, to make it even more simple.

formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt
```diff
@@ -641,8 +641,12 @@ internal abstract class AbstractJsonLexer(internal val configuration: JsonConfig
                 exponentAccumulator = exponentAccumulator * 10 + digit
```
fzhinkin, line 641: BTW, it overflows too: `10e36893488147419103232` is parsed as `10`.

fzhinkin: Some overflows are easy to handle, we can bail out when `exponentAccumulator > 19` when exponent is positive.

For negative exponents, we have to go through full parsing, as `0e-100000000000` should be parsed to `0`, assuming there are no bugs left in the parser. As well as any other mantissa given small enough exponent. 😿

pdvrieze: Please note that exponents can also over/underflow given the format. As Ints and Longs have no inherent exponent this would probably be best expressed as the minimum/maximum expressed value:
- a signed int has a maximum 10 digits, so the exponent range would only be "valid" between 10 and -9.

It would make most sense to handle over/underflows in a consistent manner. This means that for int/long values underflows should be handled as invalid as overflows are also handled that way (probably also values that would have decimals).

There is the special case of a `0` mantissa, in which case the effective value is always `0e0` and the only requirement for the exponent is that it is a valid digit (which his cheaper to check than fully parsing it to a number)

Note also that doubles have the lowest exponent (subnormal/denormalized) as -1074 (which is much less than -100000000000) so at this point it is a matter of choice/definition whether the special case of `0e`<anything integer> is handled as underflow to 0 (IEEE 754 behaviour).

fzhinkin: Currently, extremely large positive exponents lead to "numeric value overflow" error, and extremely large negative exponents lead to `0.0`. Thus, the underflow is "considered" legal value.

~~I agree that underfloor handling is inconsistent here and the underflow should be an error (the same way it would be we would try to parse the same literal as `Double`).~~

Oh, the underflow is not an error when parsing string to double. So it shouldn't be an error in this case too.

PR #3270, opened 2026-09-11
Remove leading whitespace from Proguard rules.

cpovirk:
The whitespace should be harmless, but Bazel appears to have a bug that [rejects leading whitespace](https://github.com/bazelbuild/bazel/blob/304d98eaf2c089ccc4978d7b900a18631291376d/tools/jdk/proguard_allowlister.py#L59). I'm looking to fix that, but this PR works around the issue in the meantime.

(The whitespace appears to have been present since the beginning in https://github.com/Kotlin/kotlinx.serialization/pull/2092.)


fzhinkin:
@shanshin was there a special meaning behind the spaces? :)

shanshin (approved):

shanshin:
> was there a special meaning behind the spaces? :)

No, they got there by accident

